PR #4, opened 2024-08-18
update thirdparty download timeout

wxyucs:


extern/boost/boost.cmake
```diff
@@ -44,7 +44,9 @@ ExternalProject_Add(
     LOG_CONFIGURE TRUE
     LOG_BUILD TRUE
     LOG_INSTALL TRUE
-    DOWNLOAD_NO_PROGRESS 1
+    DOWNLOAD_NO_PROGRESS 0
```
jiaweizone, line 47: Why is progress display turned off for some operations and on for others?

wxyucs: depends on the tar file size, show the progress in large file downloading only

jiaweizone (approved):
LGTM

PR #6, opened 2024-08-18
add circleci workflow to check code format and run tests

wxyucs:


tests/test_diskann.cpp
```diff
@@ -329,6 +329,7 @@ TEST_CASE("DiskAnn IP Search", "[ft][diskann]") {
     REQUIRE(recall > 0.70);
 }
 
+/* FIXME: segmentation fault on some platform
```
jiaweizone, line 332: Is it possible to exclude only specific platforms?

wxyucs: its hard for now, the crash line is in the BLAS library, maybe we can use another impl of OPQ

jiaweizone: OK

jiaweizone (approved):

PR #36, opened 2024-09-24
make openblas always run with single thread

wxyucs:
diskann index building is a multi-thread task, so we should set openblas use single thread to avoid conflict.

https://github.com/OpenMathLib/OpenBLAS/wiki/Faq#how-can-i-use-openblas-in-multi-threaded-applications

### DIFF
--- a/extern/openblas/openblas.cmake
+++ b/extern/openblas/openblas.cmake
@@ -21,7 +21,7 @@ ExternalProject_Add(
         OMP_NUM_THREADS=1
         PATH=/usr/lib/ccache:$ENV{PATH}
         LD_LIBRARY_PATH=/opt/alibaba-cloud-compiler/lib64/:$ENV{LD_LIBRARY_PATH}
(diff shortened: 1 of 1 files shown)


extern/openblas/openblas.cmake
```diff
@@ -21,7 +21,7 @@ ExternalProject_Add(
         OMP_NUM_THREADS=1
         PATH=/usr/lib/ccache:$ENV{PATH}
         LD_LIBRARY_PATH=/opt/alibaba-cloud-compiler/lib64/:$ENV{LD_LIBRARY_PATH}
-        make USE_OPENMP=1 -j${NUM_BUILDING_JOBS}
+        make USE_THREAD=0 USE_LOCKING=1 -j${NUM_BUILDING_JOBS}
```
jiaweizone, line 24: image
OK,  set OpenBLAS to use single thread

jiaweizone (approved):
LGTM

PR #38, opened 2024-09-24
remove redundant judgments during the HNSW search process

inabao:


src/algorithm/hnswlib/hnswalg.cpp
```diff
@@ -1404,13 +1404,9 @@ HierarchicalNSW::searchKnn(const void* query_data,
     }
 
     MaxHeap top_candidates;
-    if (num_deleted_) {
-        top_candidates =
-            searchBaseLayerST<true, true>(currObj, query_data, std::max(ef, k), isIdAllowed);
```
jiaweizone, line 1409: Why remove this logic ?

inabao: num_deleted_ is the original logic for marking deletions in HNSW, but we no longer rely on this logic.

jiaweizone (approved):
LGTM

wxyucs (approved):
lgtm

PR #46, opened 2024-09-27
change the memory allocator of the HNSW in stl to a custom memory allocator

inabao:


src/utils.h
```diff
@@ -20,12 +20,27 @@
...
 
+#include "default_allocator.h"
 #include "spdlog/spdlog.h"
 #include "vsag/errors.h"
 #include "vsag/expected.hpp"
 
 namespace vsag {
 
+template <typename T>
+using unordered_set =
```
jiaweizone: Use `UnorderedSet` explicitly to distinguish it from the std lib

src/utils.h
```diff
@@ -20,12 +20,27 @@
...
+    std::unordered_set<T, std::hash<T>, std::equal_to<T>, vsag::AllocatorWrapper<T>>;
+
+template <typename T>
+using vector = std::vector<T, vsag::AllocatorWrapper<T>>;
```
jiaweizone: ditto: vector -> Vector

src/utils.h
```diff
@@ -20,12 +20,27 @@
...
+
+template <typename KeyType, typename ValType>
+using unordered_map = std::unordered_map<KeyType,
```
jiaweizone: ditto

src/utils.h
inabao: fixed

src/utils.h
inabao: fixed

src/utils.h
inabao: fixed

wxyucs (approved):
lgtm

jiaweizone (approved):
LGTM

Issue #49, opened 2024-09-30
 fix memory leak in hnsw


inabao:
**Describe the bug**
Currently, a lot of memory allocation in the HNSW constructor can lead to memory leaks when memory allocation fails, as some memory has already been allocated. In addition, some pointers are not managed using shared pointers, which can easily lead to memory leaks.

**To Reproduce**
The places where code leaks may occur shows as follows:
```cpp
HierarchicalNSW::HierarchicalNSW(SpaceInterface* s,
                                 size_t max_elements,
                                 vsag::Allocator* allocator,
                                 size_t M,
                                 size_t ef_construction,
                                 bool use_reversed_edges,
                                 bool normalize,
                                 size_t block_size_limit,
                                 size_t random_seed,
                                 bool allow_replace_deleted)
    : allocator_(allocator),
      link_list_locks_(max_elements, allocator),
      label_op_locks_(MAX_LABEL_OPERATION_LOCKS, allocator),
      allow_replace_deleted_(allow_replace_deleted),
      use_reversed_edges_(use_reversed_edges),
      normalize_(normalize),
      label_lookup_(allocator),
      deleted_elements_(allocator) {
    max_elements_ = max_elements;
    num_deleted_ = 0;
    data_size_ = s->get_data_size();
    fstdistfunc_ = s->get_dist_func();
    dist_func_param_ = s->get_dist_func_param();
    dim_ = data_size_ / sizeof(float);
    M_ = M;
    maxM_ = M_;
    maxM0_ = M_ * 2;
    ef_construction_ = std::max(ef_construction, M_);

    element_levels_ = (int*)allocator->Allocate(max_elements * sizeof(int));

    level_generator_.seed(random_seed);
    update_probability_generator_.seed(random_seed + 1);

    size_links_level0_ = maxM0_ * sizeof(tableint) + sizeof(linklistsizeint);
    size_data_per_element_ = size_links_level0_ + data_size_ + sizeof(labeltype);
    offsetData_ = size_links_level0_;
    label_offset_ = size_links_level0_ + data_size_;
    offsetLevel0_ = 0;

    if (use_reversed_edges_) {
        reversed_level0_link_list_ =
            (reverselinklist**)allocator->Allocate(max_elements_ * sizeof(reverselinklist*));
        memset(reversed_level0_link_list_, 0, max_elements_ * sizeof(reverselinklist*));
        reversed_link_lists_ = (vsag::UnorderedMap<int, reverselinklist>**)allocator->Allocate(
[...]

PR #57, opened 2024-10-10
fix hnsw memory leak in search

inabao:
fixed: #49

### DIFF
--- a/src/algorithm/hnswlib/hnswalg.cpp
+++ b/src/algorithm/hnswlib/hnswalg.cpp
@@ -269,12 +269,12 @@ HierarchicalNSW::getRandomLevel(double reverse_size) {
 
 MaxHeap
 HierarchicalNSW::searchBaseLayer(tableint ep_id, const void* data_point, int layer) {
-    VisitedList* vl = visited_list_pool_->getFreeVisitedList();
+    std::shared_ptr<VisitedList> vl = visited_list_pool_->getFreeVisitedList();
     vl_type* visited_array = vl->mass;
     vl_type visited_array_tag = vl->curV;
 
-    MaxHeap top_candidates;
-    MaxHeap candidateSet;
+    MaxHeap top_candidates(allocator_);
+    MaxHeap candidateSet(allocator_);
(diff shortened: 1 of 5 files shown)


src/algorithm/hnswlib/hnswalg.cpp
```diff
@@ -71,8 +71,19 @@ HierarchicalNSW::HierarchicalNSW(SpaceInterface* s,
...
+    allocator_->Deallocate(reversed_level0_link_list_);
+    allocator_->Deallocate(reversed_link_lists_);
+    allocator_->Deallocate(molds_);
+    allocator_->Deallocate(link_lists_);
+}
+
 bool
 HierarchicalNSW::init_memory_space() {
+    // release the memory allocated by the init_memory_space function that was called earlier
+    reset();
```
jiaweizone, line 91: Will `init_memory_space` be called multiple times?

inabao: In very rare cases, calling `init_memory_space` may result in a `bad_alloc` issue. Therefore, before initializing memory, it is necessary to release any previously allocated memory that may exist.

jiaweizone: OK, Also need to clean released memory to avoid double free

src/algorithm/hnswlib/hnswalg.cpp
```diff
@@ -71,8 +71,19 @@ HierarchicalNSW::HierarchicalNSW(SpaceInterface* s,
     rev_size_ = 1.0 / mult_;
 }
 
+void
+HierarchicalNSW::reset() {
+    allocator_->Deallocate(element_levels_);
```
jiaweizone, line 76: Here need clean freed memory pointer (set to nullptr)

inabao: done

jiaweizone (approved):
LGTM

LHT129 (approved):
lgtm

Issue #40, opened 2024-09-25
Add a new Layer "DataCell" to manage data access 


LHT129:
I think high level index should ignore the specific implement of quantization and io type, So the quantization and io components can be managed by new layer DataCell which provide different data access interface to the index layer. just like the following show

|Layer| Description|Implement|
|-------|--------|---------|
|Index Layer | ANN algorithm & architecture.          |                       Graph (HNSW,NNDescent) + Reorder,  IVF(PQ,SQ,fastscan), HNSW_IVF, MultiIndex ......|
|DataCell Layer| data access, use quantization, io and others    | FlattenDC, GraphNeighborhoodDC, MixtureDC .......|
|Operator Layer| quantizations, IO and other basic components                                                 | PQ,SQ,MemoryIO,SSDIO,......|

this abstract layered approach may facilitate scalability and code reuse
currently we only have index layer, so the issue will provide Datacell Layer and Operator Layer in this feature.
the related pull requests:

- #25 
- #34 
- #37 
- #41 
- #35 
- #59 
- #60 
- #61 
- #62 
- #79 
- #80 
- #96 
- #111 
- #114 
- #125 
todo tasks:
- support mixture DataCell
- adapt Graph DataCell for HNSW
- implement HNSW + SQ4 + SQ8 Index

LHT129:
![image](https://github.com/user-attachments/assets/faaa8e38-c78a-497d-aeb7-a62da2815fec)

PR #62, opened 2024-10-11
Add Graph Interface

LHT129:
issue: #40

### DIFF
--- a/src/data_cell/graph_interface.h
+++ b/src/data_cell/graph_interface.h
@@ -1,0 +1,122 @@
+
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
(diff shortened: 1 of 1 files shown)


ShawnShawnYou (approved):

inabao (approved):

src/data_cell/graph_interface.h
```diff
@@ -0,0 +1,122 @@
...
+        this->max_capacity_ = std::max(this->max_capacity_, total_count_ + 1);
+        this->InsertNeighborsById(total_count_ + 1, neighbor_ids);
+        IncreaseTotalCount(1);
+        return total_count_;
+    }
+
+    virtual void
+    IncreaseTotalCount(uint64_t count) {
+        std::unique_lock<std::mutex> lock(global_);
+        total_count_ += count;
```
jiaweizone, line 80: Can we use atomic increment here?

src/data_cell/graph_interface.h
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    virtual void
+    SetTotalCount(uint64_t total_count) {
+        this->total_count_ = total_count;
+    };
+
+    virtual void
+    SetMaxCapacity(uint64_t capacity) {
+        this->max_capacity_ = std::max(capacity, this->total_count_);
```
jiaweizone, line 110: `max_capacity_` equals `total_count_` ?

src/data_cell/graph_interface.h
LHT129: now use lock in external Index, we will apply atomic in the datacell layer or add concurrent interface in future

src/data_cell/graph_interface.h
LHT129: max_capacity_  >=  total_count_, equal is allowed

PR #71, opened 2024-10-17
using typing.h to define LabelType and InnerIdType

LHT129:


inabao (approved):

src/typing.h
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <cstdint>
+
+namespace vsag {
+
+using InnerIdType = uint32_t;  // inner id's type; index's vector count may less than 2^31 - 1
+using LabelType = uint64_t;    // external id's type
```
wxyucs, line 23: @LHT129  ID type should change to signed integer to avoid type-casting while calling from library interfaces

wxyucs (commented):
lgtm

wxyucs (approved):

PR #94, opened 2024-10-24
use parallel for ci_test

LHT129:
- introduce Makefile new tag: test_asan_parallel
- mark [diskann] & [hnsw] tags run parallel

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           key: fork-cache-{{ checksum "CMakeLists.txt" }}
           paths:
             - ./build
-      - run: make test_asan
+      - run: make test_asan_parallel
 
   main-branch-check:
     docker:
@@ -44,8 +44,9 @@ jobs:
       - restore_cache:
           keys:
             - main-ccache-{{ checksum "CMakeLists.txt" }}
-      - run: make test_asan
+      - run: make asan
       - save_cache:
           key: main-ccache-{{ checksum "CMakeLists.txt" }}
           paths:
             - ./build
+      - run: make test_asan_parallel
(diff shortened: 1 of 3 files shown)


src/bitset_impl_test.cpp
```diff
@@ -22,6 +22,10 @@
 
 using namespace roaring;
 
+TEST_CASE("false") {
+    REQUIRE(1 == 0);
```
jiaweizone: force failure this task ?

LHT129: just for test. remove now

inabao (approved):

test_asan_bg.sh
```diff
@@ -0,0 +1,35 @@
...
+  pids+=($!)
+done
+
+for pid in "${pids[@]}"
+do
+  wait $pid
+  exit_codes+=($?)
+done
+
+all_successful=true
```
jiaweizone: here can direct return `exit_codes` ?

LHT129: exit codes is an array. we may add some logger to detail which case fail in future

test_asan_bg.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
wxyucs: move this file in to `scripts/`

test_asan_bg.sh
```diff
@@ -0,0 +1,40 @@
...
+done
+
+./build/tests/functests -d yes ${UT_FILTER} --allow-running-no-tests ${othertag} &
+pids+=($!)
+
+for pid in "${pids[@]}"
+do
+  wait $pid
+  exit_codes+=($?)
+done
```
wxyucs: what is the behavior here when the second process exit before the first process

LHT129: no problem, I have checked

test_asan_bg.sh
LHT129: done

wxyucs:
lgtm

wxyucs (approved):

PR #98, opened 2024-10-28
support cosine in diskann

inabao:
#99


extern/diskann/DiskANN/src/disk_utils.cpp
```diff
@@ -1039,29 +1039,6 @@ void create_disk_layout(const T *data, uint32_t npts, uint32_t ndims, const std:
         std::ifstream reorder_data_reader;
 
         uint32_t npts_reorder_file = 0, ndims_reorder_file = 0;
-        if (reorder_data_file != std::string(""))
```
jiaweizone, line 1042: Why remove reorder logic?

extern/diskann/DiskANN/src/in_mem_data_store.cpp
```diff
@@ -258,6 +264,11 @@ template <typename data_t> void InMemDataStore<data_t>::get_vector(location_t lo
     {
         loc = _loc_to_memory_index[loc];
         memcpy(dest, _data + loc * this->_dim, this->_dim * sizeof(data_t));
+        if (_compute_norms)
+        for (int i = 0; i < this->_dim; ++i)
```
jiaweizone: indent, add `{}` to `if` clause

inabao: done

extern/diskann/DiskANN/src/disk_utils.cpp
inabao: This is the original reorder logic of DiskANN, which we did not rely on.

wxyucs (approved):
lgtm

PR #102, opened 2024-10-29
chore: compile example only when required

jiacai2050:
Skip build example by default, so users depending on vsag can skip build them too.


wxyucs (approved):
lgtm

jiaweizone (approved):
LGTM

PR #104, opened 2024-10-30
add check for number of elements in diskann

inabao:


src/index/diskann.cpp
```diff
@@ -226,6 +226,8 @@ DiskANN::build(const DatasetPtr& base) {
         auto vectors = base->GetFloat32Vectors();
         auto ids = base->GetIds();
         auto data_num = base->GetNumElements();
+        CHECK_ARGUMENT(data_num >= 2,
+                       fmt::format("base.num_elements({}) must be greater than 2", data_num));
```
wxyucs: is it related to any bug or issue?

inabao: It is not relate any bugs. We have constrained this matter in functests, but we haven't done any checks in the code, which is not acceptable.

wxyucs: should we add some unit tests or func tests to cover this new behavior?

inabao: I have added the related unit test.

wxyucs: @inabao I saw that the minimum number of elements is changed to 1. What is the consideration?

inabao: `>1` == `>=2`, so we should keep the minimum number of elements equal to 2 and greater than 1

wxyucs (approved):
lgtm

ShawnShawnYou (approved):

PR #109, opened 2024-11-01
move common params check to class IndexCommonParam

LHT129:
- common params(dim,metric_type,data_type...) should be checked for all index
- add param value's datatype check for hnsw and common params

### DIFF
--- a/src/index/index_common_param.cpp
+++ b/src/index/index_common_param.cpp
@@ -1,0 +1,86 @@
+
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/index/index_common_param.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // Check MetricType
+    CHECK_ARGUMENT(params.contains(PARAMETER_METRIC_TYPE),
+                   fmt::format("parameters must contains {}", PARAMETER_METRIC_TYPE));
+    const auto metric_obj = params[PARAMETER_METRIC_TYPE];
+    CHECK_ARGUMENT(metric_obj.is_string(),
+                   fmt::format("parameters[{}] must string type", PARAMETER_METRIC_TYPE));
+    std::string metric = metric_obj;
+    if (metric == METRIC_L2) {
```
inabao, line 58: Perhaps we don't need to require strict string equality here; we can ignore case.

LHT129: we can discuss the strict equality on another pull request.

inabao (approved):
LGTM

jiaweizone (approved):
LGTM

PR #136, opened 2024-11-13
use shared lock for global mutex of hgraph

LHT129:


wxyucs (approved):
lgtm

src/index/hgraph_index.cpp
```diff
@@ -246,60 +248,20 @@ HGraphIndex::hnsw_add(const DatasetPtr& data) {
...
-            }
-
-            if (need_lock) {
+                this->add_one_point(datas + i * dim_, level, inner_id);
                 entry_point_id_ = inner_id;
+                add_lock.unlock();
+            } else {
+                add_lock.unlock();
+                std::shared_lock<std::shared_mutex> rlock(this->global_mutex_);
+                this->add_one_point(datas + i * dim_, level, inner_id);
```
jiaweizone, line 264: add_one_point is thread safe ?

LHT129: no need global mutex, but have lock for per node's neighbors

src/index/hgraph_index.cpp
```diff
@@ -236,6 +236,8 @@ HGraphIndex::hnsw_add(const DatasetPtr& data) {
     auto cur_count = this->bottom_graph_->TotalCount();
     vsag::Vector<std::shared_mutex>(total + cur_count, allocator_).swap(this->neighbors_mutex_);
```
jiaweizone, line 237: swap function will destroy all original `neighbors_mutex_`. it is unsafe for other running thread to hold `neighbors_mutex_` at this time ?

src/index/hgraph_index.cpp
jiaweizone: OK

src/index/hgraph_index.cpp
LHT129: We don't support concurrent add & search currently, only support muti-thread search. To avoid user use add & search at same time, I want to introduce a feature list for all index. only support enabled features, detail in issue #154

jiaweizone: Only for background build index ?

LHT129: for concurrency, only support search

jiaweizone (approved):
LGTM

PR #140, opened 2024-11-15
add resize for hgraph

LHT129:
- visitlist for hgraph is fixed, now make it flexible
- fix typo error on graph interface

### DIFF
--- a/src/index/hgraph_index.cpp
+++ b/src/index/hgraph_index.cpp
@@ -17,6 +17,8 @@
 
 #include <fmt/format-inl.h>
 
+#include <memory>
+
 #include "common.h"
 #include "data_cell/sparse_graph_datacell.h"
 #include "hnsw.h"
@@ -234,7 +236,7 @@ HGraphIndex::hnsw_add(const DatasetPtr& data) {
     auto* ids = data->GetIds();
     auto* datas = data->GetFloat32Vectors();
     auto cur_count = this->bottom_graph_->TotalCount();
-    vsag::Vector<std::shared_mutex>(total + cur_count, allocator_).swap(this->neighbors_mutex_);
+    this->resize(total + cur_count);
 
     std::mutex add_mutex;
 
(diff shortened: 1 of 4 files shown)


inabao (approved):
LGTM

src/index/hgraph_index.cpp
```diff
@@ -671,5 +673,14 @@ HGraphIndex::calc_distance_by_id(const float* vector, int64_t id) const {
         }
     }
 }
+void
+HGraphIndex::resize(uint64_t new_size) {
+    auto cur_size = this->bottom_graph_->MaxCapacity();
+    if (new_size > cur_size) {
+        vsag::Vector<std::shared_mutex>(new_size, allocator_).swap(this->neighbors_mutex_);
```
jiaweizone, line 700: Is this thread-safe?

src/index/hgraph_index.cpp
```diff
@@ -671,5 +673,14 @@ HGraphIndex::calc_distance_by_id(const float* vector, int64_t id) const {
...
+        pool_ = std::make_shared<hnswlib::VisitedListPool>(1, new_size, allocator_);
```
jiaweizone: Need copy original data to new `pool_`

src/index/hgraph_index.cpp
LHT129: no concurrent in resize

src/index/hgraph_index.cpp
LHT129: no need to copy, the original visitlist is discarded

src/index/hgraph_index.cpp
```diff
@@ -209,7 +211,7 @@ HGraphIndex::hnsw_add(const DatasetPtr& data) {
     auto* ids = data->GetIds();
     auto* datas = data->GetFloat32Vectors();
     auto cur_count = this->bottom_graph_->TotalCount();
-    vsag::Vector<std::shared_mutex>(total + cur_count, allocator_).swap(this->neighbors_mutex_);
+    this->resize(total + cur_count);
```
wxyucs, line 239: is it ok if we call hnsw_add frequently

LHT129: we need a strategy for resize, less than 1000 increase 100, more than 1000 increase 1000? like this

src/index/hgraph_index.cpp
jiaweizone: maybe other thread is  lock `neighbors_mutex_` ?

LHT129: yeah, lock neighbors

jiaweizone: can simply write to `this->neighbors_mutex_.resize(new_size)` ?

jiaweizone (approved):
LGTM

Issue #154, opened 2024-11-21
Introduce index_feature for all index


LHT129:
The characteristics of each index are often different. Users often want to know the specific characteristics of the index, such as whether it supports adding data from empty index, whether it supports deletion, whether it supports concurrent search and insertion, etc.

In fact, the internal functions of each index should also be implemented according to the characteristic granularity, which is also convenient for testing and verification, and is more friendly to developers.

Therefore, we provide an enum of all characteristics in the external interface, and at the same time provide interface methods for printing the characteristics of the current index for a specific index

```C++
enum class IndexFeature {
     SUPPORT_ADD_WITHOUT_TRAIN = 1,
     SUPPOET_DELETE_BY_ID = 2,
     ...
}
```
```c++
std::vector<Feature> Index::GetFeatures();
bool Index::CheckFeature(IndexFeature feature);
```

the related pull requests:
#199
#210
#284

todo:
- add feature object 
- introduce it to all kinds of index
- use the feature to check and reconstruct some test

wxyucs:
sounds good, will it also available in the pyvsag?

wxyucs:
done

PR #155, opened 2024-11-21
introduce the IndexFeature for user

LHT129:
issue: #154

### DIFF
--- a/include/vsag/index_feature.h
+++ b/include/vsag/index_feature.h
@@ -1,0 +1,58 @@
+
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace vsag {
+
+enum class IndexFeature {
+    NEED_TRAIN, /**< Indicates that the class needs to be trained */
+
+    SUPPORT_BUILD,                     /**< Supports building the index */
+    SUPPORT_ADD,                       /**< Supports adding new elements */
+    SUPPORT_ADD_FROM_EMPTY,            /**< Supports adding elements to an empty index */
+    SUPPORT_KNN_SEARCH,                /**< Supports K-nearest neighbor search */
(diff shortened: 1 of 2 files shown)


include/vsag/index_feature.h
```diff
@@ -0,0 +1,57 @@
...
+    SUPPORT_BATCH_SEARCH,
+    SUPPORT_METRIC_TYPE_L2,
+    SUPPORT_METRIC_TYPE_INNER_PRODUCT,
+    SUPPORT_METRIC_TYPE_COSINE,
+    SUPPORT_SERIALIZE_FILE,
+    SUPPORT_SERIALIZE_BINARY_SET,
+    SUPPORT_DESERIALIZE_FILE,
+    SUPPORT_DESERIALIZE_BINARY_SET,
+    SUPPORT_DESERIALIZE_READER_SET,
+    SUPPORT_RESET,
```
wxyucs: what is this feature RESET?

include/vsag/index_feature.h
```diff
@@ -0,0 +1,57 @@
...
+    SUPPORT_SEARCH_ONE_WITH_MULTI_THREAD,
+    SUPPORT_BATCH_SEARCH_WITH_MULTI_THREAD,
+
+    SUPPORT_ADD_CONCURRENT,
+    SUPPORT_SEARCH_CONCURRENT,
+    SUPPORT_DELETE_CONCURRENT,
+    SUPPORT_ADD_SEARCH_CONCURRENT,
+    SUPPORT_ADD_DELETE_CONCURRENT,
+    SUPPORT_SEARCH_DELETE_CONCURRENT,
+    SUPPORT_ADD_SEARCH_DELETE_CONCURRENT,
```
wxyucs: should we make all of them true in the future?

include/vsag/index_feature.h
LHT129: clear all data in index

include/vsag/index_feature.h
LHT129: not for all, some index may static without add and search concurrently. some feature can be enable by pass different params to factory. but may bring some performance loss. so user can disable these feature, just avoiding the concurrent between such behaviors in their own process with vsag

include/vsag/index_feature.h
```diff
@@ -0,0 +1,58 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace vsag {
+
+enum class IndexFeature {
```
jiaweizone, line 20: IndexFeature -> IndexFeatures ?

LHT129: easily misunderstood as a collection, it is actually an enum class. We will create another method to check which features in current feature collection.

jiaweizone (approved):
LGTM

PR #163, opened 2024-11-26
implement for hgraph_range

LHT129:


src/index/hgraph_index.cpp
```diff
@@ -134,26 +134,26 @@ HGraphIndex::knn_search(const DatasetPtr& query,
         // check query vector
         CHECK_ARGUMENT(query->GetNumElements() == 1, "query dataset should contain 1 vector only");
 
-        auto ep = this->entry_point_id_;
+        InnerSearchParam search_param;
+        search_param.ep_ = this->entry_point_id_;
+        search_param.ef_ = 1;
```
inabao, line 150: It is best to set the default value of ef to k.

src/index/hgraph_index.cpp
```diff
@@ -662,27 +744,32 @@ HGraphIndex::calc_distance_by_id(const float* vector, int64_t id) const {
 }
 void
 HGraphIndex::add_one_point(const float* data, int level, InnerIdType inner_id) {
-    auto ep = this->entry_point_id_;
     MaxHeap result(allocator_);
+
+    InnerSearchParam param{
+        .ep_ = this->entry_point_id_,
```
inabao, line 761: public fields do not need to end with `_`

LHT129: rename all public member soon

src/index/hgraph_index.cpp
LHT129: 1 for high level graph

ShawnShawnYou (approved):

PR #169, opened 2024-11-28
introduce sq8 uniform quantizer

LHT129:


inabao (approved):
LGTM

src/quantization/sq8_uniform_quantizer.h
```diff
@@ -0,0 +1,278 @@
...
+    uint64_t offset_sum_{0};
+};
+
+template <MetricType metric>
+SQ8UniformQuantizer<metric>::SQ8UniformQuantizer(int dim, Allocator* allocator)
+    : Quantizer<SQ8UniformQuantizer<metric>>(dim, allocator) {
+    lower_bound_ = std::numeric_limits<DataType>::max();
+    diff_ = std::numeric_limits<DataType>::lowest();
+
+    this->code_size_ = 0;
```
jiaweizone, line 92: minor: `this` is not necessary?

src/quantization/sq8_uniform_quantizer.h
```diff
@@ -0,0 +1,278 @@
...
+        sum_type sum2 = *((sum_type*)(codes2 + offset_sum_));
+
+        result = lower_bound_ * (sum1 + sum2) + (diff_ / 255.0) * (diff_ / 255.0) * result -
+                 lower_bound_ * lower_bound_;
+
+        result = 1 - result;
+
+    } else {
+        logger::error("unsupported metric type");
+        result = 0;
```
jiaweizone, line 224: `0` can indicate an error ?

src/quantization/sq8_uniform_quantizer.h
LHT129: for template class this is necessary for var in basic class

src/quantization/sq8_uniform_quantizer.h
LHT129: same behavior as sq4 uniform, actually should process the wrong metric by throw exception

src/quantization/sq8_uniform_quantizer.h
jiaweizone: OK，for template class is necessary

src/quantization/sq8_uniform_quantizer.h
jiaweizone: Do we need a special state to indicate failure ？

LHT129: actually the metric error have special state at param check. in this code, we will not meet the metric error, but as an individual component, we suggest quantizer can process all errors by itself. So in future, we will have specific state and method to deal with this errors

jiaweizone: OK

jiaweizone (approved):
LGTM

ShawnShawnYou (approved):

PR #171, opened 2024-11-28
add boundary checks for data when parsing the conjugate graph

inabao:
fixed: #172


src/impl/conjugate_graph.cpp
```diff
@@ -153,6 +153,10 @@ ConjugateGraph::Serialize(std::ostream& out_stream) const {
 tl::expected<void, Error>
 ConjugateGraph::Deserialize(const Binary& binary) {
     auto func = [&](uint64_t offset, uint64_t len, void* dest) -> void {
+        if (len + offset > binary.size) {
+            throw std::runtime_error(
+                fmt::format("offset({}) + len({}) > size({})", offset, len, binary.size));
+        }
```
wxyucs, line 159: Is there a related issue or bug?

LHT129: done

PR #175, opened 2024-12-02
fix illegal instruction on platform which has avx only

LHT129:


wxyucs (approved):
lgtm

wxyucs:
@LHT129 I think we should release a patch version after this pull request is merged into the 0.11 branch

jiaweizone (approved):
LGTM

Issue #176, opened 2024-12-02
illegal instruction on platform which has avx only


LHT129:
Describe the bug :
illegal instruction error when run on platform without avx2 but have avx only, which the program is compiled in avx2

Environment :
- compile platform (have avx2)
- platform not support avx2
![image](https://github.com/user-attachments/assets/0f3992ae-8cef-4b80-941e-350c04c0de77)

ralated pull request:
#175 
#177

PR #177, opened 2024-12-02
fix illegal instruction on platform which has avx only

LHT129:
fixed: #176

### DIFF
--- a/src/simd/CMakeLists.txt
+++ b/src/simd/CMakeLists.txt
@@ -6,6 +6,9 @@ set (SIMD_SRCS
         sq4_simd.cpp
         sq4_uniform_simd.cpp
         sq8_uniform_simd.cpp
+        sse.cpp
+        avx.cpp
+        avx512.cpp
 )
 if (DIST_CONTAINS_SSE)
     set (SIMD_SRCS ${SIMD_SRCS} sse.cpp)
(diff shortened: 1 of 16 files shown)


ShawnShawnYou (approved):

src/simd/CMakeLists.txt
```diff
@@ -50,7 +55,7 @@ endmacro ()
 
 simd_add_definitions (DIST_CONTAINS_SSE -DENABLE_SSE=1)
 simd_add_definitions (DIST_CONTAINS_AVX -DENABLE_AVX=1)
-simd_add_definitions (DIST_CONTAINS_AVX2 -DENABLE_AVX2=1)
+#simd_add_definitions (DIST_CONTAINS_AVX2 -DENABLE_AVX2=1)
```
jiaweizone, line 58: remove unused line

src/simd/normalize.cpp
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+#include "normalize.h"
+
+#include "simd_status.h"
+
+namespace vsag {
+
+static NormalizeType
+SetNormalize() {
```
jiaweizone, line 23: GetNormalize ?

src/simd/normalize.cpp
```diff
@@ -0,0 +1,60 @@
...
+#if defined(ENABLE_SSE)
+        return sse::Normalize;
+#endif
+    }
+    return generic::Normalize;
+}
+NormalizeType Normalize = SetNormalize();
+
+static DivScalarType
+SetDivScalar() {
```
jiaweizone, line 42: GetDivScalar ?

LHT129: Get is better, but all simd define use Set now, I'd like to fix it in another PR

src/simd/normalize.cpp
LHT129: ditto

src/simd/CMakeLists.txt
LHT129: we have a very useful annotation for this line to fix the avx bug, so shound not remove this now 
image

src/simd/normalize.cpp
jiaweizone: OK

jiaweizone (approved):
LGTM

PR #191, opened 2024-12-07
use SafeAllocator instead Allocator

LHT129:
- inner Allocator must be safe
- hold the unique_ptr in common_param
- AllocatorWrapper use SafeAllocator

### DIFF
--- a/src/factory/factory.cpp
+++ b/src/factory/factory.cpp
@@ -70,9 +70,6 @@ Factory::CreateIndex(const std::string& origin_name,
             logger::debug("created a diskann index");
             return std::make_shared<DiskANN>(diskann_params, index_common_params);
         } else if (name == INDEX_HGRAPH) {
-            if (allocator == nullptr) {
-                index_common_params.allocator_ = DefaultAllocator::Instance().get();
-            }
             logger::debug("created a hgraph index");
             JsonType hgraph_params;
             if (parsed_params.contains(INDEX_PARAM)) {
--- a/src/index/hnsw.h
+++ b/src/index/hnsw.h
@@ -47,7 +47,7 @@ namespace vsag {
 
 class HNSW : public Index {
 public:
-    HNSW(HnswParameters hnsw_params, const IndexCommonParam& index_common_param);
+    HNSW(HnswParameters hnsw_params, IndexCommonParam& index_common_param);
 
     virtual ~HNSW() {
         alg_hnsw_ = nullptr;
@@ -276,7 +276,7 @@ private:
     bool is_init_memory_ = false;
     DataTypes type_;
 
-    std::shared_ptr<SafeAllocator> allocator_;
+    std::unique_ptr<SafeAllocator> allocator_;
 
     mutable std::mutex stats_mutex_;
     mutable std::map<std::string, WindowResultQueue> result_queues_;
--- a/src/index/hnsw_zparameters.h
+++ b/src/index/hnsw_zparameters.h
@@ -27,7 +27,7 @@ namespace vsag {
 struct HnswParameters {
 public:
     static HnswParameters
-    FromJson(JsonType& hnsw_param_obj, IndexCommonParam index_common_param);
+    FromJson(JsonType& hnsw_param_obj, const IndexCommonParam& index_common_param);
 
 public:
     // required vars
@@ -47,7 +47,7 @@ protected:
 struct FreshHnswParameters : public HnswParameters {
 public:
     static HnswParameters
-    FromJson(JsonType& hnsw_param_obj, IndexCommonParam index_common_param);
+    FromJson(JsonType& hnsw_param_obj, IndexCommonParam& index_common_param);
 
 private:
     FreshHnswParameters() = default;
--- a/src/index/index_common_param.cpp
+++ b/src/index/index_common_param.cpp
@@ -18,6 +18,7 @@
 #include <fmt/format-inl.h>
 
 #include "common.h"
+#include "default_allocator.h"
 #include "vsag/constants.h"
 
 namespace vsag {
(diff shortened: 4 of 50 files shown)


src/index/hgraph_index.cpp
```diff
@@ -169,13 +179,14 @@ HGraphIndex::knn_search(const DatasetPtr& query,
         while (search_result.size() > k) {
             search_result.pop();
         }
+        const auto& allocator = this->common_param_.allocator_.get();
 
         auto dataset_results = Dataset::Make();
-        dataset_results->Dim(search_result.size())->NumElements(1)->Owner(true, allocator_);
+        dataset_results->Dim(search_result.size())->NumElements(1)->Owner(true, allocator);
```
inabao, line 173: The safe allocator cannot be directly passed out from here because the lifecycle of the Dataset cannot be guaranteed. However, you can obtain the externally passed allocator by using the safe allocator.

src/index/hgraph_index.cpp
```diff
@@ -449,12 +461,13 @@ HGraphIndex::range_search(const DatasetPtr& query,
                 search_result.pop();
             }
         }
+        const auto& allocator = this->common_param_.allocator_.get();
 
         auto dataset_results = Dataset::Make();
-        dataset_results->Dim(search_result.size())->NumElements(1)->Owner(true, allocator_);
-        auto* ids = (int64_t*)allocator_->Allocate(sizeof(int64_t) * search_result.size());
```
inabao, line 455: ditto

src/index/index_common_param.cpp
```diff
@@ -18,14 +18,20 @@
...
 IndexCommonParam
 IndexCommonParam::CheckAndCreate(JsonType& params, Allocator* allocator) {
     IndexCommonParam result;
-    result.allocator_ = allocator;
+
+    if (allocator != nullptr) {
+        result.allocator_ = std::make_unique<SafeAllocator>(allocator);
+    } else {
+        result.allocator_ = std::make_unique<SafeAllocator>(new DefaultAllocator(), /*owner*/ true);
+    }
```
inabao, line 34: If we really need to make this change, we can remove the static keyword from DefaultAllocator. This part needs to be confirmed by the @wxyucs.

src/factory/factory.cpp
```diff
@@ -70,9 +70,6 @@ Factory::CreateIndex(const std::string& origin_name,
             logger::debug("created a diskann index");
             return std::make_shared<DiskANN>(diskann_params, index_common_params);
         } else if (name == INDEX_HGRAPH) {
-            if (allocator == nullptr) {
-                index_common_params.allocator_ = DefaultAllocator::Instance().get();
-            }
```
wxyucs, line 75: why can we remove this check now?

src/index/hnsw.h
```diff
@@ -276,7 +276,7 @@ class HNSW : public Index {
     bool is_init_memory_ = false;
     DataTypes type_;
 
-    std::shared_ptr<SafeAllocator> allocator_;
+    std::unique_ptr<SafeAllocator> allocator_;
```
wxyucs, line 279: HNSW object will hold the allocator instance now?

src/index/index_common_param.cpp
wxyucs: @inabao you mean we don't need to make DefaultAllocator singleton anymore?

src/safe_allocator.h
```diff
@@ -46,28 +43,37 @@ class SafeAllocator : public Allocator {
...
+        return allocator_;
     }
 
 public:
-    ~SafeAllocator() override = default;
+    ~SafeAllocator() override {
+        if (owner_) {
+            delete allocator_;
+        }
+    };
```
wxyucs, line 67: when should SafeAllocator release the allocator_?

LHT129: set owner flag for default allocator

src/index/hnsw.h
LHT129: use std::move to hold

src/factory/factory.cpp
LHT129: allocator will not be nullptr, it's initialize at common param parse

src/index/index_common_param.cpp
LHT129: I think singleton is not necessary

inabao: If we create a default allocator every time we create an index, then we won't need a global singleton. This relates to the design of resources that we discussed earlier—whether we need to create a default resource each time we create an index instead of using a global resource.

> @inabao you mean we don't need to make DefaultAllocator singleton anymore?

src/factory/factory.cpp
jiaweizone: For params checking, we need throw an invalid_argument exception when `allocator` not set ?

src/index/hgraph_index.cpp
```diff
@@ -169,13 +179,14 @@ HGraphIndex::knn_search(const DatasetPtr& query,
         while (search_result.size() > k) {
             search_result.pop();
         }
+        const auto& allocator = this->common_param_.allocator_.get();
```
jiaweizone, line 170: minor: We can add an field `shared_ptr<SafeAllocator> allocator_` to hold `common_param_.allocator` ?

src/index/hnsw_zparameters.h
```diff
@@ -47,7 +47,7 @@ struct HnswParameters {
 struct FreshHnswParameters : public HnswParameters {
 public:
     static HnswParameters
-    FromJson(JsonType& hnsw_param_obj, IndexCommonParam index_common_param);
+    FromJson(JsonType& hnsw_param_obj, IndexCommonParam& index_common_param);
```
jiaweizone, line 50: use `const IndexCommonParam&` ?

src/index/index_common_param.cpp
jiaweizone: The management logic for `Allocator` lifecycles is scattered across many places, requiring considerable code for maintenance. I think we need more centralized and simplified logic to manage Allocators. Let's discuss this later :)

PR #199, opened 2024-12-10
introduce feature list

LHT129:
#154

### DIFF
--- a/include/vsag/index_feature.h
+++ b/include/vsag/index_feature.h
@@ -17,8 +17,8 @@
 
 namespace vsag {
 
-enum class IndexFeature {
-    NEED_TRAIN, /**< Indicates that the class needs to be trained */
+enum IndexFeature {
+    NEED_TRAIN = 1, /**< Indicates that the class needs to be trained */
 
     SUPPORT_BUILD,                     /**< Supports building the index */
     SUPPORT_ADD,                       /**< Supports adding new elements */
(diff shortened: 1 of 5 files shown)


src/index_feature_list_test.cpp
```diff
@@ -0,0 +1,31 @@
...
+TEST_CASE("Test set&check", "[ut][index_feature_list]") {
+    using namespace vsag;
+    IndexFeatureList list;
+    auto count = static_cast<uint32_t>(IndexFeature::INDEX_FEATURE_COUNT);
+    for (uint32_t i = 0; i < count; ++i) {
+        REQUIRE(list.CheckFeature(static_cast<IndexFeature>(i)) == false);
+        list.SetFeature(static_cast<IndexFeature>(i));
+        REQUIRE(list.CheckFeature(static_cast<IndexFeature>(i)) == true);
+        list.SetFeature(static_cast<IndexFeature>(i), false);
+        REQUIRE(list.CheckFeature(static_cast<IndexFeature>(i)) == false);
```
inabao, line 29: We need to verify that the current feature's settings do not affect other features.

LHT129: done

inabao (approved):
LGTM

include/vsag/index_feature.h
```diff
@@ -17,8 +17,8 @@
 
 namespace vsag {
 
-enum class IndexFeature {
-    NEED_TRAIN, /**< Indicates that the class needs to be trained */
+enum IndexFeature {
+    NEED_TRAIN = 0, /**< Indicates that the class needs to be trained */
```
wxyucs: start from 1, 0 is reserved

wxyucs (changes requested):
@LHT129 please add an example and an issue of this feature

include/vsag/index_feature.h
LHT129: why reserve 0, I have make the last one is the count of the enum, if start with 1,  when I cast 0 to a IndexFeature type will cause error.

wxyucs: reserve of 0 is useful for exposing the problems with uninitiated variables or unexpected type casting

LHT129: done

Issue #203, opened 2024-12-10
Introduce vsag::Engine to manage resource instead  origin vsag::Factory


LHT129:
Regarding the management of Allocator, we are currently quite confused, especially in various special scenarios, such as static global default allocator, multiple indexes sharing one allocator, etc.

Therefore, we plan to introduce `vasg::Engine`, a new data structure, to manage resources such as allocator, thread pools, etc. Index generation can only be completed by `vasg::Engine`, so that we can ensure that all resources are valid before the `vsag::Engine` shutdown, and the resources sharing problems can also be solved very well.

Of course, the resources required for `vasg::Engine` initialization can be imported from outside.  The default resource implementation can be provided by vsag if the resource is not imported external.

The following PR will solve this problem:
#194 
#209

PR #209, opened 2024-12-12
implement of the engine and rewrite factory

LHT129:
issue: #203

- allocator shared_ptr hold in Index and Dataset
- the allocator ptr always point to SafeAllocator
- allocator ptr is used inner index
- only Engine/Resource can create Allocator Object

### DIFF
--- a/include/vsag/engine.h
+++ b/include/vsag/engine.h
@@ -17,16 +17,19 @@
 
 #include <memory>
 
+#include "dataset.h"
 #include "index.h"
 #include "resource.h"
 
 namespace vsag {
 class Engine {
 public:
-    explicit Engine(Resource* resource = nullptr);
+    explicit Engine();
+
+    explicit Engine(Resource* resource);
 
     void
-    Shutdown();
+    Shutdown();  // like ~Engine(), but will warn whether the resources are still reference outside
 
     tl::expected<std::shared_ptr<Index>, Error>
     CreateIndex(const std::string& name, const std::string& parameters);
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -53,14 +53,14 @@ next_multiple_of_power_of_two(uint64_t x, uint64_t n) {
 HGraph::HGraph(const JsonType& index_param, const vsag::IndexCommonParam& common_param) noexcept
     : index_param_(index_param),
       common_param_(common_param),
-      label_lookup_(common_param.allocator_),
-      label_op_mutex_(MAX_LABEL_OPERATION_LOCKS, common_param_.allocator_),
-      neighbors_mutex_(0, common_param_.allocator_),
-      route_graphs_(common_param.allocator_),
-      labels_(common_param.allocator_) {
+      label_lookup_(common_param.allocator_.get()),
+      label_op_mutex_(MAX_LABEL_OPERATION_LOCKS, common_param.allocator_.get()),
+      neighbors_mutex_(0, common_param.allocator_.get()),
+      route_graphs_(common_param.allocator_.get()),
+      labels_(common_param.allocator_.get()) {
     this->dim_ = common_param.dim_;
     this->metric_ = common_param.metric_;
-    this->allocator_ = common_param.allocator_;
+    this->allocator_ = common_param.allocator_.get();
 }
 
 tl::expected<void, Error>
--- a/src/factory/factory.cpp
+++ b/src/factory/factory.cpp
@@ -13,25 +13,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "vsag/factory.h"
+
 #include <algorithm>
-#include <cctype>
 #include <cstdint>
 #include <exception>
 #include <fstream>
 #include <ios>
 #include <memory>
 #include <mutex>
-#include <stdexcept>
 #include <string>
 
-#include "index/diskann.h"
-#include "index/diskann_zparameters.h"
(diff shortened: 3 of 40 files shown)


include/vsag/engine.h
```diff
@@ -17,20 +17,26 @@
...
     void
-    Shutdown();
+    Shutdown();  // like ~Engine(), but will warn whether the resources are still reference outside
 
     tl::expected<std::shared_ptr<Index>, Error>
     CreateIndex(const std::string& name, const std::string& parameters);
 
+    //    tl::expected<DatasetPtr, Error>
+    //    CreateDataset();
+
```
inabao: delete unnecessary modifications

src/algorithm/hgraph.cpp
```diff
@@ -43,14 +43,14 @@ empty_binaryset() {
...
-      labels_(common_param.allocator_) {
+      label_lookup_(common_param.allocator_.get()),
+      label_op_mutex_(MAX_LABEL_OPERATION_LOCKS, common_param.allocator_.get()),
+      neighbors_mutex_(0, common_param.allocator_.get()),
+      route_graphs_(common_param.allocator_.get()),
+      labels_(common_param.allocator_.get()) {
     this->dim_ = common_param.dim_;
     this->metric_ = common_param.metric_;
-    this->allocator_ = common_param.allocator_;
+    this->allocator_ = common_param.allocator_.get();
```
jiaweizone, line 63: HGraph can hold the allocator shared_ptr ?

src/factory/factory.cpp
```diff
@@ -13,87 +13,30 @@
...
-    } catch (const std::invalid_argument& e) {
-        LOG_ERROR_AND_RETURNS(
-            ErrorType::INVALID_ARGUMENT, "failed to create index(invalid argument): ", e.what());
-    } catch (const std::exception& e) {
-        LOG_ERROR_AND_RETURNS(
-            ErrorType::UNSUPPORTED_INDEX, "failed to create index(unknown error): ", e.what());
-    }
+    Resource resource(allocator);
+    Engine e(&resource);
+    return e.CreateIndex(origin_name, parameters);
```
jiaweizone, line 39: Here only need to hold `allocator` object ?

src/io/memory_io.h
```diff
@@ -36,7 +36,7 @@ class MemoryIO : public BasicIO<MemoryIO> {
     }
 
     MemoryIO(const JsonType& io_param, const IndexCommonParam& common_param)
-        : allocator_(common_param.allocator_) {
+        : allocator_(common_param.allocator_.get()) {
```
jiaweizone, line 39: ditto

src/safe_allocator.h
```diff
@@ -17,18 +17,21 @@
...
 #include "vsag/allocator.h"
 
 namespace vsag {
 
 class SafeAllocator : public Allocator {
 public:
-    explicit SafeAllocator(Allocator* raw_allocator, bool owner = false)
-        : raw_allocator_(raw_allocator), owner_(owner) {
+    static std::shared_ptr<Allocator>
+    FactoryDefaultAllocator() {
```
jiaweizone, line 28: FactoryDefaultAllocator -> NewDefaultAllocator ?

src/algorithm/hgraph.cpp
LHT129: Hgraph Index hold the shared ptr

src/factory/factory.cpp
LHT129: maybe threadpool or other resource in future

src/io/memory_io.h
LHT129: the MemoryIO's lifetime is shorter than Index, so only raw pointer is enough

src/safe_allocator.h
LHT129: Both are possible, but the former is a verb-object structure and is more suitable as a function name.

src/algorithm/hgraph.cpp
jiaweizone: It is recommended use shared_ptr to avoid possible memory issues and keep consistent allocator object passing logic

LHT129: if lifetime is shorter than Index, only raw pointer is enough, too much shared_ptr's construction will influence the performance

wxyucs (commented):
@LHT129 we need to add an example for engine.

include/vsag/engine.h
LHT129: done

src/factory/factory.cpp
```diff
@@ -13,87 +13,30 @@
...
 
-#include "index/diskann.h"
-#include "index/diskann_zparameters.h"
-#include "index/hgraph_index.h"
-#include "index/hgraph_zparameters.h"
-#include "index/hnsw.h"
-#include "index/hnsw_zparameters.h"
-#include "index/index_common_param.h"
-#include "vsag/vsag.h"
+#include "ThreadPool.h"
```
inabao, line 27: why include it

inabao (approved):
LGTM

wxyucs (approved):
lgtm

ShawnShawnYou (approved):

src/algorithm/hgraph.cpp
jiaweizone: Ok, we'll discuss this later

jiaweizone (approved):
LGTM

PR #233, opened 2024-12-18
add pyvsag package tool

### LINKED ISSUE #110: pyvsag  cannot be used
**Describe the bug**
image

**To Reproduce**
```cpp
// paste codes here

```

**Environment**
- OS(e.g. `Ubuntu 22.04`): 
- vsag version(e.g. `v0.11.0` or `7f94a27`): 
- compiler version(e.g. `GCC 11.4.0`):
- interface(`cpp` or `python`):

**Expected behavior**

**Screenshots**

**Additional context**

wxyucs:
Close #110

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,9 @@
 CMAKE_GENERATOR ?= "Unix Makefiles"
 CMAKE_INSTALL_PREFIX ?= "/usr/local/"
 COMPILE_JOBS ?= 6
-VSAG_CMAKE_ARGS = -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DNUM_BUILDING_JOBS=${COMPILE_JOBS} -DENABLE_TESTS=1 -DENABLE_PYBINDS=1 -G ${CMAKE_GENERATOR} -S. -Bbuild
+DEBUG_BUILD_DIR ?= "./build/"
+RELEASE_BUILD_DIR ?= "./build-release/"
+VSAG_CMAKE_ARGS = -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DNUM_BUILDING_JOBS=${COMPILE_JOBS} -DENABLE_TESTS=1 -DENABLE_PYBINDS=1 -G ${CMAKE_GENERATOR} -S.
 UT_FILTER = ""
 ifdef CASE
   UT_FILTER = $(CASE)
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -105,8 +93,33 @@ test_cov: cov            ## Build and run unit tests with code coverage enabled.
...
+	cmake --build ${RELEASE_BUILD_DIR} --parallel ${COMPILE_JOBS}
 
 .PHONY: install
 install:                 ## Build and install the release version of vsag.
-	cmake --install build/
+	cmake --install ${RELEASE_BUILD_DIR}/
+
+
+PARAM1 := "-DNUM_BUILDING_JOBS=${COMPILE_JOBS} -DENABLE_PYBINDS=1 -S. -B${RELEASE_BUILD_DIR} -DCMAKE_BUILD_TYPE=Release"
+PARAM2 := "--build ${RELEASE_BUILD_DIR} --parallel ${COMPILE_JOBS}"
```
LHT129, line 120: why use these parameters, rather than place them in L125

python/example.c
LHT129, line 1: what is the meaning of this file

Makefile
wxyucs: These three parameters have different uses in the pyvsag build script. Write them separately and wrap them with quotas to make it clear.

python/example.c
wxyucs: @LHT129  setup.py only marks the wheel as compatible with a specific Python version when built with c/cpp source codes. So this file is just a dummy file to trigger the compatible mark.

 Without this file, the wheel's name will be like `pyvsag-0.0.10-py3-none-any.whl`, but what we expected is `pyvsag-0.0.10-cp310-cp310-linux_x86_64.whl`, which means that this wheel can only run with Python 3.10.

used in python/setup.py:L70

scripts/build_pyvsag_multiple_version.sh
jiacai2050, line 1: Can this script run in GitHub CI?

python/example.c
jiacai2050: IMO `pyvsag-0.0.10-py3-none-any.whl` is the preferred version, since we have no strict python version requirements?

wxyucs: @jiacai2050 actually, the pyvsag package cannot be compatible with multiple Python versions now because it is compiled with a specific of CPython and numpy  (like 3.8) which does not provide ABI compatibility.

scripts/build_pyvsag_multiple_version.sh
wxyucs: yes, after weeks of testing.

jiacai2050 (approved):
LGTM

LHT129 (approved):
LGTM

jiaweizone (approved):
LGTM

PR #248, opened 2024-12-23
support inserting and querying vectors with nan values

inabao:


tests/test_index.cpp
```diff
@@ -254,6 +254,79 @@ TestIndex::TestSerializeFile(const IndexPtr& index_from,
         }
     }
 }
+void
+TestIndex::TestSearchWithNan(const TestIndex::IndexPtr& index,
+                             const TestDatasetPtr& dataset,
+                             const std::string& search_param,
+                             float recall,
```
wxyucs: recall - > expected_recall

tests/fixtures/test_dataset.h
```diff
@@ -25,7 +25,11 @@ class TestDataset {
 public:
     using DatasetPtr = vsag::DatasetPtr;
 
-    TestDataset(uint64_t dim, uint64_t count, std::string metric_str = "l2");
+    static std::shared_ptr<TestDataset>
+    CreateTestDataset(uint64_t dim, uint64_t count, std::string metric_str = "l2");
```
wxyucs, line 29: std::string -> const std::string &

tests/fixtures/test_dataset.h
```diff
@@ -25,7 +25,11 @@ class TestDataset {
...
+
+    static std::shared_ptr<TestDataset>
+    CreateNanDataset(std::string metric_str);
```
wxyucs: std::string -> const std::string &

wxyucs (changes requested):

tests/test_index.cpp
inabao: done

tests/fixtures/test_dataset.h
inabao: done

tests/fixtures/test_dataset.h
inabao: done

wxyucs (approved):
lgtm

PR #249, opened 2024-12-24
update the example of custom memory allocator

wxyucs:


examples/cpp/custom_memory_allocator.cpp
```diff
@@ -15,52 +15,75 @@
...
+        return addr;
     }
 
     void
     Deallocate(void* p) override {
+        if (sizes_.find(p) == sizes_.end())
+            return;
+        vsag::Options::Instance().logger()->Debug("deallocate " + std::to_string(sizes_[p]) +
+                                                  " bytes.");
+        sizes_.erase(p);
```
jiaweizone, line 42: minor: reduce dup find cost
```C++
auto iter = sizes_.find(p);
if (iter == sizes_.end() {
    return;
}
sizes_.erase(iter);
```

jiaweizone (approved):
LGTM

examples/cpp/custom_memory_allocator.cpp
wxyucs: done

LHT129 (approved):
LGTM

PR #260, opened 2024-12-26
enable coverage test for github action

LHT129:


wxyucs (approved):
lgtm

.github/workflows/lcov.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+      - name: Install
+        run: |
+          python -m pip install --upgrade pip
+          pip install lcov_cobertura
```
jiaweizone, line 24: can install to VSAG docker image ?

scripts/check_cov.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+line_coverage=`lcov --summary coverage/coverage.info | grep "lines......" | awk '/lines/ { print $2 }' | cut -d '%' -f 1`
+line_coverage=$(printf "%.0f" $line_coverage)
+if [ "$line_coverage" -gt 84 ]; then
```
jiaweizone, line 5: line coverage can not set to 90% ?

LHT129: this is no used, codecov bot have method to set threshold, We will enable soon

.github/workflows/lcov.yml
LHT129: good idea, but only install for coverage now

inabao (approved):
LGTM

.github/workflows/lcov.yml
jiaweizone: OK

jiaweizone (approved):
LGTM

PR #301, opened 2025-01-06
enable examples in the compilation by default

wxyucs:


examples/cpp/custom_memory_allocator.cpp
```diff
@@ -64,7 +64,7 @@ main() {
     // vsag::Options::Instance().logger()->SetLevel(vsag::Logger::kDEBUG);
 
     ExampleAllocator allocator;
-    vsag::Resource resource(&allocator);
+    vsag::Resource resource(&allocator, nullptr);
```
wxyucs, line 67: fix compile issue after enable the examples

inabao (approved):
LGTM

PR #302, opened 2025-01-07
add logger in tests and fix some linter warnings

wxyucs:


inabao (approved):
LGTM

tests/fixtures/test_logger.h
```diff
@@ -81,6 +118,64 @@ class TestLogger : public vsag::Logger {
...
+
+public:
+    virtual int
+    overflow(int ch) override {
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (ch != EOF) {
+            *this->pptr() = (char)ch;
+            this->pbump(1);
+        }
+        this->flush();
```
ShawnShawnYou, line 144: why not flush first and then set *pptr() = ch ?

wxyucs: we have an extra space to place this character so we can flush them together

tests/fixtures/test_logger.h
```diff
@@ -81,6 +118,64 @@ class TestLogger : public vsag::Logger {
...
+    uint64_t size_;
+};
+
+extern TestLogger test_logger;
+extern std::basic_ostream<char> trace;
+extern std::basic_ostream<char> debug;
+extern std::basic_ostream<char> info;
+extern std::basic_ostream<char> warn;
+extern std::basic_ostream<char> error;
+extern std::basic_ostream<char> critical;
```
wxyucs, line 180: `logger::debug( ... )` provides better performance, but we prefer to use `logger::debug << ... ` in test codes to output variables simply.

tests/fixtures/test_logger.cpp
```diff
@@ -16,9 +16,25 @@
...
 
-}  // namespace fixtures
+TestLogger test_logger;
+
+LoggerStream trace_buff(&test_logger, vsag::Logger::kTRACE);
+LoggerStream debug_buff(&test_logger, vsag::Logger::kDEBUG);
+LoggerStream info_buff(&test_logger, vsag::Logger::kINFO);
+LoggerStream warn_buff(&test_logger, vsag::Logger::kWARN);
+LoggerStream error_buff(&test_logger, vsag::Logger::kERR);
+LoggerStream critical_buff(&test_logger, vsag::Logger::kCRITICAL);
```
jiaweizone, line 32: critical_buff not use ?

wxyucs: fixed

jiaweizone (approved):
LGTM

PR #311, opened 2025-01-10
upgrade the lcov version in github ci

wxyucs:
there is a coverage results collection issue in v2.2 of lcov, so we decided to upgrade the version lcov.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,10 +1,14 @@
 
 # Configuration for CodeCov
 codecov:
+  max_report_age: off
   require_ci_to_pass: no
   notify:
     require_ci_to_pass: no
     wait_for_ci: false
+fixes:
+  - "/__w/vsag/::/"
+
 coverage:
   precision: 2
   round: down
@@ -22,7 +26,7 @@ coverage:
         if_ci_failed: error
 
 comment:
-  layout: "reach, diff, flags, files"
+  layout: "header, diff, files, footer"
   behavior: default
   require_changes: false
   branches:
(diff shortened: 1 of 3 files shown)


scripts/collect_cpp_coverage.sh
```diff
@@ -13,17 +14,25 @@ else
...
      --parallel 8 \
-     --directory . \
+     --directory vsag \
      --capture \
+     --substitute "s#${PARENT_DIR}/##g" \
      --ignore-errors mismatch,mismatch \
      --ignore-errors count,count \
      --output-file ${COVERAGE_DIR}/coverage.info
-
-lcov --remove ${COVERAGE_DIR}/coverage.info '/usr/*' '*/build/*' '*/vsag/tests/*' --ignore-errors inconsistent,inconsistent --output-file ${COVERAGE_DIR}/coverage.info
```
inabao, line 25: Do we need to check the coverage under the tools directory?

.github/codecov.yml
```diff
@@ -1,10 +1,14 @@
 
 # Configuration for CodeCov
 codecov:
+  max_report_age: off
   require_ci_to_pass: no
   notify:
     require_ci_to_pass: no
     wait_for_ci: false
+fixes:
+  - "/__w/vsag/::/"
```
jiaweizone, line 10: Ok，Here change root source path ?

scripts/collect_cpp_coverage.sh
wxyucs: we just run unittest and functest in CI, so there is no coverage result about the codes in tools/

.github/codecov.yml
wxyucs: yes, use the relative path instead, the origin result uses the absolute path which makes the vsag/ directory deeper and meaningless in the pages of codecov.io

jiaweizone (approved):
LGTM

inabao (approved):
LGTM

PR #313, opened 2025-01-10
add new interface for dataset to get/set num_results

inabao:


src/dataset_impl.h
```diff
@@ -105,6 +107,22 @@ class DatasetImpl : public Dataset {
         return 0;
     }
 
+    DatasetPtr
+    NumResults(int64_t num_results) override {
+        // To maintain compatibility, we currently ensure that the values in NumResults and Dim are the same.
+        this->data_[NUM_RESULTS] = num_results;
+        this->data_[DIM] = num_results;
```
jiaweizone, line 114: Have any `dim` and `num_results` are used together ?

inabao: no, `dim` is input while `num_results` is output

PR #315, opened 2025-01-10
enable bugprone-* options in clang-tidy

wxyucs:


ShawnShawnYou (approved):

LHT129 (approved):
LGTM

src/algorithm/hgraph.cpp
```diff
@@ -839,8 +847,14 @@ HGraph::add_one_point(const float* data, int level, InnerIdType inner_id) {
...
 HGraph::resize(uint64_t new_size) {
     auto cur_size = this->neighbors_mutex_.size();
-    auto new_size_power_2 =
-        next_multiple_of_power_of_two(new_size, this->resize_increase_count_bit_);
+    uint64_t new_size_power_2 = 0;
+    try {
+        new_size_power_2 =
+            next_multiple_of_power_of_two(new_size, this->resize_increase_count_bit_);
+    } catch (std::runtime_error& err) {
+        logger::critical("failed to resize(illegal new size): ", err.what());
```
jiaweizone: need throw exception ?

inabao (approved):
LGTM

src/algorithm/hgraph.cpp
wxyucs: this function will be called by an class constructor, so throwing an exception is not a good idea

jiaweizone: Add a return value to indicate operation status？

wxyucs: @jiaweizone currently, the constructor of HGraph may throw an exception when allocate memory fails, which is unrecoverable, so I remove the exception catch statement. if the value of resize_increase_count_bit_ is incorrect, an exception will be raised directly in programming

jiaweizone: OK

jiaweizone (commented):
LGTM

jiaweizone (approved):

Issue #324, opened 2025-01-13
support sparse vector search


wxyucs:
general tasks:
- [x] support sparse dtype
- [x] support sparse quantizer

graph-based method:
- [x] index-level adapt (dtype, param...) in HGraph

SINDI method:
- [x] first version of SOTA performance

ShawnShawnYou:
done

PR #325, opened 2025-01-13
add sparse dtype in dataset

ShawnShawnYou:
* support sparse dataset
* #324

### DIFF
--- a/include/vsag/dataset.h
+++ b/include/vsag/dataset.h
@@ -25,17 +25,18 @@
 
 namespace vsag {
 
-class Dataset;
-using DatasetPtr = std::shared_ptr<Dataset>;
-
 struct SparseVector {
     uint32_t dim_;   // dim of vector
-    uint32_t* ids_;  // contains offsets[num] ids
-    float* vals_;    // contains offsets[num] vals
+    uint32_t* ids_;  // contains ids with size of dim_
+    float* vals_;    // contains vals with size of dim_
 
     SparseVector() : dim_{0}, ids_{nullptr}, vals_{nullptr} {
     }
 };
+
+class Dataset;
+using DatasetPtr = std::shared_ptr<Dataset>;
+using SparseVectorPtr = struct SparseVector*;
 
 /**
  * @class Dataset
@@ -189,14 +190,14 @@ public:
      * @return DatasetPtr A shared pointer to the dataset with updated sparse vectors.
      */
     virtual DatasetPtr
-    SparseVector(const struct SparseVector* sparse_vectors) = 0;
+    SparseVector(const SparseVectorPtr sparse_vectors) = 0;
 
     /**
      * @brief Retrieves the sparse struct of the dataset.
      *
      * @return const SparseVector to the array of sparse vectors.
      */
-    virtual const struct SparseVector*
+    virtual const SparseVectorPtr
     GetSparseVector() const = 0;
 
     /**
--- a/include/vsag/dataset.h
+++ b/include/vsag/dataset.h
@@ -28,15 +28,12 @@ namespace vsag {
 class Dataset;
 using DatasetPtr = std::shared_ptr<Dataset>;
 
-struct SparseVectors {
-    uint32_t num;  // num of vectors
-    uint32_t*
-        offsets;  // vectors offset, len(offsets) = num + 1, offsets[i + 1] - offsets[i] = dim(vector)
-                  // e.g. [0, 2, 5, 10] means len(vec1) = 2 - 0 == 2, len(vec2) = 5 - 2 == 3
-    uint32_t* ids;  // contains offsets[num] ids
-    float* vals;    // contains offsets[num] vals
+struct SparseVector {
+    uint32_t dim_;   // dim of vector
+    uint32_t* ids_;  // contains offsets[num] ids
+    float* vals_;    // contains offsets[num] vals
 
-    SparseVectors() : num(0), offsets{nullptr}, ids{nullptr}, vals{nullptr} {
+    SparseVector() : dim_{0}, ids_{nullptr}, vals_{nullptr} {
     }
 };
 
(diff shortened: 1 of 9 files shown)


include/vsag/dataset.h
```diff
@@ -28,6 +28,18 @@ namespace vsag {
 class Dataset;
 using DatasetPtr = std::shared_ptr<Dataset>;
 
+struct SparseVectors {
+    uint32_t num;  // num of vectors
```
ZhitaoShen: dimension of vectors?

include/vsag/dataset.h
```diff
@@ -28,6 +28,18 @@ namespace vsag {
...
+    uint32_t*
+        offsets;  // vectors offset, len(offsets) = num + 1, offsets[i + 1] - offsets[i] = dim(vector)
+                  // e.g. [0, 2, 5, 10] means len(vec1) = 2 - 0 == 2, len(vec2) = 5 - 2 == 3
+    uint32_t* ids;  // contains offsets[num] ids
```
ZhitaoShen: may we have the map from tokens to ids for storing the original tokens?

include/vsag/dataset.h
ShawnShawnYou: number of vectors, num == 3 means there are 3 vectors

include/vsag/dataset.h
ShawnShawnYou: Can it be stored outside? Since the map won't affect the search process and result.

include/vsag/dataset.h
ZhitaoShen: we don't need to store the number of dimensions of a sparse vector?

ShawnShawnYou: the dimensions is stored in offsets, e.g., offsets[i + 1] - offsets[i] = dim(vectors[i])

inabao (approved):
LGTM

include/vsag/dataset.h
```diff
@@ -25,6 +25,15 @@
...
 
+struct SparseVector {
+    uint32_t dim_;   // dim of vector
+    uint32_t* ids_;  // contains ids with size of dim_
+    float* vals_;    // contains vals with size of dim_
+
+    SparseVector() : dim_{0}, ids_{nullptr}, vals_{nullptr} {
+    }
+};
+
```
nedchu, line 38: Do we need to ensure `ids_` is ordered?
This will lead to better perf.

ShawnShawnYou: we don't require the `ids_` in dataset is ordered since we can sort them during the process of  `Build` or `Add Vector`. It can be viewed as a pre-process of the vector inside the index.

tests/fixtures/fixtures.cpp
```diff
@@ -51,6 +51,59 @@ get_common_used_dims(uint64_t count, int seed) {
...
+
+    for (int i = 0; i < count; i++) {
+        sparse_vectors[i].dim_ = distrib_dim(rng);
+        sparse_vectors[i].ids_ = new uint32_t[sparse_vectors[i].dim_];
+        sparse_vectors[i].vals_ = new float[sparse_vectors[i].dim_];
+        for (int d = 0; d < sparse_vectors[i].dim_; d++) {
+            sparse_vectors[i].ids_[d] = distrib_id(rng);
+            sparse_vectors[i].vals_[d] = distrib_real(rng);
+        }
+    }
```
nedchu: make ids_ ordered?

include/vsag/dataset.h
nedchu: Better add a comment on `ids_` saying that `ids_` must be ordered inside the index.

tests/fixtures/fixtures.h
```diff
@@ -15,20 +15,44 @@
...
 namespace fixtures {
 
 std::vector<int>
 get_common_used_dims(uint64_t count = -1, int seed = 369);
 
+template <typename T>
+T*
+CopyVector(const std::vector<T>& vec) {
+    auto result = new T[vec.size()];
+    memcpy(result, vec.data(), vec.size() * sizeof(T));
```
jiaweizone, line 39: Here is shallow copy ?

include/vsag/dataset.h
```diff
@@ -25,6 +25,15 @@
 
 namespace vsag {
 
+struct SparseVector {
+    uint32_t dim_;   // dim of vector
```
jiaweizone: `dim_` is array size or sparse vector dim size ?

ShawnShawnYou: Both are the same. e.g., dim_ == 2 means there are two ids and values are non-zero, which is also the array size.

tests/fixtures/fixtures.h
ShawnShawnYou: right, this func is only used in test. It's used to create a `T*` from `std::vector<vsag::SparseVector> GenerateSparseVectors` to test the `Set Vector` of dataset.  The resource(e.g., `ids_`, `vals_`) is still be hold in the original `std::vector<vsag::SparseVector>`

tests/fixtures/fixtures.cpp
ShawnShawnYou: We don't require the user input data to be ordered since we will sort the `ids_` in the quantizer, which can be implemented in another PR.

include/vsag/dataset.h
```diff
@@ -25,6 +25,15 @@
...
+    uint32_t* ids_;  // contains ids with size of dim_
+    float* vals_;    // contains vals with size of dim_
+
```
LHT129, line 34: why use column storage rather than raw storage which is benefit for sort?

LHT129 (changes requested):

wxyucs (approved):
lgtm

include/vsag/dataset.h
jiaweizone: change to `len_`

ShawnShawnYou: fixed

include/vsag/dataset.h
ShawnShawnYou: done

jiaweizone (approved):
LGTM

LHT129 (approved):
LGTM

PR #345, opened 2025-01-19
support build sparse graph in ODescent

inabao:


LHT129 (approved):
LGTM

src/impl/odescent_graph_builder.cpp
```diff
@@ -88,34 +96,23 @@ ODescent::SaveGraph(std::stringstream& out) {
     // Note: at this point, either _nd == _max_points or any frozen points have
     // been temporarily moved to _nd, so _nd + _num_frozen_points is the valid
     // location limit.
-    auto final_graph = GetGraph();
     for (uint32_t i = 0; i < data_num_; i++) {
```
wxyucs: avoid implicit type conversion

ShawnShawnYou (approved):

src/impl/odescent_graph_builder.cpp
inabao: done

wxyucs (approved):
lgtm

PR #372, opened 2025-02-05
enable modernize-* checks in clang-tidy

wxyucs:


src/utils.h
```diff
@@ -31,7 +31,7 @@
 namespace vsag {
 
 struct SlowTaskTimer {
-    explicit SlowTaskTimer(const std::string& name, int64_t log_threshold_ms = 0);
+    explicit SlowTaskTimer(std::string name, int64_t log_threshold_ms = 0);
```
jiaweizone, line 34: minor: why remove const& ?

jiaweizone (approved):
LGTM

src/utils.h
wxyucs: @jiaweizone it's changed to std::move() in utils.cpp:L26

inabao (approved):

Issue #369, opened 2025-01-27
hgraph index When the query result is 0, a bad_alloc error is reported


Carrot-77:
**Describe the bug**
hgraph index When the query result is 0, a bad_alloc error is reported
Since the query result data is 0, the result memory of size 0 is applied, resulting in the application failure and error bad_alloc

**To Reproduce**
Codes to reproduce the behavior:
```cpp
nlohmann::json hnswsq_parameters{{"base_quantization_type", "sq8"}, 
                                            {"max_degree", max_degree}, 
                                            {"ef_construction", ef_construction},
                                            {"build_thread_count", 1}};
nlohmann::json index_parameters{{"dtype", dtype}, {"metric_type", metric}, {"dim", dim}, {"index_param", hnswsq_parameters}}; 
auto index = vsag::Factory::CreateIndex("hgraph", index_parameters.dump(), vsag_allocator);
HnswIndexHandler* hnsw = static_cast<HnswIndexHandler*>(index_handler); 
auto dataset = vsag::Dataset::Make();
dataset->Dim(dim)
    ->NumElements(11000)
    ->Ids(ids)
    ->Float32Vectors(vector_list)
    ->Owner(false);
ret = hnsw->build_index(dataset);
query->NumElements(1)->Dim(dim)->Float32Vectors(query_vector)->Owner(false);
// The data filtered by the filter is 100%
ret = hnsw->knn_search(query, topk, search_parameters.dump(), dist, ids, result_size, filter);
```

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS: e.g. `Ubuntu 22.04`
- vsag version: e.g. `v0.11.0` or `7f94a27`
- compiler version: e.g. `GCC 11.4.0`
- interface: `cpp` or `python`

**Expected behavior**
Verify the returned result like hnsw

![Image](https://github.com/user-attachments/assets/6cf918dd-2159-4129-ae28-f2d821038201)

PR #379, opened 2025-02-10
fix unexpected behavior of HGraph::KnnSearch when filter ratio 100%

wxyucs:
fixes: #369

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -165,6 +165,13 @@ HGraph::KnnSearch(const DatasetPtr& query,
             search_result.pop();
         }
 
+        // return an empty dataset directly if searcher returns nothing
+        if (search_result.empty()) {
+            auto result = Dataset::Make();
+            result->Dim(0)->NumElements(1);
+            return result;
+        }
+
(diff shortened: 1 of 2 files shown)


LHT129 (approved):
LGTM

src/algorithm/hgraph.cpp
```diff
@@ -165,6 +165,13 @@ HGraph::KnnSearch(const DatasetPtr& query,
             search_result.pop();
         }
 
+        // return an empty dataset directly if searcher returns nothing
+        if (search_result.empty()) {
+            auto result = Dataset::Make();
+            result->Dim(0)->NumElements(1);
```
jiaweizone, line 171: empty-result why not set both dim and num_elements to `0` ？

wxyucs: @jiaweizone  the result is like the `numpy.array`, with the shape `(number_of_queries, k)`. NumElements is an alias of the first dimension, Dim is an alias of the second dimension. so, in this case, the first dimension of result dataset should be `1`, and the second dimension of result dataset should be `0`.

jiaweizone (approved):
LGTM

PR #407, opened 2025-02-17
fix compiler issues when using clang make libcxx

### LINKED ISSUE #405: compiler issues when make libcxx using clang in 0.13
**Describe the bug**
compile error when using clang 
1. can't find std::__countr_zero 
2. fmt::format meets std::string

**To Reproduce**
```
make libcxx CMAKE_INSTALL_PREFIX=`pwd`/install
```

**Environment**
- vsag version: v0.13.1-1-ga9a0f5d9-dirty
- compiler version: c++ (GCC) 9.2.1 20200522 

**Expected behavior**
1. implement `std::__countr_zero`
2. replace `std::string param_str` with `constexpr static const char* param_str`

**Screenshots**
1. 
Image
2. 
Image

ShawnShawnYou:
fix #405

std::countr_zero available since C++20: https://en.cppreference.com/w/cpp/numeric/countr_zero

### DIFF
--- a/src/io/memory_block_io.h
+++ b/src/io/memory_block_io.h
@@ -80,9 +80,22 @@ public:
     }
 
 private:
+    inline int
+    countr_zero(uint64_t x) {
+        if (x == 0)
+            return 64;
+
+        int count = 0;
+        while ((x & 1) == 0) {
+            x >>= 1;
+            ++count;
+        }
+        return count;
+    }
+
     inline void
     update_by_block_size() {
-        block_bit_ = std::__countr_zero(block_size_);
+        block_bit_ = countr_zero(block_size_);
         in_block_mask_ = block_size_ - 1;
     }
 
--- a/src/io/buffer_io_test.cpp
+++ b/src/io/buffer_io_test.cpp
@@ -35,7 +35,7 @@ TEST_CASE("buffer io param", "[ut][buffer_io]") {
     fixtures::TempDir dir("buffer_io");
     auto path = dir.GenerateRandomFile();
     auto allocator = SafeAllocator::FactoryDefaultAllocator();
-    std::string param_str = R"(
+    constexpr static const char* param_str = R"(
     {{
         "type": "buffer_io",
         "file_path" : "{}"


inabao (commented):
add a issue

inabao (approved):
LGTM

wxyucs (approved):
lgtm

src/io/memory_block_io.h
```diff
@@ -80,9 +80,22 @@ class MemoryBlockIO : public BasicIO<MemoryBlockIO> {
...
+            x >>= 1;
+            ++count;
+        }
+        return count;
+    }
+
     inline void
     update_by_block_size() {
-        block_bit_ = std::__countr_zero(block_size_);
+        block_bit_ = countr_zero(block_size_);
```
jiaweizone, line 98: Here can change to use std::__popcount ?

ShawnShawnYou: __popcount is also be available in C++20? Besides, __countr_zero is used to count the 0 from rightmost bit until the first 1, while __popcount is used to count the number of all 1.

LHT129 (approved):
LGTM

src/io/memory_block_io.h
jiaweizone: OK，`update_by_block_size` currently is O(n) cost implement, if in hot path your can change to use compiler optimize implement, like:

```C++
#pragma once
#include <type_traits>

namespace bitops {
    template <typename T>
    inline int countr_zero(T x) noexcept {
        static_assert(std::is_unsigned_v<T>, "T must be unsigned");

        if (x == 0) return sizeof(T) * 8;

        #if defined(__GNUC__)
            if constexpr (sizeof(T) <= sizeof(unsigned)) {
                return __builtin_ctz(x);
            } else if constexpr (sizeof(T) <= sizeof(unsigned long)) {
                return __builtin_ctzl(x);
            } else {
                return __builtin_ctzll(x);
            }
        #elif defined(_MSC_VER)
            unsigned long index;
            _BitScanForward(&index, x);
            return index;
        #else
            // fallback to generic O(n) implement
            int count = 0;
            while ((x & 1) == 0) {
                x >>= 1;
                ++count;
            }
            return count;
        #endif
    }
}
```

jiaweizone (approved):
LGTM

PR #436, opened 2025-02-24
Introduce fp16 simd operators

LHT129:
issue: #437

### DIFF
--- a/src/simd/fp16_simd_test.cpp
+++ b/src/simd/fp16_simd_test.cpp
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "bf16_simd.h"
+#include "fp16_simd.h"
 
 #include <catch2/benchmark/catch_benchmark.hpp>
 #include <catch2/catch_all.hpp>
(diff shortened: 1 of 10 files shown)


inabao (approved):
LGTM

src/simd/fp16_simd_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+    std::vector<uint8_t> result(count * 2);
+    auto* fp16 = reinterpret_cast<uint16_t*>(result.data());
+    for (int64_t i = 0; i < count; ++i) {
+        fp16[i] = generic::FloatToFP16(data[i]);
+    }
+    return result;
+}
+
+TEST_CASE("Encode & Decode FP16", "[ut][simd]") {
+    auto vec_fp32 = fixtures::generate_vectors(10, 100);
```
ShawnShawnYou, line 37: use more test cases, since fixtures::generate_vectors only generate small range or float

LHT129: it's not a good idea to touch the limit of the fp16, because we will never guarantee the precise of the special number

ShawnShawnYou (approved):

PR #443, opened 2025-02-26
add different simd operator test for github action

LHT129:


.github/workflows/asan_with_simd_option.yml
```diff
@@ -0,0 +1,99 @@
...
+    runs-on: ubuntu-22.04
+    container:
+      image: vsaglib/vsag:ci-x86
+    strategy:
+      matrix:
+        simd_option: [
+#          "avx2;-DDISABLE_AVX512_FORCE=1",
+          "avx;-DDISABLE_AVX512_FORCE=1 -DDISABLE_AVX2_FORCE=1",
+          "sse;-DDISABLE_AVX512_FORCE=1 -DDISABLE_AVX2_FORCE=1 -DDISABLE_AVX_FORCE=1",
+          "generic;-DDISABLE_AVX512_FORCE=1 -DDISABLE_AVX2_FORCE=1 -DDISABLE_AVX_FORCE=1 -DDISABLE_SSE_FORCE=1",
```
wxyucs: we can disable the SIMD feature by the lowest flag, no need to list and disable all the flags, for example, DISABLE_AVX_FORCE=1 will disable `avx512`, `avx2`, and `avx` at the same time because high-level SIMD distance calculation depends on the lower-level SIMD distance calculation

LHT129: done

inabao (approved):
LGTM

wxyucs (approved):
lgtm

Issue #166, opened 2024-11-27
introduce a new index pyramid (for path-like prefilter search)


inabao:
- structure design
- use pattern
- performance ref

**Background**

- Problem: `pyramid` aims to solve the vector retrieval problem with paths.
- Example: Each vector is associated with paths such as "a/b/c", "a/b/c/d", "a/b", "a/b/c/e". When given a query path "a/b/c", it retrieves the nearest vector under "a/b/c/*" to the target vector.

**Interface Definition**
Users need to pass the vector paths through the following interface. Other behaviors of `pyramid` are defined similarly to other indices.
```
// Pass the corresponding path to the index
virtual DatasetPtr Paths(const std::string* paths) = 0;
```

**Parameter Design**
Build Parameters
```
{
    "pyramid": {
        "sub_index_type": "index_name", // Specify a sub-index
        "index_param": {
            ………… // Parameters corresponding to the sub-index
        }
    }
}
```
Search Parameters
```
{
    "pyramid": {
        ………… // Search parameters corresponding to the sub-index
    }
}
```

**Implementation Principle**

- Based on the path of the vector, we can build an index tree, where each node in the index tree contains the corresponding data volume.
- For each node in the index tree, we construct the corresponding index.
- The index can share the underlying data storage, and the redundancy of the index equals the average number of levels.

**Usage Demo**
Create Index
```
auto pyramid_build_parameters = R"(
{
    "dtype": "float32",
    "metric_type": "l2",
    "dim": 128,
    "index_param": {
        "sub_index_type": "hnsw",
        "index_param": {
            "max_degree": 16,
            "ef_construction": 100
        }
    }
}
)";
auto index = vsag::Factory::CreateIndex("pyramid", pyramid_build_parameters).value();
auto base = vsag::Dataset::Make();
base->NumElements(num_vectors)->Dim(dim)->Ids(ids)->Float32Vectors(vectors)->Paths(paths)->Owner(false);
index->Build(base);
```

Prepare a Query Vector
```
// Memory will be released by querying the dataset since owner is set to true when creating the query.
auto query_vector = new float[dim];
auto query_path = new std::string[1];
for (int64_t i = 0; i < dim; ++i) {
    query_vector[i] = distrib_real(rng);
}
query_path[0] = ......; // generate a random path for query vector.
```
Search on the Index
```
auto pyramid_search_parameters = R"(
{
    "pyramid": {
        "ef_search": 100
    }
}
)";
int64_t topk = 10;
[...]

PR #465, opened 2025-03-03
support build index for different levels

inabao:
#166

### DIFF
--- a/src/index/pyramid.cpp
+++ b/src/index/pyramid.cpp
@@ -58,13 +58,14 @@ IndexNode::IndexNode(IndexCommonParam* common_param, GraphInterfaceParamPtr grap
       children_(common_param->allocator_.get()),
       common_param_(common_param),
       graph_param_(std::move(graph_param)) {
-    graph_ = GraphInterface::MakeInstance(graph_param_, *common_param_, true);
 }
 
 void
 IndexNode::BuildGraph(ODescent& odescent) {
     std::lock_guard<std::mutex> lock(mutex_);
-    if (not ids_.empty()) {
+    // Build an index when the level corresponding to the current node requires indexing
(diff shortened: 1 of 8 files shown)


src/index/pyramid.cpp
```diff
@@ -131,6 +142,28 @@ IndexNode::Serialize(StreamWriter& writer) const {
         // calculate size of `content`
         item.second->Serialize(writer);
     }
+    StreamWriter::WriteObj(writer, has_index_);
```
jiaweizone: `has_index_` need write two times ?

src/index/pyramid.cpp
```diff
@@ -131,6 +142,28 @@ IndexNode::Serialize(StreamWriter& writer) const {
...
+}
+void
+IndexNode::InitGraph() {
+    graph_ = GraphInterface::MakeInstance(graph_param_, *common_param_, true);
```
jiaweizone: add comments for literal constant arguments, like this:
```C++
graph_ = GraphInterface::MakeInstance(graph_param_, *common_param_, /*is_sparse=*/true);
```

jiaweizone: Q: for pyramid index need fill the argument `is_sparse` to `true` ?

LHT129 (approved):
LGTM

src/index/pyramid.cpp
inabao: The subsequent serialization logic has been removed.

src/index/pyramid.cpp
inabao: done

inabao: Yes, because the nodes in the subgraphs of the pyramid are not contiguous.

jiaweizone (approved):
LGTM

PR #477, opened 2025-03-05
update performance pipeline

wxyucs:


.github/workflows/performance.yml
```diff
@@ -28,9 +28,7 @@ jobs:
         run: make release
       - name: Run Perf - Recall 90%, 95%, 99%
         run:  |
-          ./build-release/tools/eval/eval_performance .github/perf-90.yml
-          ./build-release/tools/eval/eval_performance .github/perf-95.yml
-          ./build-release/tools/eval/eval_performance .github/perf-99.yml
```
jiaweizone, line 33: disable p90/95/99 perf test ?

wxyucs: The cpu provided by GitHub actions is different each time, resulting in performance changes. 

So we move the performance regression onto private resources and run a minimal test on GitHub actions to test the performance regression procedure and scripts

jiaweizone (approved):
LGTM

LHT129 (approved):

Issue #273, opened 2024-12-27
add a performance evaluation tool


wxyucs:
- [x] same behavior, cpp eval tool
- [x] single thread
- [x] multiple thread
- [x] output report format
- [ ] ~~random dataset generation (v0.14)~~
- [x] support yaml config (v0.14)

#234

wxyucs:
done

PR #489, opened 2025-03-07
support multi-threading in eval_performance

wxyucs:
related to: #273

### DIFF
--- a/.github/perf-90.yml
+++ b/.github/perf-90.yml
@@ -1,11 +1,13 @@
 global:
-  exporters:
-    e1: # give a name you like, will not be used
-      format: "text"
-      to: "stdout"
-    e2:
-      format: "json"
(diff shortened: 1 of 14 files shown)


inabao (approved):
LGTM

tools/eval/case/build_eval_case.cpp
```diff
@@ -15,12 +15,15 @@
 
 #include "./build_eval_case.h"
 
+#include <omp.h>
```
jiaweizone: ohh! add back OMP ?

wxyucs: @jiaweizone  actually, OMP is just used in search case, so I removed this line

jiaweizone (approved):
LGTM

PR #505, opened 2025-03-17
Support IVF Index

### LINKED ISSUE #276: support IVF index
IVF index is a partition type index, which consists of a set of inverted buckets. In the retrieval stage, an inverted bucket with a certain amount of data is selected, and then scanned in the bucket to obtain several candidate points that finally meet the nearest neighbors.

Compared with HNSW type algorithms, IVF often requires a certain amount of data for training bucketing. The usual bucketing method is to use K-means clustering to generate K centroids. For the bucket selection strategy during query, we support indexing these K centroids, such as graph indexing for routing.
The vectors in the bucket support multiple encoding methods. Due to the continuous arrangement of data, the access overhead is relatively low, but the total computational effort is higher than that of graph algorithms with similar configurations.

Below we will introduce the basic design framework of IVF index

1. Construct `BucketDatacell` to manage the data in the bucket (excluding the centroid)

2. Construct a data structure called `Router` to manage the centroid and the corresponding routing method. A simple implementation of Router is composed of centroids. Its classic construction method is k-means. Of course, it also supports importing from the outside. It can contain an Index entity

3. IVF also supports a reordering mechanism

Tasks:
- [x] implement bucket datacell
- [x] implement partition strategy
- [ ] support search

LHT129:
close: #276 
- add ivf index example

### DIFF
--- a/examples/cpp/106_index_ivf.cpp
+++ b/examples/cpp/106_index_ivf.cpp
@@ -1,0 +1,91 @@
+
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <vsag/vsag.h>
+
+#include <iostream>
+
+int
+main(int argc, char** argv) {
+    vsag::init();
+
+    /******************* Prepare Base Dataset *****************/
+    int64_t num_vectors = 10000;
+    int64_t dim = 128;
+    std::vector<int64_t> ids(num_vectors);
+    std::vector<float> datas(num_vectors * dim);
+    std::mt19937 rng(47);
+    std::uniform_real_distribution<float> distrib_real;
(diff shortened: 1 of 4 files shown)


examples/cpp/106_index_ivf.cpp
```diff
@@ -0,0 +1,91 @@
...
+        "dim": 128,
+        "index_param": {
+            "buckets_count": 50,
+            "base_quantization_type": "fp32"
+        }
+    }
+    )";
+    auto index = vsag::Factory::CreateIndex("ivf", ivf_build_params).value();
+
+    /******************* Build BruteForce Index *****************/
```
wxyucs: change bruteforce to ivf

examples/cpp/106_index_ivf.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    /******************* Prepare Query Dataset *****************/
+    std::vector<float> query_vector(dim);
+    for (int64_t i = 0; i < dim; ++i) {
+        query_vector[i] = distrib_real(rng);
+    }
+    auto query = vsag::Dataset::Make();
+    query->NumElements(1)->Dim(dim)->Float32Vectors(query_vector.data())->Owner(false);
+
+    /******************* KnnSearch For BruteForce Index *****************/
```
wxyucs: change bruteforce to ivf

LHT129: done

examples/cpp/106_index_ivf.cpp
LHT129: ditto

wxyucs (approved):
lgtm

inabao (approved):
LGTM

PR #512, opened 2025-03-20
Supports getting the maximum and minimum label interface from the label

Carrot-77:


src/algorithm/hgraph.cpp
```diff
@@ -476,6 +476,18 @@ HGraph::CalDistanceById(const float* query, const int64_t* ids, int64_t count) c
     return result;
 }
 
+tl::expected<void, Error>
```
inabao: return `void`

Carrot-77: Modified

inabao (approved):
This interface can be modified to serve as the default implementation of inner_index in the future.

include/vsag/index.h
```diff
@@ -311,6 +311,17 @@ class Index {
...
     };
 
+    /**
+     * @brief Calculate the maximum and minimum labels.
+     *
+     * @param min_id The minimum id returned
+     * @param max_id The maximum id returned
+     */
+    virtual tl::expected<void, Error>
+    GetMinAndMaxId(int64_t& min_id, int64_t& max_id) const {
```
wxyucs: return the min_id and max_id via return value. Use the `tl::expected` to process the result from the interface, which will remind the user to check if there is an error. e.g., it's an empty index, we cannot return a min/max id.

And modern cpp provides the structured binding declaration so we can unpack the result more easily.

standard usage:
```
auto get_min_and_max = index->GetMinAndMaxId();
if (not get_min_and_max.has_value()) {
  ... // handling the error
}
auto [min_id, max_id] = get_min_and_max.value();
...
```

tests/test_index.cpp
```diff
@@ -515,6 +515,26 @@ TestIndex::TestBatchCalcDistanceById(const IndexPtr& index,
     }
 }
 
+void
+TestIndex::TestGetMinAndMaxId(const IndexPtr& index, const TestDatasetPtr& dataset) {
```
wxyucs, line 519: test the interface when the index is empty

wxyucs (changes requested):

tests/test_index.cpp
Carrot-77: Modified

include/vsag/index.h
Carrot-77: Modified

wxyucs (approved):
lgtm

PR #526, opened 2025-03-25
Hnsw iter filter search

### LINKED ISSUE #544: support iterative search on index

Carrot-77:
A discard heap is introduced. The function of the discard heap is to save the nodes discarded in the previous round of queries and sort them by distance. If the results in the first round of queries are insufficient, another round of queries will be performed. In the next round, the nodes in the discard heap are used as candidate sets, and the neighbors that have not been visited are visited as new output results, and the discard heap is updated. Each subsequent round of queries will query ef_search nodes. The query ends when limitK is satisfied or discard is empty.

### DIFF
--- a/src/algorithm/hnswlib/hnswalg.cpp
+++ b/src/algorithm/hnswlib/hnswalg.cpp
@@ -24,6 +24,7 @@
 
 namespace hnswlib {
 
+const static InnerIdType UNUSED_ENTRY_POINT_NODE = 0;
 HierarchicalNSW::HierarchicalNSW(SpaceInterface* s,
                                  size_t max_elements,
                                  vsag::Allocator* allocator,
@@ -419,7 +420,8 @@ HierarchicalNSW::searchBaseLayerST(InnerIdType ep_id,
                                    const void* data_point,
                                    size_t ef,
                                    const vsag::FilterPtr is_id_allowed,
-                                   const float skip_ratio) const {
+                                   const float skip_ratio,
+                                   vsag::IteratorFilterContext* iter_ctx) const {
     vsag::LinearCongruentialGenerator generator;
     VisitedListPtr vl = visited_list_pool_->getFreeVisitedList();
     vl_type* visited_array = vl->mass;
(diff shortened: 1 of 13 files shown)


tests/test_index.cpp
```diff
@@ -336,6 +336,90 @@ TestIndex::TestKnnSearch(const IndexPtr& index,
     REQUIRE(cur_recall > expected_recall * query_count * RECALL_THRESHOLD);
 }
 
+void
+TestIndex::TestKnnSearchIter(const IndexPtr& index,
+                             const TestDatasetPtr& dataset,
+                             const std::string& search_param,
+                             float expected_recall,
+                             bool expected_success) {
+    if (not index->CheckFeature(vsag::SUPPORT_RANGE_SEARCH)) {
```
inabao: why is range search?

tests/test_index.cpp
```diff
@@ -336,6 +336,90 @@ TestIndex::TestKnnSearch(const IndexPtr& index,
...
+    auto dim = queries->GetDim();
+    auto gts = dataset->ground_truth_;
+    auto gt_topK = dataset->top_k;
+    float cur_recall = 0.0f;
+    auto topk = gt_topK;
+    class MyFilter : public vsag::Filter {
+    public:
+        bool
+        CheckValid(int64_t id) const override {
+            return true;
```
inabao: why is always `true`

src/algorithm/hnswlib/hnswalg.cpp
```diff
@@ -503,11 +521,20 @@ HierarchicalNSW::searchBaseLayerST(InnerIdType ep_id,
 
                     if ((!has_deletions || !isMarkedDeleted(candidate_id)) &&
                         ((!is_id_allowed) ||
-                         is_id_allowed->CheckValid(getExternalLabel(candidate_id))))
+                         is_id_allowed->CheckValid(getExternalLabel(candidate_id)))) {
+                        if (iter_ctx != nullptr && !(*iter_ctx)->CheckPoint(candidate_id)) {
```
inabao: why is a check needed here?

src/algorithm/hnswlib/hnswalg.cpp
```diff
@@ -503,11 +521,20 @@ HierarchicalNSW::searchBaseLayerST(InnerIdType ep_id,
...
+                            continue;
+                        }
                         top_candidates.emplace(dist, candidate_id);
+                    }
 
-                    if (top_candidates.size() > ef)
+                    if (top_candidates.size() > ef) {
+                        auto cur_node_pair = top_candidates.top();
+                        if (iter_ctx != nullptr && (*iter_ctx)->CheckPoint(cur_node_pair.second)) {
```
inabao: ditto

src/algorithm/hnswlib/hnswalg.cpp
Carrot-77: The result returned in the previous round will be marked on ctx. Before getting the result, check whether this node is the result of the previous round. If it is the result of the previous round continue

src/algorithm/hnswlib/hnswalg.cpp
Carrot-77: The discarded node top_candidates.top() also needs to be verified to see if it is the return result of the previous query before writing it to discard

tests/test_index.cpp
Carrot-77: Modified

tests/test_index.cpp
Carrot-77: sorry，Modified

inabao (approved):
LGTM

wxyucs (approved):
lgtm

Issue #528, opened 2025-03-26
Suppor graph compression in graph-based vector index


nedchu:
**Is your feature request related to a problem? Please describe.**

Graph indexes occupy a lot of memory after using vector quantization (sometimes graph uses 70% of memory).
We want to reduce the memory cost of graphs with compression.
The first step is to compress graphs in disk-based vector indexes, where graphs are static and memory cost is important.

--- 

**Describe the solution you'd like**

Using Elias-Fano encoding, we can **reduce 40% of graph memory usage** on SIFT-1M and GIST-1M.
The encoding/decoding speed is 50-100 million edges per second. 

--- 

**Describe alternatives you've considered**

Any other solution that can reduce graph memory usage.

---

**Additional context**

Milestone

- [x] Elias-Fano Encoder
- [x] Graph Data Cell/Quantizer using Elias-Fano Encoder
- [x] HGraph support Compressed Graph
- [x] HGraph Print graph memory usage in eval_performance
- [x] Add performance benchmark report for graph compression

nedchu:
Benchmark results:
- Graph Memory​​: -71%
- Recall: unchanged 
- QPS: -2.14%
- Latency: +2.44%
- TPS: -0.55%

PR #536, opened 2025-03-28
add elias-fano encoder

nedchu:
Add Elias-Fano Encoder

- it can encode an ordered seq (adjacency list) and decode the whole seq
- the max number of elements is 255, as the size of our adjacency list is mostly no more than 100

closes: #548

### DIFF
--- a/src/impl/elias_fano_encoder.cpp
+++ b/src/impl/elias_fano_encoder.cpp
@@ -1,3 +1,4 @@
+
 // Copyright 2024-present the vsag project
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/impl/elias_fano_encoder.cpp
+++ b/src/impl/elias_fano_encoder.cpp
@@ -1,0 +1,144 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "elias_fano_encoder.h"
+
+#include <algorithm>
+#include <cmath>
+#include <iostream>
+
+namespace vsag {
+
+size_t
+EliasFanoEncoder::ctzll(uint64_t x) const {
+#ifdef __GNUC__
+    return __builtin_ctzll(x);
+#else
+    if (x == 0)
+        return 64;
+    int count = 0;
+    while ((x & 1) == 0) {
+        x >>= 1;
+        count++;
+    }
+    return count;
+#endif
+}
+
+void
+EliasFanoEncoder::set_low_bits(size_t index, InnerIdType value) {
+    if (low_bits_width_ == 0)
+        return;
+
+    size_t bit_pos = index * low_bits_width_;
+    size_t word_pos = bit_pos >> 6;
+    size_t shift = bit_pos & 63;
+    uint64_t mask = ((1ULL << low_bits_width_) - 1) << shift;
+    low_bits_[word_pos] = (low_bits_[word_pos] & ~mask) | ((uint64_t)value << shift);
+
+    // Handle word boundary crossing
+    if (shift + low_bits_width_ > 64 && word_pos + 1 < low_bits_.size()) {
+        size_t remaining_bits = shift + low_bits_width_ - 64;
+        mask = (1ULL << remaining_bits) - 1;
+        low_bits_[word_pos + 1] =
+            (low_bits_[word_pos + 1] & ~mask) | (value >> (low_bits_width_ - remaining_bits));
+    }
+}
+
+InnerIdType
+EliasFanoEncoder::get_low_bits(size_t index) const {
+    if (low_bits_width_ == 0)
+        return 0;
+
+    size_t bit_pos = index * low_bits_width_;
+    size_t word_pos = bit_pos >> 6;
+    size_t shift = bit_pos & 63;
+    InnerIdType value = (low_bits_[word_pos] >> shift) & ((1ULL << low_bits_width_) - 1);
+
+    // Handle word boundary crossing
+    if (shift + low_bits_width_ > 64 && word_pos + 1 < low_bits_.size()) {
+        size_t remaining_bits = shift + low_bits_width_ - 64;
+        value |= (low_bits_[word_pos + 1] & ((1ULL << remaining_bits) - 1))
+                 << (low_bits_width_ - remaining_bits);
+    }
+    return value;
+}
+
+void
+EliasFanoEncoder::encode(const Vector<InnerIdType>& values, InnerIdType max_value) {
+    clear();
+    if (values.empty())
+        return;
+
+    // Check if number of elements exceeds uint8_t maximum
+    if (values.size() > UINT8_MAX) {
+        num_elements_ = UINT8_MAX;
+        throw std::runtime_error("Error: Elias-Fano encoder, number of elements exceeds 255.");
+    } else {
+        num_elements_ = static_cast<uint8_t>(values.size());
+    }
+
+    InnerIdType universe = max_value + 1;
+
+    // Calculate low bits width
+    low_bits_width_ =
+        static_cast<uint32_t>(std::floor(std::log2(static_cast<double>(universe) / num_elements_)));
+
+    // Allocate space for high bits
+    const size_t high_bits_count = (max_value >> low_bits_width_) + num_elements_ + 1;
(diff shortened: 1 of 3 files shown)


src/impl/elias_fano_encoder.h
```diff
@@ -0,0 +1,77 @@
...
+    Vector<uint64_t> low_bits_;   // Compressed storage for low bits
+    uint8_t num_elements_{0};     // Number of elements, max 255
+    uint8_t low_bits_width_{0};   // Width of low bits
+
+    // Cross-platform implementation of ctzll (count trailing zeros)
+    static size_t
+    ctzll(uint64_t x);
+
+    static inline void
+    set_high_bit(Vector<uint64_t>& vec, size_t pos) {
```
jiaweizone: private static inline method can move to `elias_fano_encoder.cpp` ?

src/impl/elias_fano_encoder.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "typing.h"
+
+namespace vsag {
+
+/// @brief Elias-Fano Encoder for *ordered* adjacency list
+/// @note Our adjacency list size is mostly no more than 255, so we use uint8_t to store the number of elements
+class EliasFanoEncoder {
+private:
+    Vector<uint64_t> high_bits_;  // Bitmap storage for high bits
```
jiaweizone: All field definitions are placed at the end of the class

src/impl/elias_fano_encoder.h
```diff
@@ -0,0 +1,77 @@
...
+    Clear() {
+        high_bits_.clear();
+        low_bits_.clear();
+        num_elements_ = 0;
+        low_bits_width_ = 0;
+    }
+
+    [[nodiscard]] size_t
+    SizeInBytes() const {
+        return high_bits_.size() * sizeof(uint64_t) + low_bits_.size() * sizeof(uint64_t) +
```
jiaweizone: here need add `sizeof(low_bits_)` and `sizeof(high_bits_)` ?

Or we could write it like this:
```C++
return sizeof(EliasFanoEncoder) +
    high_bits_.size() * sizeof(uint64_t) + low_bits_.size() * sizeof(uint64_t)
```

src/impl/elias_fano_encoder.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Clear();
+    if (values.empty()) {
+        return;
+    }
+
+    // Check if number of elements exceeds uint8_t maximum
+    if (values.size() <= UINT8_MAX) {
+        num_elements_ = static_cast<uint8_t>(values.size());
+    } else {
+        num_elements_ = UINT8_MAX;
```
jiaweizone: why need change `num_elements_` ?

src/impl/elias_fano_encoder_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <algorithm>
+#include <catch2/catch_test_macros.hpp>
+#include <random>
+
+#include "safe_allocator.h"
+
+namespace vsag {
+
+TEST_CASE("EliasFanoEncoder, original seq equal to decoded seq", "[ut][EliasFanoEncoder]") {
```
jiaweizone, line 26: add benchmark for Elias Fano Encoder

src/impl/elias_fano_encoder.h
nedchu: done. `ctzll` and `set_high_bit` are moved to `elias_fano_encoder.cpp`

src/impl/elias_fano_encoder.h
nedchu: done

src/impl/elias_fano_encoder.h
nedchu: done. `sizeof(EliasFanoEncoder)` is added to `SizeInBytes()`.

At first, `sizeof(EliasFanoEncoder)=72`. I optimize it to `24` by

1. replace `Vector<uint64_t>` with `uint64_t*`
2. combine low bits and high bits into a single `uint64_t *bits_`

src/impl/elias_fano_encoder.cpp
nedchu: no need to change. I have removed the line.

src/impl/elias_fano_encoder_test.cpp
nedchu: will add a benchmark report when we integrate Elias-Fano Encoder into a graph index.

I record it in #528

src/impl/elias_fano_encoder.h
```diff
@@ -0,0 +1,77 @@
...
+
+private:
+    void
+    set_low_bits(size_t index, InnerIdType value);
+
+    [[nodiscard]] InnerIdType
+    get_low_bits(size_t index) const;
+
+    Allocator* allocator_;
+    uint64_t* bits_{nullptr};    // Combined storage for low bits and high bits
```
LHT129, line 71: why use uint64_t ?

src/impl/elias_fano_encoder.cpp
```diff
@@ -0,0 +1,158 @@
...
+    while ((x & 1) == 0) {
+        x >>= 1;
+        count++;
+    }
+    return count;
+#endif
+}
+
+static inline void
+set_high_bit(uint64_t* vec, size_t pos, size_t low_bits_size) {
```
jiaweizone: vec -> bits_array?

src/impl/elias_fano_encoder.cpp
```diff
@@ -0,0 +1,158 @@
...
+    // Allocate space for high bits
+    const size_t high_bits_count = (max_value >> low_bits_width_) + num_elements_ + 1;
+    high_bits_size_ = (high_bits_count + 63) / 64;
+
+    // Allocate space for low bits
+    size_t total_low_bits = static_cast<size_t>(num_elements_) * low_bits_width_;
+    low_bits_size_ = std::max<size_t>(1, (total_low_bits + 63) / 64);
+
+    // Allocate combined space for both low and high bits
+    bits_ = static_cast<uint64_t*>(
```
jiaweizone, line 117: allocator_->Reallocate ?

src/impl/elias_fano_encoder.h
nedchu: uint64_t and uint32_t are both ok. An element's low bits are usually in a single block and sometimes go across at most 2 blocks.
If we use uint8_t, an element may go across > 2 blocks and is more complex to handle.

src/impl/elias_fano_encoder.cpp
nedchu: I call `Clear()` at the start of this function, it will Deallocate the space and set `bits_ = nullptr`.
So cannot call allocator_->Reallocate here.

src/impl/elias_fano_encoder.cpp
nedchu: done

jiaweizone (approved):
LGTM

LHT129 (approved):
LGTM

src/impl/elias_fano_encoder.cpp
```diff
@@ -0,0 +1,158 @@
...
+            count++;
+            // Delete lowest 1
+            word &= (word - 1);
+        }
+    }
+
+    return result;
+}
+
+}  // namespace vsag
```
ShawnShawnYou: add new line

src/impl/elias_fano_encoder.h
```diff
@@ -0,0 +1,77 @@
...
+    uint8_t num_elements_{0};    // Number of elements, max 255
+    uint8_t low_bits_width_{0};  // Width of low bits
+    uint8_t low_bits_size_{0};   // Size of low_bits_ array
+    uint8_t high_bits_size_{0};  // Size of high_bits_ array
+};
+
+}  // namespace vsag
```
ShawnShawnYou: ditto

src/impl/elias_fano_encoder.cpp
```diff
@@ -0,0 +1,158 @@
...
+        InnerIdType high = x >> low_bits_width_;
+        InnerIdType low = x & ((1U << low_bits_width_) - 1);
+
+        set_high_bit(bits_, i + high, low_bits_size_);
+        set_low_bits(i, low);
+    }
+}
+
+Vector<InnerIdType>
+EliasFanoEncoder::DecompressAll(Allocator* allocator) const {
```
ShawnShawnYou, line 133: why not use allocator_? What's the difference between allocator_ and allocator

nedchu: The returned vector is a temp vector. 
The user is expected to allocate the memory for this vector and then destroy it soon.
allocator_ is used internally for bit_, and the class will help destroy bit_.

ShawnShawnYou (approved):

src/impl/elias_fano_encoder.cpp
nedchu: done

src/impl/elias_fano_encoder.cpp
inabao: can pass the reference of `result` as parameter here instead of passing the allocator

PR #543, opened 2025-03-28
Support Hgraph Iterator Filter Search

### LINKED ISSUE #544: support iterative search on index

Carrot-77:
related to: #544 

Iterator Filter Search is used in queries with a vector data filtering rate of 1%-50%, to achieve higher performance by compressing the number of queries. When the filtering rate is 1%-50%, iterative filtering is basically stable between 1-2 rounds of queries.

The principle：
A discard heap is introduced. The function of the discard heap is to save the nodes discarded in the previous round of queries and sort them by distance. If the results in the first round of queries are insufficient, another round of queries will be performed. In the next round, the nodes in the discard heap are used as candidate sets, and the neighbors that have not been visited are visited as new output results, and the discard heap is updated. Each subsequent round of queries will query ef_search nodes. The query ends when limitK is satisfied or discard is empty.

Here are our test results
feat-perf is our current optimization, comparing the performance of master
image

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -216,12 +216,12 @@ HGraph::KnnSearch(const DatasetPtr& query,
 
     if (iter_ctx == nullptr) {
         auto cur_count = this->bottom_graph_->TotalCount();
-        IteratorFilterContext* new_ctx = new IteratorFilterContext();
+        auto new_ctx = new IteratorFilterContext();
         new_ctx->init(cur_count, params.ef_search, allocator_);
         iter_ctx = new_ctx;
     }
 
-    IteratorFilterContext* iter_filter_ctx = static_cast<IteratorFilterContext*>(iter_ctx);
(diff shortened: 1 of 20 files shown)


src/algorithm/hgraph.cpp
```diff
@@ -190,6 +191,98 @@ HGraph::KnnSearch(const DatasetPtr& query,
     return std::move(dataset_results);
 }
 
+DatasetPtr
+HGraph::KnnSearch(const DatasetPtr& query,
+                  int64_t k,
+                  const std::string& parameters,
+                  const FilterPtr& filter,
+                  vsag::IteratorContext*& iter_ctx,
```
inabao: `IteratorContext*& iter_ctx`

Carrot-77: Modified

wxyucs (approved):
lgtm

inabao (approved):
LGTM

ShawnShawnYou (approved):

PR #554, opened 2025-04-02
fix the issue where train_data_type_ is not initialized

### LINKED ISSUE #550: evaluating performance using eval_performance on the IP dataset will causes a core dump.
**Describe the bug**
The initialization order of train_data_type_ is incorrect.

**To Reproduce**
```cpp

                if (obj->train_data_type_ == vsag::DATATYPE_FLOAT32) {
                    obj->distance_func_ = vsag::InnerProductDistance;
                } else if (obj->train_data_type_ == vsag::DATATYPE_INT8) {
                    obj->distance_func_ = vsag::INT8InnerProductDistance;
                }

```

**Environment**
- vsag version: v0.13.0-117-g135a5ded
- compiler version: c++ (GCC) 9.2.1 20200522 (Alibaba 9.2.1-3 2.17)

inabao:
fixed: #550

### DIFF
--- a/tools/eval/eval_dataset.cpp
+++ b/tools/eval/eval_dataset.cpp
@@ -153,50 +153,6 @@ EvalDataset::Load(const std::string& filename) {
         }
     }
 
-    try {
-        H5::Attribute attr = file.openAttribute("distance");
-        H5::StrType str_type = attr.getStrType();
-        std::string metric;
-        attr.read(str_type, metric);
-        if (obj->vector_type_ == DENSE_VECTORS) {
-            if (metric == "euclidean") {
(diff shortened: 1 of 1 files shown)


wxyucs (approved):
lgtm

tools/eval/eval_dataset.cpp
```diff
@@ -272,6 +228,50 @@ EvalDataset::Load(const std::string& filename) {
         }
     }
 
+    try {
```
jiaweizone, line 231: minor: the method EvalDataset::Load is too long

jiaweizone (approved):
LGTM

PR #559, opened 2025-04-03
fix coredump for searching empty HGraph index

### LINKED ISSUE #557: KnnSearch in Hgraph without Build will cause a coredump
**Describe the bug**
there is no check in HGraph when the index is empty

**Screenshots**

Image

### LINKED ISSUE #558: loss initialization of topk in HGraph when searching high level graph
**To Reproduce**
 search_param.topk is 0 as default:
```cpp
    InnerSearchParam search_param;
    search_param.ep = this->entry_point_id_;
    search_param.ef = 1;
    search_param.is_inner_id_allowed = nullptr;
    for (auto i = static_cast<int64_t>(this->route_graphs_.size() - 1); i >= 0; --i) {
        auto result = this->search_one_graph(query->GetFloat32Vectors(),
                                             this->route_graphs_[i],
                                             this->basic_flatten_codes_,
                                             search_param);
        search_param.ep = result.top().second;
    }

```

inabao:
fixed: #558 
fixed: #557

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -196,6 +196,7 @@ HGraph::KnnSearch(const DatasetPtr& query,
 
     InnerSearchParam search_param;
     search_param.ep = this->entry_point_id_;
+    search_param.topk = 1;
     search_param.ef = 1;
     search_param.is_inner_id_allowed = nullptr;
     for (auto i = static_cast<int64_t>(this->route_graphs_.size() - 1); i >= 0; --i) {
@@ -289,6 +290,7 @@ HGraph::KnnSearch(const DatasetPtr& query,
     } else {
         InnerSearchParam search_param;
         search_param.ep = this->entry_point_id_;
+        search_param.topk = 1;
         search_param.ef = 1;
         search_param.is_inner_id_allowed = nullptr;
         if (iter_filter_ctx->IsFirstUsed()) {
@@ -446,6 +448,7 @@ HGraph::RangeSearch(const DatasetPtr& query,
 
     InnerSearchParam search_param;
     search_param.ep = this->entry_point_id_;
+    search_param.topk = 1;
     search_param.ef = 1;
     for (auto i = static_cast<int64_t>(this->route_graphs_.size() - 1); i >= 0; --i) {
         auto result = this->search_one_graph(query->GetFloat32Vectors(),
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -22,6 +22,7 @@
 
 #include "common.h"
 #include "data_cell/sparse_graph_datacell.h"
+#include "dataset_impl.h"
 #include "empty_index_binary_set.h"
 #include "impl/pruning_strategy.h"
 #include "index/iterator_filter.h"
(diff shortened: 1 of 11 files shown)


src/algorithm/inner_index_interface.h
```diff
@@ -111,6 +111,11 @@ class InnerIndexInterface {
                 const std::string& parameters,
                 int64_t limited_size = -1) const {
         FilterPtr filter = nullptr;
+        if (GetNumElements() == 0) {
```
jiaweizone, line 115: The line L119 has already been checked, so there is no need to check here ？

inabao: Line L119 is the search function within the specific index; we cannot rely on its checks.

src/algorithm/inner_index_interface.cpp
```diff
@@ -42,6 +42,11 @@ InnerIndexInterface::KnnSearch(const DatasetPtr& query,
     if (filter != nullptr) {
         filter_ptr = std::make_shared<UniqueFilter>(filter);
     }
+    if (GetNumElements() == 0) {
+        auto result = Dataset::Make();
+        result->Dim(0)->NumElements(1);
+        return result;
+    }
```
wxyucs: add a new function named `Dataset::MakeEmptyResult` and put these codes in it to avoid code repetition.

wxyucs: after modified:
```cpp
if (GetNumElements() == 0) {
    return Dataset::MakeEmptyResult();
}
```

wxyucs (changes requested):

src/algorithm/inner_index_interface.cpp
inabao: done

src/algorithm/hgraph.cpp
```diff
@@ -254,6 +255,11 @@ HGraph::KnnSearch(const DatasetPtr& query,
                   const FilterPtr& filter,
                   IteratorContext*& iter_ctx,
                   bool is_last_filter) const {
+    if (GetNumElements() == 0) {
+        auto result = Dataset::Make();
+        result->Dim(0)->NumElements(1);
+        return result;
+    }
```
LHT129, line 259: why not place it on IndexImpl

tests/test_hgraph.cpp
```diff
@@ -327,6 +327,34 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::HgraphTestIndex,
     }
 }
 
+TEST_CASE_PERSISTENT_FIXTURE(fixtures::HgraphTestIndex,
+                             "HGraph Search Empty Index",
+                             "[ft][hgraph]") {
+    auto origin_size = vsag::Options::Instance().block_size_limit();
```
LHT129, line 333: add new test method on test_index.h and reuse on different idex

LHT129 (changes requested):

src/algorithm/hgraph.cpp
inabao: done

wxyucs (approved):
lgtm

LHT129 (approved):
LGTM

PR #680, opened 2025-05-06
use DistanceHeap replace MaxHeap

LHT129:


src/impl/basic_searcher.cpp
```diff
@@ -85,15 +89,15 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+DistHeapPtr
 BasicSearcher::search_impl(const GraphInterfacePtr& graph,
                            const FlattenInterfacePtr& flatten,
                            const VisitedListPtr& vl,
                            const void* query,
                            const InnerSearchParam& inner_search_param,
                            IteratorFilterContext* iter_ctx) const {
-    MaxHeap top_candidates(allocator_);
-    MaxHeap candidate_set(allocator_);
+    auto top_candidates = std::make_shared<StandardHeap<true, false>>(allocator_, -1);
```
inabao, line 99: can we initialize a size here?

LHT129: origin MaxHeap not fixed size, so we only replace with non-fixed-size StandardHeap

inabao (approved):
LGTM

wxyucs (approved):
lgtm

ShawnShawnYou (approved):

Issue #696, opened 2025-05-12
diskann static library has unexpected AVX instructions may cause SIGILL on machines without AVX


gaopy3:
**Describe the bug**
libdiskann.a has unexpected AVX instructions in std vector implementations which may cause SIGILL on machines without AVX.

**To Reproduce**
Codes to reproduce the behavior:
```shell
objdump -d libdiskann.a | grep vector.*pairIjj.*realloc_insert -C 10 | grep vzeroupper -C 10
echo "_ZNSt6vectorISt4pairIjjESaIS1_EE17_M_realloc_insertIJS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_" | c++filt
```

```
 67c:	5d                   	pop    %rbp
 67d:	c3                   	retq
 67e:	66 90                	xchg   %ax,%ax
 680:	4d 85 c0             	test   %r8,%r8
 683:	0f 85 cc 01 00 00    	jne    855 <_ZNSt6vectorISt4pairIjjESaIS1_EE17_M_realloc_insertIJS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_+0x855>
 689:	41 bc 08 00 00 00    	mov    $0x8,%r12d
 68f:	45 31 ed             	xor    %r13d,%r13d
 692:	e9 07 fa ff ff       	jmpq   9e <_ZNSt6vectorISt4pairIjjESaIS1_EE17_M_realloc_insertIJS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_+0x9e>
 697:	66 0f 1f 84 00 00 00 	nopw   0x0(%rax,%rax,1)
 69e:	00 00
 6a0:	c5 f8 77             	vzeroupper
 6a3:	eb b6                	jmp    65b <_ZNSt6vectorISt4pairIjjESaIS1_EE17_M_realloc_insertIJS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_+0x65b>
 6a5:	0f 1f 00             	nopl   (%rax)
 6a8:	41 bc 08 00 00 00    	mov    $0x8,%r12d
 6ae:	e9 b2 f9 ff ff       	jmpq   65 <_ZNSt6vectorISt4pairIjjESaIS1_EE17_M_realloc_insertIJS1_EEEvN9__gnu_cxx17__normal_iteratorIPS1_S3_EEDpOT_+0x65>
[...]

PR #698, opened 2025-05-12
disable simd instructions in diskann

wxyucs:
fixes: #696

SIMD instructions (avx,sse, ..), in diskann, are not used by vsag now, so we can disable these codes to avoid `illegal instructions` happened on some old platforms.


jiaweizone (approved):
LGTM

inabao (approved):
LGTM

PR #753, opened 2025-05-23
introduce fast bitset for attribute index

LHT129:


inabao (approved):
LGTM

wxyucs (approved):
lgtm

src/impl/bitset/sparse_bitset.cpp
```diff
@@ -51,46 +31,72 @@ BitsetImpl::Set(int64_t pos, bool value) {
...
     std::lock(mutex_, another_ptr->mutex_);
     std::lock_guard<std::mutex> lock(mutex_, std::adopt_lock);
     std::lock_guard<std::mutex> lock_other(another_ptr->mutex_, std::adopt_lock);
     r_ ^= another_ptr->r_;
 }
 
+void
+SparseBitset::Not() {
+    std::lock_guard<std::mutex> lock(mutex_);
+    r_.flipClosed(r_.minimum(), r_.maximum());
```
jiaweizone, line 79: `minimum` will throw exception when `Roaring bitmap` is empty, so we need add an empty check

src/CMakeLists.txt
```diff
@@ -6,8 +6,10 @@ add_subdirectory (quantization)
 file (GLOB CPP_SRCS "*.cpp")
 list (FILTER CPP_SRCS EXCLUDE REGEX "_test.cpp")
 
-file (GLOB CPP_CONJUGATE_GRAPH_SRCS "impl/*.cpp")
-list (FILTER CPP_CONJUGATE_GRAPH_SRCS EXCLUDE REGEX "_test.cpp")
+file (GLOB CPP_IMPL_SRCS "impl/*.cpp")
```
jiaweizone, line 9: need exclude unit test files
```CMake
file (GLOB_RECURSE CPP_IMPL_SRCS "impl/*.cpp")
list (FILTER CPP_IMPL_SRCS EXCLUDE REGEX "_test.cpp")
```

src/impl/bitset/sparse_bitset.cpp
LHT129: we will catch the excption on index layer, if we check here, we still need to throw the exception too

ShawnShawnYou (approved):

PR #773, opened 2025-05-28
Add gramma expression define  and expression visitor

yulijunzj:


LHT129 (approved):
LGTM

src/attr/expression_visitor.h
```diff
@@ -0,0 +1,350 @@
...
+#include "fmt/format.h"
+
+namespace vsag {
+class FCErrorListener final : public antlr4::BaseErrorListener {
+public:
+    FCErrorListener(const std::string& input) : input_(input) {
+    }
+
+    void
+    syntaxError(antlr4::Recognizer* recognizer,
```
wxyucs, line 33: please follow this rule of function naming: https://github.com/antgroup/vsag/blob/main/.clang-tidy#L49

wxyucs (changes requested):

src/attr/expression_visitor.h
wxyucs: override virtual function, which is generated. cannot renaming, fix in the future

wxyucs (approved):
lgtm

Issue #777, opened 2025-05-30
HGraph with Compressed Graph has Segment Fault on perf-99.yml


nedchu:
**Describe the bug**

Segment Fault when we use HGraph with Compressed Graph on perf-99.yml.
I am investigating the issue.

![Image](https://github.com/user-attachments/assets/2a04dc67-7367-4475-a4f8-a1f4f7ee762d)

Core

**To Reproduce**
Codes to reproduce the behavior:

``` bash
cp .github/perf-99.yml /tmp/perf-99-compress.yml
# add `"graph_storage_type": "compressed"` to "index_params" in the new yaml.
./build-release/tools/eval/eval_performance /tmp/perf-99-compress.yml
```

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS: Alibaba Group Enterprise Linux Server 7.2
- vsag version: v0.14.0-113-gf9537f2-dirty
- compiler version: c++ (GCC) 10.2.1 20200825 (Alibaba 10.2.1-3 2.17)

**Expected behavior**

No segment fault.

**Screenshots**

**Additional context**

Issue #778, opened 2025-05-30
Unusual Recall Degradation of HGraph with Compressed Graph


nedchu:
**Describe the bug**

Unusual recall degradation when we use HGraph with Compressed Graph

![Image](https://github.com/user-attachments/assets/ef037169-f456-4c90-9a26-bc2773ebf577)

Core

**To Reproduce**
Codes to reproduce the behavior:

``` bash
cp .github/perf-90.yml /tmp/perf-90-compress.yml
# add `"graph_storage_type": "compressed"` to "index_params" in the new yaml.
./build-release/tools/eval/eval_performance /tmp/perf-90-compress.yml
```

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS: Alibaba Group Enterprise Linux Server 7.2
- vsag version: v0.14.0-113-gf9537f2-dirty
- compiler version: c++ (GCC) 10.2.1 20200825 (Alibaba 10.2.1-3 2.17)

**Expected behavior**

No recall degradation

**Screenshots**

**Additional context**

PR #784, opened 2025-06-04
fix segment fault and recall degradation issues in compressed HGraph

nedchu:
closes: #777 #778

### DIFF
--- a/src/data_cell/compressed_graph_datacell.cpp
+++ b/src/data_cell/compressed_graph_datacell.cpp
@@ -43,15 +43,12 @@ CompressedGraphDataCell::InsertNeighborsById(InnerIdType id,
     Vector<InnerIdType> tmp(neighbor_ids.begin(), neighbor_ids.end(), allocator_);
     std::sort(tmp.begin(), tmp.end());
 
-    // TODO(deming): reset the size of neighbor_sets_[id] while tmp is empty
+    std::unique_ptr<EliasFanoEncoder> new_node;
     if (not tmp.empty()) {
-        if (neighbor_sets_[id] == nullptr) {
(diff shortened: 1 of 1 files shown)


src/data_cell/compressed_graph_datacell.cpp
```diff
@@ -65,6 +62,7 @@ CompressedGraphDataCell::GetNeighborSize(InnerIdType id) const {
 
 void
 CompressedGraphDataCell::GetNeighbors(InnerIdType id, Vector<InnerIdType>& neighbor_ids) const {
+    neighbor_ids.clear();
     if (GetNeighborSize(id) > 0) {
         neighbor_sets_[id]->DecompressAll(neighbor_ids);
```
nedchu, line 67: bug happens here

src/data_cell/compressed_graph_datacell.cpp
```diff
@@ -65,6 +62,7 @@ CompressedGraphDataCell::GetNeighborSize(InnerIdType id) const {
 
 void
 CompressedGraphDataCell::GetNeighbors(InnerIdType id, Vector<InnerIdType>& neighbor_ids) const {
+    neighbor_ids.clear();
```
jiaweizone, line 65: bug is miss clear vector ?

nedchu: if `neighbor_ids = {1, 2, 3}`  and `neighbor_set_[id]={}`, the function is intended to return empty `neighbor_ids={}`.
In the old version, the function will return `neighbor_ids = {1, 2, 3}`, causing a bug in graph construction.

jiaweizone (approved):
LGTM

PR #815, opened 2025-06-18
support for HGraph using a static schema

inabao:
- In static mode, HGraph removes locks and reverse label tables; the current static mode remains pluggable (only guaranteeing no core dump), but does not guarantee correctness.


src/label_table.h
```diff
@@ -94,6 +115,7 @@ class LabelTable {
     UnorderedMap<LabelType, InnerIdType> label_remap_;
```
LHT129: merge implement

include/vsag/constants.h
```diff
@@ -146,6 +146,7 @@ extern const char* const HGRAPH_INIT_CAPACITY;
 extern const char* const HGRAPH_GRAPH_TYPE;
 extern const char* const HGRAPH_GRAPH_STORAGE_TYPE;
 extern const char* const HGRAPH_BUILD_THREAD_COUNT;
+extern const char* const HGRAPH_STATIC;
 extern const char* const HGRAPH_PRECISE_QUANTIZATION_TYPE;
```
LHT129, line 150: I think immutable is better than static

LHT129 (changes requested):

include/vsag/constants.h
inabao: done

wxyucs (approved):
lgtm

LHT129 (approved):
LGTM

Issue #879, opened 2025-07-02
improve compression ratio of compressed graph


nedchu:
**Is your feature request related to a problem? Please describe.**

Improve the perf of compressed graph.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #880, opened 2025-07-02
improve compressed graph in compression ratio

nedchu:
closes: #879 

Two improvements

1. One allocator per adjacency list -> one allocator per compressed graph (encoder's data is now managed by compressed graph datacell). The space reduction is `|V| * 8B`. The gain cannot be seen on the eval report, as we do not serialize allocators.

<del> 2. In EliasFanoEncoder, change the type of `bits` from `uint64_t` to `uint32_t`. This can save 5% space on SIFT, due to a smaller block size. <del>


jiaweizone:
bugbot run

nedchu:
Reply to cursor bot:

For the first issue, InnerIdType is 32-bit, so the code is safe.
For the second issue, we need to be careful. The serialization is not compatible with v0.15. There are two solutions. One is that we revert an optimization (resulting in a 5% difference in space). Another is that we use the index version to check this incompatibility.

nedchu:
To ensure compatibility, I revert the second optimization. I will use index versioning to add the optimization in the future.

nedchu:
bugbot run

LHT129 (approved):
LGTM

PR #885, opened 2025-07-03
fix coredump in hgraph when using empty parameter

### LINKED ISSUE #883: [0.15.0] create hgraph index crash when index_param is empty
**Describe the bug**
create hgraph index with empty index_param causes core dump

version/0.15.0

**To Reproduce**
```cpp
  std::string hgraph_build_parameters = R"(
    {
        "dtype": "float32",
        "metric_type": "l2",
        "dim": 128,
        "index_param": {}
    }
    )";
  ::vsag::Engine engine;
  auto index = engine.CreateIndex("hgraph", hgraph_build_parameters).value();
```

inabao:
closed: #883

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -1229,7 +1229,7 @@ static const std::string HGRAPH_PARAMS_TEMPLATE =
                 "{PCA_DIM}": 0,
                 "{RABITQ_QUANTIZATION_BITS_PER_DIM_QUERY}": 32,
                 "nbits": 8,
-                "{PRODUCT_QUANTIZATION_DIM}": 0
+                "{PRODUCT_QUANTIZATION_DIM}": 32
             }
         },
         "{HGRAPH_PRECISE_CODES_KEY}": {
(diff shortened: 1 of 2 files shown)


wxyucs (approved):
lgtm

src/algorithm/hgraph.cpp
```diff
@@ -1242,7 +1242,7 @@ static const std::string HGRAPH_PARAMS_TEMPLATE =
                 "{QUANTIZATION_TYPE_KEY}": "{QUANTIZATION_TYPE_VALUE_FP32}",
                 "{SQ4_UNIFORM_QUANTIZATION_TRUNC_RATE}": 0.05,
                 "{PCA_DIM}": 0,
-                "{PRODUCT_QUANTIZATION_DIM}": 0
+                "{PRODUCT_QUANTIZATION_DIM}": 32
```
LHT129: 1 is better, some times dim may lower than 32

inabao: done

LHT129 (approved):
LGTM

PR #900, opened 2025-07-08
remove file when use "disk io" type

### LINKED ISSUE #901: remove "temporary file" with disk like IO type
when use disk like IO type "buffer io", "async io", "mmap io". we will create temporary file on search phrase. when Index is destroyed, the file is useless but occupy too much disk.  So we can remove it

LHT129:
closed: #901

### DIFF
--- a/src/io/async_io.cpp
+++ b/src/io/async_io.cpp
@@ -15,8 +15,17 @@
 
 #include "async_io.h"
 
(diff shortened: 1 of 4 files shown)


ShawnShawnYou (approved):

inabao (approved):
LGTM

PR #907, opened 2025-07-08
update str pipe list gramma

yulijunzj:
update gramma
update visitStr_value_list


LHT129 (approved):
LGTM

src/attr/expression_visitor.cpp
```diff
@@ -15,9 +15,11 @@
 
 #include "expression_visitor.h"
 
+#include "FCLexer.h"
```
wxyucs: include with project path, like #include <spdlog/spadlog.h>

yulijunzj: done

inabao (approved):
LGTM

tests/test_fc_visitor.cpp
```diff
@@ -13,7 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <antlr4-autogen/FCLexer.h>
+#include <safe_allocator.h>
+#include <vsag_exception.h>
```
wxyucs: internal header files should use "" to include

wxyucs (approved):

tests/test_fc_visitor.cpp
yulijunzj: done

PR #929, opened 2025-07-14
Improve sq8 ipc and decrease building cost in ARM.

Danbaiwq:
In ARM, use below optimizations to decrease build time (237s -> 232s), improve search QPS(2482 -> 5216) and ipc(2.08 -> 2.16) by decreasing instructions' stalls:
- Loop unrolling from 4 elements to 12 elements.
- FMA instruction performs both multiplication and addition.
- Prefetch to decrease instruction pipeline stalls.
- Pre-calculate `inv255` to avoid repeated calculations in the loop.
- Optimize `RabitQFloatBinary`. 

Before optimizing:
截屏2025-07-14 13 13 18

After optimizing:
5E47CE4B-0F78-474B-9A0B-9D7C1EEF15BB

### DIFF
--- a/src/simd/neon.cpp
+++ b/src/simd/neon.cpp
@@ -1743,10 +1743,10 @@ shuffle_16_char(const uint8x16_t* a, const uint8x16_t* b) {
 #endif
 
 void
-Prefetch(const void* data){
-    // #if defined(ENABLE_NEON)
-    //     __builtin_prefetch(data, 0, 1);
-    // #endif
+Prefetch(const void* data) {
+#if defined(ENABLE_NEON)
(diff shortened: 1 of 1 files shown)


src/simd/neon.cpp
```diff
@@ -1307,4 +1500,4 @@ BitNot(const uint8_t* x, const uint64_t num_byte, uint8_t* result) {
 #endif
 }
 
-}  // namespace vsag::neon
+}  // namespace vsag::neon
```
LHT129: format

LHT129 (approved):
LGTM

PR #931, opened 2025-07-14
Arm simd optimize for fht

jingyueob:


wxyucs:
already done in main.

PR #951, opened 2025-07-17
add new implement for attr

LHT129:


ShawnShawnYou (approved):

src/attr/executor/executor.h
```diff
@@ -49,16 +53,21 @@ class Executor {
...
+    Filter*
+    Run() {
+        return this->Run(0);
+    }
 
 public:
     bool only_bitset_{true};
 
-    FilterPtr filter_{nullptr};
+    Filter* filter_{nullptr};
```
wxyucs, line 70: @LHT129 What is the reason to use the raw pointer here?

wxyucs (approved):
lgtm

src/attr/executor/executor.h
LHT129: for high performance

PR #964, opened 2025-07-21
optimize IteratorFilterContext bitmap memory usage

shadowao:
The current memory overhead of bitmap in IteratorFilterContext is max_size * sizeof(uint16_t). This overhead is too high when there are a large number of vectors, such as tens of millions. Therefore, it is changed to a 1-bit implementation


LHT129 (approved):
LGTM

inabao (approved):
LGTM

wxyucs (approved):
lgtm

PR #969, opened 2025-07-24
make thirdparty library static

wxyucs:
- use the static library of `cpuinfo` instead of the dynamic library
- make the submodule `simd` output a static library, not a dynamic library


inabao (approved):
LGTM

LHT129 (approved):
LGTM

PR #993, opened 2025-07-30
eliminate the effect of data size on the efsearch threshold

inabao:
## Summary by Sourcery

Enforce a consistent ef_search threshold across graph and HNSW searches and consolidate parameter parsing to eliminate code duplication and the impact of data size on the threshold.

Enhancements:
- Unify ef_search threshold logic in HGraph and HNSW to use max(AMPLIFICATION_FACTOR * k, 1000) for range validation
- Consolidate parameter parsing and ef_search validation into a single location in each search overload to remove duplicate code
- Reorder validations to perform ef_search range checks before k constraints and locking operations

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -297,6 +297,14 @@ HGraph::KnnSearch(const DatasetPtr& query,
             query_dim == dim_,
             fmt::format("query.dim({}) must be equal to index.dim({})", query_dim, dim_));
     }
+
+    auto params = HGraphSearchParameters::FromJson(parameters);
+
+    auto ef_search_threshold = std::max(AMPLIFICATION_FACTOR * k, 1000L);
+    CHECK_ARGUMENT(  // NOLINT
+        (1 <= params.ef_search) and (params.ef_search <= ef_search_threshold),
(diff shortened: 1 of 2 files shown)


src/algorithm/hgraph.cpp
```diff
@@ -297,6 +297,14 @@ HGraph::KnnSearch(const DatasetPtr& query,
...
     }
+
+    auto params = HGraphSearchParameters::FromJson(parameters);
+
+    auto ef_search_threshold = std::max(AMPLIFICATION_FACTOR * k, 1000L);
+    CHECK_ARGUMENT(  // NOLINT
+        (1 <= params.ef_search) and (params.ef_search <= ef_search_threshold),
+        fmt::format("ef_search({}) must in range[1, {}]", params.ef_search, ef_search_threshold));
+
     // check k
```
LHT129, line 308: what will happen, when ef_search larger than num_element?

inabao: The function will return all vectors if the graph is connected; otherwise, it will return a connected component.

LHT129 (approved):
LGTM

ShawnShawnYou (approved):

PR #994, opened 2025-07-30
improve ivf parallel search

### LINKED ISSUE #833: Add parallel search for one query on IVF Index

LHT129:
closed: #833 

## Summary by Sourcery

Refactor IVF search to eliminate shared heap locking by using thread-local heaps and executors, then merge results after parallel execution.

Enhancements:
- Replace a single mutex-protected DistanceHeap with a vector of per-thread heaps and merge them into a final heap after parallel search
- Change attribute filtering to use a vector of per-thread Executor instances instead of a single executor
- Optimize single-thread execution path by assigning the first thread’s heap directly as the search result when parallelism is one

### DIFF
--- a/src/algorithm/ivf.cpp
+++ b/src/algorithm/ivf.cpp
@@ -683,15 +683,16 @@ IVF::search(const DatasetPtr& query, const InnerSearchParam& param) const {
         }
     }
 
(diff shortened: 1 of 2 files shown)


inabao (approved):
LGTM

ShawnShawnYou (approved):

PR #1005, opened 2025-07-31
add benchmark parameters of popular datasets

wxyucs:
## Summary by Sourcery

Introduce a new YAML benchmark configuration for SIFT datasets and refactor evaluation cases to use file streams for deserialization with enhanced exception handling.

New Features:
- Add benchs/eval-sift.yml containing benchmark parameters for SIFT datasets including HGRAPH, HNSW, IVF, and DiskANN indices

Enhancements:
- Refactor SearchEvalCase::deserialize to accept an ifstream reference and update Run() to open and close the input file stream explicitly
- Include <stdexcept> and add a catch block for std::runtime_error when retrieving memory usage detail in both search and build eval cases


LHT129 (approved):
LGTM

inabao (approved):
LGTM

tools/eval/case/search_eval_case.cpp
```diff
@@ -250,6 +256,9 @@ SearchEvalCase::process_result() {
     // TODO(deming): remove try-catch after implement GetMemoryUsageDetail
     try {
         result["memory_detail(B)"] = this->index_->GetMemoryUsageDetail();
+    } catch (std::runtime_error& e) {
+        // if GetMemoryUsageDetail not implemented
+        logger_->Debug(e.what());
```
jiaweizone: Here can change to use `logger_->Error(e.what())` ?

tools/eval/case/search_eval_case.cpp
```diff
@@ -20,6 +20,7 @@
 #include <cstdio>
 #include <fstream>
 #include <iostream>
+#include <stdexcept>
```
jiaweizone, line 23: Where is need this header file ?

wxyucs: std::runtime_error is defined in the <stdexcept> header file

tools/eval/case/search_eval_case.cpp
wxyucs: done

jiaweizone (approved):
LGTM

PR #1023, opened 2025-08-07
add file remove check on disk io type

LHT129:
## Summary by Sourcery

Add file existence checks to disk IO classes to prevent deleting pre-existing files and to error on directory paths, and refactor MMapIO by extracting its implementation into a new source file

Enhancements:
- Track file existence in MMapIO, AsyncIO, and BufferIO and only remove files created by the process
- Throw an exception when initializing IO classes with a directory path
- Move MMapIO inline method implementations to a new mmap_io.cpp and update CMakeLists to include it


inabao (approved):
LGTM

wxyucs (approved):
lgtm

PR #1038, opened 2025-08-12
update perf.yml for COHERE

inabao:
## Summary by Sourcery

Update COHERE performance workflows to use rabitq quantization with FHT, run only search, tune search parameters, and update index paths.

Enhancements:
- Run only the search stage for COHERE across perf-90, perf-95, and perf-99 configurations
- Switch Hgraph index creation to use rabitq quantization with FHT
- Adjust ef_search values to 23, 38, and 150 for perf-90, perf-95, and perf-99 respectively
- Update index path suffix for perf-90 and perf-95 to reflect rabitq_FHT index


wxyucs (approved):
lgtm

PR #1039, opened 2025-08-13
fix logic in generic sq4 compute

### LINKED ISSUE #960: Bug: Incorrect Logic ingeneric::SQ4ComputeCodes*Scalar Implementations
#### **Bug Description**

There is a logical error in the generic/scalar implementations of `SQ4ComputeCodesIP` and `SQ4ComputeCodesL2Sqr`. These functions are currently used as the ground truth for SIMD accuracy tests in `src/simd/sq4_simd_test.cpp`.

The error is that when decoding the high 4-bit nibble of a code byte (which corresponds to dimension `d+1`), the code incorrectly re-uses `diff[d]` and `lower_bound[d]` from the low nibble's dimension. The correct logic should use `diff[d+1]` and `lower_bound[d+1]` for the `d+1` dimension.

#### **Code Analysis: The Buggy Section**

The issue lies in the `if (d + 1 < dim)` block within both `SQ4ComputeCodesIP` and `SQ4ComputeCodesL2Sqr` in the generic implementation.

**File:** `src/simd/generic.cpp` 

```cpp
// Problematic code from the generic implementation
float
SQ4ComputeCodesIP(const uint8_t* RESTRICT codes1,
                  // ... other params
                  uint64_t dim) {
    // ...
    for (uint64_t d = 0; d < dim; d += 2) {
        // ... lo part is correct ...
        x_lo = (codes1[d >> 1] & 0x0f) / 15.0 * diff[d] + lower_bound[d];
        y_lo = (codes2[d >> 1] & 0x0f) / 15.0 * diff[d] + lower_bound[d];

        if (d + 1 < dim) {
            // ERROR: These lines should be using diff[d + 1] and lower_bound[d + 1]
            x_hi = ((codes1[d >> 1] & 0xf0) >> 4) / 15.0 * diff[d] + lower_bound[d];
            y_hi = ((codes2[d >> 1] & 0xf0) >> 4) / 15.0 * diff[d] + lower_bound[d];
        }
        // ...
    }
    return result;
}
```

#### **Proposed Fix**

The generic/scalar functions should be corrected to use the `d+1` index for `diff` and `lower_bound` when processing the high 4-bit nibble.

```cpp
// Corrected generic implementation
float
SQ4ComputeCodesIP(const uint8_t* RESTRICT codes1,
                  const uint8_t* RESTRICT codes2,
                  const float* RESTRICT lower_bound,
                  const float* RESTRICT diff,
                  uint64_t dim) {
    double result = 0; // Use double for better precision as a ground truth
    float x_lo = 0, x_hi = 0, y_lo = 0, y_hi = 0;

    for (uint64_t d = 0; d < dim; d += 2) {
        x_lo = (float)(codes1[d >> 1] & 0x0f) / 15.0f * diff[d] + lower_bound[d];
        y_lo = (float)(codes2[d >> 1] & 0x0f) / 15.0f * diff[d] + lower_bound[d];
        
        if (d + 1 < dim) {
            // FIX: Use d+1 for the high nibble's corresponding dimension
            x_hi = (float)((codes1[d >> 1] & 0xf0) >> 4) / 15.0f * diff[d + 1] + lower_bound[d + 1];
            y_hi = (float)((codes2[d >> 1] & 0xf0) >> 4) / 15.0f * diff[d + 1] + lower_bound[d + 1];
        } else {
            x_hi = 0;
            y_hi = 0;
        }

        result += (double)x_lo * y_lo + (double)x_hi * y_hi;
    }

    return (float)result;
}
```
*(A similar correction should be applied to `SQ4ComputeCodesL2Sqr`)*.

Thanks for looking into this

jac0626:
close #960


jac0626:
> add ut and ft to cover the bug ?

Hi @ShawnShawnYou,
Thank you for the feedback.
Based on this, here's my proposed plan:
1. Add unit tests for all scalar functions: I plan to write correctness tests that cover all scalar functions. However, since we can only validate results through manual computation, we'll be limited to simple test cases. There's still some risk involved as hand calculations are prone to human error.
2. Skip functional tests for now: I don't believe dedicated functional tests are necessary for these  math functions. Because the inherent precision loss from quantization makes them difficult to validate effectively.
I'm ready to proceed with implementing unit tests for the  scalar functions. Please let me know if this approach works for you.

LHT129 (approved):
LGTM

Issue #1043, opened 2025-08-14
Precompiled CXX11 ABI distribution incorrectly provides Pre-CXX11 ABI


wxyucs:
**Describe the bug**
Precompiled binary distributions in the Assets of releases (https://github.com/antgroup/vsag/releases) incorrectly provide Pre-CXX11 ABI.

```bash
# currently
➜  lib nm libvsag.so | grep 'CreateIndex' | c++filt
0000000000848b70 T vsag::InnerIndexInterface::FastCreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::IndexCommonParam const&)
000000000045af50 t vsag::InnerIndexInterface::FastCreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::IndexCommonParam const&) [clone .cold]
0000000000502fa0 T vsag::Engine::CreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
000000000043121e t vsag::Engine::CreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) [clone .cold]
0000000000541120 T vsag::Factory::CreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::Allocator*)
0000000000431eba t vsag::Factory::CreateIndex(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::Allocator*) [clone .cold]


# expected
➜  src  nm libvsag.so | grep 'CreateIndex' | c++filt
0000000000a64c90 T vsag::InnerIndexInterface::FastCreateIndex(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::IndexCommonParam const&)
000000000061d7d8 t vsag::InnerIndexInterface::FastCreateIndex(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, vsag::IndexCommonParam const&) [clone .cold]
[...]

PR #1044, opened 2025-08-14
fix incorrect CXX ABI in binary distribution

wxyucs:
solves: #1043

## Summary by Sourcery

Refactor the distribution script to eliminate duplication, introduce a script to validate the C++11 ABI in releases, and update build targets and Dockerfiles to generate both pre-CXX11 and CXX11 ABI binaries

New Features:
- Add check-cpp-abi.sh script to verify C++11 ABI in binary distributions

Enhancements:
- Refactor release/dist.sh to use a generic build function for different ABI targets
- Rename dist-old-abi Makefile target to dist-pre-cxx11-abi

Build:
- Add Dockerfiles and update dist.sh to produce pre-CXX11 and CXX11 ABI distribution images and archives


yulijunzj (approved):
LGTM

scripts/release/check-cpp-abi.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
jiaweizone, line 1: Where is this script file called ?

wxyucs: @jiaweizone it's not in any automatic process now, just a utility tool for debugging.

jiaweizone: OK

jiaweizone (approved):
LGTM

PR #1058, opened 2025-08-18
fix compile error in tests

### LINKED ISSUE #1057: compile fail when using clang-17 as compiler
**Environment**
- vsag version: v0.16.0-19-ga0e65497-dirty
- compiler version: clang version 17.0.6 (Alibaba Cloud Compiler 17.0.6.1-24.05.10.alios7)

Image

Image

inabao:
close #1057

## Summary by Sourcery

Fix various compile errors by adjusting compiler detection script, correcting simd feature checking, and standardizing test string template declarations.

Bug Fixes:
- Fix compiler detection in check_environment.sh by using the CC environment variable instead of hardcoded c++
- Correct SupportNEON invocation to use the function call in simd tests
- Resolve compile errors in multiple tests by declaring param_template as a constexpr static const char* instead of auto

### DIFF
--- a/scripts/check_environment.sh
+++ b/scripts/check_environment.sh
@@ -12,7 +12,7 @@ function get_vsag() {
 }
 
 function get_compiler() {
-    compiler_version=$($CC --version | head -n 1)
+    compiler_version=$(${CXX:-c++} --version | head -n 1)
     echo "- compiler version: $compiler_version"
 }
 


src/simd/rabitq_simd_test.cpp
```diff
@@ -102,7 +102,7 @@ TEST_CASE("RaBitQ SQ4U-BQ Compute Benchmark", "[ut][simd][!benchmark]") {
     if (SimdStatus::SupportSVE()) {
         BENCHMARK_SIMD_COMPUTE_SQ4(sve, RaBitQSQ4UBinaryIP);
     }
-    if (SimdStatus::SupportNEON) {
+    if (SimdStatus::SupportNEON()) {
```
wxyucs, line 105: 😱

scripts/check_environment.sh
```diff
@@ -12,7 +12,7 @@ function get_vsag() {
 }
 
 function get_compiler() {
-    compiler_version=$(c++ --version | head -n 1)
+    compiler_version=$($CC --version | head -n 1)
```
wxyucs: @inabao maybe it should be `$CXX`?

inabao: done

PR #1079, opened 2025-08-22
support mrng in hgraph

HeHuMing:
## Summary by Sourcery

Introduce an alpha parameter to the pruning heuristic to allow adjustable neighbor filtering and expand unit tests to validate its behavior and mutual connections.

New Features:
- Support an alpha parameter in select_edges_by_heuristic for customizable pruning behavior.

Enhancements:
- Incorporate the alpha multiplier into the edge selection comparison and provide a default value for backward compatibility.

Tests:
- Add unit tests covering select_edges_by_heuristic with different alpha values.
- Add a unit test for mutual_connect_new_element to verify neighbor insertion and entry point selection.

### DIFF
--- a/src/impl/pruning_strategy.cpp
+++ b/src/impl/pruning_strategy.cpp
@@ -70,7 +70,7 @@ mutually_connect_new_element(InnerIdType cur_c,
                              const MutexArrayPtr& neighbors_mutexes,
                              Allocator* allocator) {
     const size_t max_size = graph->MaximumDegree();
-    select_edges_by_heuristic(top_candidates, max_size, flatten, allocator, 1.0F);
+    select_edges_by_heuristic(top_candidates, max_size, flatten, allocator);
     if (top_candidates->Size() > max_size) {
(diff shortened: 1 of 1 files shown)


src/impl/pruning_strategy.cpp
```diff
@@ -69,7 +70,7 @@ mutually_connect_new_element(InnerIdType cur_c,
                              const MutexArrayPtr& neighbors_mutexes,
                              Allocator* allocator) {
     const size_t max_size = graph->MaximumDegree();
-    select_edges_by_heuristic(top_candidates, max_size, flatten, allocator);
```
inabao: set 1.0F as the default argument for `select_edges_by_heuristic` to avoid using a magic number.

HeHuMing: Done.
The default value of the alpha parameter is set in "pruning_strategy.h".

inabao (approved):
LGTM

PR #1096, opened 2025-08-29
feat(ivf): support calculate distance by id for IVF

### LINKED ISSUE #1094: support calculate distance by id for IVF
**Is your feature request related to a problem? Please describe.**  
In current IVF-based vector retrieval systems, distance calculation is typically limited to query vectors against indexed vectors during search operations. However, there are scenarios where users need to compute distances between a query vector and **specific vectors identified by their IDs** (e.g., for validation, debugging, or post-processing). The lack of direct ID-to-distance computation forces users to either:  
1. Perform a full search to retrieve all vectors and manually filter by ID (inefficient for large datasets), or  
2. Manually retrieve the target vectors from external storage and compute distances separately (which breaks integration with the retrieval pipeline).  

---

**Describe the solution you'd like**  
Add a new API/endpoint to the IVF system that allows users to:  
- Input a query vector and a list of target vector IDs.  
- Output the pairwise distances between the query vector and each specified ID's vector.  

This would enable direct access to distance computations for specific IDs without requiring full search operations or external data fetching. The implementation should leverage existing index structures (e.g., ID-to-vector mappings) to maintain efficiency.  

---

**Describe alternatives you've considered**  
1. **Manual external lookup**: Users fetch vectors via ID from a separate database and compute distances. This is error-prone and introduces latency due to data duplication.  
2. **Full search with filtering**: Run a search with a large `k` and filter results by ID. This is inefficient for sparse ID subsets and increases memory/compute costs.  
3. **Custom pre-processing**: Precompute and cache ID-specific vectors in memory. This adds complexity and is not scalable for dynamic datasets.  

---

**Additional context**  
- **Use cases**:  
  - Validating retrieval accuracy for specific ground-truth IDs.  
  - Debugging edge cases (e.g., why a particular ID was not retrieved).  
  - Hybrid systems that combine IVF with rule-based filters (e.g., business logic requiring specific IDs to be evaluated).  
- **Performance considerations**:  
  - For large-scale systems, ensure this feature avoids full-index scans.  
  - Consider optional batching for multiple query vectors/IDs.  
- **Compatibility**: This should not break existing APIs but could extend the `IVFIndex` class or add a utility function.  

This feature would enhance flexibility for users who need fine-grained control over vector-distance calculations while maintaining the efficiency guarantees of IVF.

inabao:
close: #1094

## Summary by Sourcery

Add support for calculating distances by ID in the IVF index, including residual and cosine metric handling, update get_location signature for const-correctness, and add a corresponding unit test.

New Features:
- Introduce IVF::CalDistanceById to compute distances for specified IDs using IVF indices

Enhancements:
- Mark get_location as const for improved const-correctness

Tests:
- Add TestCalcDistanceById unit test for the new distance calculation method

### DIFF
--- a/src/algorithm/ivf.cpp
+++ b/src/algorithm/ivf.cpp
@@ -321,7 +321,6 @@ IVF::Build(const DatasetPtr& base) {
     this->Train(base);
     // TODO(LHT): duplicate
     auto result = this->Add(base);
-    this->fill_location_map();
     return result;
 }
 
@@ -410,7 +409,7 @@ IVF::Add(const DatasetPtr& base) {
                 offset_id = bucket_->InsertVector(
                     data_ptr, buckets[idx], idx + current_num * buckets_per_data_);
             }
-            {
+            if (j == 0) {
                 std::lock_guard lock(label_lookup_mutex_);
                 location_map_[i + current_num] =
                     (static_cast<uint64_t>(buckets[idx]) << LOCATION_SPLIT_BIT) |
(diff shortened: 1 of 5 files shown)


wxyucs (approved):
lgtm

LHT129 (approved):
LGTM

Issue #1107, opened 2025-09-02
sindi returns duplicate results


Ningsir:
**Describe the bug**
sindi returns duplicate results.

<img width="1428" height="1012" alt="Image" src="https://github.com/user-attachments/assets/7fff572a-a530-4468-bea1-d2a22647bd56" />

**To Reproduce**
Codes to reproduce the behavior:
```cpp
// paste codes here

```

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS: e.g. `Ubuntu 22.04`
- vsag version: e.g. `v0.11.0` or `7f94a27`
- compiler version: e.g. `GCC 11.4.0`
- interface: `cpp` or `python`

**Expected behavior**

**Screenshots**

**Additional context**

PR #1108, opened 2025-09-02
fix(sindi): fix redundant results in the return data

inabao:
close: #1107
close: #1112
## Summary by Sourcery

Fix redundant result entries in SINDI search by sizing output buffers to the actual number of heap items and preventing zero-distance candidates from being inserted into the sparse term heap.

Bug Fixes:
- Allocate result arrays in search_impl based on min(heap.size(), k) to avoid empty placeholders in return data.
- Guard heap insertion in SparseTermDataCell to skip zero-distance entries and prevent duplicate results.


LHT129 (approved):
LGTM

PR #1118, opened 2025-09-05
refactor(headers): improve header include rules

LHT129:
- less include on .h files but more on .cpp files
- forward declare on header file

## Summary by Sourcery

Streamline header dependencies by removing unnecessary includes from headers and using forward declarations, while adding required includes to corresponding source files; standardize dimension and length parameters from size_t to uint64_t across SIMD and related modules; refactor SparseIndex by relocating constructor, destructor, and serialization logic into the implementation file and cleaning up its header

Enhancements:
- Reduce header file includes and introduce forward declarations to decouple dependencies
- Add missing includes in .cpp files to compensate for header include removals
- Replace std::size_t/size_t with uint64_t for dimension and length parameters consistently across SIMD and other code
- Move SparseIndex method implementations (constructors, destructor, Serialize/Deserialize) from header to source and clean up its header declarations

### DIFF
--- a/src/algorithm/brute_force.cpp
+++ b/src/algorithm/brute_force.cpp
@@ -24,6 +24,8 @@
 #include "data_cell/flatten_interface.h"
 #include "fmt/chrono.h"
 #include "impl/heap/standard_heap.h"
+#include "index/index_common_param.h"
+#include "index_feature_list.h"
 #include "inner_string_params.h"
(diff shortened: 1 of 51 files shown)


wxyucs (approved):
lgtm

ShawnShawnYou (approved):

Issue #1126, opened 2025-09-05
Build: Missing #include <cstdint> in src/common.h


wxyucs:
**Describe the bug**
```
In file included from /work/src/parameter.h:18,
                 from /work/src/algorithm/brute_force_parameter.h:17,
                 from /work/src/algorithm/brute_force_parameter.cpp:16:
/work/src/common.h:48:24: error: ‘uint32_t’ does not name a type
   48 | constexpr static const uint32_t GENERATE_SEARCH_K = 50;
      |                        ^~~~~~~~
/work/src/common.h:19:1: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
   18 | #include "vsag_exception.h"
  +++ |+#include <cstdint>
   19 |
/work/src/common.h:49:24: error: ‘uint32_t’ does not name a type
   49 | constexpr static const uint32_t UPDATE_CHECK_SEARCH_K = 10;
      |                        ^~~~~~~~
/work/src/common.h:49:24: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
/work/src/common.h:50:24: error: ‘uint32_t’ does not name a type
   50 | constexpr static const uint32_t GENERATE_SEARCH_L = 400;
      |                        ^~~~~~~~
/work/src/common.h:50:24: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
/work/src/common.h:51:24: error: ‘uint32_t’ does not name a type
   51 | constexpr static const uint32_t UPDATE_CHECK_SEARCH_L = 100;
      |                        ^~~~~~~~
/work/src/common.h:51:24: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
/work/src/common.h:55:24: error: ‘uint32_t’ does not name a type
   55 | constexpr static const uint32_t ESTIMATE_DOC_TERM = 100;
      |                        ^~~~~~~~
/work/src/common.h:55:24: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
/work/src/common.h:56:24: error: ‘uint32_t’ does not name a type
   56 | constexpr static const uint32_t ESTIMATE_DOC_DIM = 50000;
      |                        ^~~~~~~~
[...]

PR #1127, opened 2025-09-05
fix: add missing <cstdint> include in commen.h

wxyucs:
The file src/common.h uses uint32_t without including the necessary header, causing build failures on stricter compilers like GCC 14.

fixes: #1126

## Summary by Sourcery

Bug Fixes:
- Add <cstdint> include in src/common.h to resolve undefined uint32_t errors

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <cstdint>
+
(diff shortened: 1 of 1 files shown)


inabao (approved):
LGTM

PR #1137, opened 2025-09-10
refactor(kmeans): split kmeans compile

LHT129:
## Summary by Sourcery

Extract kmeans_cluster into a standalone 'cluster' object library and update build configuration and includes to use the new module

Enhancements:
- Extract kmeans_cluster.cpp into a dedicated 'cluster' module separate from the main impl library
- Update source files to include kmeans_cluster.h from the new cluster directory

Build:
- Add cluster subdirectory and CMakeLists to define the 'cluster' object library
- Modify impl CMakeLists to exclude kmeans_cluster.cpp and link against the new 'cluster' target


inabao (approved):
LGTM

PR #1143, opened 2025-09-10
typo(test): replace "[]" with "()" for catch2 test

LHT129:
- catch2 use [] to tag test, so we should not use [Daily] as a common string

## Summary by Sourcery

Tests:
- Use parentheses instead of square brackets for test case titles in catch2 TEST_CASE declarations to avoid tag conflicts in hgraph and ivf tests


inabao (approved):
LGTM

wxyucs (approved):
lgtm

PR #1157, opened 2025-09-15
feat(SINDI): support get memory usage

ShawnShawnYou:
## Summary by Sourcery

Provide a unified GetMemoryUsage method for all indices by defaulting it to the serialized size, remove redundant per-index overrides, and add tests to verify its output stays within expected bounds.

New Features:
- Expose GetMemoryUsage for all index types via default interface implementation

Enhancements:
- Move GetMemoryUsage logic into InnerIndexInterface to return CalSerializeSize instead of throwing
- Remove manual GetMemoryUsage overrides in HGraph and Pyramid

Tests:
- Add tests to check GetMemoryUsage falls within acceptable memory estimation ranges

### DIFF
--- a/src/algorithm/hgraph.h
+++ b/src/algorithm/hgraph.h
@@ -138,11 +138,6 @@ public:
     uint64_t
     EstimateMemory(uint64_t num_elements) const override;
 
-    int64_t
-    GetMemoryUsage() const override {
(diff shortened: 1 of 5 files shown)


inabao (approved):
LGTM

wxyucs (approved):
lgtm

PR #1201, opened 2025-09-28
feat: support use old serial format

wxyucs:


src/index_common_param.cpp
```diff
@@ -116,6 +116,10 @@ IndexCommonParam::CheckAndCreate(JsonType& params, const std::shared_ptr<Resourc
         fill_extra_info_size(result, params[EXTRA_INFO_SIZE]);
     }
 
+    if (params.Contains(PARAMETER_USE_OLD_SERIAL_FORMAT)) {
+        result.use_old_serial_format_ = true;
```
LHT129, line 121: GetBool()?

wxyucs: done

src/algorithm/hgraph.cpp
```diff
@@ -799,39 +800,39 @@ HGraph::Deserialize(StreamReader& reader) {
...
         this->total_count_ = this->basic_flatten_codes_->TotalCount();
 
         if (this->use_attribute_filter_ and this->attr_filter_index_ != nullptr) {
-            this->attr_filter_index_->Deserialize(buffer_reader);
+            this->attr_filter_index_->Deserialize(reader);
         }
     } else {  // create like `else if ( ver in [v0.15, v0.17] )` here if need in the future
         logger::debug("parse with new version format");
 
+        BufferStreamReader buffer_reader(
```
jiaweizone, line 833: Why old format not need `BufferStreamReader` ?

wxyucs: @jiaweizone The OceanBase's reader cannot provide the correct length, so an error will occur in deserialization if here use the BufferStreamReader

PR #1225, opened 2025-10-13
improvement: share vt in hgraph's different layer

LHT129:
- one vt used for one search (refresh in different levels)

## Summary by Sourcery

Share a pooled VisitedList across multiple levels in HGraph searches by adding an optional parameter to search_one_graph and reusing the same instance in SearchWithRequest to reduce pool allocations.

Enhancements:
- Add optional VisitedListPtr parameter to search_one_graph to allow passing and reusing an existing visited list across calls
- Modify search_one_graph to reset provided visited lists instead of always taking a new one from the pool and to return only newly allocated lists
- Update SearchWithRequest to allocate a single visited list for all graph levels and return it once after the bottom graph search

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -476,8 +476,16 @@ DistHeapPtr
 HGraph::search_one_graph(const void* query,
                          const GraphInterfacePtr& graph,
                          const FlattenInterfacePtr& flatten,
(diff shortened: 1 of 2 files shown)


wxyucs (approved):
lgtm

inabao (approved):
LGTM

PR #1239, opened 2025-10-15
refactor: avoiding similar declare in simd

LHT129:
## Summary by Sourcery

Refactor SIMD headers to eliminate repetitive function declarations by introducing macros that generate per-architecture namespaces, and add a Prefetch implementation in the AVX512 backend.

Enhancements:
- Abstract SIMD function prototypes into DECLARE_* macros across multiple headers to remove duplicate namespace declarations for fp32, SQ4, SQ8, basic, bit, PQFS, INT8, normalize, SQ4/SQ8 uniform, BF16, and FP16 modules
- Replace individual per-architecture declarations with macro invocations and undefine macros after use
- Add a Prefetch function implementation in avx512.cpp

### DIFF
--- a/src/simd/avx.cpp
+++ b/src/simd/avx.cpp
@@ -88,6 +88,11 @@ PQDistanceFloat256(const void* single_dim_centers, float single_dim_val, void* r
 #endif
 }
 
+void
+Prefetch(const void* data) {
+    sse::Prefetch(data);
+}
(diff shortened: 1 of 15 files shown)


inabao (approved):
LGTM

ShawnShawnYou (approved):

Issue #1263, opened 2025-10-22
SINDI param CheckCompatibility


cubicc:
**Describe the bug**
here are two related issues within the SINDI::Deserialize function that lead to incorrect behavior when loading a serialized index.
- **Incorrect Parameter Parsing** : When deserializing an index, the stored index parameters ( index_param ) are read as a JSON string, but are then passed directly to the SINDIParameter::FromJson method without being parsed into a JSON object first. It fails to find any parameters, causing all SINDIParameter members to fall back to their default values. 
- **Silent Compatibility Check Failure** : Instead of throwing an exception to halt the process, the code just go on. 
```cpp
    if (not deserialize_without_footer_) {
        auto footer = Footer::Parse(reader);
        auto metadata = footer->GetMetadata();
        JsonType jsonify_basic_info = metadata->Get("basic_info");
        // Check if the index parameter is compatible
        {
            auto param = jsonify_basic_info[INDEX_PARAM];
            SINDIParameterPtr index_param = std::make_shared<SINDIParameter>();
            index_param->FromJson(param);
            this->create_param_ptr_->CheckCompatibility(index_param);
        }
    }

```
**To Reproduce**
Codes to reproduce the behavior:

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS: Debian GNU/Linux 11
- vsag version: v0.17.0-12-gdf134a3-dirty
- compiler version: c++ (Debian 10.2.1-6) 10.2.1 20210110

**Expected behavior**
if parameter mismatch, it should throw an excption like:
```cpp
[2025-10-22 15:49:50.679] [error] SINDI index parameter not match, current: {
    "doc_prune_ratio": 0.0,
    "term_id_limit": 1000000,
    "use_reorder": false,
    "window_size": 100000
}, new: {
    "doc_prune_ratio": 0.30000001192092896,
    "term_id_limit": 10000000,
    "use_reorder": true,
    "window_size": 10000
}

```

**Screenshots**

**Additional context**

PR #1266, opened 2025-10-22
fix(sindi): fix parameter checkCompatibility in deserialize 

cubicc:
fix: #1263

## Summary by Sourcery

Validate compatibility of index parameter during SINDI deserialization and throw an exception on mismatch with a detailed error message.

Bug Fixes:
- Honor the return value of CheckCompatibility and throw an exception if the index parameter is incompatible
- Parse the index parameter with FromString/GetString instead of FromJson to match the updated method signature

Enhancements:
- Log a descriptive error message showing current and new parameter values on mismatch


cubicc:
close #1263

PR #1276, opened 2025-10-24
refactor(ivf): move the residual processing logic in IVF down to the bucket level

inabao:
## Summary by Sourcery

Refactor residual processing logic from the IVF layer into the bucket level by centralizing centroid subtraction, query normalization, and distance adjustments within BucketDataCell and updating related interfaces.

Enhancements:
- Move all residual computations (centroid retrieval, vector subtraction, and inner‐product bias adjustment) into BucketDataCell methods (InsertVector, Train, QueryOneById, ScanBucketById, FactoryComputer).
- Update BucketInterface and BucketDataCell to hold metric and use_residual flags, remove the centroid parameter from InsertVector, and add a SetStrategy method for partition strategies.
- Simplify IVF by removing its residual logic, invoking SetStrategy on the bucket, and delegating Train, Add, Search, and distance calculations entirely to the bucket implementation.
- Introduce a raw_query buffer in Computer to support normalization when computing residuals.

Tests:
- Replace the dynamic search_param template in test_brute_force with a constexpr static string.

Chores:
- Update the 106_index_ivf example to enable use_residual and adjust default parameters.

### DIFF
--- a/src/algorithm/ivf.cpp
+++ b/src/algorithm/ivf.cpp
@@ -241,6 +241,9 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
             FlattenInterface::MakeInstance(param->precise_codes_param, common_param);
     }
     this->use_residual_ = param->bucket_param->use_residual_;
+    if (this->use_residual_) {
+        this->bucket_->SetStrategy(partition_strategy_);
+    }
 
     this->thread_pool_ = common_param.thread_pool_;
     if (param->build_thread_count > 1 and this->thread_pool_ == nullptr) {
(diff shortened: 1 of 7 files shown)


src/algorithm/ivf.cpp
```diff
@@ -307,7 +309,7 @@ IVF::InitFeatures() {
 
     if (name == QUANTIZATION_TYPE_VALUE_FP32 and
         this->bucket_->GetMetricType() != MetricType::METRIC_TYPE_COSINE and
-        not this->use_residual_) {
+        not bucket_->use_residual_) {
         this->index_feature_list_->SetFeature(IndexFeature::SUPPORT_GET_DATA_BY_IDS);
```
LHT129, line 313: use interface rather than use member of bucket

src/algorithm/ivf.cpp
```diff
@@ -240,7 +240,9 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
         this->reorder_codes_ =
             FlattenInterface::MakeInstance(param->precise_codes_param, common_param);
     }
-    this->use_residual_ = param->bucket_param->use_residual_;
+    if (param->bucket_param->use_residual_) {
+        this->bucket_->SetStrategy(partition_strategy_);
```
LHT129, line 244: ditto

src/datacell/bucket_interface.h
```diff
@@ -82,6 +80,11 @@ class BucketInterface {
     virtual void
     MergeOther(const BucketInterfacePtr& other, InnerIdType bias) = 0;
 
+    virtual void
+    SetStrategy(IVFPartitionStrategyPtr strategy) {
```
LHT129: const ref&

src/quantization/computer.h
```diff
@@ -38,7 +39,7 @@ template <typename T>
 class Computer : public ComputerInterface {
 public:
     explicit Computer(const T* quantizer, Allocator* allocator)
-        : quantizer_(quantizer), allocator_(allocator){};
+        : quantizer_(quantizer), allocator_(allocator), raw_query_(allocator){};
```
LHT129, line 43: why use raw_query_, buf_ can be used

src/algorithm/ivf.cpp
inabao: done

src/datacell/bucket_interface.h
inabao: done

src/quantization/computer.h
inabao: raw_query_ store raw query while the buf_ store the result after quantization

LHT129: why not use ref of the query, but use the copy

LHT129 (approved):
LGTM

ShawnShawnYou (approved):

wxyucs (approved):
lgtm

PR #1361, opened 2025-11-24
refactor(engine): add catching for new error types

inabao:
## Summary by Sourcery

Bug Fixes:
- Return a specific NO_ENOUGH_MEMORY error when index creation fails due to std::bad_alloc instead of treating it as a generic unknown error


LHT129 (approved):
LGTM

Issue #1393, opened 2025-11-27
HNSW indices created in version 0.16 cannot be parsed by version 0.14 code.


inabao:
**Describe the bug**
The main issue is that HNSW lacks critical key data during the serialization process.

**Screenshots**

<img width="2592" height="1046" alt="Image" src="https://github.com/user-attachments/assets/5fdf1839-cb9c-43bc-8d11-819ad7a0ae6c" />

PR #1394, opened 2025-11-27
fix(deserialize): fix HNSW in 0.14 cannot parse indexes in 0.16

inabao:
close: #1393

## Summary by Sourcery

Bug Fixes:
- Fix deserialization compatibility so newer empty HNSW and DiskANN indexes can be correctly parsed by older versions of the library.


wxyucs (approved):
lgtm

PR #1403, opened 2025-12-06
feat(hgraph): support deduplication under different quantization

inabao:
## Summary by Sourcery

Refine duplicate vector handling in HGraph search to correctly detect duplicates when using different quantization schemes and align tests with the new behavior.

Enhancements:
- Change duplicate-id detection to compare the query against the nearest candidate via quantized distance instead of relying on raw distance thresholds during traversal.
- Extend inner search parameters to carry the query id when duplicate compression is enabled so search can resolve duplicates consistently across quantizations.

Tests:
- Update HGraph duplicate tests to use shared quantization test cases and skip unsupported bf16 configurations for inner-product and cosine metrics.


wxyucs (approved):
lgtm

ShawnShawnYou (approved):
LGTM

PR #1409, opened 2025-12-09
fix(LocalFileReader): fix thread safety issues in LocalFileReader

inabao:
## Summary by Sourcery

Bug Fixes:
- Prevent data races when lazily initializing the LocalFileReader thread pool by guarding it with a mutex.

### DIFF
--- a/src/factory/factory.cpp
+++ b/src/factory/factory.cpp
@@ -71,7 +71,7 @@ public:
     void
     AsyncRead(uint64_t offset, uint64_t len, void* dest, CallBack callback) override {
         {
-            std::lock_guard<std::mutex> lock(mutex_);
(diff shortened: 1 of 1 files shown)


src/factory/factory.cpp
```diff
@@ -70,8 +70,11 @@ class LocalFileReader : public Reader {
 
     void
     AsyncRead(uint64_t offset, uint64_t len, void* dest, CallBack callback) override {
-        if (not pool_) {
-            pool_ = SafeThreadPool::FactoryDefaultThreadPool();
+        {
+            std::lock_guard<std::mutex> lock(mutex_);
```
wxyucs: use std::scoped_lock instead

inabao: done

PR #1420, opened 2025-12-11
refactor(pyramid): optimize the reorder of the pyramid

inabao:
## Summary by Sourcery

Optimize pyramid index search by decoupling result reordering from base search and tightening range-based result filtering.

Bug Fixes:
- Ensure range search enforces a non-negative radius and slightly expands the effective radius to avoid boundary issues in distance comparisons.
- Adjust range search result trimming to respect both the requested limit and the search radius when discarding candidates.
- Update a range search test expectation to align with the new boundary behavior of results near the top-k threshold.

Enhancements:
- Always search using base codes and apply optional result reordering as a post-processing step when enabled, improving separation between core search and reordering logic.

Tests:
- Relax a range search assertion to tolerate off-by-one differences between k-NN and range search results under the new radius handling.

### DIFF
--- a/src/algorithm/pyramid.cpp
+++ b/src/algorithm/pyramid.cpp
@@ -27,6 +27,8 @@
 #include "utils/util_functions.h"
 namespace vsag {
 
+const static float RADIUS_EPSILON = 1.1F;
+
 std::vector<std::string>
 split(const std::string& str, char delimiter) {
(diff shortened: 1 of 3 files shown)


wxyucs (approved):
lgtm

ShawnShawnYou (approved):
LGTM

PR #1421, opened 2025-12-12
enhance(sindi): improve sindi performance with filter

Ningsir:
## Summary by Sourcery

Optimize sparse term search performance in SINDI by tightening window iteration with filter-aware bounds and introducing a no-prune heap insertion path that leverages full distance buffers.

New Features:
- Add a filter API to expose all valid IDs and propagate it through inner ID wrapper filters.
- Introduce InsertHeapWithoutPrune variants for sparse term data cells supporting both KNN and range search modes with optional filtering.

Enhancements:
- Adjust SINDI search to restrict processed windows based on pre-filter valid ID ranges.
- Switch SINDI heap insertion to choose between pruned and non-pruned paths based on document retain ratio and filter presence.

### DIFF
--- a/src/algorithm/sindi/sindi.cpp
+++ b/src/algorithm/sindi/sindi.cpp
@@ -404,7 +404,8 @@ SINDI::get_min_max_window_id(const FilterPtr& filter) const {
         int64_t max_inner_id = INT64_MIN;
         int64_t id;
         for (int i = 0; i < valid_count; i++) {
-            if (__builtin_expect(label_table_->CheckLabel(valid_ids[i]), 1)) {
+            if (__builtin_expect(static_cast<long>(label_table_->CheckLabel(valid_ids[i])), 1) !=
+                0) {
                 id = label_table_->GetIdByLabel(valid_ids[i]);
                 min_inner_id = std::min(min_inner_id, id);
                 max_inner_id = std::max(max_inner_id, id);


src/data_cell/sparse_term_datacell_test.cpp
```diff
@@ -128,7 +128,7 @@ TEST_CASE("SparseTermDatacell Basic Test", "[ut][SparseTermDatacell]") {
         inner_param.radius = dists[pos];
         MaxHeap heap(allocator.get());
 
-        data_cell->InsertHeap<RANGE_SEARCH, PURE>(dists.data(), computer, heap, inner_param, 0);
```
ShawnShawnYou, line 131: add tests to verify that InsertHeapByTermLists() and InsertHeapByDists() produce identical results

Ningsir: done

src/algorithm/sindi/sindi.cpp
```diff
@@ -158,21 +158,59 @@ SINDI::search_impl(const SparseTermComputerPtr& computer,
...
         term_list->Query(dists.data(), computer);
 
         // insert heap
-        if (inner_param.is_inner_id_allowed) {
-            term_list->InsertHeap<mode, WITH_FILTER>(
-                dists.data(), computer, heap, inner_param, window_start_id);
+        // When using pruning strategy (doc_retain_ratio_ < 1.0), use InsertHeapByTermLists
+        // which can reduce data access by applying term-level pruning during iteration.
+        // Otherwise, use InsertHeapByDists for better performance when all documents are retained.
+        if (doc_retain_ratio_ < 1.0) {
```
inabao: This part needs to be controlled via parameters, as uniformly applying InsertHeapByDists for filtering scenarios may not be optimal.

src/algorithm/sindi/sindi.cpp
```diff
@@ -158,21 +158,59 @@ SINDI::search_impl(const SparseTermComputerPtr& computer,
 
     // window iteration
     std::vector<float> dists(window_size_, 0.0);
-
-    for (auto cur = 0; cur < window_term_list_.size(); cur++) {
+    auto filter = inner_param.is_inner_id_allowed;
+    int64_t min_window_id = 0;
+    int64_t max_window_id = static_cast<int64_t>(window_term_list_.size() - 1);
+    // get min and max window id
+    if (filter) {
```
inabao: Put this code into a function to reduce the complexity of the main function.

Ningsir: move to get_min_max_window_id now

src/algorithm/sindi/sindi.cpp
Ningsir: add param use_term_lists_heap_insert

PR #1439, opened 2025-12-18
feat(platform): Add support for macOS (Darwin)

stuBirdFly:
This MR introduces comprehensive macOS compatibility to the vsag project, enabling successful builds and tests on Apple Silicon (arm64) Macs.

🔧 Key Changes:
Fixed linker issues:
Replaced Linux-specific --whole-archive with macOS-compatible -force_load (or removed where unnecessary).
Resolved duplicate symbol errors from fmt by ensuring consistent static-library usage (removed accidental header-only instantiations).
Resolved missing LAPACKE dependency:
On macOS, linked against the system-provided Accelerate framework instead of liblapacke, which is not available by default.
Updated CMake logic to auto-select Accelerate on APPLE.
Improved compiler detection:
Updated Clang detection to include AppleClang (CMake ID is AppleClang, not Clang).
Ensured correct debug flags (-gdwarf-4) and warning suppressions work on both LLVM Clang and Apple Clang.
Fixed ExternalProject integration:
Adjusted HDF5 and other dependencies to avoid premature include_directories/link_directories that cause “search path not found” warnings on macOS.
Cleaned up duplicate library linking:
Removed redundant static library references (e.g., fmt, roaring, io) to eliminate linker warnings.

PR #1451, opened 2025-12-19
refactor(size_t): use uint64_t instead size_t

LHT129:


LHT129:
implement in another PR : #1439

PR #1464, opened 2025-12-23
refactor(hgraph): modify the default value of factor

inabao:


src/algorithm/hgraph_parameter.h
```diff
@@ -75,7 +75,7 @@ class HGraphSearchParameters : public IndexSearchParameter {
 
 public:
     int64_t ef_search{30};
-    float topk_factor{0.0F};
+    float topk_factor{2.0F};
     bool use_reorder{false};
```
LHT129, line 79: why change the default value. when user ignore this config, we need use "efsearch" as reorder count. Now we use 2 * topk as reorder count? I think we need split the two search parameters, allow topk_factor * topk > efsearch

PR #1550, opened 2026-01-27
fix(hgraph): refactor label remap handling in SetImmutable

### LINKED ISSUE #1555: After setting the index to SetImmutable, CalcDistanceById cannot be called normally.
Image
This code sets the index to an immutable state while releasing memory, causing the hash table to be destroyed. Subsequent ID lookups (CalcDistanceById) will fall back from O(1) hash lookups to O(N) linear traversal scans when the corresponding vector is found based on the ID.
The following is a comparison of the time consumption of CalcDistanceById before and after calling SetImmutable  :

Image

misaka0714:
close #1555 

This pull request makes targeted changes to the `HGraph` class in `hgraph.cpp`, specifically around how label remapping and immutability are handled. The updates ensure that label remapping is consistently managed, even after the graph is set to immutable, and remove unnecessary cleanup logic.

Label remapping and immutability handling:

* Removed the check for `immutable_` before updating `label_remap_` in `deserialize_label_info`, allowing label remapping to occur regardless of immutability status.
* Stopped clearing and disabling `label_remap_` and `use_reverse_map_` in `SetImmutable()`, so label remapping data is now preserved when the graph becomes immutable.

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -1272,9 +1272,7 @@ HGraph::deserialize_label_info(StreamReader& reader) const {
         StreamReader::ReadObj(reader, key);
         InnerIdType value;
         StreamReader::ReadObj(reader, value);
-        if (not this->immutable_) {
-            this->label_table_->label_remap_.emplace(key, value);
-        }
+        this->label_table_->label_remap_.emplace(key, value);
     }
 }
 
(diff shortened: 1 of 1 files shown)


inabao (approved):
LGTM

wxyucs:
@inabao this pull request cannot cherry-pick to the branch 0.17 (CONFLICT), please create a new pull request to the branch 0.17 .

PR #1563, opened 2026-02-05
refactor(duplicate): refactor record array for duplicate

inabao:


src/impl/label_table.cpp
LHT129, line 1: unit test for label table

wxyucs (approved):
lgtm

LHT129 (approved):

PR #1587, opened 2026-02-10
fix(memory_usage): conditionally call cal_memory_usage only on actual capacity changes

### LINKED ISSUE #1582: Build performance rollback for HGRAPH/SINDI 0.18 Version
**Describe the bug**

For cohere 768D1M dataset, 16 thread , hgraph.
Before  build time is 321.03  but switch to 0.18,  build time is 617.40
Sindi is the same.

**To Reproduce**

**Environment**
- OS:  `Linux`
- vsag version: `v0.18.0`
- compiler version: `GCC9`
- interface: `cpp`

**Expected behavior**

**Screenshots**

**Additional context**

LHT129:
fixes: #1582 

Previously cal_memory_usage() was called unconditionally in resize() and Add() methods, even when no actual capacity or window changes occurred. This caused unnecessary memory usage recalculations. Now it's only invoked when:
- resize() actually increases capacity (brute_force, hgraph)
- Add() actually creates new windows (sindi)

### DIFF
--- a/src/algorithm/brute_force.cpp
+++ b/src/algorithm/brute_force.cpp
@@ -621,8 +621,8 @@ BruteForce::resize(uint64_t new_size) {
     if (cur_size < new_size_power_2) {
         this->inner_codes_->Resize(new_size_power_2);
         this->max_capacity_.store(new_size_power_2);
+        this->cal_memory_usage();
     }
-    this->cal_memory_usage();
 }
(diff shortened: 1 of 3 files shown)


inabao (approved):
LGTM

wxyucs (approved):
lgtm

Issue #1601, opened 2026-02-25
Feature: Add WARP Index for Multi-Vector Document Retrieval


inabao:
## Background

In modern information retrieval, ColBERT (Contextualized Late Interaction over BERT) has emerged as an effective approach for document retrieval. Unlike traditional single-vector representations, ColBERT represents each document and query as multiple vectors (one per token), enabling fine-grained matching through "late interaction".

The key similarity metric for ColBERT-style retrieval is **MaxSim** (Maximum Similarity), which computes the sum of maximum similarities between each query vector and all document vectors.

## Feature Description

Add a new index type called **WARP** (Weighted Aggregation of Representative Points) that supports:

1. **Multi-Vector Document Storage**: Each document can contain multiple vectors instead of a single aggregated vector
2. **MaxSim Similarity Search**: Efficient computation of ColBERT-style maxsin similarity
3. **Standard API Compatibility**: Works with existing VSAG index interfaces

### Similarity Metric

For a query with `m` vectors `q_1, q_2, ..., q_m` and a document with `n` vectors `d_1, d_2, ..., d_n`:

```
MaxSim(q, d) = sum_i(max_j(similarity(q_i, d_j)))
```

Where `similarity` can be inner product (IP), cosine similarity, or L2 distance.

### New Dataset API

Add `VectorCounts` field to `Dataset` class to specify the number of vectors per document:

```cpp
auto dataset = vsag::Dataset::Make();
dataset->NumElements(num_docs)
       ->Dim(dim)
       ->Ids(ids.data())
       ->Float32Vectors(datas.data())
       ->VectorCounts(vector_counts.data())  // Number of vectors per document
       ->Owner(false);
```

### Index Parameters

```json
{
    "dtype": "float32",
    "metric_type": "ip",
    "dim": 128
}
```

### Example Usage

```cpp
// Create WARP index
auto index = vsag::Factory::CreateIndex("warp", build_parameters).value();

// Build with multi-vector documents
index->Build(multi_vector_dataset);

// Search with multi-vector query
auto result = index->KnnSearch(multi_vector_query, topk, search_parameters);
```

## Implementation Components

| Component | File | Description |
|-----------|------|-------------|
| Index Implementation | `src/algorithm/warp.cpp` | Core WARP algorithm implementation |
| Index Header | `src/algorithm/warp.h` | WARP class definition |
| Parameter | `src/algorithm/warp_parameter.cpp` | Index parameter handling |
[...]

PR #1602, opened 2026-02-25
feat(warp): add multi-vector interface and brute force  implementation

inabao:
link to #1601


src/factory/engine.cpp
```diff
@@ -84,4 +83,4 @@ Engine::CreateThreadPool(uint32_t num_threads) {
 
 }  // namespace vsag
 
-// NOLINTEND(readability-else-after-return )
+// NOLINTEND(readability-else-after-return )
```
LHT129, line 86: new blank

src/algorithm/warp.h
```diff
@@ -0,0 +1,159 @@
+
```
LHT129, line 1: like sindi ，create new dir

examples/cpp/110_index_warp.cpp
```diff
@@ -0,0 +1,117 @@
...
+        datas.push_back(distrib_real(rng));
+    }
+
+    // Create dataset with VectorCounts for multi-vector support
+    auto base = vsag::Dataset::Make();
+    base->NumElements(num_docs)
+        ->Dim(dim)
+        ->Ids(ids.data())
+        ->Float32Vectors(datas.data())
+        ->VectorCounts(vector_counts.data())  // Specify number of vectors per document
```
LHT129, line 58: I suggest not reuse Float32Vectors, How to find the specific "doc" vector? Why not place the value as the SparseVector

PR #1588, opened 2026-02-10
fix(iter_filter): add bound check for iter filter context

inabao:
close: #1579 
close: #1580

Issue #1580, opened 2026-02-10
HGraph Iterative KnnSearch return failed to init: max size or ef_search is empty


shadowao:
**Describe the bug**
HGraph Iterative KnnSearch  return error : `failed to init: max size or ef_search is empty`

use bottom_graph_->TotalCount() to init IteratorFilterContext, but GetNumElements may be empty when there some add operation concurently

<img width="2542" height="684" alt="Image" src="https://github.com/user-attachments/assets/bc80bd07-d84e-437d-aeb6-8906015e5b59" />

<img width="2318" height="476" alt="Image" src="https://github.com/user-attachments/assets/3f44e7e8-8a18-4bc2-9e5e-5ed419d419dd" />

<img width="2352" height="762" alt="Image" src="https://github.com/user-attachments/assets/93745932-7626-48c5-b2bb-c9ec1d4afab3" />

**To Reproduce**
nop

**Environment**
Please run `bash scripts/check_environment.sh` and paste the output here:
- OS:  `Linux`
- vsag version: `v0.18.0`
- compiler version: e.g. `GCC9`
- interface: `cpp`

**Expected behavior**

**Screenshots**

**Additional context**

Issue #1579, opened 2026-02-10
[HNSW] IteratorFilterContext uses stale max_elements during concurrent add/search, can cause out-of-bounds access


stuBirdFly:
Problem
When running add and search concurrently, IteratorFilterContext is initialized using alg_hnsw_->getMaxElements() outside of any resize_mutex_ protection. If add triggers resizeIndex concurrently, max_elements_ increases, but the IteratorFilterContext bitset is sized with the old value. Later during search, SetPoint(inner_id) can write past the allocated bitset.
Code path
HNSW::knn_search initializes IteratorFilterContext before entering searchKnn:
if (iter_ctx != nullptr && *iter_ctx == nullptr) {    auto* filter_context = new IteratorFilterContext();    filter_context->init(alg_hnsw_->getMaxElements(), params.ef_search, search_allocator);    *iter_ctx = filter_context;}
IteratorFilterContext::SetPoint directly uses inner_id to index the bitset:
list_[byte_pos(inner_id)] |= (1 << bit_pos(inner_id));
Why this is a bug
getMaxElements() can change during concurrent add (via resizeIndex). The bitset is allocated with an old size, but search may encounter newer ids, causing out-of-bounds access.
Impact
Potential memory corruption / crashes during concurrent add + search, especially under load.
Repro idea
Run high-concurrency stress test with add (large enough to trigger resize) and search with iter_ctx enabled. Under contention, search can hit new internal ids and write past the bitset.
Suggested fix
Initialize IteratorFilterContext under a consistent snapshot of max_elements_, or re-check/resize it inside searchKnn after acquiring resize_mutex_. Alternatively, prevent add from running concurrently with search.

PR #1611, opened 2026-02-27
fix(iter_filter): add bound check for iter filter context (#1588)

inabao:
cp #1588 to 0.17
link #1579
link #1580


LHT129 (approved):
LGTM

PR #1613, opened 2026-02-27
feat(index): add ShrinkAndRepair API with timeout

LHT129:
Introduce ShrinkAndRepair(double timeout_ms = ...) to the public index API and switch the return to tl::expected<void, Error> for structured error handling.


inabao (approved):
LGTM

Issue #1643, opened 2026-03-10
Hgraph core at vsag::avx512::FP32ComputeIP(float const*, float const*, unsigned long)


shadowao:
**Describe the bug**
Hgraph core at `vsag::avx512::FP32ComputeIP(float const*, float const*, unsigned long)`

```
Thread 1 (Thread 0x7fbc91d79240 (LWP 3113334)):
#0  0x00007fbd0afc2055 in raise () from /usr/lib64/libpthread.so.0
#1  0x000055bcd4ddb11b in oceanbase::common::coredump_cb (sig=11, sig_code=1, sig_addr=0x0, context=<optimized out>) at ./deps/oblib/src/lib/signal/ob_signal_handlers.cpp:267
#2  <signal handler called>
#3  0x000055bcd8c15f59 in vsag::avx512::FP32ComputeIP(float const*, float const*, unsigned long) ()
#4  0x000055bcd8f1a4ea in vsag::FP32Quantizer<(vsag::MetricType)1>::ComputeImpl(unsigned char const*, unsigned char const*) ()
#5  0x000055bcd8ccfbec in vsag::FlattenDataCell<vsag::FP32Quantizer<(vsag::MetricType)1>, vsag::MemoryBlockIO>::ComputePairVectors(unsigned int, unsigned int) ()
#6  0x000055bcd8febbd2 in vsag::mutually_connect_new_element(unsigned int, std::shared_ptr<vsag::DistanceHeap> const&, std::shared_ptr<vsag::GraphInterface> const&, std::shared_ptr<vsag::FlattenInterface> const&, std::shared_ptr<vsag::MutexArray> const&, vsag::Allocator*, float) ()
#7  0x000055bcd8fbbc62 in vsag::HGraph::graph_add_one(void const*, int, unsigned int) ()
#8  0x000055bcd8fbc1b4 in vsag::HGraph::add_one_point(void const*, int, unsigned int) ()
#9  0x000055bcd8fc7108 in vsag::HGraph::Add(std::shared_ptr<vsag::Dataset> const&) ()
#10 0x000055bcd8b6e2e8 in vsag::IndexImpl<vsag::HGraph>::Add(std::shared_ptr<vsag::Dataset> const&) ()
#11 0x000055bcd69f70d2 in oceanbase::common::obvsag::HnswIndexHandler::add_index (this=0xa01bfc08040, incremental=...) at ./deps/oblib/src/lib/vector/ob_vsag_adaptor.cpp:290
[...]

PR #1650, opened 2026-03-10
fix(datacell): update max_capacity after io operation completes

inabao:
cp #1642 to 0.15
link: #1643

Move max_capacity assignment after io->Write() to ensure data is persisted before updating capacity.


wxyucs (approved):
lgtm

LHT129 (approved):
LGTM

PR #1712, opened 2026-03-18
fix: use quoted project includes in search_param.h

wxyucs:
## Summary
- align `include/vsag/search_param.h` with the rest of the public headers by switching internal includes from angle brackets to quoted includes
- avoid potential build issues in stricter include path configurations while keeping the API unchanged

## Testing
- `make debug`


inabao (approved):
LGTM

PR #1713, opened 2026-03-18
perf(pyramid): trim flat search path overhead

wxyucs:
## Summary
- remove redundant visited-list marking from the Pyramid FLAT search path because brute-force search does not revisit neighbors
- mark the small Pyramid search heuristic helpers inline to reduce hot-path call overhead

## Testing
- make debug
- build/tests/unittests "[pyramid]"
- build/tests/unittests

### DIFF
--- a/src/algorithm/pyramid.cpp
+++ b/src/algorithm/pyramid.cpp
@@ -39,7 +39,7 @@ split(const std::string& str, char delimiter) {
     return vec;
 }
 
-static uint64_t
+static inline uint64_t
 get_suitable_max_degree(int64_t data_num) {
     if (data_num < 100'000) {
         return 16;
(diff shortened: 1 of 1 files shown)


src/algorithm/pyramid.cpp
```diff
@@ -836,10 +836,6 @@ Pyramid::search_node(const IndexNode* node,
         auto computer = codes->FactoryComputer(query->GetFloat32Vectors());
         codes->Query(dists.data(), computer, ids_ptr, id_count);
 
-        for (const auto& id : node->ids_) {
-            vl->Set(id);
-        }
-
         for (int i = 0; i < id_count; ++i) {
             results->Push(dists[i], ids_ptr[i]);
             if (results->Size() > search_param.ef) {
```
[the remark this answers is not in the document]
wxyucs: Good catch. I restored the FLAT-path visited marking in 85aeef1e so traversal semantics stay unchanged, and kept the helper-function inlining optimization from this PR.

src/algorithm/pyramid.cpp
inabao: This is a redundant tag.

wxyucs: Removed the redundant `inline` specifiers in ff6edc15 and kept the rest of the Pyramid search-path behavior unchanged.

PR #1745, opened 2026-03-24
fix(diskann): normalize data for COSINE metric in PQ quantization

inabao:
- Add should_normalize parameter to generate_pq_data_from_pivots to normalize data during PQ code generation when using COSINE metric
- Normalize query vectors in cal_distance_by_ids for COSINE metric to ensure correct distance calculation against normalized base vectors
- Add cosine metric to DiskANN test cases


LHT129 (approved):
LGTM

PR #1773, opened 2026-03-30
perf(build): skip boost library compilation, only copy headers

inabao:
VSAG only uses boost::dynamic_bitset which is header-only. DiskANN core library uses dynamic_bitset, apps directory is not built. All boost libraries are unnecessary for compilation.

This reduces boost build time from 10-30 minutes to seconds.


LHT129 (approved):
LGTM

PR #1776, opened 2026-03-30
fix: make VsagException public inherit from std::exception

### LINKED ISSUE #1777: VsagException uses private inheritance from std::exception
## Bug Description

The `VsagException` class in `src/vsag_exception.h` uses private inheritance from `std::exception`:

```cpp
class VsagException : std::exception {  // private inheritance (class default)
```

## Problem

Private inheritance causes the following issues:

1. **Cannot catch with std::exception handler**: Users cannot catch `VsagException` using the standard pattern `catch (std::exception& e)`, which is the typical way to handle exceptions in C++.

2. **Breaks exception handling expectations**: When code throws `VsagException`, it propagates up the call stack but cannot be caught by generic `std::exception` handlers, leading to uncaught exceptions and program termination.

3. **Observed behavior**: As seen in production crash logs, when `VsagException` is thrown (e.g., from `ReaderIO::ReadImpl`), it may not be caught properly by exception handlers expecting `std::exception`:

```
terminate called after throwing an instance of 'vsag::VsagException'
*** SIGABRT received
```

## Expected Behavior

`VsagException` should be catchable via `std::exception&`, following standard C++ exception handling conventions.

## Solution

Change inheritance to public:

```cpp
class VsagException : public std::exception {
```

## Impact

- No breaking changes to existing code
- Enables standard exception handling patterns
- All existing `catch (VsagException& e)` handlers continue to work

inabao:
## Summary

- Fix VsagException to use public inheritance from std::exception instead of private inheritance
- This allows VsagException to be caught by `catch (std::exception& e)` handlers, following standard C++ exception handling patterns

## Problem

The class was using private inheritance (default for class):
```cpp
class VsagException : std::exception {  // private inheritance
```

This prevented catching VsagException with std::exception handler, breaking standard exception handling.

## Solution

Changed to public inheritance:
```cpp
class VsagException : public std::exception {
```

## Impact

- No breaking changes - all existing code behavior unchanged
- Adds ability to catch VsagException via std::exception handler
- Follows C++ standard practice for exception classes

Fixes #1777

### DIFF
--- a/src/factory/engine.cpp
+++ b/src/factory/engine.cpp
@@ -59,11 +59,11 @@ Engine::CreateIndex(const std::string& origin_name, const std::string& parameter
     } catch (const std::bad_alloc& e) {
         LOG_ERROR_AND_RETURNS(
             ErrorType::NO_ENOUGH_MEMORY, "failed to create index(not enough memory): ", e.what());
+    } catch (const vsag::VsagException& e) {
+        LOG_ERROR_AND_RETURNS(e.error_.type, "failed to create index: " + e.error_.message);
     } catch (const std::exception& e) {
         LOG_ERROR_AND_RETURNS(
             ErrorType::UNSUPPORTED_INDEX, "failed to create index(unknown error): ", e.what());
-    } catch (const vsag::VsagException& e) {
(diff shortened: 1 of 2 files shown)


LHT129 (approved):
LGTM

PR #1802, opened 2026-04-07
fix(hnsw): synchronize parallel add graph updates

inabao:
## Summary
- make neighbor-side updates in `HierarchicalNSW::mutuallyConnectNewElement()` atomic so concurrent inserts do not overwrite each other from stale snapshots
- keep a newly inserted point hidden until its link lists are initialized, and serialize `getRandomLevel()` access to avoid shared RNG races
- align `addPoint()` lock ordering with other read paths so TSAN no longer reports a lock-order inversion with `max_level_mutex_`

## Test Plan
- `make fmt`
- `./build-release/examples/cpp/322_feature_hnsw_concurrent_add_recall`
- `TSAN_OPTIONS="halt_on_error=1 second_deadlock_stack=1 history_size=7" ./build/examples/cpp/322_feature_hnsw_concurrent_add_recall`

Closes #1801


LHT129 (approved):
LHTM

PR #1823, opened 2026-04-13
[Backport 0.18] chore: change default value of ENABLE_INTEL_MKL to OFF and trigger Python Build X86 on CMake changes (#1806)

wxyucs:
## Summary
- backport the `#1806` CI/build adjustments to `0.18`, including changing the default `ENABLE_INTEL_MKL` value to `OFF` in the branch's `CMakeLists.txt` and `Makefile`
- explicitly enable MKL in the x86 ASan workflows so sanitizer jobs keep using MKL after the default flip, while preserving the `0.18` workflow layout
- remove the broken CentOS MKL/libomp package setup from the Python wheel config and dependency script, and carry over the related developer documentation update

## Notes
- `0.18` does not have the same `pr-ci.yml` and `cmake/VSAGOptions.cmake` structure as `main`, so this backport applies the equivalent behavior in the branch's existing files instead of introducing those `main`-only files


LHT129 (approved):
LGTM

Issue #1842, opened 2026-04-15
nullptr crash when init IteratorFilterContext fail in HNSW


shadowao:
**Describe the bug**
During the initialization process of IteratorFilterContext (https://github.com/antgroup/vsag/blob/main/src/index/iterator_filter.cpp#L43), if the memory allocation fails, the return value is not checked here (https://github.com/antgroup/vsag/blob/main/src/index/hnsw.cpp#L272) , resulting in the error being ignored and the execution continuing, ultimately leading to a crash due to accessing a null pointer.

```
vsag::IteratorFilterContext::init(unsigned int, long, vsag::Allocator*) [clone .cold] at xpointer.c:?
vsag::HNSW::knn_search(std::shared_ptr<vsag::Dataset> const&, long, std::string const&, std::shared_ptr<vsag::Filter> const&, vsag::Allocator*, vsag::IteratorContext**, bool) const at ??:?
vsag::HNSW::KnnSearch(std::shared_ptr<vsag::Dataset> const&, long, vsag::SearchParam&) const at ??:?
tl::expected<std::shared_ptr<vsag::Dataset>, vsag::Error>::has_value() const at ./build_sanity/deps/oblib/src/lib/./deps/3rd/usr/local/oceanbase/deps/devel/include/vsag/expected.hpp:1981
oceanbase::common::obvsag::knn_search(void*&, float*, int, long, float const*&, long const*&, long&, int, bool, char const*&, void*, bool, bool, float, void*&, bool, void*, float) at ./build_sanity/deps/oblib/src/lib/./deps/oblib/src/lib/vector/ob_vsag_adaptor.cpp:1305 (discriminator 2)
oceanbase::common::obvectorutil::knn_search(void*, float*, int, long, float const*&, long const*&, char const*&, long&, int, void*, bool, bool, float, void*, bool, void*&, bool, float) at ./build_sanity/deps/oblib/src/lib/./deps/oblib/src/lib/vector/ob_vector_util.cpp:278
oceanbase::share::ObVectorIndexSegQueryHandler::knn_search(oceanbase::share::ObVsagQueryResult&, int) at ./build_sanity/src/share/./src/share/vector_index/ob_vector_index_segment.cpp:2166 (discriminator 4)
[...]

PR #1844, opened 2026-04-16
fix: propagate iterator filter init failures in hnsw

wxyucs:
## Summary
- handle `IteratorFilterContext::init()` failures in `HNSW::knn_search()` instead of continuing with a null context
- use RAII for the temporary iterator context during initialization and preserve the original allocation error
- add a targeted regression test that forces iterator filter initialization to fail and verifies `NO_ENOUGH_MEMORY` is returned

## Testing
- `./build/tests/unittests \"iterator filter init allocation failure\"`

#1842


LHT129 (approved):
LGTM

PR #1871, opened 2026-04-20
refactor(simd): unify SIMD dispatch cascades via shared macros

### LINKED ISSUE #1870: [Refactor] Unify duplicated SIMD dispatch cascades in src/simd
## Description

The SIMD dispatch code under `src/simd/` contains ~47 nearly-identical
cascades for selecting the best available implementation of each
function (avx512 / avx2 / avx / sse / sve / neon / generic). The same
~30-line `if (Support<ISA>()) { #if defined(ENABLE_<ISA>) return ...; #endif }`
pattern is copy-pasted across 13 files, adding up to well over a
thousand lines of boilerplate.

## Background

Every time a new SIMD-backed function is added, the same seven-way
cascade has to be written again by hand. The pattern is error-prone
(easy to forget a fallback, mis-order clauses, or skip an `#ifdef`)
and obscures the few places where the dispatch actually differs
(e.g. the AVX512VPOPCNTDQ variant used by rabitq binary IP, or the
SSE/SVE/NEON-only Prefetch).

## Requirements

- [ ] Introduce shared macros that encapsulate the standard dispatch
      cascade.
- [ ] Rewrite the standard cascades in `src/simd/*_simd.cpp` to use the
      macros, keeping behavior byte-identical.
- [ ] Keep the genuinely non-standard cases as explicit code so their
      uniqueness is visible at the call site.

## Technical Details

Two critical properties of the original code must be preserved:

1. Symbols from an ISA that is not enabled at compile time must not
   appear in the translation unit.
2. When `SimdStatus::Support<ISA>()` returns true at runtime but
   `ENABLE_<ISA>` is not set at compile time, control must fall
   through to the next clause and ultimately to `generic::`.

## Acceptance Criteria

- [ ] `make release` and `make debug` both succeed.
- [ ] Exported symbols of `libvsag.so` are byte-identical before and
      after the change (verified via `nm -D --defined-only | sort` diff).
- [ ] `make fmt` / `make lint` pass.
- [ ] Net reduction in source lines of ~1000+.

## Related

- Original task file (local): tasks/in_progress/2026-03-13-统一-simd-分发代码消除重复.md

## Notes

Non-standard dispatch functions (e.g. `NormalizeWithCentroid`,
`InverseNormalizeWithCentroid`, `RaBitQFloatSQIP`, `FlipSign`,
`RaBitQSQ4UBinaryIP`, `Prefetch`) are intentionally left outside of the
generic macro to avoid hiding their unique ISA support matrices behind
a misleading abstraction.

LHT129:
## Summary

Remove ~1430 lines of duplicated SIMD-selection code from `src/simd/`
by introducing three small macros that capture the standard dispatch
patterns. Behavior is byte-identical: the exported symbol table of
`libvsag.so` is unchanged.

Fixes #1870

## Changes

- Add `src/simd/simd_dispatch.h` with three macros:
  - `VSAG_DEFINE_SIMD_DISPATCH` - standard seven-way cascade
    (avx512 / avx2 / avx / sse / sve / neon / generic).
  - `VSAG_DEFINE_SIMD_DISPATCH_VPOPCNTDQ` - variant that prefers
    AVX512VPOPCNTDQ over AVX512, used by rabitq binary inner-product.
    AVX2/AVX/SSE are intentionally skipped (no implementations exist).
  - `VSAG_DEFINE_SIMD_DISPATCH_PREFETCH` - narrow dispatch for
    `Prefetch`, which only has SSE/SVE/NEON variants.
- Rewrite the 13 dispatch source files (`basic_func.cpp`,
  `bf16_simd.cpp`, `bit_simd.cpp`, `fp16_simd.cpp`,
  `fp32_simd.cpp`, `int8_simd.cpp`, `normalize.cpp`,
  `pqfs_simd.cpp`, `rabitq_simd.cpp`, `sq4_simd.cpp`,
  `sq4_uniform_simd.cpp`, `sq8_simd.cpp`, `sq8_uniform_simd.cpp`)
  so each dispatch is registered with a single macro call.
- Keep genuinely non-standard cases as explicit code rather than
  introducing one-off macros:
  - `NormalizeWithCentroid` / `InverseNormalizeWithCentroid` -
    generic-only.
  - `RaBitQFloatSQIP` - generic-only.
  - `FlipSign` - AVX512 / SVE / NEON only (no AVX2/AVX/SSE).

## Design Notes

Two critical properties of the original hand-written code are
preserved:

1. **No references to disabled ISAs.** Each `VSAG_SIMD_DISPATCH_BODY_<ISA>`
   expands to the `return` statement only when `ENABLE_<ISA>` is
   defined, and to nothing otherwise. A build without, say,
   `ENABLE_AVX2` therefore never mentions `avx2::Name` in the
   translation unit.
2. **Runtime-check / compile-flag mismatch falls through.** When
   `SimdStatus::Support<ISA>()` is true at runtime but `ENABLE_<ISA>`
   is not defined at compile time, the body macro is empty, control
   falls to the next `if`, and ultimately reaches
   `return generic::FnName;`. This matches the previous hand-written
   behavior exactly.

## Verification

- `make release` and `make debug` both succeed on x86_64.
- `make fmt` passes (clang-format-15).
- `nm -D --defined-only build-release/src/libvsag.so | awk '{print $3}' | sort`
  is **byte-identical** before and after this change (0-line diff
  against the `main` baseline at `d7ae0283`). All 52054 exported
  symbols match.

## Net Change

- +76 / -1434 lines in `src/simd/*.cpp`
- +177 lines in the new `src/simd/simd_dispatch.h`
- **Net reduction: ~1181 lines**


wxyucs (approved):
lgtm

PR #1878, opened 2026-04-20
fix(sindi): bypass buffered reads for footerless deserialize

wxyucs:
## Summary
- bypass `BufferStreamReader` when SINDI deserializes legacy footerless data
- add a regression test covering `deserialize_without_footer=true` with the default buffer setting

## Testing
- `./tests/functests \"SINDI Serialize File\"`
- `./tests/functests \"SINDI Deserialize Without Footer Defaults To NonBuffered Read\"`


LHT129 (approved):
LGTM

PR #1906, opened 2026-04-21
chore: make fmt and lint compatible with macOS (Homebrew llvm@15)

### LINKED ISSUE #1905: chore: make fmt and lint compatible with macOS (Homebrew llvm@15)
## Problem

`make fmt` and `make lint` fail on macOS for two reasons:

### 1. Toolchain binary naming

On macOS, Homebrew `llvm@15` installs tools **without** the `-15` suffix (e.g. `clang-format` instead of `clang-format-15`). The scripts `scripts/format/format-cpp.sh` and `scripts/linters/run-clang-tidy-15.sh` hard-code the Linux `apt`-style suffixed binary names, causing an immediate `command not found` error.

### 2. `grep -oP` not supported on macOS

Both scripts use `grep -oP` (Perl-compatible regex) to extract the version number. macOS ships BSD grep which does not support `-P`, so the version check always fails.

### 3. `std::min` type deduction failure on macOS ARM

On macOS ARM (`unsigned long long` vs `unsigned long` for `size_type`), `std::min` cannot deduce a common type between `uint64_t` and `container::size_type`, causing a compile error reported by `clang-tidy`:

```
src/impl/label_table.h:280:21: error: no matching function for call to 'min' [clang-diagnostic-error]
        auto size = std::min(static_cast<uint64_t>(max_count), deleted_ids_.size());
```

## Solution

- Add Homebrew fallback paths (`/opt/homebrew/...` for Apple Silicon, `/usr/local/opt/...` for Intel) in both scripts.
- Replace `grep -oP` with POSIX-compatible `sed` for version extraction.
- Change `std::min(...)` to `std::min<uint64_t>(...)` with an explicit template argument (consistent with other call sites in the codebase).

## Environment

- macOS (Apple Silicon / ARM)
- Homebrew llvm@15 (`brew install llvm@15`)

jac0626:
## What

Fix `make fmt` and `make lint` failures on macOS caused by three separate issues.

Closes #1905

### 1. Toolchain binary naming
Homebrew `llvm@15` installs binaries **without** a version suffix (`clang-format`, not `clang-format-15`). The scripts only looked for the suffixed names (Linux `apt` convention), causing immediate `command not found` errors on macOS.

### 2. `grep -oP` incompatibility
Both scripts used `grep -oP` (Perl-compatible regex) for version extraction. macOS ships BSD grep which does not support `-P`.

### 3. `std::min` type deduction failure on macOS ARM
On Apple Silicon, `uint64_t` resolves to `unsigned long long` while container `size_type` is `unsigned long`. `std::min` cannot deduce a common type, producing a `clang-diagnostic-error` in `label_table.h`.

## How

| File | Change |
|------|--------|
| `scripts/format/format-cpp.sh` | Add Homebrew fallback paths for Apple Silicon (`/opt/homebrew`) and Intel (`/usr/local/opt`); replace `grep -oP` with portable `sed` |
| `scripts/linters/run-clang-tidy-15.sh` | Same fixes for `clang-tidy` and `clang-apply-replacements` |
| `src/impl/label_table.h` | `std::min` → `std::min<uint64_t>` (explicit template arg, consistent with all other call sites) |

## Test

Verified `make fmt` and `make lint` pass on macOS 14 (Apple Silicon, Homebrew llvm@15).

### DIFF
--- a/scripts/format/format-cpp.sh
+++ b/scripts/format/format-cpp.sh
@@ -5,15 +5,35 @@
 
 REQUIRED_VERSION="15"
 
-# Check if clang-format-15 is available
-if ! command -v clang-format-15 &> /dev/null; then
-    echo "ERROR: clang-format-15 is not installed!"
-    echo "Please install it with: sudo apt-get install clang-format-15"
+# Find clang-format-15 binary: check the suffixed name first, then use
+# 'brew --prefix' for dynamic resolution on macOS, and finally fall back
+# to stable Homebrew opt symlinks.
+CLANG_FORMAT=""
+if command -v clang-format-15 &> /dev/null; then
+    CLANG_FORMAT="clang-format-15"
+elif command -v brew &> /dev/null; then
+    BREW_LLVM_PREFIX=$(brew --prefix llvm@15 2>/dev/null || true)
+    if [ -n "$BREW_LLVM_PREFIX" ] && [ -x "$BREW_LLVM_PREFIX/bin/clang-format" ]; then
+        CLANG_FORMAT="$BREW_LLVM_PREFIX/bin/clang-format"
+    fi
+fi
+if [ -z "$CLANG_FORMAT" ] && [ -x "/opt/homebrew/opt/llvm@15/bin/clang-format" ]; then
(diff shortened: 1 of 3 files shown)


wxyucs (approved):
lgtm

PR #1932, opened 2026-04-25
fix: fallback async_io param and warn when libaio is unavailable

### LINKED ISSUE #1931: async_io fallback should also fallback parameter type and warn when libaio is unavailable
## Background
When libaio is unavailable, AsyncIO implementation falls back to BufferIO at compile time.

## Problem
The fallback was inconsistent in parameter parsing:
- runtime IO implementation fell back, but
- async_io parameter parsing could still produce async-specific parameter semantics silently.

This can lead to confusing behavior and harder troubleshooting.

## Expected Behavior
When parsing io parameters with type async_io and libaio is unavailable:
1. Parameter object should fallback to BufferIOParameter.
2. Serialized parameter type should reflect buffer_io.
3. A warning should be logged to make the fallback observable.

## Proposed Fix
- Add conditional fallback in IOParameter::GetIOParameterByJson for async_io.
- Add one-time warning log for fallback path.
- Add regression assertions in AsyncIO tests for both HAVE_LIBAIO branches.

jac0626:
## Summary
- fallback async_io parameter parsing to BufferIOParameter when libaio is unavailable
- add a one-time warning log to make the async_io -> buffer_io fallback visible
- add regression assertions to cover both HAVE_LIBAIO branches in AsyncIO parameter test

## Why
Without this, fallback behavior can be partially silent and confusing in environments that compile without libaio support.

Fixes #1931

### DIFF
--- a/src/io/async_io_test.cpp
+++ b/src/io/async_io_test.cpp
@@ -17,7 +17,9 @@
 
 #include <memory>
 
+#include "async_io_parameter.h"
 #include "basic_io_test.h"
+#include "buffer_io_parameter.h"
 #include "impl/allocator/safe_allocator.h"
 #include "unittest.h"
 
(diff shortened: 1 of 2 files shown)


wxyucs (approved):
lgtm

PR #1934, opened 2026-04-26
docs: reorganize documentation navigation

wxyucs:
## Summary
- reorder quick-start entries so index creation appears before search
- group parameter reference with index docs
- add a dedicated performance and tuning section and align English/Chinese resource entries

## Testing
- mdbook build docs/docs/zh -d /tmp/vsag-docs-zh
- mdbook build docs/docs/en -d /tmp/vsag-docs-en


LHT129 (approved):
LGTM

Issue #1950, opened 2026-04-29
Track macOS CI test failures exposed during platform support


jac0626:
While adding macOS CI/build support in PR #1930, the macOS build jobs pass, but ASan-based macOS test jobs expose several test/runtime issues that should be fixed independently from the CI wiring change.

Related workflow run:

- Test macOS Functests: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256867?pr=1930
- Test macOS Unittests: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256873?pr=1930

Current findings:

- Dataset DeepCopy path test is using `memcmp` on `std::string` objects. This compares implementation-specific object bytes instead of copy semantics. It fails on macOS/libc++ when equal short strings have identical object representation.
- HNSW feedback functest assumes the first search round must produce at least one error (`error_fix > 0`). On macOS/arm64/ASan the first round can already reach full recall, so the test fails even though there is no feedback error to fix.
- ASan/UBSan logs also report existing runtime undefined behavior, including HNSW unaligned access, quantizer NaN-to-uint8 conversion, and null-pointer offset in tests. These are real follow-up risks but should be handled in separate focused issues/PRs.

Planned sub-issues:

- [x] #1951
- [x] #1953
- [ ] #1960
- [x] #1966

Scope note: this parent issue tracks test/runtime problems discovered while enabling macOS support. Each fix should stay small and reviewable, and should not be mixed into the macOS CI/build PR unless it only changes CI/build plumbing.

jac0626:
done

PR #1952, opened 2026-04-29
fix: avoid memcmp on dataset path strings

### LINKED ISSUE #1951: Fix Dataset DeepCopy path test semantics on macOS
Parent tracking issue: #1950

The macOS ASan unittest job fails in `Dataset Copy and Append Test / Deep Copy`:

- Job: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256873?pr=1930
- Failure: `src/dataset_impl_test.cpp:280`
- Assertion: `REQUIRE(AreAllPointersDifferent(original->GetPaths(), copy->GetPaths(), num_elements))`

Root cause:

`AreAllPointersDifferent` currently uses `std::memcmp` on each `std::string` object. That compares implementation-specific object bytes, not whether `DatasetImpl::DeepCopy` created an independent path array. On macOS/libc++, equal short strings can have identical object representation, so the test reports a false failure. Linux/libstdc++ may not expose this because the object layout differs.

Suggested fix:

- Keep the fix test-only.
- For `Paths`, assert that the source and copy arrays are different pointers.
- Assert each copied string has the same value.
- Do not use `memcmp` on `std::string` objects.

Non-goals:

- Do not change `DatasetImpl::DeepCopy` behavior unless investigation shows a real ownership bug.
- Do not mix sanitizer UB fixes into this PR.

jac0626:
## Summary

Fixes #1951. Part of #1950.

This keeps the Dataset DeepCopy fix test-only. The previous assertion used `memcmp` on `std::string` objects to infer whether `Paths` had been deep-copied. That compares implementation-specific object bytes, which can be identical for equal short strings on macOS/libc++, causing a false failure.

The updated assertion checks the intended semantics directly: the copied path array must be a different array pointer, and every copied string value must match the original.

## Validation

- `/opt/homebrew/opt/llvm@15/bin/clang-format -i src/dataset_impl_test.cpp`
- `git diff --check`
- `./build/tests/unittests "Dataset Copy and Append Test"`

Note: rebuilding from a clean `upstream/main` checkout on local macOS is blocked by the existing AppleClang/libc++ configuration issue being addressed separately in PR #1930, so the local executable run used the available build artifact. Full PR CI should validate the change on the current supported Linux path; macOS CI validation depends on the macOS support PR.

### DIFF
--- a/src/dataset_impl_test.cpp
+++ b/src/dataset_impl_test.cpp
@@ -250,6 +250,21 @@ AreAllPointersDifferent(T* original, T* copy, uint64_t num_elements) {
     return true;
 }
 
+bool
+ArePathArraysDeepCopied(const std::string* original,
+                        const std::string* copy,
+                        uint64_t num_elements) {
+    if (original == copy) {
+        return false;
+    }
+    for (uint64_t i = 0; i < num_elements; ++i) {
+        if (original[i] != copy[i]) {
+            return false;
+        }
+    }
+    return true;
+}
+
 TEST_CASE("Dataset Copy and Append Test", "[ut][Dataset]") {
     std::random_device rd;
     std::mt19937 gen(rd());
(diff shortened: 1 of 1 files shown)


wxyucs (approved):
lgtm

PR #1954, opened 2026-04-29
test: stabilize hnsw feedback functest

### LINKED ISSUE #1953: Stabilize HNSW feedback functest when initial recall is already perfect
Parent tracking issue: #1950

The macOS ASan functest job fails in `int8 + freshhnsw + feedback + update`:

- Job: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256867?pr=1930
- Failure: `tests/test_index_old.cpp:1108`
- Assertion: `REQUIRE(error_fix > 0)`
- Observed expansion: `0 > 0`
- Logged recall before the failure: `Recall: 1.0000`

Root cause:

The test assumes that the first search round must produce at least one wrong top-1 result, so feedback must fix at least one error. That assumption depends on random data, platform behavior, compiler/runtime differences, and sanitizer configuration. On macOS/arm64/ASan, the first round can already reach perfect recall, so there is no error to fix and `error_fix` remains zero. This makes the test fail even though feedback did not regress correctness.

Suggested fix:

- Keep the fix test-only.
- Do not require `error_fix > 0` unconditionally.
- Preserve the intended feedback/update coverage by checking stable semantics, for example:
  - post-feedback/update recall should not regress from the first round;
  - if the first round was imperfect, feedback should improve or reach expected recall;
  - if the first round was already perfect, the test should pass without requiring artificial errors.

Non-goals:

- Do not change HNSW feedback implementation unless investigation finds a real behavior bug.
- Do not mix UBSan unaligned-access fixes into this PR.

jac0626:
## Summary

Fixes #1953. Part of #1950.

The HNSW feedback functests assumed the first search round must produce at least one wrong result, then required `error_fix > 0` and strict recall improvement. On macOS/arm64/ASan the first round can already reach perfect recall, so there is no error to fix and the test fails even though correctness did not regress.

This updates the assertions to keep the intended feedback coverage when initial recall is imperfect, while allowing the already-perfect case to pass. The final non-remove path still requires perfect recall, and the second round must not regress from the first round. The same stable assertion pattern is applied to the related remove/non-remove feedback test in the same file.

## Validation

- `/opt/homebrew/opt/llvm@15/bin/clang-format -i tests/test_index_old.cpp`
- `git diff --check`

Local macOS execution is blocked from a clean `upstream/main` checkout by existing compiler/toolchain configuration issues: system AppleClang hits the AppleClang/libc++ detection issue addressed in PR #1930, while Homebrew LLVM 15 fails C compiler detection locally because it cannot link `libSystem`. Full runtime validation should come from PR CI.

### DIFF
--- a/tests/test_index_old.cpp
+++ b/tests/test_index_old.cpp
@@ -1105,8 +1105,12 @@ TEST_CASE("int8 + freshhnsw + feedback + update", "[ft][feedback][update][hnsw]"
     logger->Debug("====summary====");
     logger->Debug(fmt::format(R"(Error fix: {})", error_fix));
 
-    REQUIRE(error_fix > 0);
-    REQUIRE(recall[0] < recall[1]);
+    const bool initial_recall_is_perfect = fixtures::time_t(recall[0]) == fixtures::time_t(1.0f);
+    if (not initial_recall_is_perfect) {
+        REQUIRE(error_fix > 0);
+        REQUIRE(recall[0] < recall[1]);
+    }
+    REQUIRE(recall[1] >= recall[0]);
     REQUIRE(fixtures::time_t(recall[1]) == fixtures::time_t(1.0f));
 }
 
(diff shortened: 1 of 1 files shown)


wxyucs (approved):
lgtm

PR #1967, opened 2026-05-06
docs: update eval_performance tool documentation

wxyucs:
## Summary

Sync the `eval_performance` documentation (both `tools/eval/README{,_zh}.md` and the website pages under `docs/docs/{en,zh}/src/resources/eval.md`) with the actual implementation in `tools/eval/`.

### Corrections (the previous docs were factually wrong)

- **Output formats**: removed `markdown` (not supported); the real list is `table` / `text` / `json` / `line_protocol` (`tools/eval/exporter/formatter.cpp`).
- **Export destinations**: documented the URL-scheme based `to` values that the code actually parses — `stdout`, `file://<path>`, `influxdb://<host>:<port>/<path>?<query>`. Old docs claimed `to: file` with a `path` var, which doesn't work.
- **`exporters` is a YAML map, not a list** — `tools/eval/main.cpp` requires `IsMap()` and iterates with named keys.
- **Config-file invocation**: `eval_performance my.yaml` (positional). Removed the incorrect `--config` flag mentioned on the website.
- **Build instructions**: clarified that the tool requires `VSAG_ENABLE_TOOLS=ON` / `-DENABLE_TOOLS=ON`.
- **InfluxDB token**: noted that `vars.token` must include the `Token ` prefix because the value is sent verbatim as the `Authorization` header (`tools/eval/exporter/exporter_influxdb.h`).

### Newly documented features

- `http_server` block in `global` (embedded HTTP monitor for live progress / metrics).
- `influxdb://` exporter (with `vars.token` for InfluxDB auth).
- `--search-query-count` and `--delete-index-after-search` CLI flags.
- `num_threads_building` / `num_threads_searching`, available both globally and per-case.
- Per-case `disable_*` keys in YAML.

### Intentionally NOT documented: `appendfile://`

The codebase contains an `AppendfileExporter` registered under the `appendfile://` scheme (`tools/eval/exporter/exporter.cpp`), but the constructor strips only 6 characters (`file_path.substr(6)` in `tools/eval/exporter/exporter_appendfile.h:44`), the same as the `file://` exporter. Since `appendfile://` is 13 characters long, an input like `appendfile:///tmp/x` is parsed as `file:///tmp/x`, which is incorrect.

To avoid advertising a broken feature, this PR deliberately does not surface `appendfile://` in user-facing documentation. Once the prefix-stripping bug in `AppendfileExporter` is fixed (and ideally covered by a test), a follow-up change can re-introduce the documentation for `appendfile://<path>`.

## Test Plan

Documentation-only change. Verified the documented behavior against the source under `tools/eval/` (`main.cpp`, `eval_config.cpp`, `exporter/exporter.cpp`, `exporter/formatter.cpp`, `exporter/exporter_influxdb.h`, `exporter/exporter_appendfile.h`, `monitor/http_server_monitor.h`, `eval_template.yaml`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,6 +5,20 @@
 VSAG is a high-performance vector indexing library for similarity search, written primarily in C++
 with Python bindings provided by `pyvsag`.
 
+## Where To Learn About The Project
+
+Before making non-trivial changes, consult the project documentation:
+
+- The official documentation site: <https://vsag.io/docs> (English and Chinese versions are
+  available, including index parameters, best practices, performance references, and the
+  `eval_performance` tool guide).
+- The in-repo `docs/` directory, which mirrors the website source under `docs/docs/{en,zh}/src/`
+  and also contains design notes (e.g. `docs/hgraph.md`, `docs/ivf.md`, `docs/sindi.md`,
+  `docs/eval_performance.md`, `docs/dataset_format.md`).
+
+When updating user-facing behavior, keep both the website docs and any related in-repo READMEs
+(`tools/eval/README.md`, `tools/eval/README_zh.md`, etc.) in sync.
+
 ## What To Optimize For
 
 - Preserve API compatibility unless a breaking change is explicitly intended.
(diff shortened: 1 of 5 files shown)


jiaweizone (approved):
LGTM

PR #1971, opened 2026-05-06
chore: harden C++ release workflow

wxyucs:
## Summary
- Keep C++ release publishing as a manual workflow that builds/tests before creating the GitHub Release.
- Add release target validation, contents write permissions, tag-scoped concurrency, and prerelease support.
- Continue using the existing manylinux Docker packaging script and fail if no tarball assets are produced.

## Testing
- ruby YAML parser: `.github/workflows/release.yml` OK
- `git diff --check -- .github/workflows/release.yml`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,10 @@
-# .github/workflows/release.yml
-name: Create Release
+name: Build and Publish Release
 
 on:
   workflow_dispatch:
     inputs:
       branch:
-        description: 'The branch to create the release from'
+        description: 'The branch, tag, or commit SHA to release from'
         required: true
         default: 'main'
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -34,22 +45,60 @@ jobs:
...
+          fi
+
+          if gh release view "$TAG_NAME" --repo "$GITHUB_REPOSITORY" >/dev/null 2>&1; then
+            echo "GitHub Release '$TAG_NAME' already exists. Refusing to overwrite existing release assets."
+            exit 1
+          fi
+
+          release_sha="$(git rev-parse HEAD)"
+          git tag "$TAG_NAME" "$release_sha"
+          echo "sha=$release_sha" >> "$GITHUB_OUTPUT"
```
jiaweizone, line 84: Does the VSAG build artifact include the Git commit hash? That would make it easier to troubleshoot issues later and identify the exact version.

wxyucs: Addressed in 911f873f. Each C++ release tarball now gets a `vsag/BUILD_INFO` file before upload. It includes the release tag, full commit SHA, source ref, repository, workflow run URL, and UTC build timestamp, so the artifact can be traced back to the exact build source even when `git describe` resolves to only the release tag.

LHT129 (approved):
LGTM

PR #1982, opened 2026-05-07
fix: compilation failure on newer gcc version

pkusunjy:
GCC 12+ has dropped support for `avx512pf` and `avx512er`, causing compilation failure. This commit removes these flags by default while maintaining backward compatibility for default toolchains.


wxyucs (approved):
lgtm

PR #1995, opened 2026-05-07
feat: add MultiVector dataset support and migrate WARP to MultiVectors API

xfmeng17:
## Summary

- Introduce `MultiVector` struct and associated constants for per-document variable-length dense sub-vectors
- Add MultiVector virtual interfaces to `Dataset` and implement them in `DatasetImpl` (setter/getter, DeepCopy, Append, destructor)
- Migrate WARP index to exclusively consume `MultiVectors`, removing the legacy `VectorCounts` path
- Adapt tests and `110_index_warp` example to use the MultiVectors dataset API
- Fix a bug in `DatasetImpl::Append` where validation was hoisted before state mutation to prevent OOB access

## Test plan

- [x] Unit tests pass: `make test`
- [x] Functional tests pass for WARP index
- [x] Coverage threshold maintained (>= 90%)

### DIFF
--- a/examples/cpp/110_index_warp.cpp
+++ b/examples/cpp/110_index_warp.cpp
@@ -34,7 +34,7 @@ main(int argc, char** argv) {
     std::uniform_int_distribution<int> vec_count_dist(5, 10);
 
     // Generate MultiVector array: each document owns its own vectors buffer
-    std::vector<vsag::MultiVector> multi_vectors(num_elements);
+    vsag::MultiVector* multi_vectors = new vsag::MultiVector[num_elements];
     uint64_t total_vectors = 0;
     for (int64_t i = 0; i < num_elements; ++i) {
         ids[i] = i;
@@ -47,14 +47,14 @@ main(int argc, char** argv) {
         total_vectors += len;
     }
 
-    // Create dataset with MultiVectors API
+    // Create dataset with MultiVectors API (Owner(true) transfers memory ownership to Dataset)
     vsag::DatasetPtr base = vsag::Dataset::Make();
     base->NumElements(num_elements)
         ->Dim(dim)
         ->Ids(ids.data())
-        ->MultiVectors(multi_vectors.data())
+        ->MultiVectors(multi_vectors)
         ->MultiVectorDim(dim)
-        ->Owner(false);
+        ->Owner(true);
 
     std::cout << "Created multi-vector dataset with " << num_elements << " elements (documents)"
               << std::endl;
@@ -70,7 +70,8 @@ main(int argc, char** argv) {
         "dim": 128
     }
     )";
-    std::shared_ptr<vsag::Index> index = vsag::Factory::CreateIndex("warp", warp_build_parameters).value();
+    std::shared_ptr<vsag::Index> index =
+        vsag::Factory::CreateIndex("warp", warp_build_parameters).value();
 
     /******************* Build WARP Index *****************/
     if (auto build_result = index->Build(base); build_result.has_value()) {
@@ -114,11 +115,6 @@ main(int argc, char** argv) {
     for (int64_t i = 0; i < result->GetDim(); ++i) {
         std::cout << "  Document " << result->GetIds()[i]
                   << ": score = " << result->GetDistances()[i] << std::endl;
-    }
-
-    // Cleanup multi-vector buffers
-    for (int64_t i = 0; i < num_elements; ++i) {
-        delete[] multi_vectors[i].vectors_;
     }
 
     return 0;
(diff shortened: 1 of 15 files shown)


examples/cpp/110_index_warp.cpp
```diff
@@ -23,44 +23,43 @@ main(int argc, char** argv) {
...
+    // Create dataset with MultiVectors API
+    vsag::DatasetPtr base = vsag::Dataset::Make();
+    base->NumElements(num_elements)
         ->Dim(dim)
         ->Ids(ids.data())
-        ->Float32Vectors(datas.data())
-        ->VectorCounts(vector_counts.data())  // Specify number of vectors per document
+        ->MultiVectors(multi_vectors.data())
+        ->MultiVectorDim(dim)
         ->Owner(false);
```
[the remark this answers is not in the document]
xfmeng17: Good catch. Updated to use `Owner(true)` with heap-allocated `MultiVector` array (`new                                
  vsag::MultiVector[num_elements]`) so the Dataset destructor handles both the per-element `vectors_` buffers and the   
  top-level array. Fixed in b539032e.

include/vsag/dataset.h
```diff
@@ -405,6 +422,52 @@ class Dataset : public std::enable_shared_from_this<Dataset> {
...
+     *
+     * @details
+     * This defines the number of floats per sub-vector in each MultiVector element.
+     * It is independent of Dim(), which describes the dense single-vector dimensionality.
+     *
+     * @param dim The sub-vector dimensionality (must be > 0 when MultiVectors is used).
+     * @return DatasetPtr A shared pointer to the dataset with updated multi-vector dim.
+     */
+    virtual DatasetPtr
+    MultiVectorDim(int64_t dim) = 0;
```
LHT129, line 462: what is different between MultiVectorDim and Dim

src/algorithm/warp.cpp
```diff
@@ -61,32 +68,45 @@ WARP::EstimateMemory(uint64_t num_elements) const {
 std::vector<int64_t>
 WARP::Build(const vsag::DatasetPtr& data) {
     this->Train(data);
-    return this->Add(data);
+    std::vector<int64_t> failed_ids = this->Add(data);
+    return failed_ids;
```
LHT129, line 72: why modify this code

include/vsag/dataset.h
```diff
@@ -35,6 +35,23 @@ struct SparseVector {
...
+ * The sub-vector dimensionality is defined at the Dataset level via
+ * MultiVectorDim(), not per-element.
+ *
+ * @note When Owner(true) is set on the Dataset, each element's vectors_
+ * pointer must be independently allocated (not an offset into a shared
+ * buffer), because the Dataset destructor will free each vectors_ separately.
+ */
+struct MultiVector {
+    uint32_t len_ = 0;          // number of sub-vectors in this document
+    float* vectors_ = nullptr;  // flat array of len_ * MultiVectorDim floats
```
LHT129, line 52: prefer using vector instead vector_ for public member

LHT129 (changes requested):

include/vsag/dataset.h
xfmeng17: Dim is the dimension for the normal single-vector dataset: each element contains one dense vector with `Dim` floats.

MultiVectorDim is the dimension of each sub-vector inside a multi-vector document. A multi-vector document may contain `len_` sub-vectors, and each sub-vector has `MultiVectorDim` floats.

So they are independent. For example, `Dim` may be 1024 for a normal dense-vector dataset, while `MultiVectorDim` may be 32 for a multi-vector dataset.

src/algorithm/warp.cpp
xfmeng17: No behavior change here.

I expanded it only to make the return value explicit. `Add(data)` returns `std::vector<int64_t>` as failed ids, and I sometimes forget what the returned value represents when reading this code. Naming it `failed_ids` makes the intent clearer.

include/vsag/dataset.h
xfmeng17: I used `vectors_` to keep it consistent with the existing public member naming style in `struct SparseVector`.

For example, `SparseVector` already uses members like `len_`, `ids_`, and `vals_`, so `MultiVector` follows the same convention with `len_` and `vectors_`.

LHT129 (approved):
LGTM

xfmeng17:
Hi @LHT129  @wxyucs @inabao @jiaweizone, all review comments have been addressed on my side. Could you please help approve/merge this PR into main when you have time? Thanks!

LHT129 (approved):
LGTM

PR #1997, opened 2026-05-07
fix: guard scalar quantizer NaN encoding

### LINKED ISSUE #1966: Handle NaN scalar quantizer encoding before uint8 casts
Parent tracking issues: #1950, #1960

The macOS ASan/UBSan jobs report NaN-to-uint8 conversions in scalar quantizer encoding. This was observed while enabling macOS CI, but it is a general undefined-behavior risk rather than a macOS-only semantic issue.

CI evidence:

- Test macOS Functests: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256867?pr=1930
- Test macOS Unittests: https://github.com/antgroup/vsag/actions/runs/25094707114/job/73530256873?pr=1930

Log excerpts:

```text
/Users/runner/work/vsag/vsag/src/quantization/scalar_quantization/scalar_quantizer.cpp:108:39: runtime error: nan is outside the range of representable values of type 'unsigned char'
/Users/runner/work/vsag/vsag/src/quantization/scalar_quantization/sq8_uniform_quantizer.cpp:120:39: runtime error: nan is outside the range of representable values of type 'unsigned char'
/Users/runner/work/vsag/vsag/src/quantization/scalar_quantization/sq4_uniform_quantizer.cpp:132:39: runtime error: nan is outside the range of representable values of type 'unsigned char'
```

Code locations:

- `src/quantization/scalar_quantization/scalar_quantizer.cpp:108`: casts `MAX_CODE_PER_DIM * delta` to `uint8_t`.
- `src/quantization/scalar_quantization/sq8_uniform_quantizer.cpp:120`: casts `255.0F * delta` to `uint8_t`.
- `src/quantization/scalar_quantization/sq4_uniform_quantizer.cpp:132`: casts `15.0F * delta` to `uint8_t`.

Initial root cause:

`delta` can become NaN, for example from zero-range training (`diff_ == 0`) or non-finite input values. The existing comparisons against `< 0.0F` and `> 0.999F` do not catch NaN, so the code eventually casts NaN to `uint8_t`, which is undefined behavior.

Suggested fix:

- Define explicit zero-range encoding semantics for scalar quantizers.
- Check `std::isfinite(delta)` before casting.
- Clamp or handle non-finite values explicitly; do not rely on `static_cast<uint8_t>`.
- Add regression tests for zero-range training and any non-finite input behavior the project chooses to support.

Non-goals:

- Do not silently change ranking/distance semantics without a regression test.
- Do not mix HNSW or generic deserialize/bitset UBSan fixes into this PR.

jac0626:
## What changed

Fixes #1966.

- Guard scalar quantizer encoding so `NaN` deltas are clamped before casting to `uint8_t`.
- Define zero-range SQ4/SQ8 uniform encoding as code `0`, avoiding division by zero before the cast.
- Add regression coverage for ordinary SQ NaN-delta encoding and SQ4/SQ8 uniform zero-range encoding.

## Why

UBSan reported `nan is outside the range of representable values of type 'unsigned char'` in scalar quantizer encoding. Existing `< 0` / `> 0.999` clamping does not catch `NaN`, and SQ4/SQ8 uniform quantizers could also divide by zero when trained on a zero-range dataset.

## Validation

- `/opt/homebrew/opt/llvm@15/bin/clang-format -i ... && git diff --check`
- `clang++ -fsyntax-only` for the six changed translation units using the existing compile commands

Full local `unittests` were not completed on this macOS checkout because upstream/main's current macOS build path is blocked by local toolchain/third-party configuration issues; PR CI should provide the full Linux sanitizer validation.

### DIFF
--- a/src/quantization/scalar_quantization/scalar_quantization_utils.h
+++ b/src/quantization/scalar_quantization/scalar_quantization_utils.h
@@ -16,12 +16,14 @@
 
 namespace vsag {
 
-static inline float
+inline constexpr float kScalarQuantizationUpperClamp = 0.999F;
+
+inline float
 ClampScalarQuantizationDelta(float delta) {
     if (!(delta >= 0.0F)) {
         return 0.0F;
     }
-    if (delta > 0.999F) {
+    if (delta > kScalarQuantizationUpperClamp) {
         return 1.0F;
     }
     return delta;
--- a/src/quantization/scalar_quantization/scalar_quantization_utils.h
+++ b/src/quantization/scalar_quantization/scalar_quantization_utils.h
@@ -1,0 +1,30 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace vsag {
+
+static inline float
+ClampScalarQuantizationDelta(float delta) {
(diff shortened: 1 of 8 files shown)


LHT129 (approved):
LGTM

PR #2013, opened 2026-05-09
fix: avoid NaN recall in HGraph duplicate range-search functest

### LINKED ISSUE #1996: HGraph Duplicate range search can produce NaN recall in PR functests
## Summary

`Test X86 Functests` can intermittently fail in `(PR) HGraph Duplicate` because the range-search recall check becomes `NaN`.

This has been observed on unrelated PR branches, so it should be tracked independently from the current sanitizer/undefined-behavior fixes.

## Observed failures

- PR #1985 rerun: https://github.com/antgroup/vsag/actions/runs/25493821628/job/74813592675
  - Failing case: `(PR) HGraph Duplicate`
  - Parameters: `metric_type: ip`, `dim: 192`, `base_quantization_str: fp16`, `recall: 0.98`, `duplicate_pos: suffix`
  - Assertion: `tests/test_index.cpp:535`
  - Expansion: `nanf > 83.300003052f`

- PR #1989: https://github.com/antgroup/vsag/actions/runs/25487244946/job/74792631072
  - Failing case: `(PR) HGraph Duplicate`
  - Parameters: `metric_type: cosine`, `dim: 192`, `base_quantization_str: fp16`, `recall: 0.98`, `duplicate_pos: prefix`
  - Assertion: `tests/test_index.cpp:535`
  - Expansion: `nanf > 83.300003052f`

## Current understanding

In `TestIndex::TestRangeSearch`, recall is accumulated with a denominator based on `std::min(gt_topK - 1, res.value()->GetDim())`.

If `RangeSearch` returns an empty result for a query, the denominator becomes `0`, so the recall calculation can produce `NaN`. The test then fails with a hard-to-read comparison such as `nanf > 83.300003052f`.

## Why this appears flaky

- `(PR) HGraph Duplicate` samples parts of the test configuration randomly, including metric, dimension, and quantization.
- HGraph/ODescent graph construction uses runtime randomness, so the graph and recall behavior can vary between runs.
- The issue has been seen in x86 ASan GitHub Actions jobs, but did not reproduce locally on macOS arm64 after repeated random and forced-parameter runs.

## Suggested direction

- Make `TestRangeSearch` handle empty range-search results explicitly instead of letting the recall calculation become `NaN`.
- Investigate why HGraph duplicate range search can return empty results for the affected fp16 configurations.
- Consider making the affected test configuration deterministic or logging enough seed/configuration data to reproduce failures reliably.

jac0626:
## Summary
Fix flaky (PR) HGraph Duplicate failures where range-search recall in test code can become NaN.

## Reproduction
In TestIndex::TestRangeSearch, recall was accumulated as:
- numerator: intersection count
- denominator: min(gt_topK - 1, res_dim)

When RangeSearch returned empty result (res_dim = 0), denominator became 0, producing 0/0 -> NaN, and the assertion failed as NaN > threshold is false.

## Root Cause
The test did not guard against zero denominator for empty range-search results.

## Fix
In tests/test_index.cpp, compute denominator first and only accumulate per-query recall when denominator > 0.

## Validation
- Local minimal reproduction confirmed original formula yields 0/0 -> NaN.
- Full project test build in this container is currently blocked by unrelated OpenBLAS build error in make debug.

Closes #1996

### DIFF
--- a/tests/test_index.cpp
+++ b/tests/test_index.cpp
@@ -518,11 +518,17 @@ TestIndex::TestRangeSearch(const IndexPtr& index,
         if (limited_size > 0) {
             REQUIRE(res.value()->GetDim() <= limited_size);
         }
+        auto res_dim = res.value()->GetDim();
         auto result = res.value()->GetIds();
         auto gt = gts->GetIds() + gt_topK * i;
-        auto val = Intersection(gt, gt_topK - 1, result, res.value()->GetDim());
-        cur_recall += static_cast<float>(val) /
-                      static_cast<float>(std::min(gt_topK - 1, res.value()->GetDim()));
+        auto gt_count = std::max<int64_t>(0, gt_topK - 1);
+        auto val = Intersection(gt, gt_count, result, res_dim);
+        auto denominator = std::min(gt_count, res_dim);
+        // Skip recall calculation if denominator is 0 to avoid NaN when RangeSearch returns
+        // empty result.
+        if (denominator > 0) {
+            cur_recall += static_cast<float>(val) / static_cast<float>(denominator);
+        }
     }
     if (cur_recall <= expected_recall * query_count) {
         WARN(fmt::format("cur_result({}) <= expected_recall * query_count({})",


wxyucs (approved):
lgtm

PR #2014, opened 2026-05-09
test: cover vsag_c_api edge paths

LHT129:
## Summary
- add C API tests for null-handle wrapper behavior across build, search, clone, update, and serialize APIs
- add failure-path tests for invalid search parameters, missing ids, update edge cases, and deserialize-on-non-empty-index behavior
- keep production code unchanged and validate the new cases with make test CASE=[vsag_c_api]

### DIFF
--- a/src/vsag_c_api_test.cpp
+++ b/src/vsag_c_api_test.cpp
@@ -47,6 +47,7 @@ static constexpr const char* hgraph_search_parameters = R"(
 
 static constexpr int64_t num_vectors = 500;
 static constexpr int64_t dim = 128;
+thread_local std::string deserialize_read_func_path;
 
 class VsagTestCase {
 public:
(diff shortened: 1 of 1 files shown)


LHT129:
## Review Feedback Addressed

### 1. Null handle returning VSAG_SUCCESS (High Priority) — **Fixed**

`vsag_index_add`, `vsag_index_train`, `vsag_index_knn_search`, `vsag_index_knn_search_with_filter`, `vsag_index_range_search`, and `vsag_index_range_search_with_filter` now correctly return `VSAG_INTERNAL_ERROR` when passed a `nullptr` index handle. Previously these functions silently returned `VSAG_SUCCESS` for null handles — this has been fixed both in the C API implementation and in the test assertions.

### 2. `vsag_index_calculate_distance_by_ids` returning VSAG_SUCCESS for missing IDs (Medium Priority) — **By Design, Added Comment**

This is intentional behavior: the C API uses `-1` as a sentinel distance value for invalid/missing IDs (consistent with the underlying `CalDistanceById` implementation). A comment has been added to the test to document this design choice.

### 3. `stat()` return value not checked (Copilot) — **Fixed**

Added a check for `stat()` return value in the size callback, returning 0 on failure.

### 4. Read callback missing file-open check (Copilot) — **Fixed**

Added `ifile.is_open()` check before reading from the file in the deserialize read callback.

All changes are in the latest push. CI should pass with the updated code.

LHT129:
## Remaining Review Comments Resolved

### Copilot: `stat()` return value not checked — **Fixed**
Added stat() return value check in the size callback, returning 0 on failure.

### Copilot: Read callback missing file-open check — **Fixed**
Added `ifile.is_open()` check before reading from the file in the deserialize read callback.

All review feedback has been addressed. ✅

PR #2051, opened 2026-05-14
feat: add HGraph duplicate distance threshold

### LINKED ISSUE #2050: Support distance-threshold deduplication in HGraph
## Background
HGraph duplicate detection currently relies on exact code `memcmp` equality after Searcher picks the nearest candidate. This is too strict for near-duplicate vectors that should collapse into the same duplicate group.

## Proposal
- add a `duplicate_distance_threshold` HGraph build parameter
- keep duplicate ownership on the nearest candidate only
- when the threshold is greater than `0`, deduplicate by nearest-candidate distance
- when the threshold is `0`, preserve the current `memcmp` fallback behavior

## Notes
This keeps duplicate decision logic inside `BasicSearcher`, while `HGraph` only passes the setting through and consumes `duplicate_id`.

LHT129:
Closes #2050

## Summary
- add `duplicate_distance_threshold` as an HGraph build parameter and propagate it through the search insertion path
- keep duplicate ownership on the nearest candidate while allowing distance-threshold deduplication with `memcmp` fallback at zero
- add unit coverage for parameter mapping and Searcher duplicate-threshold behavior, plus docs updates

## Testing
- make fmt
- ./build/tests/unittests -d yes '[ut][HGraphParameter]' --allow-running-no-tests
- ./build/tests/unittests -d yes '[ut][BasicSearcher][duplicate]' --allow-running-no-tests

### DIFF
--- a/docs/docs/en/src/indexes/hgraph.md
+++ b/docs/docs/en/src/indexes/hgraph.md
@@ -74,6 +74,7 @@ and `docs/hgraph.md` in the repository.
 | `base_pq_dim` | int | `1` | Number of PQ subspaces. When using `pq` / `pqfs`, set this explicitly instead of relying on the default. |
 | `build_thread_count` | int | `100` | Threads used to parallelise build |
 | `support_duplicate` | bool | `false` | Enable duplicate-ID detection on insert |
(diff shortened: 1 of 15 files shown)


wxyucs (approved):
lgtm

Issue #1957, opened 2026-04-30
Support SINDI Disk Algorithm DiskSindi


Roxanne0321:
## Motivation

SINDI is a pure in-memory sparse vector index. For large-scale datasets (tens of millions of docs), keeping all posting lists and rerank vectors in RAM is prohibitively expensive. DiskSINDI is the disk-backed counterpart: it keeps only lightweight metadata resident in memory and loads posting/rerank payloads from disk on demand, achieving 87-96% memory savings at the cost of controlled latency increase.

## Architecture

- `sindi` remains the pure in-memory implementation.
- `disksindi` is a read-only disk index (V1 does not support online insert/update).
- Both inherit `InnerIndexInterface` directly.
- DiskSINDI reuses `SparseVectorDataCell` (implementing `FlattenInterface`) for rerank flat storage, enabling pluggable IO backends.

### On-disk File Layout

All segments are written into a single sequential binary file:

```
+----------------------------------------------------------------------+
| DiskSINDIManifest (segment offsets)                                  |
| cur_element_count (int64_t)                                          |
| quantization params (optional, 3×float when use_quantization=true)   |
+----------------------------------------------------------------------+
| Term Dictionary  [DiskTermEntry × (term_id_limit + 1)]               |
+----------------------------------------------------------------------+
| Posting Payload Area                                                 |
|   term_0 payload | term_1 payload | ... | term_N payload             |
+----------------------------------------------------------------------+
| Rerank Flat Segment (optional, when use_reorder=true)                |
+----------------------------------------------------------------------+
| Label Table Segment                                                  |
+----------------------------------------------------------------------+
| Term ID Mapper Segment (optional, when remap_term_ids=true)          |
+----------------------------------------------------------------------+
| Footer (param JSON, metadata, format marks)                          |
+----------------------------------------------------------------------+
```

### Core Data Structures

```cpp
struct DiskSINDIManifest {
    uint64_t term_dict_offset, term_dict_size;
    uint64_t posting_payload_offset, posting_payload_size;
    uint64_t rerank_flat_offset, rerank_flat_size;
    uint64_t label_table_offset, label_table_size;
};

struct DiskTermEntry {
    uint64_t posting_payload_offset;  // term payload start within posting area
    uint32_t posting_payload_size;    // term payload byte length
    uint32_t term_num;                // number of postings for this term
};

struct TermBuffer {                   // per-term posting in window-first layout
    vector<uint32_t> window_offsets;  // size = window_count + 1
[...]

PR #2056, opened 2026-05-14
feat(sindi): use sparse vector datacell and remove sparse index

Roxanne0321:
## Summary
- replace SINDI rerank forward storage with `SparseVectorDataCell`
- keep legacy SINDI rerank deserialization compatible without writing a new magic marker
- remove the standalone `SparseIndex` implementation, registration, constants, and functional tests
- update SINDI and datacell tests to cover the new exact-baseline path and sparse vector retrieval
- sync affected English and Chinese docs for sparse-index references

## Validation
- `./build/tests/unittests "[SparseDataCell]"`
- `./build/tests/unittests "[SINDI]"`
- `make release`

## Labels To Add
- `kind/improvement`
- `version/1.0`


Roxanne0321:
Related to #1957

LHT129 (commented):
Solid, careful migration of SINDI's rerank store from the standalone SparseIndex to a SparseVectorDataCell, with thoughtful serialization versioning (sentinel + version) and a backward-compat legacy reader for both the datacell's own format and SINDI's old rerank payload. The 32->64-bit offset upgrade and packed DocLocation are well reasoned and tested, and the legacy byte layout reads back exactly (int64 count, then per-element uint32 len + len ids + len vals). Two behavioral concerns are worth addressing before merge: (1) the precise-distance CalcDistanceById path now throws on a missing id instead of returning -1 as the old SparseIndex did, and (2) SINDI::Deserialize now calls Footer::Parse unconditionally, dropping the deserialize_without_footer_ short-circuit. Plus minor cleanup nits. The datacell resource/lock handling looks correct. (Flagged from the diff — please confirm against full file context.)

src/algorithm/sindi/sindi.cpp
```diff
@@ -649,7 +770,10 @@ SINDI::CalcDistanceById(const DatasetPtr& vector,
     std::shared_lock rlock(this->global_mutex_);
 
     if (use_reorder_ && calculate_precise_distance) {
-        return this->rerank_flat_index_->CalcDistanceById(vector, id);
+        auto inner_id = this->label_table_->GetIdByLabel(id);
```
LHT129: Behavior regression for missing ids. The old precise path delegated to rerank_flat_index_->CalcDistanceById, which used TryGetIdByLabel and returned -1.0F when the label was absent (this was exercised by the deleted test_sparse_index.cpp 'returns -1 for missing id' case). This new path calls label_table_->GetIdByLabel(id) directly, which throws VsagException for a missing/removed label. Note the batch CalDistanceById below was correctly migrated to TryGetIdByLabel -> -1; this single-id path should do the same for parity.

src/algorithm/sindi/sindi.cpp
```diff
@@ -512,8 +622,9 @@ void
 SINDI::Deserialize(StreamReader& reader) {
     std::scoped_lock wlock(this->global_mutex_);
 
-    if (not deserialize_without_footer_) {
-        auto footer = Footer::Parse(reader);
+    bool has_datacell_rerank_format = false;
+    auto footer = Footer::Parse(reader);
```
LHT129: Footer::Parse(reader) now appears to be called unconditionally, removing the previous `if (not deserialize_without_footer_)` short-circuit (the flag now only gates a debug log). Footer::Parse does PushSeek(Length()-16) and seeks back on failure, so this relies on the StreamReader supporting backward Seek/Length for every deserialize path — including the streaming/no-footer scenarios the flag was originally introduced to avoid. Please confirm all readers used with deserialize_without_footer_=true (e.g. BufferStreamReader-backed streaming loads) can seek to end-of-stream and back without side effects; otherwise this is a silent regression for that path.

src/algorithm/sindi/sindi.cpp
```diff
@@ -601,9 +719,13 @@ SINDI::EstimateMemory(uint64_t num_elements) const {
...
     }
 
-    // size of rerank index is same as sindi
     if (use_reorder_) {
-        mem *= 2;
+        mem += num_elements *
+               (sizeof(uint32_t) + avg_doc_term_length_ * (sizeof(uint32_t) + sizeof(float)));
+
+        const auto block_size = Options::Instance().block_size_limit();
+        const auto offset_bytes = num_elements * (sizeof(uint64_t) + sizeof(uint32_t));
```
LHT129, line 1326: EstimateMemory for the reorder case was rewritten from `mem *= 2` to an explicit datacell sizing: payload + block-rounded offset table using num_elements * (sizeof(uint64_t)+sizeof(uint32_t)) (12 bytes/entry, matching the packed DocLocation). Looks correct — just double-check avg_doc_term_length_ here is in the same units as the datacell's per-doc term count so the payload term is not under/over-counted.

src/algorithm/sparse_distance.h
```diff
@@ -0,0 +1,70 @@
...
+            sum += vals1[i] * vals2[j];
+            i++;
+            j++;
+        }
+    }
+
+    return 1 - sum;
+}
+
+}  // namespace vsag
```
LHT129: Missing trailing newline at EOF (the diff shows '\ No newline at end of file'). The project requires every committed text file to end with a newline and CI enforces formatting. Also: get_distance defined here appears unused after the migration — cal_distance_by_id_unsafe in sindi.cpp inlines its own merge-join rather than calling it, and only sort_sparse_vector is consumed. Consider dropping get_distance or routing the SINDI rerank through it to avoid a second copy of the same merge logic.

src/algorithm/sindi/sindi.cpp
```diff
@@ -15,15 +15,126 @@
...
 
 namespace vsag {
+
+namespace {
+
+constexpr const char* SINDI_RERANK_FLAT_FORMAT_KEY = "sindi_rerank_flat_format";
+constexpr int64_t SINDI_RERANK_FLAT_FORMAT_DATACELL = 2;
+
+float
+cal_distance_by_id_unsafe(const FlattenInterfacePtr& flat,
```
wxyucs: rename to calc_distance_by_id_unsafe

src/algorithm/sparse_distance.h
wxyucs: add a newline at EOF

wxyucs (changes requested):

src/algorithm/sindi/sindi.cpp
Roxanne0321: Good catch. I restored the old missing-label behavior on the precise reorder path by using TryGetIdByLabel and returning -1.0F, and added a SINDI unit assertion for the single-id path.

src/algorithm/sindi/sindi.cpp
Roxanne0321: Confirmed this was a real regression risk. I restored the deserialize_without_footer_ short-circuit so Footer::Parse is skipped for no-footer streaming loads; normal footer-enabled loads still parse metadata and fall back to the legacy path when no footer is present.

src/algorithm/sindi/sindi.cpp
Roxanne0321: Checked. avg_doc_term_length_ is the per-document sparse term count used by the datacell payload sizing, so the payload term is in the same units. No code change needed here.

src/algorithm/sparse_distance.h
Roxanne0321: Fixed. The file now has a trailing newline, and I removed the unused get_distance helper so SINDI keeps a single active rerank distance implementation.

src/algorithm/sindi/sindi.cpp
Roxanne0321: Fixed: renamed the helper to calc_distance_by_id_unsafe.

src/algorithm/sparse_distance.h
Roxanne0321: Fixed: sparse_distance.h now ends with a newline.

Roxanne0321:
Compatibility and performance validation for the latest PR update (`b434d841`):

**PR changes covered**
- SINDI reorder/rerank storage is moved from the standalone `SparseIndex` to `SparseVectorDataCell`.
- The standalone sparse index implementation and its tests are removed.
- SparseVectorDataCell serialization keeps a sentinel + version field. I kept both because a version-only field would be ambiguous with the legacy `uint32_t current_offset_`; the sentinel cleanly distinguishes old/new layouts and the version remains the future extension point.
- Added compatibility handling for upstream-main `use_reorder=true` indexes, footerless deserialization, and empty SINDI index serialization.

**Local targeted validation**
- `make release -j 8`: passed
- `cmake --build build-release --target eval_performance unittests -j 6`: passed
- `cmake --build build-release --target functests -j 6`: passed
- `build-release/tests/functests 'SINDI Serialize File' -d yes`: passed, 125600 assertions / 1 test case
- `build-release/tests/unittests '[SINDI]' -d yes`: passed, 15681 assertions / 11 test cases
- `build-release/tests/unittests '[SparseDataCell]' -d yes`: passed, 14079 assertions / 4 test cases

**Compatibility test**
- Dataset: splade-1m
- Index producer: upstream main binary with `use_reorder=true`
- Index consumer: this PR branch binary
- Script: `BLOCKS=1 WARMUP_BLOCKS=1 experiments/compare_sindi_reorder_two_bins.sh`
- Result: this PR successfully deserialized and searched the upstream-main reorder index. Recall matched main exactly in all measured runs.

**Search performance, excluding warmup**

| Version | Runs | QPS Mean | Recall Mean | Avg Latency Mean | P99 Latency Mean |
| --- | ---: | ---: | ---: | ---: | ---: |
| upstream main | 2 | 1114.59 | 0.8121719202 | 0.9029 ms | 2.0688 ms |
| PR | 2 | 1167.94 | 0.8121719202 | 0.8562 ms | 1.3094 ms |

PR vs main QPS mean: +4.79%. No SINDI reorder search performance regression was observed in this small compatibility/performance run.

A detailed local report was also generated at `experiments/results/pr2056_sindi_datacell_test_report_20260611.md`.

wxyucs (approved):
lgtm

PR #2086, opened 2026-05-20
feat(hgraph): add support_force_remove switch

### LINKED ISSUE #2085: [feat](hgraph): add dedicated support_force_remove parameter for force remove
### Affected area

hgraph

### Motivation

`support_remove` and `RemoveMode::FORCE_REMOVE` currently represent different concerns and should not share the same public knob.

For HGraph, `support_remove` is about graph delete-tracking metadata used by mark-remove recovery paths, while `support_force_remove` is the dedicated switch for `RemoveMode::FORCE_REMOVE` and the extra synchronization it requires. Keeping these semantics separate makes the API easier to understand and avoids implying that every remove-related workload needs the heavier force-remove path.

This split is useful for users who:
- need `MARK_REMOVE` semantics without enabling force-remove-specific locking
- want a clear, explicit opt-in before using `RemoveMode::FORCE_REMOVE`
- need to avoid conflating force remove with existing graph delete metadata semantics

Using `support_remove` for both purposes would continue to mix two different behaviors and would add unnecessary lock overhead to non-force-remove workloads.

### Proposed solution

Add a dedicated HGraph build parameter `support_force_remove` and scope it only to `RemoveMode::FORCE_REMOVE` plus its extra synchronization.

Expected user-facing behavior:
- `support_remove` continues to control graph delete-tracking metadata for mark-remove-related behavior
- `support_force_remove` independently controls whether `RemoveMode::FORCE_REMOVE` is allowed
- HGraph add/search/range/remove paths only take the extra force-remove synchronization when `support_force_remove: true`
- calling `Remove(..., RemoveMode::FORCE_REMOVE)` on an index built without `support_force_remove: true` returns an invalid-argument error
- docs and examples describe `support_remove` and `support_force_remove` as separate knobs with separate semantics

This should be implemented and documented consistently across:
- external/public constants
- HGraph parameter mapping and compatibility checks
- HGraph runtime locking and force-remove guard rails
- regression and parameter-mapping tests
- HGraph and lifecycle documentation

### Alternatives considered

Reusing `support_remove` for `FORCE_REMOVE` as well:
- rejected because it mixes graph delete metadata with physical graph rewrite semantics

Always enabling the force-remove synchronization path:
- rejected because non-force-remove workloads would pay extra locking cost without needing `RemoveMode::FORCE_REMOVE`

additive only (no break)

## References

- `src/algorithm/hgraph_parameter.h:68`
- `src/algorithm/hgraph_parameter.cpp:134`
- `src/algorithm/hgraph_parameter.cpp:153`
- `src/algorithm/hgraph_parameter.cpp:205`
- `src/algorithm/hgraph.cpp:487`
- `src/algorithm/hgraph.cpp:866`
- `src/algorithm/hgraph.cpp:1074`
- `src/algorithm/hgraph.cpp:1373`
- `src/algorithm/hgraph.cpp:2195`
- `src/algorithm/hgraph.h:374`
- `include/vsag/constants.h:168`
- `src/constants.cpp:150`
- `src/inner_string_params.h:169`
- `src/algorithm/hgraph_parameter_test.cpp:210`
- `tests/test_hgraph_remove.cpp:420`
- `docs/docs/en/src/indexes/hgraph.md:78`
- `docs/docs/en/src/indexes/hgraph.md:79`
- `docs/docs/en/src/advanced/index_lifecycle.md:33`
- `docs/docs/en/src/advanced/index_lifecycle.md:36`
- `docs/hgraph.md:219`
- `docs/hgraph.md:225`

### Additional context

This draft is based on the current workspace changes, which already update code, tests, examples, and docs around the HGraph remove-parameter split.

## Related issues

Issue searches run:
- `gh issue list --repo antgroup/vsag --state all --limit 5 --search "hgraph support_force_remove OR force remove OR support_remove in:title,body"`
- `gh issue list --repo antgroup/vsag --state all --limit 5 --search "hgraph remove FORCE_REMOVE in:title,body"`
- `gh issue list --repo antgroup/vsag --state all --limit 10 --search "\"support_force_remove\""`
- `gh issue list --repo antgroup/vsag --state all --limit 10 --search "hgraph mark remove force remove in:title,body"`

Matches found:
- [#2073](https://github.com/antgroup/vsag/issues/2073) OPEN, `[bug](index deletion): core dump while adding/updating vector with deleting vector in parallel`
  Relevant because it touches concurrent deletion behavior, but it is a bug report rather than a feature request for a dedicated parameter split.
- [#2079](https://github.com/antgroup/vsag/issues/2079) CLOSED, `[bug](hgraph): repeated force remove reuses stale label mappings`
  Relevant because it covers force-remove correctness, but it does not request separating `support_force_remove` from `support_remove`.

No direct existing feature issue was found for introducing a dedicated `support_force_remove` parameter or for separating it from `support_remove`.

---
_Drafted with AI assistant: OpenCode:gpt-5.4_

LHT129:
## Summary
- add a dedicated `support_force_remove` HGraph parameter and map it through the public and internal parameter layers
- gate force-remove locking and `RemoveMode::FORCE_REMOVE` on `support_force_remove` while keeping `support_remove` for existing delete-metadata semantics
- update regression tests, examples, and EN/ZH docs to describe the split behavior

## Verification
- `cmake -S . -B build-release -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON`
- `cmake --build build-release --target functests unittests`
- `./build-release/tests/unittests "*support_force_remove*"`
- `./build-release/tests/functests "HGraph ForceRemove Requires support_force_remove"`

Closes: #2085

### DIFF
--- a/docs/docs/en/src/advanced/index_lifecycle.md
+++ b/docs/docs/en/src/advanced/index_lifecycle.md
@@ -31,12 +31,12 @@ supports it and which mode it supports.
 `Remove` deletes vectors by id. HGraph supports two deletion modes with different requirements:
 
 - `RemoveMode::MARK_REMOVE` (the default) only writes a tombstone via the label table and works
-  regardless of `support_remove`. The id is filtered out of subsequent searches, but the underlying
-  graph node and vector storage are kept.
+  regardless of `support_force_remove`. The id is filtered out of subsequent searches, but the
+  underlying graph node and vector storage are kept.
 - `RemoveMode::FORCE_REMOVE` physically rewrites the graph and reclaims the slot. This mode is
-  only available when the index was built with `support_remove: true` in `index_param` (which
-  causes the graph data cell to allocate the delete-tracking metadata). Calling `FORCE_REMOVE` on
-  an index built without `support_remove: true` will fail.
+  only available when the index was built with `support_force_remove: true` in `index_param`. That
+  flag enables the force-remove path and its extra synchronization; calling `FORCE_REMOVE` on an
+  index built without `support_force_remove: true` will fail.
 
 ```json
 {
@@ -47,13 +47,13 @@ supports it and which mode it supports.
         "base_quantization_type": "sq8",
         "max_degree": 16,
         "ef_construction": 100,
-        "support_remove": true
+        "support_force_remove": true
     }
 }
 ```
 
 The JSON snippet above is only required if you intend to use `FORCE_REMOVE`. For `MARK_REMOVE`
-alone you can omit the `support_remove` flag.
+alone you can omit the `support_force_remove` flag.
 
 ```json
 {
@@ -63,8 +63,7 @@ alone you can omit the `support_remove` flag.
     "index_param": {
         "base_quantization_type": "sq8",
         "max_degree": 16,
-        "ef_construction": 100,
-        "support_remove": true
+        "ef_construction": 100
     }
 }
 ```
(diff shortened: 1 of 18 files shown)


LHT129:
@gemini-code-assist The comment about the commented-out `add_mutex_` lock in `insert_persistent_codes` (line 1838) was based on an older commit `dcb187a7`. In the current commits (`97aab97d` and later), the lock handling has been properly conditionized:

```cpp
void HGraph::insert_persistent_codes(const void* data, InnerIdType inner_id) {
    std::shared_lock<std::shared_mutex> add_lock;
    if (not this->support_force_remove()) {
        add_lock = std::shared_lock<std::shared_mutex>(this->add_mutex_);
    }
    // ... insertion code ...
}
```

When `support_force_remove` is **disabled**:
- The read lock on `add_mutex_` is taken in `insert_persistent_codes`, protecting concurrent access.

When `support_force_remove` is **enabled**:
- No lock is taken here because `add_one_point` already holds a write lock on `add_mutex_` earlier in the call chain (before `insert_codes`), which provides exclusive protection for the insertion path.

Regarding the `build_by_odescent` call: it executes sequentially during the build phase and does not run concurrently with other `Add()` operations, so the race condition concern does not apply in that path. The lock conditionization is scoped specifically to force-remove workloads.

wxyucs (approved):
lgtm

PR #2107, opened 2026-05-26
chore: add Apache 2.0 license headers to remaining source files

wxyucs:
## Summary

A repo-wide scan of `src/` and `include/` for the Apache 2.0 license header turned up three tracked files that still lack one:

- `src/reader.cpp`
- `src/version.h.in` — template; the generated `src/version.h` is gitignored and inherits the header on rebuild
- `src/impl/blas/blas_compat.h`

After this PR every tracked `*.cpp` / `*.h` under `src/` and `include/` carries the standard `// Copyright 2024-present the vsag project` block (matching the format used across the rest of the tree, e.g. `src/dataset_impl.cpp`). No code change.

## Test plan

- [x] `find src include \( -name "*.cpp" -o -name "*.h" \) -exec grep -L "Licensed under the Apache License" {} \;` — returns only the gitignored generated `src/version.h`, which gets the header from `version.h.in` on the next build.

### DIFF
--- a/src/impl/blas/blas_compat.h
+++ b/src/impl/blas/blas_compat.h
@@ -1,3 +1,17 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


LHT129 (approved):
LGTM

PR #2108, opened 2026-05-26
docs: add SECURITY.md vulnerability disclosure policy

wxyucs:
## Summary

VSAG is approaching its v1.0 milestone but does not yet ship a `SECURITY.md`. Add one at the repository root so the file is picked up both by the GitHub repo's "Security" tab and by tooling that scans for it (OpenSSF Scorecard, GitHub's repo health checklist, dependency scanners that surface "no disclosure policy" findings).

The policy:

- Points reporters at GitHub's private vulnerability reporting flow (`/security` tab) so we do not have to maintain a separate inbox and reporters do not have to leak details over plaintext email.
- States response-time targets (5 business days to ack, 10 to triage, 90 days target for a patched release) in line with widely-used coordinated disclosure practice.
- Describes which release lines receive security fixes by referring to the existing SemVer policy and the GitHub Releases page, instead of hard-coding a table that ages out as new minors ship.
- Calls out what is explicitly out of scope (perf regressions, raw third-party `extern/` findings, exploits requiring already-elevated privileges) so we don't get noise reports here.

No code change.

## Test plan

- [x] File renders on GitHub.
- [ ] Maintainer review: adjust response-time targets, supported-versions phrasing, or add a fallback email contact if desired.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,77 @@
+# Security Policy
+
+The VSAG team takes the security of the library seriously. This document
+describes how to report vulnerabilities and which release lines we keep
+patched.
+
+## Reporting a Vulnerability
+
+**Please do not file a public GitHub issue for security-sensitive reports.**
+
+The preferred channel is GitHub's private vulnerability reporting:
+
+1. Navigate to the repository's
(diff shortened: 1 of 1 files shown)


wxyucs:
No need now.

PR #2112, opened 2026-05-28
feat(rabitq): support split hybrid IO and 3+5 base reorder

LightWant:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

None.

## What Changed

This PR extends the RaBitQ split-code path in two steps:

- Keeps the existing split datacell hybrid IO work: one-bit/filter code and supplement code can use different IO backends, and `rabitq_version=split_1bit_7bit` can be derived for `rabitq_split` configs.
- Adds configurable split filter bits so x+y layouts can filter with x bits and reorder with x+y bits. The main new path is `3+5bits`: 3-bit lower-bound/filter distance, 5-bit supplement scan during base-code reorder.
- Adds optimized 3-bit filter computation through SIMD/query LUT helpers.
- Adds split-code reorder hinting so 3+5 reorder can reuse the 3-bit filter distance and only scan the supplement bits for candidates.
- Adds runtime `rabitq_error_rate` parsing for lower-bound experiments and search statistics for filter/reorder distance and IO-count analysis.
- Updates RaBitQ docs and tests for split filter-bit parameters and split datacell behavior.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build --target unittests --parallel 8
./build/tests/unittests "RaBitQ Quantizer Parameter CheckCompatibility"
./build/tests/unittests "RaBitQ Split Code Storage"
./build/tests/unittests "[RaBitQSplitDataCell]"
./build/tests/unittests "[HGraphParameter]"
cmake --build build-release --target eval_performance --parallel 8

LC_ALL=C.utf8 LANG=C.utf8 LANGUAGE= ./build-release/tools/eval/eval_performance \
  /root/vsag-rabitq-split-hybrid-io/benchs/datasets/gist1m_rabitq_3plus5_m64_efc300.yml
```

Key benchmark observation on GIST1M with `max_degree=64` and `ef_construction=300`:
3+5 recall exceeded 1+7 recall at the tested `ef_search` points while scanning fewer supplement/reorder candidates, though QPS remains lower due to the heavier 3-bit filter path.

## Compatibility Impact

- API/ABI compatibility: Adds optional RaBitQ parameters and search statistics; existing default 1-bit split behavior remains supported.
- Behavior changes: Split RaBitQ indexes can choose `rabitq_bits_per_dim_filter` for x+y filtering/reorder layouts.

- Performance impact: Adds optimized 3-bit filter kernels and reduces supplement reads for 3+5 reorder; 3-bit filter computation is still heavier than 1-bit filtering in GIST1M experiments.
- Concurrency/thread-safety impact: No intended concurrency model changes.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/en/src/quantization/rabitq.md`, `docs/docs/zh/src/quantization/rabitq.md`, `docs/rabitq_split_1bit_7bit.md`

## Risk and Rollback

- Risk level: Medium
- Rollback plan: Revert the RaBitQ split/hybrid IO and split filter-bit commits to return split storage and reorder to the previous 1+7 behavior.

## Checklist

- [x] I have linked the relevant issue (required for `kind/bug` and `kind/feature`; not required for `kind/improvement`)


src/datacell/rabitq_split_datacell.h
```diff
@@ -329,8 +363,19 @@ class RaBitQSplitDataCell : public FlattenInterface {
...
+    void
+    InitIO(const IOParamPtr& one_bit_io_param, const IOParamPtr& supplement_io_param) {
+        this->one_bit_cell_->InitIO(SuffixIOParam(one_bit_io_param, "_onebit"));
+        // When a dedicated supplement IO param is supplied, do not append a
+        // suffix because the caller (typically the factory) has already picked
+        // an independent file path / IO type.
+        this->supplement_cell_->InitIO(supplement_io_param != nullptr
+                                           ? supplement_io_param
+                                           : SuffixIOParam(one_bit_io_param, "_supplement"));
     }
```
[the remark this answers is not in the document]
LightWant: Thanks for catching this — the failure mode you describe is correct: in the single-arg `InitIO` path the supplement cell would receive a `MemoryBlockIOParameter` while it is actually an `AsyncIO` storage, which crashes inside the dynamic_cast.

Fixed in dd312ab0 with the approach you suggested:

1. Added `std::string supplement_io_type_{};` to `RaBitQSplitDataCell`.
2. In the 4-arg constructor, when a dedicated `supplement_io_param` is supplied, record `supplement_io_param->GetTypeName()`.
3. In `InitIO(const IOParamPtr&)`, route through a new helper `RebuildSupplementIOParam` that — when `supplement_io_type_` is non-empty — clones the one-bit param's JSON, overwrites `type` with the recorded supplement type, suffixes the optional `file_path`, and reconstructs an IOParameter via the factory. When the recorded type is empty (homogeneous IO), it falls back to the original `SuffixIOParam` behaviour for backward compatibility.
4. The 2-arg `InitIO` overload also refreshes `supplement_io_type_` so subsequent single-arg calls (e.g. from `Deserialize`) reconstruct the correct subtype.

Verified locally:
- `./build/tests/unittests "[RaBitQSplitDataCell]"` — 4 cases / 7775 assertions pass, including the hybrid IO Serialize/Deserialize round-trip case.
- `./build/examples/cpp/323_index_hgraph_rabitq_split` — memory / disk / hybrid modes return identical top-10 results.

Marking as resolved.

LHT129 (commented):
Overall solid design — the dual template approach is clean and the auto-derive of rabitq_version is a nice UX win. Test coverage is thorough. A few suggestions inline.

src/datacell/rabitq_split_datacell.h
```diff
@@ -443,52 +493,93 @@ class RaBitQSplitDataCell : public FlattenInterface {
...
     uint64_t supplement_code_size_{0};
+    // Type name (e.g. "async_io") of the dedicated supplement IO when the
+    // caller supplies a separate `supplement_io_param` at construction time.
+    // Empty string means "supplement shares the same IO type as the one-bit
+    // storage" (the legacy single-IO behaviour). Recorded so that the
+    // single-parameter `InitIO(const IOParamPtr&)` overload (e.g. invoked
+    // from Deserialize) can rebuild a parameter of the correct concrete
+    // IOParameter subclass for `supplement_cell_` instead of feeding it the
+    // mismatched one-bit IO parameter type.
+    std::string supplement_io_type_{};
```
LHT129, line 684: `supplement_io_type_` is set at construction time and used by single-arg `InitIO()` to rebuild the correct IO parameter, but it is NOT written to the serialization stream.

This works today because the deserializing object is always constructed with the same params (same constructor → same type string). However, if future code ever constructs a "blank" instance and deserializes into it, `InitIO` will silently use the wrong IO type.

Consider persisting this field in `Serialize`/`Deserialize` for robustness.

src/datacell/rabitq_split_datacell.h
```diff
@@ -61,6 +89,13 @@ class RaBitQSplitDataCell : public FlattenInterface {
           InnerIdType id_count,
           QueryContext* ctx = nullptr) override {
         auto* comp = static_cast<Computer<RaBitQuantizer<metric>>*>(computer.get());
+        if constexpr (not OneBitIOTmpl::InMemory or not SupplementIOTmpl::InMemory) {
+            if (id_count > 1) {
```
LHT129, line 184: In hybrid mode (block_memory_io + async_io), this forces a `MultiRead` on the in-memory one-bit storage too, losing the zero-copy direct-pointer path.

For the primary hybrid use case (one-bit hot in RAM), you could split into:
- Direct-read loop for one-bit (in-memory, zero-copy)
- Batch `MultiRead` only for supplement (disk)

This would matter for large `id_count` during reorder.

src/datacell/flatten_datacell_test.cpp
```diff
@@ -145,3 +148,240 @@ TEST_CASE("RaBitQSplitDataCell direct split compute", "[ut][RaBitQSplitDataCell]
...
+            "rabitq_bits_per_dim_base": 4
+        }
+    })";
+    auto param_json = JsonType::Parse(param_str);
+    auto param = std::make_shared<FlattenDataCellParameter>();
+    param->FromJson(param_json);
+
+    IndexCommonParam common_param;
+    common_param.allocator_ = allocator;
+    common_param.dim_ = dim;
```
LHT129, line 341: The hybrid test creates `/tmp/vsag_rabitq_split_hybrid_ut_<pid>_*` files but never removes them. Consider adding RAII cleanup (or a scope guard) so CI does not accumulate leftover temp files over time.

src/datacell/flatten_datacell_parameter.cpp
```diff
@@ -31,13 +32,43 @@ FlattenDataCellParameter::FromJson(const JsonType& json) {
...
+        this->supplement_io_parameter = IOParameter::GetIOParameterByJson(supplement_json);
+    }
+
     CHECK_ARGUMENT(
         json.Contains(QUANTIZATION_PARAMS_KEY),
         fmt::format("flatten interface parameters must contains {}", QUANTIZATION_PARAMS_KEY));
-    this->quantizer_parameter =
-        QuantizerParameter::GetQuantizerParameterByJson(json[QUANTIZATION_PARAMS_KEY]);
+
+    // When the caller asks for the split codes layout, force the quantizer
```
LHT129, line 56: Silently overriding a user-supplied `rabitq_version` is the correct behavior here, but a `LOG_WARNING` when the user explicitly set a *different* value would aid debuggability (e.g. "user passed X, overriding to split_1bit_7bit because codes_type=rabitq_split").

LHT129 (commented):
Overall solid design — the dual template approach is clean and the auto-derive of rabitq_version is a nice UX win. Test coverage is thorough. A few suggestions inline.

src/datacell/flatten_interface.cpp
```diff
@@ -186,8 +186,43 @@ make_instance(const FlattenInterfaceParamPtr& param, const IndexCommonParam& com
...
+                if (base_type == IO_TYPE_VALUE_BLOCK_MEMORY_IO and
+                    supp_type == IO_TYPE_VALUE_ASYNC_IO) {
+#if HAVE_LIBAIO
+                    return std::make_shared<RaBitQSplitDataCell<metric, MemoryBlockIO, AsyncIO>>(
+                        param->quantizer_parameter,
+                        param->io_parameter,
+                        param->supplement_io_parameter,
+                        common_param);
+#else
+                    return std::make_shared<RaBitQSplitDataCell<metric, MemoryBlockIO, BufferIO>>(
```
wxyucs, line 217: This no-libaio fallback still passes `param->supplement_io_parameter`, whose type name is `async_io`, into `RaBitQSplitDataCell<..., BufferIO>`. `BufferIO(const IOParamPtr&)` does a `dynamic_pointer_cast<BufferIOParameter>` and then dereferences it, so an `AsyncIOParameter` becomes null here and the hybrid split config crashes instead of falling back to buffer IO. Please either instantiate the `AsyncIO` alias path consistently or rebuild/convert the supplement IO param to `buffer_io` before constructing the `BufferIO`-backed cell.

wxyucs (changes requested):
Thanks for the split datacell/hybrid IO work. I found one correctness issue in the no-libaio fallback path that should be fixed before merging. The other robustness/perf points already raised by LHT129 look reasonable to address separately.

src/datacell/rabitq_split_datacell.h
LightWant: Fixed in dd3ae922: `supplement_io_type_` is now persisted with `StreamWriter::WriteString` and restored with `StreamReader::ReadString` before the split storages are deserialized, so the single-arg `InitIO()` path keeps the correct supplement IO type after deserialize.

src/datacell/rabitq_split_datacell.h
LightWant: Addressed in dd3ae922: for the hot-one-bit/cold-supplement hybrid path, `Query()` / `QueryWithDistanceHint()` now use a dedicated supplement-only `MultiRead` path. The one-bit code is read directly from the in-memory storage, and IO stats count only the supplement reads.

src/datacell/flatten_datacell_test.cpp
LightWant: Fixed in dd3ae922: the hybrid IO unit test now has RAII cleanup for the `/tmp/vsag_rabitq_split_hybrid_ut_<pid>_*` files, and it cleans both before and after the test scope.

src/datacell/flatten_datacell_parameter.cpp
LightWant: Fixed in dd3ae922: when `codes_type=rabitq_split` overrides a different supplied `rabitq_version`, it now emits a `logger::warn` with the original and replacement versions.

src/datacell/flatten_interface.cpp
LightWant: Fixed in dd3ae922: the no-libaio fallback now rebuilds the supplement IO parameter as `buffer_io` before constructing `RaBitQSplitDataCell<..., BufferIO>`, and also accepts the factory-level `async_io -> buffer_io` fallback type. This avoids passing an `AsyncIOParameter` into `BufferIO`.

LightWant:
Updated in d2c434c0:

- Reworked the public x+y split config to use `base_quantization_type=rabitq`, `precise_quantization_type=rabitq`, `rabitq_bits_per_dim_base=x`, and `rabitq_bits_per_dim_precise=y`; omitting y keeps the existing standard RabitQ path.
- Updated docs, example, and parameter tests for the new config style.
- Renamed the split datacell storage member to `x_bit_cell_`, removed the standalone one-bit/supplement storage headers and tests, and kept the shared split code storage in `rabitq_split_datacell.h`.
- Added AVX2/AVX512 implementations for the query-LUT multi-bit RaBitQ dot path, plus tests.
- Clarified `filter_dist` and added split code size assertions.

Local verification:
- `make fmt`
- `git diff --check`
- `cmake --build build --target unittests functests --parallel 6`
- `./build/tests/unittests "RaBitQ FP32 multi-bit lookup SIMD Batch4 Compute Codes"`
- `./build/tests/unittests "RaBitQ Split Code Storage"`
- `./build/tests/unittests "HGraph maps RaBitQ*"`
- `./build/tests/unittests "[RaBitQSplitDataCell]"`
- `./build/tests/functests "[rabitq_split]"`
- `make lint` (clang-tidy-15 entry passed; current script selected 0 files)

LightWant:
Pushed CI fix in 7058c4a7.

The failing `Install GitHub CLI` step depended on the unauthenticated `https://api.github.com/repos/cli/cli/releases/latest` call. A transient GitHub API/server response makes `wget` exit 8 before the tarball URL is built.

Changed the three affected workflows to:
- reuse an existing `gh` if the container already has one,
- otherwise install pinned `gh` v2.75.0 directly from the release asset,
- add wget retry/timeout,
- use `install -m 0755` instead of `mv`.

Checked locally: `git diff --check`; verified the old latest-API installer path is removed from the affected workflows.

LightWant:
Documentation consolidation pushed in 43fa6aab.

- Merged the useful current-state content into `docs/docs/{en,zh}/src/quantization/rabitq_split.md`.
- Removed the historical `1+x` evolution narrative; the new pages describe only the current x+y design.
- Covered x/y configuration, bit-plane/datacell layout, x-bit lower bound, query LUT + AVX2/AVX512 path, y-only candidate reorder, hybrid IO, and serialization.
- Added both pages to the English/Chinese SUMMARY and quantization indexes.
- Updated the RaBitQ overview and HGraph search parameter tables to link to the new page and document the runtime `rabitq_error_rate` override.
- Deleted the tracked top-level guides under `docs/`; the third local x+y note was untracked and has also been removed after consolidation.

Validation: `git diff --check`, local Markdown link checks, EN/ZH navigation checks, and no remaining tracked Markdown references to the removed guides or `1+x` narrative. `mdbook` is not installed in the local environment, so the full book build was not run.

wxyucs (approved):
lgtm

Issue #2094, opened 2026-05-25
[feat](hgraph): add source ID support and cache export/import


LHT129:
### Affected area
hgraph

### Motivation
Currently, VSAG only supports integer-based IDs (labels) for vectors. Some use cases require string-based source identifiers that are more meaningful in downstream applications. Additionally, after building an index, users may want to export a lightweight cache containing source ID to neighbor mappings for fast lookups without traversing the graph.

This feature benefits users who:
- Need to associate human-readable string IDs with vectors
- Want to persist neighbor relationships for faster retrieval by source ID
- Need to share source ID mappings across multiple processes or machines

### Proposed solution
Add source ID support to the Dataset API and implement cache export/import for HGraph:

1. **Dataset API**: Add `SourceID(const std::string* source_id)` and `GetSourceID()` methods.

2. **Index API**: Add `ExportCache(std::ostream&)` and `ImportCache(std::istream&)` virtual methods.

3. **HGraphCache**: New class storing:
   - Mapping from inner_id to source_id string
   - Mapping from source_id to neighbor inner_ids
   
4. **HGraph Integration**: Store source IDs during `Add()`, `fullfill_cache()` populates neighbor info, serialize/deserialize via ExportCache/ImportCache.

Pseudocode:
```cpp
// Set source IDs when building
auto dataset = Dataset::Make();
std::vector<std::string> sids = {"img_001", "img_002"};
dataset->SourceID(sids.data())->NumElements(2)->...;
index->Add(dataset);

// Export/Import cache
std::ofstream out("cache.bin");
index->ExportCache(out);
```

### Alternatives considered
- Store source IDs as extra_info: Requires deserialization for each lookup, less efficient.
- Hash source IDs to integers: Loses semantic meaning and may cause collisions.

### API / ABI compatibility impact
additive only (no break)

### References
- include/vsag/dataset.h:471-485
- include/vsag/index.h:935-944
- src/algorithm/hgraph/hgraph_cache.h:1-52
- src/algorithm/hgraph/hgraph_cache.cpp:1-79

## Related issues
_None found._

PR #2114, opened 2026-05-28
feat(hgraph): accelerate Build by warm-starting from an imported cache

### LINKED ISSUE #2113: [feat](hgraph): accelerate Build by warm-starting from an imported cache
### Affected area
hgraph

### Motivation
After #2094 introduced `ExportCache(out)` / `ImportCache(in)` for HGraph and a
build-time cache keyed by `source_id` (string) -> selected neighbor inner ids,
the cache today is only used for runtime lookup. The exported cache, however,
contains a strong prior about graph structure: for any input vector whose
`source_id` is already present in the cache, its previously selected neighbours
are very likely to be a high-quality starting set for the new graph.

In production, users rebuild HGraph indexes regularly when the underlying
dataset shifts only slightly (e.g. periodic refresh of a recommendation index,
re-quantisation, swapping the bottom flatten codes, hyper-parameter tuning).
Each rebuild from scratch redoes O(N * ef_construction) work even though the
new graph is, structurally, very close to the previous one. We want a build
path that warm-starts from a previously exported cache and refines, instead of
building from scratch.

This benefits:
- HGraph users who periodically rebuild large indexes over near-identical
  datasets (faster build, same recall target).
- Workflows that swap the base flatten/quantization but want to keep the graph
  topology approximately stable.
- Two-pass build pipelines (cheap pilot build -> ExportCache -> full build with
  warm-start) for very large datasets.

### Proposed solution
Make `Build()` automatically take an accelerated path when `ImportCache()` has
populated the index's internal cache before the call. Concretely:

1. `Index::ExportCache` / `Index::ImportCache` are promoted to virtual methods
   returning `tl::expected<void, Error>`, with proper docstrings (additive,
   compatible).
2. `IndexImpl` forwards `ExportCache` / `ImportCache` to the inner index using
   the standard `SAFE_CALL` wrapper.
3. `HGraph::Build(data)` checks whether a cache has already been imported. If
   so it calls a new `build_with_cache(data)` path that:
   - **Warm-start.** For each input row, look up the cached neighbours by
     `source_id`. Hits get their neighbours seeded directly into the bottom
     graph; misses get an empty neighbour list. Inputs are partitioned into
     `hit_ids` and `missed_ids`.
   - **Two-phase refine.** Run a refine loop in two stages:
     - First on `hit_ids` with `use_self_as_entry = true` (so the warm-started
       neighbours act as the search frontier; small number of rounds).
     - Then on `missed_ids` with `use_self_as_entry = false` (entry from the
       global entry point; larger number of rounds, similar to a normal
       insertion).
   - Each refine round is internally split into:
     1. Parallel search + heuristic-pruning select (returns selected neighbours
        and their distances).
     2. Serial neighbour writeback under per-node mutex.
     3. Sharded reverse-edge install with distance reuse and O(M) dedup, to
        keep the bottom graph symmetric without N*M global locking.
   - Build the upper route graphs via ODescent over the sampled level ids
     (existing code path).
4. New unit tests under `tests/test_hgraph.cpp`:
   - End-to-end smoke: baseline build -> `ExportCache` -> fresh index ->
     `ImportCache` -> `Build` (mix of hits) -> sanity-check `KnnSearch` returns
     the self vector with ~zero distance.
   - All-missed: import a cache with one set of source_ids, then `Build` with a
     completely disjoint set of source_ids so every node takes the missed
     branch. Verifies the missed path does not crash and the index is
     searchable.

User-facing API surface (additive only):

```cpp
// Build a baseline index.
auto baseline = vsag::Factory::CreateIndex("hgraph", param).value();
baseline->Build(dataset_with_source_ids);

std::stringstream cache;
baseline->ExportCache(cache);     // existing in #2094

// Reuse the cache to accelerate a later Build on a similar dataset.
auto warmed = vsag::Factory::CreateIndex("hgraph", param).value();
warmed->ImportCache(cache);       // existing in #2094
warmed->Build(new_dataset);       // NEW: takes the warm-start + refine path
```

No new public method or parameter is introduced; the new behaviour is keyed
purely off whether `ImportCache` was called before `Build`.

### Alternatives considered
- **Add an explicit `BuildFromCache(data, cache)` API.** Rejected: doubles the
  surface area and forces every binding (`pyvsag`, `vsag` Node binding) to
  expose another entry point. Keying off the imported cache is API-additive.
- **Reuse the cached neighbours as final neighbours without refine.** Rejected:
  exported neighbours come from a graph built over a slightly different
  dataset / quantisation; using them as-is degrades recall. The two-phase
  refine guarantees the final graph respects the new flatten codes.
- **Single-phase refine over hits and misses uniformly.** Rejected: hits start
  with a strong frontier so few rounds suffice; misses need cold-start
  exploration. Splitting them lets each phase run with the right number of
  rounds and entry-point choice.

additive only (no break)

### References
- include/vsag/index.h:932-958
- src/algorithm/hgraph/hgraph.h:191-432
- src/algorithm/hgraph/hgraph_build.cpp:78-95
- src/algorithm/hgraph/hgraph_build.cpp:716-1330
- src/index/index_impl.h
- tests/test_hgraph.cpp (new test cases for cache-accelerated build)
- Prior cache framework: #2094

## Related issues
- #2094 [feat](hgraph): add source ID support and cache export/import (closed; introduces the cache primitives this proposal builds on)

---
_Drafted with AI assistant: CodeFuse:claude-sonnet-4.5_

LHT129:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

Fixes: #2113

## What Changed

- Promoted `Index::ExportCache` / `Index::ImportCache` to virtual methods
  returning `tl::expected<void, Error>` and added docstrings; `IndexImpl`
  forwards them to the inner index via `SAFE_CALL`.
- `HGraph::Build()` now detects a previously imported cache and dispatches
  to a new `build_with_cache()` path that warm-starts each input vector's
  neighbours from `cache_->neighbors_` (keyed by `source_id`) and refines
  them, instead of building from scratch.
- The refine pipeline runs in two phases: hit nodes use
  `use_self_as_entry = true` for a few rounds; missed nodes use
  `use_self_as_entry = false` for more rounds. Each round is split into
  parallel search + heuristic-pruning select, serial neighbour writeback
  under per-node mutex, and a sharded reverse-edge install with distance
  reuse and O(M) dedup.
- Upper route graphs are still built via ODescent over the sampled level
  ids.
- Added new helpers `collect_refine_candidates`,
  `select_refine_neighbors_with_distances`, `refine_nodes_two_phase`, and
  `has_loaded_cache()` on `HGraph`.
- `tests/test_hgraph.cpp` gets two new cache-accelerated build test cases
  (`[ft][hgraph][cache][pr]`): an end-to-end mostly-hits smoke test and
  an all-missed test (disjoint `source_id`s) that exercises the cold
  branch end-to-end.

The new behaviour is purely additive: it is keyed off whether
`ImportCache()` was called before `Build()`. No new public API or
parameter is introduced.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
Local iteration of the cache-accelerated build path used the two new test
cases under tests/test_hgraph.cpp tagged [ft][hgraph][cache][pr]:
  - "HGraph ExportCache + ImportCache + Build acceleration smoke test"
  - "HGraph ExportCache + ImportCache + Build miss-only path"
Both cover:
  - Build baseline -> ExportCache -> fresh index -> ImportCache -> Build
    on the same / disjoint source_ids.
  - KnnSearch sanity check that the warmed index returns the self vector
    with ~zero distance.
Full make fmt / lint / test / cov runs will be triggered through CI on
this PR.
```

## Compatibility Impact

- API/ABI compatibility: additive only. `Index::ExportCache` /
  `Index::ImportCache` change from non-virtual stubs to virtual methods
  returning `tl::expected<void, Error>`; existing callers that ignored the
  previous void return continue to compile because the throwing default
  implementation is preserved for indexes that don't override.
- Behavior changes: when `ImportCache()` has been called before `Build()`
  on an HGraph index, `Build()` now takes the warm-start + two-phase
  refine path instead of the from-scratch path. Indexes that never call
  `ImportCache()` see no behavioural change.

- Performance impact: improved on workloads that periodically rebuild a
  near-identical HGraph index — Build skips the cold construction loop
  and reuses cached neighbours as warm-starts.
- Concurrency/thread-safety impact: refine writebacks happen under
  per-node `LockGuard(neighbors_mutex_, id)`; reverse-edge install is
  sharded by `neighbor_id % shard_count` to remove the global lock.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other:

The user-facing API surface (`ExportCache` / `ImportCache`) was already
introduced and documented as part of #2094; this PR only adds an internal
acceleration path that triggers automatically when a cache has been
imported. No user-visible doc page needs to change.

## Risk and Rollback

- Risk level: low. The new code path is gated by `has_loaded_cache()`;
  every existing call site that does not call `ImportCache()` before
  `Build()` continues to take the original path unchanged.
- Rollback plan: reverting this commit restores the previous behaviour
  (Build always from scratch). The on-disk cache format produced by
  `ExportCache()` is unchanged and remains compatible with the cache
  introduced in #2094.

## Checklist

### DIFF
--- a/include/vsag/index.h
+++ b/include/vsag/index.h
@@ -932,14 +932,31 @@ public:
         throw std::runtime_error("Index not support check id exist");
     }
 
-    void
-    ExportCache(std::ostream& out_stream) {
-        throw std::runtime_error("Index not support export cache");
+    /**
+     * @brief Export the index's build-time cache (e.g. graph neighbors keyed by
+     * source id) into the given stream so it can be reused to accelerate a
+     * later Build().
+     *
+     * @param out_stream Destination stream to write the cache blob into.
+     * @return tl::expected<void, Error>; an Error is returned if the index
+     * does not support exporting a build cache or the export fails.
+     */
+    virtual tl::expected<void, Error>
+    ExportCache(std::ostream& out_stream) const {
+        throw std::runtime_error("Index doesn't support ExportCache");
     }
 
-    void
+    /**
+     * @brief Import a previously exported build-time cache. Once imported, the
+     * next Build() call will use the cache to accelerate construction.
+     *
+     * @param in_stream Source stream containing a blob produced by ExportCache.
+     * @return tl::expected<void, Error>; an Error is returned if the index
+     * does not support importing a build cache or the import fails.
+     */
+    virtual tl::expected<void, Error>
     ImportCache(std::istream& in_stream) {
-        throw std::runtime_error("Index not support import cache");
+        throw std::runtime_error("Index doesn't support ImportCache");
     }
 
 public:
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -18,6 +18,8 @@
 #include <random>
 #include <shared_mutex>
 #include <string>
+#include <string_view>
+#include <unordered_map>
 
 #include "../inner_index_interface.h"
 #include "common.h"
@@ -191,7 +193,7 @@ public:
     SetImmutable() override;
 
     void
-    ExportCache(std::ostream& out_stream) override;
+    ExportCache(std::ostream& out_stream) const override;
 
     void
     ImportCache(std::istream& in_stream) override;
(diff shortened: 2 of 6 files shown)


wxyucs (approved):
lgtm

PR #2141, opened 2026-06-04
feat: use tl::unexpected in Index default implementations

### LINKED ISSUE #2140: feat: use tl::unexpected in Index default implementations
## Summary

Replace throw std::runtime_error with LOG_ERROR_AND_RETURNS in Index class virtual method default implementations that return tl::expected<T, Error>.

## Problem

The Index base class had ~36 virtual methods with default implementations that used  to indicate unsupported operations, even though the return type was . This created an inconsistency in error handling - callers had to handle both exceptions and expected error returns.

## Solution

Changed default implementations to use  which returns a proper  value. This allows callers to consistently use  and  to check results.

Methods with non-expected return types (GetIndexType, GetNumberRemoved, etc.) and CheckFeature (returns bool) still use throw as appropriate.

## Changes

- Modified  to use LOG_ERROR_AND_RETURNS in default implementations
- Unified error message format to "Index does not support <operation>"

## Testing

- Release build passes
- Existing unit tests in src/index/index_impl_test.cpp validate the expected error behavior

LHT129:
## Summary
- Replace throw std::runtime_error with LOG_ERROR_AND_RETURNS in Index class virtual method default implementations
- Provides consistent error handling via tl::expected<T, Error>

Fixes #2140

### DIFF
--- a/include/vsag/index.h
+++ b/include/vsag/index.h
@@ -1,4 +1,3 @@
-
 // Copyright 2024-present the vsag project
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -90,7 +89,7 @@ public:
      */
     [[nodiscard]] virtual IndexType
     GetIndexType() const {
-        throw std::runtime_error("Index not support GetIndexType");
+        throw std::runtime_error("Index does not support GetIndexType");
     }
 
 public:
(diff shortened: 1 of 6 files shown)


LHT129:
Thank you for the review! I have addressed the formatting issues by removing the trailing semicolons after function bodies and updated the documentation comments to reflect the new tl::unexpected-based error handling. 

Regarding the LOG_ERROR_AND_RETURNS macro suggestion: The macro depends on Options::Instance().logger(), which causes linking errors in header-only default implementations. After consulting with the maintainers, we decided to use return tl::unexpected(Error(...)) directly to avoid these linking issues while still providing the expected-based error handling.

For CheckFeature: The task description indicates that methods returning bool should keep throwing exceptions, but CheckFeature is specifically a feature capability check method where returning false (indicating the feature is not supported) is more semantically correct than throwing an exception. This aligns with the common pattern in feature detection APIs.

wxyucs (approved):
lgtm

PR #2158, opened 2026-06-08
perf(io): optimize CRC32 with lookup table and reduce ReaderIO heap allocation

jfeng18:
## Summary

- **CRC32 checksum**: replace bit-by-bit implementation (8 branch-iterations per byte) with a 256-entry lookup table (1 table lookup per byte), ~8x speedup for serialization/deserialization checksum computation. Also fixes a latent sign-extension bug for non-ASCII bytes.
- **ReaderIO::MultiReadImpl**: replace `std::vector<uint64_t>` heap allocation with a 1KB stack buffer for counts <= 128, with `std::unique_ptr` heap fallback for larger counts to maintain exception safety.

## Test plan

- [x] Footer tests: 21 assertions passed
- [x] ReaderIO tests: 1,031 assertions passed
- [x] Full unit test suite: 441 cases / 85M assertions passed (release)
- [x] Adversarial review: CRC32 table verified against Python-generated reference (256 entries, all match), known CRC32 check value 0xCBF43926 confirmed
- [ ] CI format/lint check


jfeng18:
Hi maintainers, could you please add `kind/improvement` and `version/1.0` labels? I don't have label permissions as a fork contributor. Thanks!

wxyucs:
Hi @jfeng18, please check the review comments and resolve them.

src/io/reader_io.cpp
```diff
@@ -91,11 +92,20 @@ ReaderIO::MultiReadImpl(uint8_t* datas,
...
+    constexpr uint64_t kStackThreshold = 128;
+    uint64_t local_buf[kStackThreshold];
+    std::unique_ptr<uint64_t[]> heap_buf;
+    uint64_t* real_offsets;
+    if (count <= kStackThreshold) {
+        real_offsets = local_buf;
+    } else {
+        heap_buf.reset(new uint64_t[count]);
+        real_offsets = heap_buf.get();
+    }
```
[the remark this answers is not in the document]
jfeng18: Good point — fixed in commit 97726ea2. Now using `allocator_->Allocate/Deallocate` for consistency with the rest of the IO layer.

src/storage/serialization.h
```diff
@@ -156,19 +156,52 @@ class Footer {
     virtual ~Footer() = default;
 
 private:
-    // TODO(wxyu): optimize performance
+    static const uint32_t*
+    crc32_table() {
+        static const uint32_t table[256] = {
```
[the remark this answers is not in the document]
jfeng18: Done — changed to `static constexpr uint32_t table[256]` in the same commit (97726ea2).

jfeng18:
Hi @wxyucs, review comments addressed in commit 97726ea2:
- **Use allocator_ (high)**: Fixed — now uses `allocator_->Allocate/Deallocate` instead of raw `new[]/delete[]`
- **static constexpr (medium)**: Fixed — changed to `static constexpr uint32_t table[256]`

LHT129 (commented):
CRC32 lookup table is correct (canonical zlib reflected table for poly 0xEDB88320; table[0]=0x00000000, table[1]=0x77073096, table[128]=0xEDB88320, table[255]=0x2D02EF8D all match) and the table loop is the exact equivalent of the bit-by-bit reflected algorithm. The static table is constexpr/constant-initialized, so it is thread-safe.

HOWEVER there is a serious backward-compat concern: the OLD code did `crc ^= byte` where `byte` is a signed `const char&`, so bytes >= 0x80 sign-extend into the high bits of crc and produce a NON-canonical checksum; the new code masks with uint8_t and computes the canonical CRC32. They differ (by 0x00FFFFFF after the first high-bit byte) for any payload containing bytes >= 0x80 on platforms where char is signed (x86/x64 Linux/macOS default). Footer checksums written by the old code will then fail validation under the new code. This should be treated as a serialization/footer format-version change with migration, not a transparent perf optimization — unless the checksum is never persisted across versions (please confirm).

Separately, the ReaderIO stack/heap optimization is correct (the count<=128 boundary is fine) but not exception-safe: if MultiRead throws, heap_buf leaks. Recommend an RAII guard.

src/storage/serialization.h
```diff
@@ -156,19 +156,57 @@ class Footer {
...
-        const uint32_t polynomial = 0xEDB88320;
+        const uint32_t* table = crc32_table();
         uint32_t crc = 0xFFFFFFFF;
-
         for (const char& byte : bytes) {
-            crc ^= byte;
-            for (uint64_t j = 0; j < 8; ++j) {
-                crc = (crc >> 1) ^ ((crc & 1) == 1 ? polynomial : 0);
-            }
+            crc = table[(crc ^ static_cast<uint8_t>(byte)) & 0xFF] ^ (crc >> 8);
```
LHT129, line 242: Not bit-for-bit identical to the old impl. The old loop did `crc ^= byte` with `byte` a signed `const char&`; bytes >= 0x80 sign-extend to 0xFFFFFFxx and XOR 0xFFFFFF00 into crc's high bits, and the 8 right-shifts propagate them, so the old code produced a NON-canonical CRC32 for high-bit data. This line correctly masks via static_cast<uint8_t>(byte), producing canonical CRC32. Concretely with crc=0xFFFFFFFF, byte=0x80: old -> table[0x7F]; new -> table[0x7F] ^ 0x00FFFFFF — they differ by 0x00FFFFFF. Footers written by the old code will fail validation under the new code (and vice versa) for essentially all binary data. If this checksum is persisted, bump the serialization/footer format version and handle migration, or document that existing data is invalidated.

src/io/reader_io.cpp
```diff
@@ -91,11 +91,24 @@ ReaderIO::MultiReadImpl(uint8_t* datas,
...
+        real_offsets = local_buf;
+    } else {
+        heap_buf = static_cast<uint64_t*>(allocator_->Allocate(count * sizeof(uint64_t)));
+        real_offsets = heap_buf;
+    }
     for (uint64_t i = 0; i < count; ++i) {
         real_offsets[i] = start_ + offsets[i];
     }
-    return reader_->MultiRead(datas, sizes, real_offsets.data(), count);
+    bool ret = reader_->MultiRead(datas, sizes, real_offsets, count);
```
LHT129: Not exception-safe: if reader_->MultiRead(...) throws, heap_buf (from allocator_->Allocate) leaks since the Deallocate below is skipped; the old std::vector freed it on unwind. Use an RAII guard so deallocation runs on all paths. The stack-vs-heap selection and the count<=128 boundary (indices 0..127 fit local_buf[128]) are correct.

src/storage/serialization.h
jfeng18: Good catch. Yes, the new code produces standard CRC32 while the old bit-by-bit code had a sign-extension bug for bytes >= 0x80. This is backward compatible because the checksummed content is JSON metadata (ASCII-only, all bytes < 0x80), so both old and new produce identical results for all existing serialized data.

src/io/reader_io.cpp
jfeng18: Fixed in commit d7fe7ee1 — now using `std::unique_ptr` with a custom deleter that calls `allocator_->Deallocate`, ensuring cleanup on both normal return and exception paths.

jfeng18: Good catch. You're right — if `MultiRead` throws, the allocator-allocated buffer leaks since `Deallocate` is skipped. I'll add an RAII guard (similar to the fix in PR #2214) or use a unique_ptr with custom deleter to ensure cleanup on all paths.

src/storage/serialization.h
jfeng18: Excellent analysis. You're absolutely correct — the sign-extension behavior difference means old and new CRC32 produce different checksums for high-bit bytes. This breaks backward compatibility for persisted footers.

I'll either:
1. Revert to the old (non-canonical) CRC32 to maintain compatibility, or
2. Bump the serialization format version and document the breaking change

Which approach do you recommend? Given this is an optimization PR, option 1 (preserve compatibility) seems safer unless we want to use this as an opportunity to fix the CRC32 implementation.

src/io/reader_io.cpp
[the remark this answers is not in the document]
jfeng18: Good point. Using the custom allocator is more consistent with the rest of the codebase. I'll update the heap allocation to use `allocator_->Allocate()` with a custom deleter as you suggested.

src/storage/serialization.h
jfeng18: Verified this instead of assuming: the footer JSON contains user-controlled strings (`index_name`, index params) and nlohmann's `dump()` keeps UTF-8 bytes unescaped, so bytes >= 0x80 can occur in persisted footers — my earlier ASCII-only argument was wrong. Fixed in b716964: `Footer::Parse` now verifies the canonical CRC32 first and falls back to the legacy sign-extending value, so footers written by old versions still load; the write path stays canonical, matching `StreamHeader`. Added discriminating tests: a legacy-checksum footer with non-ASCII metadata must parse, and a checksum matching neither algorithm is still rejected. Trade-off: dual-accept raises the theoretical false-accept bound from 2^-32 to at most 2^-31, acceptable for a load-time check.

PR #2164, opened 2026-06-08
fix(quantization): guard TransformQuantizer ProcessQueryImpl against repeated allocation

### LINKED ISSUE #2163: [bug](quantizer): memory leak in TransformQuantizer::ProcessQueryImpl on repeated SetQuery calls
## Description

`TransformQuantizer::ProcessQueryImpl` unconditionally allocates memory to `computer.inner_computer_->buf_` without checking if it is already non-null. If `SetQuery()` is called more than once on the same `Computer` object, the previously allocated buffer is leaked.

## Root Cause

In `src/quantization/transform_quantization/transform_quantizer.h`, the allocation:

```cpp
computer.inner_computer_->buf_ =
    reinterpret_cast<uint8_t*>(this->allocator_->Allocate(this->query_code_size_));
```

overwrites the pointer without freeing the old allocation. All other quantizer implementations (FP32, INT8, SQ8Uniform, ProductQuantizer, PQFastScan, RaBitQ) guard this with `if (computer.buf_ == nullptr)`.

## Impact

Each repeated `SetQuery` call leaks `query_code_size_` bytes (typically hundreds of bytes to several KB depending on transform chain configuration). Affects all indexes using TransformQuantizer (PCA-configured HGraph/IVF).

## Fix

Add a nullptr guard before allocation, consistent with other quantizer implementations.

LHT129:
## Summary

- Add nullptr check before allocating `inner_computer->buf_` in `TransformQuantizer::ProcessQueryImpl`, preventing memory leak on repeated `SetQuery` calls
- Consistent with all other quantizer implementations (FP32, INT8, SQ8Uniform, etc.)
- Add regression test exercising repeated `SetQuery` on the same Computer object

Fixes: #2163

### DIFF
--- a/src/quantization/transform_quantization/transform_quantizer.h
+++ b/src/quantization/transform_quantization/transform_quantizer.h
@@ -284,8 +284,10 @@ TransformQuantizer<QuantTmpl, metric>::ProcessQueryImpl(
     const float* query, Computer<TransformQuantizer>& computer) const {
     // 0. allocate
     try {
-        computer.inner_computer_->buf_ =
-            reinterpret_cast<uint8_t*>(this->allocator_->Allocate(this->query_code_size_));
+        if (computer.inner_computer_->buf_ == nullptr) {
+            computer.inner_computer_->buf_ =
+                reinterpret_cast<uint8_t*>(this->allocator_->Allocate(this->query_code_size_));
+        }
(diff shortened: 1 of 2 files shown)


LHT129:
Closing in favor of #2165, which addresses this issue.

PR #2168, opened 2026-06-08
fix: guard extra_info pointer arithmetic against nullptr in BruteForce::Add

### LINKED ISSUE #2167: [bug](bruteforce): Add performs undefined pointer arithmetic on nullptr extra_info
In `src/algorithm/bruteforce/bruteforce.cpp`, `BruteForce::Add` computes `extra_info + j * extra_info_size` without checking if `extra_info` is `nullptr`.

When `GetExtraInfos()` returns `nullptr` (the common case when no extra info is provided), this expression is undefined behavior per the C++ standard (pointer arithmetic on null pointer).

The same pattern is already correctly handled in `warp.cpp` (line 189):
```cpp
extra_info == nullptr ? nullptr : extra_info + i * extra_info_size
```

**Affected code**: `src/algorithm/bruteforce/bruteforce.cpp`, lines 126 and 132 (in the `Add` method).

**Impact**: Undefined behavior that may cause issues with aggressive compiler optimizations or sanitizers.

LHT129:
Fixes: #2167

## Summary

When `extra_info` is `nullptr` (the common case when no extra info is provided), `BruteForce::Add` performs `nullptr + j * extra_info_size`, which is undefined behavior in C++.

This PR adds a null check before the pointer arithmetic, consistent with the pattern already used in `warp.cpp:189`.

## Changes

- `src/algorithm/bruteforce/bruteforce.cpp`: Extract `ei_ptr` with null check before passing to `GeneralEnqueue` / `add_func`

## Test plan

- [x] Existing bruteforce functional tests pass (28 test cases, 715363 assertions)
- [x] Release build succeeds

### DIFF
--- a/src/algorithm/bruteforce/bruteforce.cpp
+++ b/src/algorithm/bruteforce/bruteforce.cpp
@@ -118,18 +118,17 @@ BruteForce::Add(const DatasetPtr& data, AddMode mode) {
                 continue;
             }
         }
+        const auto* ei_ptr = extra_info == nullptr ? nullptr : extra_info + j * extra_info_size;
         if (this->thread_pool_ != nullptr) {
             auto future = this->thread_pool_->GeneralEnqueue(add_func,
                                                              vectors + j * dim_,
                                                              label,
                                                              attrs == nullptr ? nullptr : attrs + j,
-                                                             extra_info + j * extra_info_size);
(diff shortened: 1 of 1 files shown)


wxyucs (approved):
lgtm

PR #2171, opened 2026-06-08
fix(storage): add bounds validation in Footer::Parse

### LINKED ISSUE #2170: fix(storage): heap out-of-bounds read in Footer::Parse on corrupted files
## Description

`Footer::Parse` in `src/storage/serialization.cpp` reads `metadata_string_length` from the binary footer data without validating that the value falls within the buffer bounds. If a serialized index file is truncated or corrupted, this can result in a heap out-of-bounds read when constructing `metadata_string` or copying the checksum.

## Steps to Reproduce

Load a VSAG index from a file where the footer region has been truncated to fewer than 20 bytes, or where `metadata_string_length` has been corrupted to a value exceeding the actual buffer size.

## Expected Behavior

`Footer::Parse` should return `nullptr` gracefully when the footer data is malformed.

## Actual Behavior

Undefined behavior (heap OOB read) — potentially crashes or reads garbage data.

## Proposed Fix

Add bounds validation before the `memcpy` and `std::string` construction:
1. Check `length >= 20` (minimum footer overhead: 8B magic + 8B length + 4B checksum)
2. Check `metadata_string_length <= length - 20`

PR: #TBD

jfeng18:
## Summary

Add bounds validation in `Footer::Parse` to prevent heap out-of-bounds read on corrupted/truncated serialized index files.

Fixes: #2170

## Changes

- Validate `length >= 20` before reading inner fields (minimum overhead: 8B magic + 8B metadata_length + 4B checksum)
- Validate `metadata_string_length <= length - 20` before constructing the metadata string
- Remove unused `string_buffer` allocation (dead code)

## Test plan

- [x] Footer tests: 21 assertions passed
- [x] Full unit tests: 441 cases / 85M assertions passed (release)
- [x] `make fmt` clean

### DIFF
--- a/src/storage/serialization.cpp
+++ b/src/storage/serialization.cpp
@@ -80,8 +80,16 @@ Footer::Parse(StreamReader& reader) {
     }
     // no popseek, continue to parse
 
+    if (length < 20) {
+        reader.PopSeek();
+        return nullptr;
+    }
     uint64_t metadata_string_length = 0;
     memcpy(&metadata_string_length, meta_buffer.data() + 8, 8);
+    if (metadata_string_length > length - 20) {
+        reader.PopSeek();
+        return nullptr;
+    }
     std::string metadata_string(meta_buffer.data() + 16, metadata_string_length);
     uint32_t checksum;
     memcpy(&checksum, meta_buffer.data() + 16 + metadata_string_length, 4);


jfeng18:
Hi maintainers, could you please add `kind/bug` and `version/1.0` labels? I don't have label permissions as a fork contributor. Thanks!

src/storage/serialization.cpp
```diff
@@ -80,9 +80,16 @@ Footer::Parse(StreamReader& reader) {
...
     // no popseek, continue to parse
 
+    if (length < 20) {
+        reader.PopSeek();
+        return nullptr;
+    }
     uint64_t metadata_string_length = 0;
-    std::vector<char> string_buffer(length);
     memcpy(&metadata_string_length, meta_buffer.data() + 8, 8);
+    if (metadata_string_length > length - 20) {
```
LHT129, line 89: Bounds check is correct. Passing gives `metadata_string_length <= length - 20`, simultaneously protecting the string read `[16, 16+L)` and the checksum read `[16+L, 20+L)` (end <= length). Boundary case `L == length-20` correctly allowed, and `16 + metadata_string_length + 4` cannot overflow since it is <= length. Relies on `meta_buffer.size() >= length`; please confirm meta_buffer is sized to `length` upstream.

src/storage/serialization.cpp
```diff
@@ -80,9 +80,16 @@ Footer::Parse(StreamReader& reader) {
     }
     // no popseek, continue to parse
 
+    if (length < 20) {
```
LHT129, line 83: Good ordering: this `length < 20` early-return runs before any `length - 20` evaluation, so the unsigned subtraction below cannot underflow.

LHT129 (commented):
The bounds-validation fix is correct and complete for the accesses it guards.

- Underflow: the `length < 20` guard runs before any `length - 20` evaluation, so the unsigned subtraction cannot underflow.
- Overflow: passing the check implies `metadata_string_length <= length - 20`, hence `16 + metadata_string_length <= length - 4` and `20 + metadata_string_length <= length`, so neither wraps uint64_t.
- Off-by-one: the boundary `metadata_string_length == length - 20` is allowed; the checksum read then spans `[length-4, length)`, exactly the last 4 bytes (exclusive end == length).
- Coverage: both post-check accesses are guarded — the string ctor reads `[16, 16+L)` (end < length) and the checksum memcpy reads `[16+L, 20+L)` (end <= length). No other access follows the check.

Correctness rests on `meta_buffer.size() >= length` (the check uses `length` as the buffer extent), which isn't shown in this diff but is strongly implied by the removed `string_buffer(length)` and the existing `data()+8`/`data()+16` reads. Worth a sanity check that meta_buffer is sized to `length` upstream; under that assumption the fix fully closes the OOB read. Nice cleanup removing the dead string_buffer vector.

jfeng18:
Hi @LHT129, thanks for the review! Confirmed that `meta_buffer` is sized to `length` (line 72: `std::vector<char> meta_buffer(length)`), so all accesses are within bounds.

Could you also help add `kind/bug` and `version/1.0` labels? Mergify is blocking on them.

jfeng18:
Hi, could a maintainer please add the `kind/bug` and `version/1.0` labels? The Mergify merge protection is currently blocked on these. Thanks!

src/storage/serialization.cpp
jfeng18: Confirmed — `meta_buffer` is constructed as `std::vector<char> meta_buffer(length)` at line 72, so `meta_buffer.size() == length` holds. All subsequent accesses stay within `[0, length)`: the bounds check ensures `16 + metadata_string_length + 4 <= length`.

jfeng18:
Gentle ping — could a maintainer please add the required `kind/*` and `version/1.0` labels when you get a chance? CI is green on this one. Thanks!

wxyucs (approved):
lgtm

src/storage/serialization.cpp
jfeng18: Thanks for the verification. Yes, `meta_buffer` is sized to `length` by the caller (confirmed in `Deserialize` at line 68 where it's allocated to `footer.meta_size_`). The bounds check correctly protects both the metadata string read `[16, 16+L)` and the checksum read `[16+L, 20+L)`.

src/storage/serialization.cpp
jfeng18: Correct. The early return on line 83 ensures we never reach the subtraction on line 89 when `length < 20`, preventing underflow.

jfeng18:
Closing — superseded by #2326 which provides stricter validation (exact equality check, correct 36-byte minimum, stream-length upper bound). Thanks @LHT129 for the more thorough fix!

PR #2178, opened 2026-06-09
fix(allocator): use throw instead of throw e to prevent exception slicing

### LINKED ISSUE #2177: Fix exception slicing in Allocator::New template method
## Description

In `include/vsag/allocator.h`, the `Allocator::New` template method uses `throw e;` to rethrow a caught exception in the catch block. This causes **exception slicing**: the derived exception type (e.g. `std::bad_alloc`, `std::runtime_error`) is copied as `std::exception`, losing the original type information.

## Impact

- Callers catching specific exception types (e.g. `catch (std::bad_alloc&)`) will not match.
- The original `what()` message may be lost depending on implementation.
- Callers cannot perform precise error handling based on exception type.

## Fix

Change `throw e;` to `throw;` (rethrow without argument), which preserves the original exception object and its full type information.

## Reference

- [C++ Core Guidelines E.15](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-throw): Throw by value, catch by reference

LHT129:
Fixes #2177

## Summary

Fix exception slicing in `Allocator::New` template method by changing `throw e;` to `throw;`.

## Problem

The catch block in `Allocator::New` (`include/vsag/allocator.h`) used `throw e;` which copies the caught `std::exception&` by value. This slices any derived exception type (e.g. `std::bad_alloc`, `std::runtime_error`) down to `std::exception`, preventing callers from catching specific exception types.

## Fix

Changed to `throw;` (parameterless rethrow) which preserves the original exception object and its complete type information.

## Testing

- All allocator-related unit tests pass (10 test cases, 27 assertions)
- Release build succeeds
- `make fmt` clean

## Reference

- [C++ Core Guidelines E.15](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-throw)

### DIFF
--- a/include/vsag/allocator.h
+++ b/include/vsag/allocator.h
@@ -101,7 +101,7 @@ public:
             return (T*)::new (p) T(std::forward<Args>(args)...);
         } catch (std::exception& e) {
             Deallocate(p);
-            throw e;
+            throw;
         }
     }
 


wxyucs (approved):
lgtm

PR #2183, opened 2026-06-09
fix(c_api): return error when index is NULL in vsag_index_add and vsag_index_train

### LINKED ISSUE #2182: [bug](c_api): vsag_index_add and vsag_index_train silently return success when index is NULL
When `vsag_index_add` or `vsag_index_train` receives a NULL index pointer, the functions silently return the global `success` object instead of reporting an error. This is inconsistent with `vsag_index_build`, which correctly returns `make_error("index is NULL")`.

This makes it impossible for C API callers to detect that the operation was silently skipped due to a null handle.

**Fix**: Replace `return success;` with `return make_error("index is NULL");` in both functions, matching the existing pattern used by `vsag_index_build`, `vsag_index_clone`, and `vsag_index_export_model`.

**Files affected**: `src/vsag_c_api.cpp` (lines 136, 160)

LHT129:
Fixes #2182

## Summary

When `vsag_index_add` or `vsag_index_train` received a NULL index pointer, they silently returned the global `success` object instead of reporting an error. This made it impossible for C API callers to detect that the operation was skipped.

## Changes

- Replace `return success;` with `return make_error("index is NULL");` in `vsag_index_add` (line 136)
- Replace `return success;` with `return make_error("index is NULL");` in `vsag_index_train` (line 160)

This matches the existing error handling pattern used by `vsag_index_build`, `vsag_index_clone`, and `vsag_index_export_model`.

## Test

All 13 C API unit tests pass (8279 assertions).

### DIFF
--- a/src/vsag_c_api.cpp
+++ b/src/vsag_c_api.cpp
@@ -133,7 +133,7 @@ vsag_index_add(vsag_index_t index,
             auto add_result = vsag_index->index_->Add(dataset);
             VSAG_CHECK_RESULT(add_result);
         }
-        return success;
+        return make_error("index is NULL");
     } catch (const std::exception& e) {
         return make_error(e);
     }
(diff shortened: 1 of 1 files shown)


wxyucs (approved):
lgtm

PR #2195, opened 2026-06-09
perf(sindi): grow sparse term list safely

Roxanne0321:
## Problem

When `remap_term_ids` is enabled, compact term ids grow globally while each SINDI window owns a fresh `SparseTermDataCell`. The previous exact-capacity resize strategy could repeatedly migrate large term arrays as later windows encountered larger compact term ids, causing severe build slowdown.

## Changes

- Grow `SparseTermDataCell` term lists geometrically instead of resizing exactly to every requested capacity.
- Raise the SINDI `term_id_limit` upper bound from 10,000,000 to 50,000,000.
- Update SINDI parameter tests and English/Chinese index docs for the new limit.

## Effect

On the sv750w build probe, the first 7 batches improved from an average `add_sec` of 287.802s before the change to 3.400s after the change. Window 6 improved from 489.486s to 3.588s. The optimized run completed all 7,256,421 docs with `build_done elapsed_sec=1736.506`.

## Tests

- `make release`


src/datacell/sparse_term_datacell.cpp
```diff
@@ -283,9 +283,13 @@ SparseTermDataCell::ResizeTermList(InnerIdType new_term_capacity) {
     if (new_term_capacity <= term_capacity_) {
         return;
     }
-    Vector<std::unique_ptr<Vector<uint16_t>>> new_ids(new_term_capacity, allocator_);
-    Vector<std::unique_ptr<Vector<uint8_t>>> new_datas(new_term_capacity, allocator_);
-    Vector<uint32_t> new_sizes(new_term_capacity, 0, allocator_);
+    InnerIdType new_capacity = term_capacity_ == 0 ? new_term_capacity : term_capacity_;
+    while (new_capacity < new_term_capacity) {
+        new_capacity *= 2;
+    }
```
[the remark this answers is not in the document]
Roxanne0321: Valid finding. I added an overflow guard before doubling and fall back to the requested capacity when the next double would overflow, so normal geometric growth is preserved while the loop cannot wrap or hang.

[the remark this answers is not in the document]
Roxanne0321: Valid finding. Addressed with the same overflow guard in ResizeTermList: normal requests still use geometric growth, and only near the InnerIdType limit do we cap at the requested capacity to avoid wraparound.

src/datacell/sparse_term_datacell_test.cpp
```diff
@@ -84,10 +84,10 @@ TEST_CASE("SparseTermDatacell Basic Test", "[ut][SparseTermDatacell]") {
...
         data_cell->InsertVector(sparse_vectors[i], i);
     }
-    REQUIRE(data_cell->term_capacity_ == exp_id_size);
-    REQUIRE(data_cell->term_ids_.size() == exp_id_size);
-    REQUIRE(data_cell->term_datas_.size() == exp_id_size);
-    for (auto i = 0; i < data_cell->term_capacity_; i++) {
+    REQUIRE(data_cell->term_capacity_ >= exp_id_size);
+    REQUIRE(data_cell->term_ids_.size() == data_cell->term_capacity_);
+    REQUIRE(data_cell->term_datas_.size() == data_cell->term_capacity_);
+    for (auto i = 0; i < exp_id_size; i++) {
```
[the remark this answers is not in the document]
Roxanne0321: Valid point. Added assertions for the over-allocated tail range to verify term_ids_ and term_datas_ stay nullptr after resizing.

src/algorithm/sindi/sindi_parameter.cpp
```diff
@@ -26,8 +26,8 @@ SINDIParameter::FromJson(const JsonType& json) {
         term_id_limit = json[SPARSE_TERM_ID_LIMIT].GetInt();
 
         CHECK_ARGUMENT(
-            (0 < term_id_limit and term_id_limit <= 10'000'000),
-            fmt::format("term_id_limit must in (0, 10'000'000], but now is {}", term_id_limit));
+            (0 < term_id_limit and term_id_limit <= 50'000'000),
+            fmt::format("term_id_limit must in (0, 50'000'000], but now is {}", term_id_limit));
```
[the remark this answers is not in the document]
Roxanne0321: Valid cleanup. Updated the message to use clearer grammar: term_id_limit must be in the allowed range, but got the provided value.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -283,9 +283,17 @@ SparseTermDataCell::ResizeTermList(InnerIdType new_term_capacity) {
     if (new_term_capacity <= term_capacity_) {
         return;
     }
-    Vector<std::unique_ptr<Vector<uint16_t>>> new_ids(new_term_capacity, allocator_);
-    Vector<std::unique_ptr<Vector<uint8_t>>> new_datas(new_term_capacity, allocator_);
-    Vector<uint32_t> new_sizes(new_term_capacity, 0, allocator_);
+    InnerIdType new_capacity = term_capacity_ == 0 ? new_term_capacity : term_capacity_;
+    while (new_capacity < new_term_capacity) {
```
LHT129, line 287: Minor: when term_capacity_ == 0 (first allocation) new_capacity is initialized to exactly new_term_capacity and the while-loop body never runs, so the first allocation gets no growth headroom. If the first insert path requests a small capacity that is then grown one-by-one, the first few resizes happen at full cost before doubling kicks in. Not a correctness issue and amortized cost is still fine; seeding with max(new_term_capacity, some_floor) would smooth the first-grow case. Optional.

LHT129 (commented):
The growth logic is correct and this is a genuine performance improvement, not a regression. The new doubling strategy makes ResizeTermList grow geometrically (amortized O(1) per insert) instead of resizing to exactly the requested capacity each time, which previously could degrade to O(n^2) total work when callers grew by small increments. The overflow guard before `new_capacity *= 2` is correct, element moves via std::move cover all existing entries (destination is always >= old size), new_sizes is zero-initialized, and the old buffers are swapped out and freed exactly once with no dangling unique_ptrs. The parameter cap was raised 10M -> 50M consistently across code, tests, and docs, and the unit tests were correctly updated to assert capacity is >= the expected size with the tail entries null. Two non-blocking notes below: a minor missed-headroom case on first allocation and a pre-existing (unchanged) thread-safety concern. Overall sound.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -283,9 +283,17 @@ SparseTermDataCell::ResizeTermList(InnerIdType new_term_capacity) {
...
+        if (new_capacity > std::numeric_limits<InnerIdType>::max() / 2) {
+            new_capacity = new_term_capacity;
```
LHT129, line 289: Overflow guard is correct: testing new_capacity > max/2 before new_capacity *= 2 prevents the doubling from overflowing, and falling back to new_capacity = new_term_capacity still honors the request. Given term_id_limit is capped at 50M and InnerIdType is uint32, this branch is effectively unreachable in practice but is good defensive code. No change needed.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -294,7 +302,7 @@ SparseTermDataCell::ResizeTermList(InnerIdType new_term_capacity) {
     term_ids_.swap(new_ids);
     term_datas_.swap(new_datas);
     term_sizes_.swap(new_sizes);
-    term_capacity_ = new_term_capacity;
+    term_capacity_ = new_capacity;
 }
```
LHT129, line 306: Thread safety (pre-existing, not introduced here): ResizeTermList still reallocates and frees the old term_ids_/term_datas_/term_sizes_ buffers via swap. Any concurrent reader holding a raw pointer/reference into those buffers (e.g. a mid-flight search dereferencing term_datas_[i].get()) would hit a use-after-free once this returns. This PR does not change that contract, so it is not a regression, but if InsertVector/grow can run concurrently with search it should be confirmed that an external lock or epoch scheme protects this path.

src/datacell/sparse_term_datacell.cpp
Roxanne0321: Thanks, this is a fair observation, but I am keeping this unchanged for now. When the first request is small, the next growth quickly becomes 1 -> 2 -> 4 and remains geometric; when the first request is large, allocating exactly the requested capacity avoids adding extra headroom. Since this is optional and not a correctness issue, the narrower change seems preferable here.

src/datacell/sparse_term_datacell.cpp
Roxanne0321: Agreed, no code change needed here. The guard keeps the doubling from wrapping, and the fallback to the requested capacity still satisfies the resize request.

src/datacell/sparse_term_datacell.cpp
Roxanne0321: Thanks for calling this out. I agree this is a pre-existing concurrency contract issue rather than a regression from this PR, so I am not changing it in this focused performance/capacity update.

wxyucs (approved):
lgtm

PR #2197, opened 2026-06-09
fix(rabitq): use fabs to skip near-zero dims in EncodeExtendRaBitQ

### LINKED ISSUE #2196: Bug: EncodeExtendRaBitQ skips all negative dimensions in t-value optimization
## Description

In `src/quantization/rabitq_quantization/rabitq_quantizer.cpp`, the `compute_next_t_for_dim` lambda inside `EncodeExtendRaBitQ` uses the condition:

```cpp
if (oi < 1e-3) {
    return std::numeric_limits<double>::infinity();
}
```

This incorrectly skips **all** negative dimensions (since any negative value < 1e-3), preventing them from participating in the t-value optimization search via the priority queue.

## Impact

For normalized vectors with negative components (which is the common case), the multi-bit RaBitQ encoding optimization ignores approximately half of the dimensions, degrading encoding quality and search recall.

## Expected Behavior

The condition should skip dimensions with near-zero absolute values (to avoid division by near-zero), not all negative values:

```cpp
if (std::fabs(oi) < 1e-3) {
    return std::numeric_limits<double>::infinity();
}
```

## Location

`src/quantization/rabitq_quantization/rabitq_quantizer.cpp:385`

LHT129:
## Summary

Fix a bug in `EncodeExtendRaBitQ` where the condition `oi < 1e-3` incorrectly skips all negative-valued dimensions from participating in the t-value optimization loop.

The fix changes the condition to `std::fabs(oi) < 1e-3` so that only dimensions with near-zero absolute values are skipped (avoiding division by near-zero), regardless of sign.

## Problem

For normalized vectors with negative components, the multi-bit RaBitQ encoding optimization was ignoring approximately half of the dimensions, degrading encoding quality and search recall.

## Changes

- `src/quantization/rabitq_quantization/rabitq_quantizer.cpp`: Replace `oi < 1e-3` with `std::fabs(oi) < 1e-3` in `compute_next_t_for_dim` lambda (line 385)

Closes #2196

### DIFF
--- a/src/quantization/rabitq_quantization/rabitq_quantizer.cpp
+++ b/src/quantization/rabitq_quantization/rabitq_quantizer.cpp
@@ -382,7 +382,7 @@ RaBitQuantizer<metric>::EncodeExtendRaBitQ(const float* o_prime,
 
     auto compute_next_t_for_dim = [&](size_t i) -> double {
         auto oi = double(o_prime[i]);
-        if (oi < 1e-3) {
+        if (std::fabs(oi) < 1e-3) {
             return std::numeric_limits<double>::infinity();
         }
 


wxyucs (approved):
lgtm

PR #2198, opened 2026-06-09
ci: add aarch64 CI docker images and run arm jobs in containers

wxyucs:
## Summary

Mirror the existing x86 CI image chain for aarch64 so the arm CI jobs can pull a prebuilt image instead of running `scripts/deps/install_deps_ubuntu.sh` on every run.

- **New images** (counterparts of `Dockerfile.x86.{base,openblas}`):
  - `docker/Dockerfile.aarch64.base` → `vsaglib/vsag:ci-aarch64-base` — shared toolchain (clang-15, gcc, cmake, ninja, gfortran, libomp-15, libaio, liburing, curl…), no BLAS.
  - `docker/Dockerfile.aarch64.openblas` → `vsaglib/vsag:ci-aarch64-openblas` — layers system OpenBLAS + LAPACKE on the base.
  - No MKL variant: Intel oneAPI MKL is x86-only, so OpenBLAS is the sole aarch64 BLAS backend (matches `install_deps_ubuntu.sh`, which never installed MKL on aarch64).
- **`update-ci-image.yml`**: split into `build-x86` (unchanged) and a new `build-aarch64` job that builds/pushes the two images on a native `ubuntu-22.04-arm` runner — no QEMU, so the published layers are genuinely arm64.
- **`asan_build_and_test.yml`** and **`daily_test.yml`**: the aarch64 jobs now run `container: vsaglib/vsag:ci-aarch64-openblas`, drop the per-job `Prepare Env` / `install_deps` step, and reuse the `/opt:/useless` free-disk pattern from the x86 jobs.

The `ci-aarch64-openblas` image covers every package `install_deps_ubuntu.sh` installs for aarch64, so it is a complete replacement. The Python wheel jobs (`python_build_and_test.yml`, `pr-ci.yml`) still use `install_deps` on the host, intentionally unchanged (same as x86).

## Notes / validation

- The `update-ci-image` workflow runs on this PR with `push: false`, so it validates that both aarch64 Dockerfiles actually **build on the arm runner** (incl. that `clang-15` / `libomp-15` exist for arm64 jammy) before merge.
- First-run bootstrapping is the same as when the x86 images were introduced: the images are only pushed once `docker/**` lands on `main`. The merge push triggers both `update-ci-image` and `asan_build_and_test` at once, so the first aarch64 job run may briefly race ahead of the image push and need a re-run; steady-state afterward.
- Build behavior is preserved: aarch64 jobs still build OpenBLAS from source (no `-DUSE_SYSTEM_OPENBLAS=ON`); the system OpenBLAS in the image is available if we later want to opt into the faster path.

### DIFF
--- a/.github/workflows/update-ci-image.yml
+++ b/.github/workflows/update-ci-image.yml
@@ -8,14 +8,16 @@ on:
     paths: [ 'docker/**' ]
 
 jobs:
-  # The CI images form a small chain: ci-x86-base is the shared toolchain
+  # The x86 CI images form a small chain: ci-x86-base is the shared toolchain
   # layer, and ci-x86-mkl / ci-x86-openblas are built FROM it. We build all
   # three in a single job so PR runs (push: false) can validate end-to-end:
   # the variant builds resolve `FROM vsaglib/vsag:ci-x86-base` against the
   # base image we just loaded into the local Docker daemon, not whatever
-  # tag happens to be on Docker Hub.
-  build:
-    name: Build CI Images
+  # tag happens to be on Docker Hub. The aarch64 images are built by a separate
+  # job below; they live on their own ci-aarch64-* tags and must be built on a
+  # native arm64 runner so the resulting layers are genuinely arm64.
+  build-x86:
+    name: Build CI Images (x86)
     runs-on: ubuntu-latest
     steps:
       # Docker Hub credentials only flow into runs triggered from the
@@ -66,3 +68,41 @@ jobs:
           file: ./docker/Dockerfile.x86.openblas
           push: ${{ github.event_name == 'push' }}
           tags: vsaglib/vsag:ci-x86-openblas
+
+  # aarch64 counterpart of the job above. There is no MKL variant on aarch64
+  # (Intel oneAPI MKL is x86-only), so the chain is just ci-aarch64-base ->
+  # ci-aarch64-openblas. We run on a native arm64 runner instead of emulating
+  # via QEMU so the published layers are real arm64 and image builds stay fast.
+  build-aarch64:
+    name: Build CI Images (aarch64)
+    runs-on: ubuntu-22.04-arm
+    steps:
+      # Same fork-safety reasoning as the x86 job: secrets (and therefore the
+      # Docker Hub login) only flow into push events / same-repo PRs. Fork PRs
+      # build with push: false and resolve the openblas variant's
+      # `FROM vsaglib/vsag:ci-aarch64-base` against the locally loaded base.
+      - name: Login to Docker Hub
+        if: github.event_name == 'push'
+        uses: docker/login-action@v3
+        with:
+          username: ${{ vars.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      # The plain `docker` driver loads each build straight into the host
+      # daemon, so the openblas build can resolve `FROM vsaglib/vsag:ci-aarch64-base`
+      # against the base we just built rather than pulling from Docker Hub.
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
+      - name: Build ci-aarch64-base
+        uses: docker/build-push-action@v6
+        with:
+          file: ./docker/Dockerfile.aarch64.base
+          push: ${{ github.event_name == 'push' }}
+          tags: vsaglib/vsag:ci-aarch64-base
+      - name: Build ci-aarch64-openblas
+        uses: docker/build-push-action@v6
+        with:
+          file: ./docker/Dockerfile.aarch64.openblas
+          push: ${{ github.event_name == 'push' }}
+          tags: vsaglib/vsag:ci-aarch64-openblas
--- a/docker/Dockerfile.aarch64.base
+++ b/docker/Dockerfile.aarch64.base
@@ -1,0 +1,39 @@
+
+# Base CI image for VSAG on aarch64.
+#
+# This image carries the common toolchain shared by every aarch64 CI variant
+# (compilers, build tools, runtime headers, etc.) but intentionally does NOT
+# install any BLAS implementation. BLAS backends are layered on top of this
+# image in the variant Dockerfiles:
+#
+#   * docker/Dockerfile.aarch64.openblas -> vsaglib/vsag:ci-aarch64-openblas
+#
+# Unlike x86_64 there is no Intel MKL variant here: Intel oneAPI MKL ships
+# x86-only, so on aarch64 OpenBLAS is the sole BLAS backend. Keeping the base
+# image free of any BLAS implementation still lets us:
+#   * share the build context across future aarch64 variants;
+#   * make it trivial to add another backend (e.g. Arm Performance Libraries)
+#     later.
+#
+# `FROM ubuntu:22.04` is a multi-arch manifest; building this Dockerfile on an
+# aarch64 host (the ubuntu-22.04-arm runner in docker/update-ci-image.yml)
+# selects the arm64 base automatically, so no QEMU emulation is required.
+
+FROM ubuntu:22.04
+
+# Install the whole toolchain in a single layer and drop the apt cache so the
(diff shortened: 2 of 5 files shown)


LHT129 (approved):
LGTM

docker/Dockerfile.aarch64.base
```diff
@@ -0,0 +1,33 @@
...
+# image free of any BLAS implementation still lets us:
+#   * share the build context across future aarch64 variants;
+#   * make it trivial to add another backend (e.g. Arm Performance Libraries)
+#     later.
+#
+# `FROM ubuntu:22.04` is a multi-arch manifest; building this Dockerfile on an
+# aarch64 host (the ubuntu-22.04-arm runner in docker/update-ci-image.yml)
+# selects the arm64 base automatically, so no QEMU emulation is required.
+
+FROM ubuntu:22.04
```
LHT129, line 22: FROM ubuntu:22.04 is a floating tag. For reproducible CI images consider pinning to a digest (FROM ubuntu:22.04@sha256:...). Not a regression since Dockerfile.x86.base does the same, but the comment above promises arm64 selection via the multi-arch manifest, which only stays stable if the tag isn't re-pushed underneath you.

docker/Dockerfile.aarch64.base
```diff
@@ -0,0 +1,33 @@
...
+# update and install basic deps
+RUN apt update && apt install -y wget gpg sudo
+
+# install build tools
+RUN apt install -y git g++ cmake ninja-build gdb ccache build-essential lcov gfortran
+RUN apt install -y clang-15 clangd-15 clang-format-15 python3-pip
+RUN ln -s /usr/bin/clang-format-15 /usr/bin/clang-format
+
+# install vsag deps (everything except a BLAS backend)
+RUN apt install -y python3-dev libomp-15-dev libaio-dev libcurl4-openssl-dev liburing-dev
```
LHT129: Each apt install runs in its own RUN layer with no `apt clean` / `rm -rf /var/lib/apt/lists/*` and no --no-install-recommends, so the apt index and recommended packages are baked into the image, inflating its size. Consider chaining the installs into a single RUN ending with cleanup. (Matches the existing x86 Dockerfiles, so consistent rather than a regression.)

docker/Dockerfile.aarch64.openblas
```diff
@@ -0,0 +1,20 @@
...
+# from-source OpenBLAS build and simply ignore the pre-installed packages.
+#
+# This is the aarch64 counterpart of docker/Dockerfile.x86.openblas. There is no
+# MKL counterpart on aarch64 because Intel oneAPI MKL is x86-only.
+
+FROM vsaglib/vsag:ci-aarch64-base
+
+# Install system OpenBLAS + LAPACKE. liblapacke-dev pulls in the C headers
+# needed for VSAG's CMake `find_path(LAPACKE_INCLUDE NAMES lapacke.h)` probe.
+RUN apt update && apt install -y libopenblas-dev liblapacke-dev
```
LHT129: Same cleanup note as the base image: `apt update && apt install ...` leaves /var/lib/apt/lists/* in the final layer. Appending `&& apt clean && rm -rf /var/lib/apt/lists/*` would trim the image. Optional / matches x86 convention.

.github/workflows/update-ci-image.yml
```diff
@@ -66,3 +68,41 @@ jobs:
...
           push: ${{ github.event_name == 'push' }}
           tags: vsaglib/vsag:ci-x86-openblas
+
+  # aarch64 counterpart of the job above. There is no MKL variant on aarch64
+  # (Intel oneAPI MKL is x86-only), so the chain is just ci-aarch64-base ->
+  # ci-aarch64-openblas. We run on a native arm64 runner instead of emulating
+  # via QEMU so the published layers are real arm64 and image builds stay fast.
+  build-aarch64:
+    name: Build CI Images (aarch64)
+    runs-on: ubuntu-22.04-arm
```
LHT129, line 78: build-aarch64 correctly runs on ubuntu-22.04-arm and (deliberately) omits the Set up QEMU step the x86 job uses, since the runner is natively arm64. Looks right. Minor: the image tags here (ci-aarch64-base/-openblas) and the docker/* actions (@v3/@v6) are floating refs; the rest of the repo does the same, so just flagging for awareness.

LHT129 (commented):
Solid, well-reasoned CI change. The aarch64 image chain (ci-aarch64-base -> ci-aarch64-openblas) mirrors the existing x86 setup, the build-aarch64 job correctly runs on a native arm64 runner and omits QEMU (the comments explain this clearly), and using driver: docker so the openblas variant's FROM resolves the locally-built base is consistent with the x86 job. Dropping the per-job install_deps step (now baked into the image) is applied consistently across the aarch64 jobs. A few non-blocking notes: the new Dockerfiles inherit the existing x86 pattern of no apt cache cleanup / no --no-install-recommends (image bloat), and CI images / base image / actions are pinned to floating tags rather than digests. These match prior convention so are not regressions, but worth flagging. No secret-exposure issues: Docker Hub login is correctly gated on push events only.

docker/Dockerfile.aarch64.base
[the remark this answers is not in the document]
wxyucs: Thanks — adopted most of this. The toolchain install is now a single `RUN` using `apt-get` (not `apt`) and ends with `apt-get clean && rm -rf /var/lib/apt/lists/*`, so the apt index is no longer baked into the layer.

I deliberately left `--no-install-recommends` off: the x86 CI images don't use it, this package list was validated to produce a working VSAG build/test toolchain, and dropping recommended packages risks silently removing an implicit runtime dependency. Since the goal here is a reliable, cacheable CI image rather than a minimal-size production image, I'd rather not introduce that risk without separately verifying each package. Applied in the current revision.

docker/Dockerfile.aarch64.openblas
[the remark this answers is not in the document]
wxyucs: Done for the `apt-get` switch and the single-layer cleanup (`apt-get clean && rm -rf /var/lib/apt/lists/*`). As with the base image I kept `--no-install-recommends` off for now, to stay consistent with the x86 images and to avoid dropping an implicit dependency of `libopenblas-dev` / `liblapacke-dev` without separate verification. Applied in the current revision.

docker/Dockerfile.aarch64.base
wxyucs: Good point on reproducibility. I'm going to keep the floating `ubuntu:22.04` tag for now:

- It matches `Dockerfile.x86.base`, so x86 and aarch64 stay consistent; pinning only aarch64 would be asymmetric.
- The floating tag is what lets the image pick up upstream base-image security patches automatically.
- These images are rebuilt and re-pushed by `update-ci-image.yml`, so a re-pushed `22.04` just means the next rebuild picks up the newer base — acceptable for a CI toolchain image.

If we want digest pinning for reproducibility, I think it's better as a follow-up that pins x86 and aarch64 together (and wires in a refresh path), rather than diverging here. Leaving as-is.

docker/Dockerfile.aarch64.base
wxyucs: Done — the installs are now chained into a single `RUN` ending with `apt-get clean && rm -rf /var/lib/apt/lists/*`, so the apt index is no longer baked into the image. I left `--no-install-recommends` off to stay consistent with the x86 images and to avoid dropping an implicit dependency without separate verification. Applied in the current revision.

docker/Dockerfile.aarch64.openblas
wxyucs: Done — appended `&& apt-get clean && rm -rf /var/lib/apt/lists/*` to the single install layer. Applied in the current revision.

.github/workflows/update-ci-image.yml
wxyucs: Thanks for confirming the QEMU omission is intentional (native arm64 runner). On the floating refs: agreed, and as you noted the rest of the repo (image tags + `docker/*` actions) uses the same convention, so I'm keeping it consistent here rather than pinning just this workflow. Noted for awareness.

PR #2208, opened 2026-06-10
fix(pyramid): set duplicate_query_id in add_one_point for duplicate detection

### LINKED ISSUE #2207: [bug](pyramid): duplicate detection always fails due to unset duplicate_query_id
## Description

In `Pyramid::add_one_point`, when `support_duplicate_` is true, the code sets `search_param.find_duplicate = true` but does not set `search_param.duplicate_query_id`. The default value (`std::numeric_limits<InnerIdType>::max()`) causes the duplicate detection logic in `BasicSearcher` to always skip both detection paths:

1. `duplicate_distance_threshold` defaults to 0, so the threshold-based branch is skipped
2. `duplicate_query_id` defaults to max, which is always `>= flatten->TotalCount()`, so the CompareVectors-based branch is also skipped

As a result, `duplicate_id` remains -1, and duplicate vectors are never detected. They are added as independent nodes via `mutually_connect_new_element`, degrading graph quality and causing intermittent failures in the Pyramid Duplicate Test.

## Fix

Set `search_param.duplicate_query_id = inner_id` after setting `find_duplicate = true`, matching the pattern used by HGraph in `hgraph_build.cpp:474`.

## References

- CI failure: https://github.com/antgroup/vsag/actions/runs/27201949367/job/80332714372
- HGraph reference: `src/algorithm/hgraph/hgraph_build.cpp:474`
- Detection logic: `src/impl/searcher/basic_searcher.cpp:518-534`

LHT129:
## Summary

Fixes #2207

In `Pyramid::add_one_point`, when `support_duplicate_` is true, `search_param.find_duplicate` was set but `search_param.duplicate_query_id` was left at its default value (`std::numeric_limits<InnerIdType>::max()`). This caused the duplicate detection logic in `BasicSearcher::Search` to always fail:

- `duplicate_distance_threshold = 0` skips the threshold branch
- `duplicate_query_id = max` is always `>= TotalCount()`, skipping the CompareVectors branch

Duplicate vectors were never detected and were added as independent graph nodes, degrading NSW graph quality and causing intermittent failures in the Pyramid Duplicate Test.

## Changes

- Set `search_param.duplicate_query_id = inner_id` after `find_duplicate = true`, matching the pattern used by HGraph (`hgraph_build.cpp:474`)

## Test Plan

- [x] All 6 Pyramid duplicate test cases pass (27162 assertions)
- [x] Pyramid Duplicate Test passes consistently across 5 consecutive runs (previously flaky)
- [x] Release build passes

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -798,6 +798,7 @@ Pyramid::add_one_point(IndexNode* node, InnerIdType inner_id, const float* vecto
         search_param.hops_limit = 10000;  // Add hops limit to prevent infinite loop
         if (support_duplicate_) {
             search_param.find_duplicate = true;
+            search_param.duplicate_query_id = inner_id;
         }
         auto codes = use_reorder_ ? precise_codes_ : base_codes_;
         bool update_entry_point;


wxyucs (approved):
lgtm

PR #2280, opened 2026-06-16
fix(io): add RAII destructor to DirectIOObject and simplify AsyncIO ownership

jfeng18:
## Summary

Add proper RAII lifecycle management to `DirectIOObject`, eliminating manual memory management in `AsyncIO`.

### Changes

**`src/io/direct_io_object.h`**:
- Add `~DirectIOObject()` that calls `free(align_data)`
- Delete copy constructor and copy assignment (prevent double-free)
- Add move constructor and move assignment (transfer ownership)

**`src/io/async_io.cpp`**:
- `DirectReadImpl`: Transfer ownership to caller by nulling `align_data` before return; error path relies on RAII destructor instead of manual `Release()`
- `MultiReadImpl`: Remove manual `ReleaseImpl()` loop and exception-path `Release()` loops; `std::vector<DirectIOObject>` destructor handles cleanup automatically

### Safety Analysis

- **No double-free**: Success paths null `align_data` before obj destruction; `free(nullptr)` is no-op
- **No use-after-free**: Returned pointers remain valid until `ReleaseImpl()` is called by the caller
- **No leaks**: Exception paths are covered by RAII destructor
- **Backward compatible**: `ReleaseImpl()` still works for external callers via `BasicIO::Release()`

### Verification

- Release build: passed
- Unit tests: 439/441 passed (2 pre-existing failures in int8_quantizer_test unrelated to this change)
- Code review: PASS (no blocking issues)

### DIFF
--- a/src/io/async_io/async_io.cpp
+++ b/src/io/async_io/async_io.cpp
@@ -114,10 +114,7 @@ AsyncIO::DirectReadImpl(uint64_t size, uint64_t offset, bool& need_release) cons
         // destructor automatically frees align_data
         throw VsagException(ErrorType::INTERNAL_ERROR, fmt::format("pread error {}", ret));
     }
-    auto* result = obj.data;
-    obj.align_data = nullptr;  // transfer ownership to caller
-    obj.data = nullptr;
-    return result;
+    return obj.TakeData();
 }
 
 void
--- a/src/io/async_io/async_io.cpp
+++ b/src/io/async_io/async_io.cpp
@@ -111,10 +111,13 @@ AsyncIO::DirectReadImpl(uint64_t size, uint64_t offset, bool& need_release) cons
     DirectIOObject obj(size, offset);
     auto ret = IOSyscall::PRead(this->rfd_, obj.align_data, obj.size, obj.offset);
     if (ret < 0) {
-        obj.Release();
+        // destructor automatically frees align_data
         throw VsagException(ErrorType::INTERNAL_ERROR, fmt::format("pread error {}", ret));
     }
-    return obj.data;
+    auto* result = obj.data;
+    obj.align_data = nullptr;  // transfer ownership to caller
+    obj.data = nullptr;
+    return result;
 }
 
 void
@@ -146,9 +149,7 @@ AsyncIO::MultiReadImpl(uint8_t* datas, uint64_t* sizes, uint64_t* offsets, uint6
         int submitted = io_submit(context->ctx_, static_cast<int64_t>(count), cb);
         if (submitted < 0) {
             io_context_pool->ReturnOne(context);
-            for (auto& obj : objs) {
-                obj.Release();
-            }
+            // objs destructor automatically frees all align_data
             throw VsagException(ErrorType::INTERNAL_ERROR, "io submit failed");
         }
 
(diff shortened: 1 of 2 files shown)


jfeng18:
Note on classification: the actual DirectIOObject leak was already fixed in #2205 (merged); this PR is a follow-up refactor replacing the manual Release with RAII (destructor + move-only). So it is `kind/improvement`, not `kind/bug` — no issue link required. Could a maintainer please add `kind/improvement` and `version/1.0`? Thanks!

src/io/async_io.cpp
```diff
@@ -157,16 +157,12 @@ AsyncIO::MultiReadImpl(uint8_t* datas, uint64_t* sizes, uint64_t* offsets, uint6
                                        &timeout);
         if (num_events != count) {
             io_context_pool->ReturnOne(context);
-            for (auto& obj : objs) {
-                obj.Release();
-            }
             throw VsagException(ErrorType::INTERNAL_ERROR, "io async read failed");
         }
```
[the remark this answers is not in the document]
jfeng18: Thanks for the detailed analysis — the core insight is correct: on the `num_events != count` branch the buffers are freed while the kernel may still have reads in-flight (and `io_getevents` does not cancel on timeout), so a completion can DMA into freed memory. I dug into it and want to record the scope precisely:

**Pre-existing, not introduced by this PR.** On `upstream/main` the same branch already does `for (auto& obj : objs) obj.Release();` (i.e. `free(align_data)`) over all `count` buffers including the in-flight ones. This PR only replaces that explicit free loop with the RAII destructor — it frees the same buffers at the same point (the throw), so the behavior is identical. This PR's RAII refactor is self-contained and correct on its own (copy deleted; move nulls the source; `ReleaseImpl`'s mask reconstructs exactly the `aligned_alloc` base; `vector(count)` default-constructs in place with no copies/moves), so it does not introduce, fix, or worsen this issue.

**On the proposed `io_destroy` + `io_setup` remedy** — directionally right (`io_destroy` blocks until in-flight IO drains), but as a drop-in it has sharp edges: the context is pooled and reused, so re-`io_setup` must `memset(&ctx_, 0)` first (else EINVAL) and use the original queue depth (`DEFAULT_REQUEST_COUNT`, not `count`); and `io_setup` can return EAGAIN under the very load that caused the timeout — if that's unchecked and the context is still returned to the pool, it permanently poisons the pool. A simpler, safer fix is to **drain via `io_getevents`** (reap the remaining `submitted` in-flight iocbs before freeing), which leaves the context clean and needs no teardown/re-setup. There's also a related gap: `io_submit` is only checked for `< 0`, so a partial submit (`0 <= submitted < count`) hits the same path.

Since this is a pre-existing memory-safety issue in the libaio error path (orthogonal to the RAII cleanup this PR is about), I'll track it as a **separate `kind/bug`** with an ASAN-backed repro rather than expand this PR.

src/io/async_io/async_io.cpp
```diff
@@ -111,10 +111,13 @@ AsyncIO::DirectReadImpl(uint64_t size, uint64_t offset, bool& need_release) cons
...
     if (ret < 0) {
-        obj.Release();
+        // destructor automatically frees align_data
         throw VsagException(ErrorType::INTERNAL_ERROR, fmt::format("pread error {}", ret));
     }
-    return obj.data;
+    auto* result = obj.data;
+    obj.align_data = nullptr;  // transfer ownership to caller
+    obj.data = nullptr;
+    return result;
```
LHT129: return the obj.data but the obj is releasd

LHT129 (changes requested):

src/io/async_io/async_io.cpp
jfeng18: Thanks for catching this! The code was technically safe (`result` is saved before nulling, and `free(nullptr)` in the destructor is a no-op), but the manual member nulling was indeed unclear and easy to misread.

Added a `TakeData()` method to `DirectIOObject` that encapsulates the ownership transfer. `DirectReadImpl` now just does `return obj.TakeData();` — the intent is self-documenting and the internal members are no longer exposed.

LHT129 (approved):
LGTM

PR #2288, opened 2026-06-17
[backport 0.18] fix(io): block until all in-flight async reads complete

### LINKED ISSUE #2281: [bug](hgraph): avoid freeing async_io MultiRead buffers while reads are in flight
### Affected area

hgraph

### Interface

cpp

### VSAG version

- Reviewed against current `origin/main` @ `e0cfcc66` (fetched 2026-06-16).
- The crash-path IO code (`src/io/async_io.cpp`) is unchanged across surrounding history and **still unfixed on current `origin/main`** — `grep -rn io_cancel src/` returns nothing.
- The reproducer is a standalone libaio program and does not depend on the vsag build.

### Summary

`AsyncIO::MultiReadImpl` (`src/io/async_io.cpp`) submits up to
`IOContext::DEFAULT_REQUEST_COUNT` (100) O_DIRECT reads via `io_submit`, then waits
with `io_getevents(min_nr = nr = count, timeout = {1, 0})` (a 1-second timeout).

When fewer than `count` events are ready — which happens on a **1s timeout**
(transient host IO stall) or when the syscall is interrupted and returns **EINTR** —
the error branch does:

```cpp
if (num_events != count) {
    io_context_pool->ReturnOne(context);   // context returned to the shared pool
    for (auto& obj : objs) obj.Release();   // frees every DMA target buffer (align_data)
    throw VsagException(...);               // ...but the in-flight requests are NOT cancelled
}
```

It **frees the DMA target buffers and returns the libaio context to the pool while
one or more reads are still in flight in the kernel.** There is no `io_cancel` and no
drain — `io_cancel` does not appear anywhere in `src/`.

When a straggler read completes later, the kernel DMAs file bytes into memory that has
already been `free`d and (typically) reused by a subsequent `MultiReadImpl` call as its
`std::vector<DirectIOObject> objs` or output buffer. This **overwrites a live
`objs[i].data` pointer field with raw on-disk vector-code bytes**. The next iteration's
`memcpy(cur_data, objs[i].data, sizes[i])` (`async_io.cpp:166`) then dereferences a
non-canonical wild pointer → **SIGSEGV**, taking down the whole process.

This is a classic libaio lifetime bug: **a buffer handed to `io_submit` must not be
freed, and its context must not be reused, until the request is known to have completed
or been successfully `io_cancel`ed.**

This was hit in a deployment using `base_quantization_type=rabitq,
precise_quantization_type=sq8, precise_io_type=async_io, use_reorder=true`, where the
reorder stage issues batched `MultiRead` calls. A transient storage stall that pushes a
single in-flight read past the 1s `io_getevents` timeout is enough to trigger it.

### Minimal reproducer

The corrupting write is performed by the **kernel via O_DIRECT DMA**, so it cannot be
caught by AddressSanitizer/Valgrind (verified: ASan stays silent while the kernel DMAs
into freed/poisoned memory). The smallest deterministic reproducer is therefore a
standalone libaio program that performs the exact lifecycle the error path performs:
submit → abandon (no completion drain, no `io_cancel`) → free → reuse memory →
late drain → dereference the overwritten pointer.

Mapping to current `origin/main`:

| Reproducer step | `AsyncIO::MultiReadImpl` equivalent |
|---|---|
| `dma = amalloc(BLK)` | `objs[i].Set(...)` allocates `DirectIOObject::align_data` |
| `io_prep_pread(..., dma, ...)` + `io_submit` | `io_prep_pread(cb[i], ..., obj.align_data, ...)` and `io_submit(...)` |
| no completion wait before `free(dma)` | the `num_events != count` branch: at least one request is still in flight, but `obj.Release()` frees the DMA buffer |
| `slots = amalloc(BLK)` filled with aligned pointer-like values | a later allocation reuses the freed block for live pointer-containing memory, e.g. another `DirectIOObject::data` field |
| late `io_getevents(...)` | the abandoned kernel request eventually completes and DMAs file bytes into the reused block |
| `memcpy(dst, (void*)slots[0], 768)` | `memcpy(cur_data, objs[i].data, sizes[i])` dereferences the corrupted pointer |

The PoC deliberately skips waiting for a real 1-second timeout/EINTR because that is
timing-dependent. It starts from the same state that the real error branch creates:
a submitted O_DIRECT request whose target buffer is freed before the kernel has
finished with it.

```cpp
// g++ -O2 -g repro.cpp -laio && ./repro   (needs O_DIRECT-capable fs + libaio)
#include <libaio.h>
#include <fcntl.h>
#include <unistd.h>

#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <cstdint>
#include <vector>

static const int ALIGN = 512;
static const size_t BLK = 4096;
static void* amalloc(size_t n){ void* p; if (posix_memalign(&p, ALIGN, n)) abort(); return p; }

int main() {
    const char* path = "/tmp/aio_uaf_crash.bin";
    { int w = open(path, O_CREAT|O_WRONLY|O_TRUNC, 0644);
      std::vector<uint8_t> b(BLK*64, 0xAB); (void)!write(w, b.data(), b.size()); fsync(w); close(w); }
    int fd = open(path, O_RDONLY|O_DIRECT);
    io_context_t ctx; memset(&ctx, 0, sizeof(ctx)); io_setup(128, &ctx);

    // 1) DMA target buffer == DirectIOObject.align_data
    void* dma = amalloc(BLK); memset(dma, 0, BLK);

    // 2) Submit an O_DIRECT read, then ABANDON it (mirrors the num_events!=count path:
    //    no io_getevents to completion, no io_cancel).
    struct iocb cb; memset(&cb, 0, sizeof(cb)); struct iocb* p = &cb;
    io_prep_pread(&cb, fd, dma, BLK, 0);
    if (io_submit(ctx, 1, &p) != 1) return 2;

    // 3) Free the DMA target while the request is still in flight (== obj.Release()).
    free(dma);

    // 4) A later MultiReadImpl reuses that memory and fills it with valid, aligned
    //    "pointer" values (stand-in for objs[i].data: always aligned, low byte 0x00).
    uintptr_t* slots = (uintptr_t*)amalloc(BLK);
    for (size_t i = 0; i < BLK/sizeof(uintptr_t); ++i) slots[i] = 0x00007f0000000000ULL;

    // 5) Drain the abandoned request: the kernel DMAs file bytes (0xAB) over the
    //    freed-and-reused memory.
    struct io_event ev[1]; struct timespec t = {2, 0};
    if (io_getevents(ctx, 1, 1, ev, &t) != 1) return 3;

    // 6) The "pointer" is now 0xabababababababab -- non-canonical, low byte 0xAB,
    //    not aligned. memcpy from such a value faults exactly like async_io.cpp:166.
    printf("slot[0] = 0x%016lx (low byte 0x%02lx)\n",
           (unsigned long)slots[0], (unsigned long)(slots[0] & 0xff));
    fflush(stdout);

    if (slots[0] != 0xABABABABABABABABULL) return 4;

    uint8_t dst[768];
    printf("memcpy(dst, (void*)0x%016lx, 768) -> expect SIGSEGV\n",
           (unsigned long)slots[0]);
    fflush(stdout);
    memcpy(dst, (void*)slots[0], sizeof(dst));
    printf("unexpectedly survived; dst[0]=%u\n", (unsigned)dst[0]);
    return 0;
}
```

Typical output (5/5 on the repro host):

```
slot[0] = 0xabababababababab (low byte 0xab)
memcpy(dst, (void*)0xabababababababab, 768) -> expect SIGSEGV
Segmentation fault (core dumped)
```

This matches the observed wild-pointer signature exactly: source pointer with a
random low byte, not aligned to `code_size_` (768), while candidate IDs were verified
**in range** (so it is not an out-of-bounds-ID bug). The merged reproducer now closes
the loop in one process: late kernel DMA overwrites a pointer-like field, and the next
`memcpy` dereferences it and exits with SIGSEGV (`rc=139`, core file generated).

In-tree trigger (non-deterministic, needs a slow disk or a signal to force the 1s
timeout / EINTR): hgraph KnnSearch with the config above, i.e. rabitq base + sq8
precise + `precise_io_type=async_io` + `use_reorder=true`, batched reorder reads going
through `FlattenDataCell::query` → `MultiRead`.

### Environment

```shell
- OS: CentOS Linux 7
- vsag version: v0.18.0-278-g48ac285a
- compiler version: g++ (GCC) 10.2.1 20210130 (Red Hat 10.2.1-11)

# Additional repro-host details not captured by the script:
- Docker image: vsaglib/vsag:manylinux2014-pre
- libaio: libaio.so.1.0.0 (libc6,x86-64) => /lib64/libaio.so.1.0.0
- kernel: 4.19.91-011.ali4000.alios7.x86_64
- CPU: Intel(R) Xeon(R) Processor @ 2.90GHz

# Source review note:
- Current official origin/main reviewed at e0cfcc66 (fetched 2026-06-16).
- The standalone reproducer does not link against vsag, so it does not depend on
  the vsag checkout version printed by check_environment.sh.
```

### Expected behavior

On an `io_getevents` timeout / partial completion / EINTR, the in-flight requests must
be cancelled or drained before their buffers are freed and before the `IOContext` is
returned to the pool. The query should fail with a recoverable `VsagException`
(caught by `SAFE_CALL` in `IndexImpl::KnnSearch`), degrading to a single failed query —
never a process-wide SIGSEGV.

### Actual behavior

The error path frees still-in-flight DMA buffers and reuses the context with no
`io_cancel`/drain. A late kernel DMA corrupts freed-and-reused memory, overwriting an
`objs[i].data` pointer with file bytes; a subsequent `memcpy` dereferences the wild
pointer and the whole process crashes:

```
SIGSEGV (0xb), si_code: 128 (SI_KERNEL), si_addr: 0x0
C  [libc.so.6+0x165c60]  __memmove_avx_unaligned_erms+0x170
vsag::AsyncIO::MultiReadImpl                                   async_io.cpp:166
vsag::BasicIO<AsyncIO>::MultiRead                              basic_io.h:123
vsag::FlattenDataCell<ScalarQuantizer<L2,8>, AsyncIO>::query   flatten_datacell.h:338
vsag::FlattenReorder::Reorder                                  flatten_reorder.cpp:50
vsag::HGraph::reorder                                          hgraph_build.cpp:649
vsag::HGraph::SearchWithRequest / KnnSearch                    hgraph_search.cpp
vsag::IndexImpl<HGraph>::KnnSearch                             index_impl.h:299
```

`SAFE_CALL` cannot catch it: the error path already threw earlier; the actual fault
happens later, in a different call, when the wild pointer is dereferenced — there is no
C++ exception at that point.

### References

- `src/io/async_io.cpp:132` — `MultiReadImpl` submit loop (`while (all_count > 0)`)
- `src/io/async_io.cpp:153` — `io_getevents` with `{1, 0}` timeout, `min_nr = nr = count`
- `src/io/async_io.cpp:158` — `num_events != count` error path: frees buffers + returns context, no `io_cancel`
- `src/io/async_io.cpp:167` — crashing `memcpy(cur_data, objs[i].data, sizes[i])`
- `src/io/io_context.h:78` — `DEFAULT_REQUEST_COUNT = 100`
- `src/io/direct_io_object.h:65` — `DirectIOObject::Set` / `align_data` allocation (`std::aligned_alloc` at :74)
- `src/datacell/flatten_datacell.h:327` — batched reorder read (`id_count > 1`) that calls `MultiRead` at :340
- `src/impl/reorder/flatten_reorder.cpp:48` — reorder candidate IDs fed to `Query` at :50

### Additional context

Suggested fix:

On the `num_events != count` path (and the partial-`io_submit` path), do **not** free
any buffer or reuse the context while requests may still be in flight. Either:

1. `io_cancel` each submitted iocb, then blocking-`io_getevents` until all `submitted`
   requests have reported completion, and only then `obj.Release()` + `ReturnOne`; or
2. simpler: drop the 1s timeout and block until all requests complete (`timeout =
   nullptr` or a sufficiently large value), so the function never abandons in-flight
   requests.

Note `io_cancel` is not guaranteed to succeed on all filesystems/devices (often returns
`-EINVAL`), so the blocking drain is the reliable backstop. If even the drain fails,
the `IOContext` and its buffers must be quarantined/leaked rather than freed-and-reused.

Secondary hardening (defense-in-depth, does not fix this bug on its own):
- Validate each `io_event.res / res2` (via `event.obj->data` back-reference) so short /
  failed reads degrade to a thrown `VsagException` instead of silently-wrong codes.
- Bounds-check reorder candidate IDs (`idx[i] < total_count_`) in the query read path
  (`flatten_datacell.h` batch branch and `GetCodesById`).
- Null-check `std::aligned_alloc` in `DirectIOObject::Set`.

LHT129:
## What problem does this PR solve?

Backport of #2287. Fixes #2281.

`AsyncIO::MultiReadImpl` frees DMA target buffers while kernel async reads are still in flight, causing wild-pointer SIGSEGV on subsequent calls.

## What is changed and how does it work?

- Remove 1-second `io_getevents` timeout, block until all submitted requests complete
- Wait on actual submitted count to handle partial `io_submit`
- Drain completed requests before releasing buffers on partial submit error

### DIFF
--- a/src/io/async_io.cpp
+++ b/src/io/async_io.cpp
@@ -138,18 +138,34 @@ AsyncIO::MultiReadImpl(uint8_t* datas, uint64_t* sizes, uint64_t* offsets, uint6
             throw VsagException(ErrorType::INTERNAL_ERROR, "io submit failed");
         }
 
-        struct timespec timeout = {1, 0};
-        auto num_events = io_getevents(context->ctx_,
-                                       static_cast<int64_t>(count),
-                                       static_cast<int64_t>(count),
-                                       context->events_,
-                                       &timeout);
-        if (num_events != count) {
+        // Wait for all submitted requests to complete, handling EINTR
+        int64_t num_events = 0;
+        while (num_events < submitted) {
+            auto ret = io_getevents(context->ctx_,
+                                    static_cast<int64_t>(submitted - num_events),
+                                    static_cast<int64_t>(submitted - num_events),
+                                    context->events_ + num_events,
+                                    nullptr);
+            if (ret < 0) {
+                if (ret == -EINTR) {
+                    continue;  // Interrupted by signal, retry
+                }
+                // Fatal error - but all requests are already submitted to kernel
+                // We must still wait for them to complete to avoid UAF
+                throw VsagException(ErrorType::INTERNAL_ERROR,
+                                    "io_getevents failed with errno " + std::to_string(-ret));
+            }
+            num_events += ret;
+        }
+
+        if (submitted != static_cast<int>(count)) {
             io_context_pool->ReturnOne(context);
             for (auto& obj : objs) {
                 obj.Release();
             }
-            throw VsagException(ErrorType::INTERNAL_ERROR, "io async read failed");
+            throw VsagException(ErrorType::INTERNAL_ERROR,
+                                "io_submit partial: requested " + std::to_string(count) +
+                                    " but submitted " + std::to_string(submitted));
         }
 
         for (int64_t i = 0; i < count; ++i) {


wxyucs (approved):
lgtm

PR #2299, opened 2026-06-18
fix(heap): bound non-fixed MemmoveHeap Push to prevent buffer overflow (topk<10)

### LINKED ISSUE #2298: [bug](ivf): KnnSearch with topk<10 overruns non-fixed MemmoveHeap buffer
### Affected area
ivf

### Interface
cpp

### VSAG version
v0.18.0-289-g42eca652 (also present on `main`; logic unchanged since the file
was moved into `src/algorithm/ivf/`)

### Summary
Calling `KnnSearch` on an IVF index with `topk < 10` can corrupt the heap and
crash with SIGSEGV. The root cause is in the **non-fixed** `MemmoveHeap`
(`MemmoveHeap<max_heap, false>`): `Pop()` decrements `cur_size_` without
shrinking `ordered_buffer_`, leaving a stale tail. The non-fixed `Push()` then
runs `std::upper_bound` over the **entire** buffer (`begin()..end()`) including
that stale tail, so the insertion index `pos` can exceed `cur_size_`. The
subsequent `memmove(..., (cur_size_ - pos) * sizeof(DistanceRecord))` underflows
to a huge `size_t`, overrunning the heap buffer.

`DistanceHeap::MakeInstanceBySize` selects `MemmoveHeap` only when
`max_size < 10` (`memmove_maxsize`), and `IVF::search` requests the **non-fixed**
variant `MakeInstanceBySize<true, false>(topk)`, then interleaves `Push` with
`if (heap->Size() > topk) heap->Pop()`. So any KNN query with `topk < 10` reaches
the buggy path; whether it overruns is data-dependent (a function of the per-bucket
distance sequence).

The **fixed-size** branch is already correct — it bounds `upper_bound` at
`begin() + cur_size_` (`src/impl/heap/memmove_heap.cpp:35-36`). Only the
non-fixed branch using `end()` is affected.

### Minimal reproducer
The following drives the real heap through the exact `IVF::search` Push/Pop
interleaving. With `topk = 5` (`< 10`) and `seed = 0`, the bad `memmove` fires
within ~90 iterations. Compiled with `-fsanitize=address`, ASan reports it
deterministically as `negative-size-param`.

```cpp
#include <limits>
#include <random>
#include "impl/heap/distance_heap.h"  // src/ is on the include path
using namespace vsag;

// (use any Allocator; SafeAllocator::FactoryDefaultAllocator() in-tree)
const int64_t topk = 5;  // < memmove_maxsize (10) -> non-fixed MemmoveHeap<true,false>
auto heap = DistanceHeap::MakeInstanceBySize<true, false>(allocator, topk);

std::mt19937 rng(0);
std::uniform_real_distribution<float> distrib_real(0.1, 0.9);
auto cur_heap_top = std::numeric_limits<float>::max();
for (uint64_t i = 0; i < 1000; ++i) {                       // mirrors ivf.cpp:802-816
    float d = distrib_real(rng);
    if (heap->Size() < (uint64_t)topk || d < cur_heap_top) heap->Push(d, (InnerIdType)i);
    if (heap->Size() > (uint64_t)topk) heap->Pop();
    if (!heap->Empty() && heap->Size() == (uint64_t)topk) cur_heap_top = heap->Top().first;
}
```

End-to-end, the official `examples/cpp/106_index_ivf.cpp` reproduces it by
changing `int64_t topk = 10;` to `topk = 5;` and issuing several random queries
(a single query is data-dependent; looping a handful of queries makes the crash
reliable).

### Environment
```shell
- OS: Ubuntu 22.04
- vsag version: v0.18.0-289-g42eca652

# Reproduced with a local AddressSanitizer build of the heap sources
# (the default build has ENABLE_ASAN=OFF).
```

### Expected behavior
`KnnSearch` with `topk < 10` returns the correct top-k results without memory
corruption, identical in behavior to `topk >= 10` (which uses `StandardHeap`).

### Actual behavior
Heap-buffer-overflow / SIGSEGV. AddressSanitizer report from the reproducer
above (`seed = 0`, `topk = 5`):

```
==ERROR: AddressSanitizer: negative-size-param: (size=-96)
    #0 __interceptor_memmove
    #1 vsag::MemmoveHeap<true, false>::Push(float, unsigned int) src/impl/heap/memmove_heap.cpp:58
    #7 vsag::MemmoveHeap<true, false>::Push(float, unsigned int) src/impl/heap/memmove_heap.cpp:56
```

`size = -96` equals `(cur_size_ - pos) * sizeof(DistanceRecord)` =
`(5 - 17) * 8 = -96`, reinterpreted as a huge `size_t`. At the crash point the
21-slot physical buffer holds only 5 logically-valid records; the 16 stale-tail
records (left by `Pop()`) include values smaller than the new distance, pushing
`upper_bound` to `pos = 17 > cur_size_ = 5`.

Secondary effect: because `Pop()` never shrinks `ordered_buffer_` and the
non-fixed `Push()` always `emplace_back`s, the buffer grows unbounded. For a
monotonically-decreasing per-bucket distance sequence, a 100k-candidate scan
grows the buffer to ~196k records (~1.5 MB) while the logical size stays at
`topk`.

### References
- `src/impl/heap/memmove_heap.cpp:51-63` — non-fixed `Push` uses `upper_bound(begin(), end(), …)` and `memmove((cur_size_ - pos) …)`
- `src/impl/heap/memmove_heap.cpp:35-43` — fixed-size branch (correct: bounds at `begin() + cur_size_`)
- `src/impl/heap/memmove_heap.h:37-39` — `Pop()` only does `cur_size_--`
- `src/impl/heap/distance_heap.cpp:25-26` — `memmove_maxsize = 10` selects `MemmoveHeap` for `max_size < 10`
- `src/algorithm/ivf/ivf.cpp:769` — `MakeInstanceBySize<true, false>(topk)` (non-fixed)
- `src/algorithm/ivf/ivf.cpp:802-816` — Push/Pop interleaving in the KNN scan loop

### Additional context
Existing tests miss this because they sit in the intersection of two blind
spots: `distance_heap_test.cpp`'s `RunBasicTest` exercises
`MemmoveHeap<true, false>` but pushes all elements then pops all — it never
interleaves `Push`/`Pop`, so the stale tail is never present during a `Push`.
IVF tests search with `top_k = 10`, which selects `StandardHeap` and never
reaches `MemmoveHeap`.

Suggested fix: bound the non-fixed branch's `upper_bound` at
`begin() + cur_size_` to match the fixed-size branch, and reuse the stale tail
slot instead of `emplace_back` when `ordered_buffer_.size() > cur_size_` (this
also bounds the buffer growth). Verified locally across 4500 cases (500 seeds x
topk 1-9) against a brute-force top-k reference: no overflow, results match, and
peak memory drops from ~1.5 MB to 96 bytes in the worst case.

## Related issues
_None found._

---
_Drafted with AI assistant: ClaudeCode:claude-opus-4-8_

jac0626:
## Change Type

- [x] Bug fix

## Linked Issue

- Fixes: #2298

## What Changed

The non-fixed `MemmoveHeap` (`MemmoveHeap<max_heap, false>`) ran
`std::upper_bound` over the **entire** `ordered_buffer_` (`begin()..end()`).
Because `Pop()` only decrements `cur_size_` without shrinking `ordered_buffer_`,
a stale tail accumulates, so the insertion index `pos` could exceed `cur_size_`
and make `memmove(..., (cur_size_ - pos) * sizeof(DistanceRecord))` underflow to
a huge `size_t`, overrunning the heap buffer.

`DistanceHeap::MakeInstanceBySize` selects `MemmoveHeap` only when
`max_size < 10`, and `IVF::search` requests the non-fixed variant
(`MakeInstanceBySize<true, false>(topk)`) then interleaves `Push` with a bounded
`Pop`. So any `KnnSearch` with `topk < 10` reaches the buggy path; whether it
overruns is data-dependent. The fixed-size branch was already correct (it bounds
`upper_bound` at `begin() + cur_size_`).

- Bound the non-fixed branch's `upper_bound` at `begin() + cur_size_`, matching
  the already-correct fixed-size branch.
- Reuse the stale tail slot instead of `emplace_back` when
  `ordered_buffer_.size() > cur_size_`, which also bounds the previously
  unbounded buffer growth.
- Tests: a regression test (`memmove_heap non-fixed Pop-then-Push (topk<10)` in
  `src/impl/heap/distance_heap_test.cpp`) replaying IVF's `Push`/`Pop`
  interleaving across `topk` 1–9 with a triggering seed, validated against a
  brute-force top-k.

## Test Evidence

- [x] `make fmt`
- [x] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
# Red -> Green (normal Debug build, ENABLE_ASAN=OFF):
#   before fix: ./build/tests/unittests "memmove_heap non-fixed Pop-then-Push (topk<10)"
#     => wrong results (0.0f leaked into top-5) + free(): invalid pointer -> SIGABRT (exit 134)
#   after fix:
./build/tests/unittests "memmove_heap non-fixed Pop-then-Push (topk<10)"
  => All tests passed (18 assertions in 1 test case)

# Regression (no behavior change on existing heap paths):
./build/tests/unittests "[distance_heap]"   => All tests passed (8106 assertions in 3 test cases)
./build/tests/unittests "[IVFNearestPartition][ut],[IVFParameter][ut]"
  => All tests passed (3035 assertions in 4 test cases)

# End-to-end: 200 real IVF KnnSearch queries at topk=5 against a rebuilt libvsag.so
#   => "OK: 200 topk=5 queries returned 5 results each, no crash"

# Cross-checked against a brute-force top-k over 4500 cases (500 seeds x topk 1-9):
#   no overflow, results match, worst-case peak buffer memory ~1.5 MB -> 96 bytes.

# clang-format-15 / clang-tidy-15: clean on memmove_heap.cpp. The test file is
# excluded from `make lint` by the repo's `(?<!_test)` source filter.
# Note: the full `make test` suite was not run end-to-end here; the targeted
# heap + IVF unit runs and the end-to-end IVF check above cover the changed paths.
```

## Compatibility Impact

- API/ABI compatibility: none (no public header changes; the change is internal
  to `src/impl/heap/`).
- Behavior changes: `KnnSearch` with `topk < 10` now returns correct top-k
  results without memory corruption, matching `topk >= 10` (`StandardHeap`).

- Performance impact: negligible — the non-fixed `Push` now searches a tighter
  range (`[begin, begin + cur_size_)`) and reuses slots, reducing reallocations.
- Concurrency/thread-safety impact: none — each IVF search thread owns its heap.

## Documentation Impact

- [x] No docs update needed

## Risk and Rollback

- Risk level: low — change is localized to the non-fixed `MemmoveHeap::Push`
  branch; the fixed-size branch and `StandardHeap` are untouched.
- Rollback plan: revert the commit; no data format or API changes are involved.

## Checklist

- [x] I have updated docs if behavior/workflow changed (n/a)

### DIFF
--- a/src/impl/heap/distance_heap_test.cpp
+++ b/src/impl/heap/distance_heap_test.cpp
@@ -15,6 +15,10 @@
 
 #include "distance_heap.h"
 
+#include <algorithm>
+#include <limits>
+#include <vector>
+
 #include "impl/allocator/safe_allocator.h"
 #include "memmove_heap.h"
 #include "standard_heap.h"
@@ -105,5 +109,60 @@ TEST_CASE_METHOD(TestDistanceHeap, "memmove_heap test", "[ut][distance_heap]") {
         RunBasicTest(heap3, false);
         MemmoveHeap<false, false> heap4(allocator.get(), max_size);
         RunBasicTest(heap4, false);
+    }
+}
+
+// Regression for the non-fixed MemmoveHeap buffer overflow hit by IVF::search
+// when topk < 10. The non-fixed heap's Pop() only decrements cur_size_ without
+// shrinking ordered_buffer_, so a stale tail accumulates. IVF interleaves
+// Push with a bounded Pop (the loop replayed below); the non-fixed Push then
+// searched the whole buffer (including the stale tail) and could compute a
+// negative memmove size, overrunning the heap. This exercises that exact
+// interleaving and validates the kept records against a brute-force top-k.
+//
+// RunBasicTest does not catch it because it pushes all elements then pops all,
+// never interleaving Push/Pop; IVF's own tests use top_k == 10, which selects
+// StandardHeap rather than MemmoveHeap.
+TEST_CASE_METHOD(TestDistanceHeap,
+                 "memmove_heap non-fixed Pop-then-Push (topk<10)",
+                 "[ut][distance_heap]") {
+    auto allocator = SafeAllocator::FactoryDefaultAllocator();
+    // seed 0 produces a distance sequence that triggers the overflow within a
+    // few dozen iterations for several of these topk values.
+    auto dists = fixtures::GenerateVectors<float>(1000, 1, 0, false);
+
+    for (int64_t topk = 1; topk < 10; ++topk) {
+        MemmoveHeap<true, false> heap(allocator.get(), topk);
+        auto cur_heap_top = std::numeric_limits<float>::max();
+
+        // Mirror the KNN scan loop in IVF::search (src/algorithm/ivf/ivf.cpp).
+        for (uint64_t i = 0; i < 1000; ++i) {
+            float d = dists[i];
+            if (heap.Size() < static_cast<uint64_t>(topk) or d < cur_heap_top) {
+                heap.Push(d, static_cast<InnerIdType>(i));
+            }
+            if (heap.Size() > static_cast<uint64_t>(topk)) {
+                heap.Pop();
+            }
+            if (not heap.Empty() and heap.Size() == static_cast<uint64_t>(topk)) {
+                cur_heap_top = heap.Top().first;
+            }
+        }
+
+        REQUIRE(heap.Size() == static_cast<uint64_t>(topk));
+
+        // The retained records must be exactly the topk smallest distances.
+        std::vector<float> got;
+        const auto* heap_data = heap.GetData();
+        for (uint64_t i = 0; i < heap.Size(); ++i) {
+            got.emplace_back(heap_data[i].first);
+        }
+        std::sort(got.begin(), got.end());
+
+        std::vector<float> expected(dists.begin(), dists.begin() + 1000);
+        std::sort(expected.begin(), expected.end());
+        expected.resize(topk);
+
+        REQUIRE(got == expected);
     }
 }
--- a/src/impl/heap/memmove_heap.cpp
+++ b/src/impl/heap/memmove_heap.cpp
@@ -54,10 +54,13 @@ MemmoveHeap<max_heap, fixed_size>::Push(float dist, InnerIdType id) {
                                     record,
                                     CompareType()) -
                    this->ordered_buffer_.begin();
-        if (cur_size_ < static_cast<int64_t>(ordered_buffer_.size())) {
-            ordered_buffer_[cur_size_] = record;
-        } else {
-            ordered_buffer_.push_back(record);
+        // Pop() only decrements cur_size_, so ordered_buffer_ may already hold a
+        // stale tail beyond cur_size_. Reuse that slot instead of emplace_back,
+        // which both bounds the buffer growth and keeps the upper_bound above
+        // (which searches only [begin, begin + cur_size_)) consistent with the
+        // memmove size below.
+        if (static_cast<int64_t>(ordered_buffer_.size()) <= cur_size_) {
+            ordered_buffer_.emplace_back(record);
         }
         // NOLINTNEXTLINE(bugprone-undefined-memory-manipulation)
         memmove(ordered_buffer_.data() + pos + 1,


LHT129 (commented):
## Code Review

### Positives
- Well-targeted, correct, and low-risk fix for a genuine buffer overflow bug
- Root cause analysis is thorough and accurate
- Fix aligns non-fixed branch upper_bound range with the already-correct fixed-size branch
- Slot-reuse optimization bounds buffer growth
- Regression test directly exercises the buggy interleaving pattern

### Issues to Address

1. Full test suite not run: PR description notes make test was not run end-to-end. Should verify no regressions in IVF and hgraph test suites.
2. min_heap non-fixed variant not tested: MemmoveHeap<false,false> with interleaved Push/Pop is not covered by the new test.

### Suggestions
- Add assert(cur_size_ > 0) in Pop() and Top() for debug builds
- Document invariant that cur_size_ tracks logical size while ordered_buffer_ may be physically larger
- Consider making Pop() overwrite stale slot (set to sentinel {0,0}) to make invariants visible
- Add test assertion that ordered_buffer_.size() stays bounded after many Push/Pop cycles
- hgraph_search.cpp also uses MakeInstanceBySize<true,false> with potentially small k - verify fix covers this path

LHT129 (approved):
LGTM

PR #2324, opened 2026-06-24
fix: initialize IVF reader IO reorder data

### LINKED ISSUE #2302: IVF reader_io search path is not usable
### Affected area
ivf

### Interface
cpp

### VSAG version
v0.18.0-289-g42eca652-dirty (42eca652744ee7f715561ebf64f027fed5ba6b91)

### Summary
IVF with `precise_io_type = reader_io` fails at search time with `ReaderIO is not initialized`, so the `reader_io` disk-reading path is not usable in the current IVF implementation.

### Minimal reproducer
```cpp
std::string params = R"(
{
  "dtype": "float32",
  "metric_type": "l2",
  "dim": 64,
  "index_param": {
    "buckets_count": 32,
    "base_quantization_type": "sq8",
    "ivf_train_type": "kmeans",
    "train_sample_count": 1500,
    "use_reorder": true,
    "precise_quantization_type": "fp32",
    "precise_io_type": "reader_io"
  }
}
)";
auto index = vsag::Factory::CreateIndex("ivf", params);
```

### Environment
- OS: Ubuntu 22.04
- vsag version: v0.18.0-289-g42eca652-dirty

### Expected behavior
`reader_io` should be usable for IVF if it is listed in the docs, or the docs should explicitly say it is unsupported.

### Actual behavior
`KnnSearch` fails with `ReaderIO is not initialized, please call Init() first.`

### References
- https://vsag.io/docs/en/indexes/ivf.html
- src/io/reader_io.cpp:42-55
- src/algorithm/ivf/ivf.cpp:260-263
- src/datacell/flatten_interface.cpp:245-247
- src/algorithm/inner_index_interface.cpp:215-244

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2302

## What Changed

- Initialize IVF reorder `ReaderIO` when deserializing from `ReaderSet`.
- Deserialize IVF `reader_io` reorder codes from the full stream position so `ReaderIO::start_` keeps the absolute offset instead of a slice-relative offset.
- Add a regression test covering IVF `precise_io_type = reader_io` with `ReaderSet` deserialization and search.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
./scripts/format/format-cpp.sh
Using clang-format version 15 (required: 15)
Code formatting completed with /opt/homebrew/opt/llvm@15/bin/clang-format

git diff --check

# macOS validation on codex/macos-build-baseline with local roaringbitmap source override:
make test CASE='"(PR) IVF Reader IO"' COMPILE_JOBS=6 EXTRA_DEFINED='-DFETCHCONTENT_SOURCE_DIR_ROARINGBITMAP=/Users/jiangchao/workspace/vsag/build/_deps/roaringbitmap-src'
All tests passed (1307 assertions in 1 test case)
```

Note: Direct `make test` on upstream/main in this macOS worktree is blocked before compiling this change by the existing AppleClang/libc++ baseline issue, so the targeted test was verified on the macOS build baseline branch.

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: IVF `ReaderSet` deserialization now supports documented `precise_io_type = reader_io`.

- Performance impact: none expected
- Concurrency/thread-safety impact: none

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other:

## Risk and Rollback

- Risk level: low
- Rollback plan: revert this commit.

## Checklist

### DIFF
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -656,7 +656,8 @@ IVF::Deserialize(StreamReader& reader) {
             this->check_reorder_codes_ready();
             if (this->use_reader_io_for_reorder()) {
                 buffer_reader.PushSeek(datacell_offsets["reorder_codes"].GetInt());
-                this->reorder_codes_->Deserialize(buffer_reader);
+                this->reorder_codes_->Deserialize(
+                    buffer_reader.Slice(datacell_sizes["reorder_codes"].GetInt()));
                 buffer_reader.PopSeek();
             } else {
                 READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -42,6 +42,59 @@
 #include "vsag_exception.h"
 
 namespace vsag {
+static uint64_t
+checked_datacell_end(uint64_t begin, uint64_t length) {
+    if (length > std::numeric_limits<uint64_t>::max() - begin) {
+        throw VsagException(ErrorType::READ_ERROR, "Serialized datacell boundary overflows");
+    }
+    return begin + length;
+}
+
+class AbsoluteBoundedStreamReader : public StreamReader {
+public:
+    AbsoluteBoundedStreamReader(StreamReader* reader, uint64_t begin, uint64_t length)
+        : StreamReader(checked_datacell_end(begin, length)),
+          reader_(reader),
+          begin_(begin),
+          end_(begin + length) {
+        reader_->Seek(begin_);
+    }
+
+    void
+    Read(char* data, uint64_t size) override {
+        auto cursor = this->GetCursor();
+        if (cursor > end_ or size > end_ - cursor) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                "Read operation exceeds serialized datacell boundary");
+        }
+        reader_->Read(data, size);
+    }
+
+    void
+    Seek(uint64_t cursor) override {
+        if (cursor < begin_ or cursor > end_) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                "Seek operation exceeds serialized datacell boundary");
+        }
+        reader_->Seek(cursor);
+    }
+
+    [[nodiscard]] uint64_t
+    GetCursor() const override {
+        return reader_->GetCursor();
+    }
+
+    [[nodiscard]] uint64_t
+    Length() override {
+        return end_;
+    }
+
+private:
+    StreamReader* const reader_{nullptr};
+    uint64_t begin_{0};
+    uint64_t end_{0};
+};
+
 static constexpr const char* IVF_PARAMS_TEMPLATE =
     R"(
     {
@@ -655,9 +708,13 @@ IVF::Deserialize(StreamReader& reader) {
         if (use_reorder_) {
             this->check_reorder_codes_ready();
             if (this->use_reader_io_for_reorder()) {
-                buffer_reader.PushSeek(datacell_offsets["reorder_codes"].GetInt());
+                auto reorder_codes_offset = datacell_offsets["reorder_codes"].GetInt();
+                auto reorder_codes_size = datacell_sizes["reorder_codes"].GetInt();
+                buffer_reader.PushSeek(reorder_codes_offset);
+                AbsoluteBoundedStreamReader reorder_codes_reader(
+                    &buffer_reader, reorder_codes_offset, reorder_codes_size);
                 // ReaderIO records absolute offsets into the owning index reader.
-                this->reorder_codes_->Deserialize(buffer_reader);
+                this->reorder_codes_->Deserialize(reorder_codes_reader);
                 buffer_reader.PopSeek();
             } else {
                 READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -32,6 +32,7 @@
 #include "index/index_impl.h"
 #include "index_feature_list.h"
 #include "inner_string_params.h"
+#include "io/reader_io_parameter.h"
 #include "ivf_nearest_partition.h"
 #include "query_context.h"
 #include "storage/serialization.h"
@@ -610,6 +611,7 @@ IVF::Deserialize(StreamReader& reader) {
         this->partition_strategy_->Deserialize(buffer_reader);
         this->label_table_->Deserialize(buffer_reader);
         if (use_reorder_) {
+            this->check_reorder_codes_ready();
             this->reorder_codes_->Deserialize(buffer_reader);
         }
 
@@ -651,7 +653,14 @@ IVF::Deserialize(StreamReader& reader) {
         READ_DATACELL_WITH_NAME(buffer_reader, "partition_strategy", this->partition_strategy_);
         READ_DATACELL_WITH_NAME(buffer_reader, "label_table", this->label_table_);
         if (use_reorder_) {
-            READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
+            this->check_reorder_codes_ready();
+            if (this->use_reader_io_for_reorder()) {
+                buffer_reader.PushSeek(datacell_offsets["reorder_codes"].GetInt());
+                this->reorder_codes_->Deserialize(buffer_reader);
+                buffer_reader.PopSeek();
+            } else {
+                READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
+            }
         }
         if (use_attribute_filter_) {
             READ_DATACELL_WITH_NAME(buffer_reader, "attr_filter_index", this->attr_filter_index_);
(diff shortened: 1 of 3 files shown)


src/algorithm/ivf/ivf.cpp
```diff
@@ -651,7 +653,15 @@ IVF::Deserialize(StreamReader& reader) {
...
-            READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
+            this->check_reorder_codes_ready();
+            if (this->use_reader_io_for_reorder()) {
+                buffer_reader.PushSeek(datacell_offsets["reorder_codes"].GetInt());
+                // ReaderIO records absolute offsets into the owning index reader.
+                this->reorder_codes_->Deserialize(buffer_reader);
+                buffer_reader.PopSeek();
+            } else {
+                READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
+            }
```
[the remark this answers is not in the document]
jac0626: Fixed in 5b7366c4. The reader_io reorder path now uses an absolute-cursor bounded reader over the reorder_codes datacell, so ReaderIO still records absolute offsets while reads and seeks cannot cross the serialized datacell boundary.

src/algorithm/ivf/ivf.cpp
```diff
@@ -651,7 +706,19 @@ IVF::Deserialize(StreamReader& reader) {
         READ_DATACELL_WITH_NAME(buffer_reader, "partition_strategy", this->partition_strategy_);
         READ_DATACELL_WITH_NAME(buffer_reader, "label_table", this->label_table_);
         if (use_reorder_) {
-            READ_DATACELL_WITH_NAME(buffer_reader, "reorder_codes", this->reorder_codes_);
+            this->check_reorder_codes_ready();
+            if (this->use_reader_io_for_reorder()) {
+                auto reorder_codes_offset = datacell_offsets["reorder_codes"].GetInt();
+                auto reorder_codes_size = datacell_sizes["reorder_codes"].GetInt();
+                buffer_reader.PushSeek(reorder_codes_offset);
```
LHT129, line 713: [suggestion] `datacell_offsets["reorder_codes"].GetInt()` and `datacell_sizes["reorder_codes"].GetInt()` return signed `int64_t` values, but are immediately passed as `uint64_t` to `PushSeek` and the `AbsoluteBoundedStreamReader` constructor. If the serialized metadata is corrupted and contains negative values, the implicit signed-to-unsigned conversion will wrap to a very large `uint64_t`, causing seeks/reads at unexpected offsets far beyond the serialized data.

Consider adding an explicit non-negative check before the conversion:

```cpp
auto reorder_codes_offset = datacell_offsets["reorder_codes"].GetInt();
auto reorder_codes_size = datacell_sizes["reorder_codes"].GetInt();
if (reorder_codes_offset < 0 or reorder_codes_size < 0) {
    throw VsagException(ErrorType::READ_ERROR,
                        "Invalid reorder_codes offset/size in serialized metadata");
}
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -41,6 +42,59 @@
...
+        }
+        reader_->Seek(cursor);
+    }
+
+    [[nodiscard]] uint64_t
+    GetCursor() const override {
+        return reader_->GetCursor();
+    }
+
+    [[nodiscard]] uint64_t
```
LHT129, line 87: [note] `AbsoluteBoundedStreamReader::Length()` returns `end_` which is the absolute end offset (`begin + length`), not the length of the bounded region. This is semantically inconsistent with `SliceStreamReader::Length()` which returns the slice length. While this does not affect the current deserialization path (neither `FlattenInterface::Deserialize` nor `ReaderIO` call `Length()`), it could cause subtle bugs if `Length()` is used by future callers expecting the bounded-region length rather than an absolute offset.

Consider changing to return the actual bounded length:

```cpp
[[nodiscard]] uint64_t
Length() override {
    return end_ - begin_;
}
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -41,6 +42,59 @@
...
+    if (length > std::numeric_limits<uint64_t>::max() - begin) {
+        throw VsagException(ErrorType::READ_ERROR, "Serialized datacell boundary overflows");
+    }
+    return begin + length;
+}
+
+class AbsoluteBoundedStreamReader : public StreamReader {
+public:
+    AbsoluteBoundedStreamReader(StreamReader* reader, uint64_t begin, uint64_t length)
+        : StreamReader(checked_datacell_end(begin, length)),
```
LHT129, line 56: [note] In the `AbsoluteBoundedStreamReader` constructor, `end_` is computed as `begin + length` without the overflow check that `checked_datacell_end` applies to the base `StreamReader`. While this is currently safe because the base-class initializer runs first and throws on overflow (preventing `end_` from being initialized with a wrapped value), the inconsistency makes the code fragile — if someone later removes or reorders the base-class check, `end_` could silently overflow.

Consider applying the same overflow check to `end_`:

```cpp
AbsoluteBoundedStreamReader(StreamReader* reader, uint64_t begin, uint64_t length)
    : StreamReader(checked_datacell_end(begin, length)),
      reader_(reader),
      begin_(begin),
      end_(checked_datacell_end(begin, length)) {
```

LHT129 (commented):
Thanks for this fix. The PR correctly addresses the IVF reader IO reorder initialization issue (#2302).

The `AbsoluteBoundedStreamReader` is a clean solution that preserves absolute cursor semantics for `ReaderIO::start_` while enforcing datacell boundary safety. The null-pointer guard via `check_reorder_codes_ready()` is also a good defensive addition.

A couple of existing review comments from @LHT129 are worth addressing before merge:
- The signed-to-unsigned conversion of `GetInt()` return values (line 713) could wrap on corrupted metadata — consider adding an explicit non-negative check.
- `AbsoluteBoundedStreamReader::Length()` returns `end_` (absolute offset) rather than `end_ - begin_` (bounded length), which is semantically inconsistent with `SliceStreamReader::Length()`.

The regression test covers the `reader_io` + `ReaderSet` deserialization + `CalDistanceById` path well.

src/algorithm/ivf/ivf.cpp
```diff
@@ -665,6 +732,16 @@ IVF::Deserialize(StreamReader& reader) {
     this->cal_memory_usage();
 }
```
LHT129, line 734: [suggestion] `SetIO` takes `std::shared_ptr<Reader>` by value, which causes an unnecessary atomic reference-count increment/decrement on every call. Since the parameter is only stored (not moved-from), it should be passed by const reference.

Current (ivf.cpp:734, ivf.h:162):
```cpp
void SetIO(const std::shared_ptr<Reader> reader) override;
```

Suggested:
```cpp
void SetIO(const std::shared_ptr<Reader>& reader) override;
```

LHT129 (commented):
Thanks for this fix. The changes look good overall:

1. **`AbsoluteBoundedStreamReader`** — correctly preserves absolute cursor semantics for `ReaderIO::start_` while enforcing datacell boundary checks on reads and seeks. The overflow check via `checked_datacell_end` is a nice touch.

2. **`check_reorder_codes_ready()`** — addresses the earlier null-pointer concern raised in review, providing a clear error message when `reorder_codes_` is unexpectedly null.

3. **`SetIO` override** — properly initializes the reorder codes IO parameter when the index is loaded via `ReaderSet`, which was the root cause of the bug.

4. **Test coverage** — the regression test covers the full `reader_io` + `ReaderSet` deserialization + search + `CalDistanceById` path, which is the exact scenario that was broken.

No blocking issues found. The existing inline comments from other reviewers cover minor improvement suggestions that are non-blocking.

LHT129 (commented):
I have completed a thorough review of this PR. The changes look good overall.

**Summary of the fix:**
- The `AbsoluteBoundedStreamReader` correctly wraps the reorder_codes datacell region so that `ReaderIO` records absolute offsets while reads/seeks are bounded to the datacell boundary.
- `check_reorder_codes_ready()` adds a proper null-guard before dereferencing `reorder_codes_`.
- `SetIO` properly propagates the reader to reorder_codes when `use_reorder_` is true.
- The regression test validates the `reader_io` deserialization path by comparing distances against a reference index.

**Existing review comments that still apply (from LHT129):**
1. `AbsoluteBoundedStreamReader::Length()` returns `end_` (absolute offset) rather than the bounded region length — semantically inconsistent with `SliceStreamReader::Length()`. Not currently called by the deserialization path, but worth fixing for future callers.
2. `end_` in the constructor is computed as `begin + length` without the same overflow check applied to the base class — currently safe because the base initializer throws first, but fragile.
3. `SetIO` takes `shared_ptr` by value — though this matches the base class signature, so changing it would require a broader refactor.

These are all `[note]`/`[suggestion]` level and do not block merging. The core fix is correct and the test coverage is adequate.

LHT129 (commented):
Thank you for this fix. The change correctly addresses the IVF reader IO reorder initialization issue (#2302) by:

1. Adding `AbsoluteBoundedStreamReader` to provide bounded reads/seeks while preserving absolute cursor semantics for `ReaderIO::start_`.
2. Adding `check_reorder_codes_ready()` null-safety guard before deserializing reorder codes.
3. Adding `SetIO` override to initialize reorder codes IO when loading from a `ReaderSet`.
4. Adding a regression test `(PR) IVF Reader IO` that covers the `reader_io` + `ReaderSet` deserialization path.

The implementation looks solid. The existing inline comments from LHT129 cover a few minor improvement suggestions (`Length()` semantics, `end_` overflow check consistency, `SetIO` parameter passing) that are worth considering but non-blocking.

LGTM overall.

PR #2342, opened 2026-06-25
ci: reuse ASan build artifacts for compatibility check

LHT129:
## Summary

Eliminate the redundant  build in the  job by reusing the  artifact.

### 
- **build-asan-x86**: add  to compile  tool alongside the ASan build
- **build-asan-x86**: remove  from the Clean step to preserve the tools artifact
- **compatibility**: change  from  to 
- **compatibility**: replace  with  step
- **compatibility**: add ,  step for consistency with other container jobs

### 
- Parameterize build directory via  environment variable (default:  for backward compatibility)

## Why

 defaults to  (), and tools are gated by  (). So adding  to  is sufficient to build the compatibility tool in the ASan build, avoiding a full redundant Release build.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -88,7 +88,7 @@ jobs:
         # ci-x86-openblas image to keep iteration latency low. The
         # build-from-source OpenBLAS path is exercised by the daily x86 ASan
         # job in daily_test.yml.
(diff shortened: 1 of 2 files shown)


LHT129:
## Review fixes (f3ce005c)

- **gemini: tool validation** — Added existence + executability check for `compatibility_tool` before running the loop
- **gemini/copilot: quoting** — Replaced inline `"$build_dir"/tools/...` with `$compatibility_tool` variable
- **copilot: /opt mount** — This pattern is already used in all other containerized jobs in the same workflow (`build-asan-x86`, test jobs, etc.). Keeping consistent. The toolcache is not needed inside the container.
- **copilot: format dep** — Intentional: `compatibility` depends on `build-asan-x86`, which itself depends on `format`. If format fails, the entire chain is skipped.

wxyucs (approved):
lgtm

PR #2354, opened 2026-06-26
fix(hgraph): refresh duplicate state on vector update

### LINKED ISSUE #2322: [bug](hgraph): UpdateVector() does not update graph topology or duplicate-group membership
### Affected area

hgraph

### Interface

cpp

### VSAG version

main @ b944730bd90c23abbdbde5680d2367a7f79a9d0e (pre-existing)

### Summary

`HGraph::UpdateVector()` updates the physical vector storage (quantized codes, raw vectors) but does **not** update graph topology or duplicate-group membership. Two independent problems result:

1. **Stale graph edges** — The updated node's neighbor list is not rebuilt, and other nodes' reverse-edges to the updated node are not removed or re-evaluated. After update, the graph still reflects the old vector position.

2. **Stale duplicate-group membership** — The updated ID remains in its previous duplicate group. Search-time duplicate expansion (`GetDuplicateIds()`) can return the updated ID with the old representative's distance, even though the vectors are no longer duplicates.

Both problems produce silently incorrect search results after `UpdateVector()`.

### Steps to reproduce

```cpp
auto index = vsag::Factory::CreateIndex(
    "hgraph",
    R"({
        "dtype": "float32",
        "metric_type": "l2",
        "dim": 2,
        "index_param": {
            "base_quantization_type": "fp32",
            "max_degree": 16,
            "ef_construction": 100,
            "support_duplicate": true
        }
    })")
    .value();

// Insert two identical vectors — they become duplicates
float duplicate_vectors[] = {1.0F, 0.0F, 1.0F, 0.0F};
int64_t duplicate_ids[] = {10, 11};
auto base = vsag::Dataset::Make()
                ->NumElements(2)->Dim(2)->Ids(duplicate_ids)
                ->Float32Vectors(duplicate_vectors)->Owner(false);
index->Build(base).value();

// Update ID 11 to a completely different vector
float updated_vector[] = {0.0F, 1.0F};
int64_t updated_id[] = {11};
auto update = vsag::Dataset::Make()
                  ->NumElements(1)->Dim(2)->Ids(updated_id)
                  ->Float32Vectors(updated_vector)->Owner(false);
index->UpdateVector(11, update, true).value();

// Search from [1,0]
float query_vector[] = {1.0F, 0.0F};
auto query = vsag::Dataset::Make()
                 ->NumElements(1)->Dim(2)
                 ->Float32Vectors(query_vector)->Owner(false);

auto result = index->KnnSearch(query, 2, R"({"hgraph":{"ef_search":32}})").value();

// Bug 1 (graph edges): ID 11 is still a neighbor of ID 10 in the graph.
//   The search may traverse the stale edge A->B and use the wrong distance.
// Bug 2 (duplicate group): ID 11 is still in ID 10's duplicate group.
//   Search expands ID 10's duplicates and assigns ID 10's distance
//   (approx 0) to ID 11, even though ID 11 now stores [0,1] and the true
//   L2 distance from [1,0] is approx 1.414.
```

### Environment

```shell
- OS: any
- vsag version: main (any commit after UpdateVector was introduced)
- compiler version: any
```

### Expected behavior

After `UpdateVector()`, the index should maintain consistency at both levels:

**Graph topology** — The updated node's neighbor list should be rebuilt (remove stale edges, search for new nearest neighbors with the updated vector), and reverse-edges in other nodes should be updated accordingly.

**Duplicate-group membership** — The updated ID should be removed from its previous duplicate group. If the new vector is still within `duplicate_distance_threshold` of an existing group member, it may be re-added; otherwise it should stand alone.

### Actual behavior

`UpdateVector()` (lines 1148-1238 of `hgraph.cpp`) only writes new codes to the physical storage layers:
- `basic_flatten_codes_->UpdateVector()`
- `high_precise_codes_->UpdateVector()` (if precise reorder is enabled)
- `raw_vector_->UpdateVector()` (if raw vectors are stored)

When `force_update` is `false`, there is a lightweight validation that checks whether **existing neighbors** are still farther than the updated node (lines 1181-1199), and returns `false` without updating if not. But even when the update proceeds, **no graph restructuring** or **duplicate-group cleanup** occurs.

The `DuplicateInterface` (`duplicate_interface.h`) currently exposes only `SetDuplicateId()` and `GetDuplicateIds()` — there is **no `RemoveId()` or equivalent operation** for removing a logical ID from a duplicate group.

### References

- `src/algorithm/hgraph/hgraph.cpp:1148-1238` — `UpdateVector()` (codes only, no graph or duplicate maintenance)
- `src/algorithm/hgraph/hgraph_build.cpp:534-554` — insertion records duplicate relation via `SetDuplicateId()`
- `src/datacell/duplicate_interface.h` — no `RemoveId()` method exposed
- `src/datacell/dense_duplicate_tracker.h` / `.cpp` — no removal operation
- `src/datacell/sparse_duplicate_tracker.h` / `.cpp` — no removal operation
- `src/impl/searcher/basic_searcher.cpp:~470` — search expands duplicate IDs using visited node distance
- #2295 — duplicate-group single-linkage transitivity (related but separate issue)
- #1810 — remove support for graph indexes (exists but not called from `UpdateVector`)

### Additional context

A complete fix likely requires:

1. Add `RemoveId(InnerIdType id)` to `DuplicateInterface` and implement in both `DenseDuplicateTracker` and `SparseDuplicateTracker`
2. In `HGraph::UpdateVector()`, after updating physical codes, call `RemoveId()` to clean old duplicate membership, then re-evaluate whether the updated vector belongs to any existing group
3. In `HGraph::UpdateVector()`, after updating physical codes, remove the node from the graph and re-insert it (or perform a localized neighbor repair) so that graph edges reflect the new vector position

Regression tests should cover:
- Updating one member of a duplicate group
- Updating the representative member of a duplicate group
- Groups with more than two IDs
- Search distance correctness after update
- Graph edge correctness after update (neighbors are valid for the new vector)
- Serialization after duplicate membership changes

_Drafted with AI assistant: Codex:gpt-5.5_

jac0626:
## Summary

- add duplicate tracker removal support for dense and sparse duplicate trackers
- refresh HGraph duplicate-member state during `UpdateVector()` by removing stale duplicate membership and reconnecting the updated member to the bottom graph
- update raw vector storage when present and add regression coverage for duplicate update search correctness

## Root Cause

`HGraph::UpdateVector()` refreshed physical vector storage but left duplicate-only IDs in their old duplicate group and outside graph topology. Search-time duplicate expansion could then return the updated ID with the representative node's distance, and the updated ID was not searchable at its new position.

## Validation

- `git diff --check`
- `make test CASE='"DenseDuplicateTracker removes duplicate ids","SparseDuplicateTracker removes duplicate ids"'`
- `./build/tests/unittests -d yes "[DenseDuplicateTracker],[SparseDuplicateTracker]" --allow-running-no-tests ""`
- `make test CASE='"HGraph UpdateVector refreshes duplicate member state"'`

Fixes: #2322

### DIFF
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -637,7 +637,9 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) 
         return false;
     }
     if (was_duplicate_member) {
-        this->bottom_graph_->RemoveDuplicateId(inner_id);
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
         std::shared_lock rlock(this->global_mutex_);
         this->graph_add_one(new_base_vec, -1, inner_id);
     }
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -585,8 +585,12 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) 
     void* new_base_vec = nullptr;
     uint64_t data_size = 0;
     get_vectors(data_type_, dim_, new_base, &new_base_vec, &data_size);
-    const bool was_duplicate_member =
-        this->support_duplicate_ && this->bottom_graph_->GetGroupId(inner_id) != inner_id;
+    std::vector<InnerIdType> duplicate_ids;
+    bool was_duplicate_member = false;
+    if (this->support_duplicate_) {
+        duplicate_ids = this->bottom_graph_->GetDuplicateIds(inner_id);
+        was_duplicate_member = this->bottom_graph_->GetGroupId(inner_id) != inner_id;
+    }
 
     if (not force_update) {
         std::shared_lock label_lock(this->label_lookup_mutex_);
@@ -642,6 +646,16 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) 
         }
         std::shared_lock rlock(this->global_mutex_);
         this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
+            Vector<int8_t> duplicate_data(data_size, allocator_);
+            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
+            if (not this->bottom_graph_->RemoveDuplicateId(duplicate_id)) {
+                return false;
+            }
+            std::shared_lock rlock(this->global_mutex_);
+            this->graph_add_one(duplicate_data.data(), -1, duplicate_id);
+        }
     }
     return update_status;
 }
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -585,6 +585,8 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) 
     void* new_base_vec = nullptr;
     uint64_t data_size = 0;
     get_vectors(data_type_, dim_, new_base, &new_base_vec, &data_size);
+    const bool was_duplicate_member =
+        this->support_duplicate_ && this->bottom_graph_->GetGroupId(inner_id) != inner_id;
 
     if (not force_update) {
         std::shared_lock label_lock(this->label_lookup_mutex_);
@@ -624,10 +626,20 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) 
 
     // note that only modify vector need to obtain unique lock
     // and the lock has been obtained inside datacell
-    auto codes = (has_precise_reorder()) ? high_precise_codes_ : basic_flatten_codes_;
     bool update_status = basic_flatten_codes_->UpdateVector(new_base_vec, inner_id);
     if (has_precise_reorder()) {
         update_status = update_status && high_precise_codes_->UpdateVector(new_base_vec, inner_id);
+    }
+    if (create_new_raw_vector_ && raw_vector_ != nullptr) {
+        update_status = update_status && raw_vector_->UpdateVector(new_base_vec, inner_id);
+    }
+    if (not update_status) {
+        return false;
+    }
+    if (was_duplicate_member) {
+        this->bottom_graph_->RemoveDuplicateId(inner_id);
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
     }
     return update_status;
 }
--- a/src/datacell/dense_duplicate_tracker.cpp
+++ b/src/datacell/dense_duplicate_tracker.cpp
@@ -55,8 +55,13 @@ DenseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
     }
 
     auto previous_id = duplicate_id;
+    uint64_t count = 0;
+    const auto max_count = duplicate_ids_.size();
     while (duplicate_ids_[previous_id] != duplicate_id) {
         previous_id = duplicate_ids_[previous_id];
+        if (++count > max_count) {
+            return false;
+        }
     }
 
     const auto next_id = duplicate_ids_[duplicate_id];
--- a/src/datacell/dense_duplicate_tracker.cpp
+++ b/src/datacell/dense_duplicate_tracker.cpp
@@ -46,6 +46,31 @@ DenseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplicat
     duplicate_ids_[group_id] = duplicate_id;
 }
 
+bool
+DenseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
+    std::scoped_lock lock(mutex_);
+
+    if (duplicate_id >= duplicate_ids_.size() || duplicate_ids_[duplicate_id] == duplicate_id) {
+        return false;
+    }
+
+    auto previous_id = duplicate_id;
+    while (duplicate_ids_[previous_id] != duplicate_id) {
+        previous_id = duplicate_ids_[previous_id];
+    }
+
+    const auto next_id = duplicate_ids_[duplicate_id];
+    duplicate_ids_[duplicate_id] = duplicate_id;
+    if (previous_id == next_id) {
+        duplicate_ids_[previous_id] = previous_id;
+        duplicate_count_--;
+        return true;
+    }
+
+    duplicate_ids_[previous_id] = next_id;
+    return true;
+}
+
 auto
 DenseDuplicateTracker::GetDuplicateIds(InnerIdType id) const -> std::vector<InnerIdType> {
     std::shared_lock lock(mutex_);
--- a/src/datacell/sparse_duplicate_tracker.cpp
+++ b/src/datacell/sparse_duplicate_tracker.cpp
@@ -130,8 +130,15 @@ SparseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
     auto previous_id = duplicate_id;
     uint64_t count = 0;
     const auto max_count = next_ids_.size();
-    while (next_ids_[previous_id] != duplicate_id) {
-        previous_id = next_ids_.at(previous_id);
+    while (true) {
+        const auto previous_iter = next_ids_.find(previous_id);
+        if (previous_iter == next_ids_.end()) {
+            return false;
+        }
+        if (previous_iter->second == duplicate_id) {
+            break;
+        }
+        previous_id = previous_iter->second;
         if (++count > max_count) {
             return false;
         }
--- a/src/datacell/sparse_duplicate_tracker.cpp
+++ b/src/datacell/sparse_duplicate_tracker.cpp
@@ -119,6 +119,31 @@ SparseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplica
     next_ids_[group_id] = duplicate_id;
 }
 
+bool
+SparseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
+    std::scoped_lock lock(mutex_);
+
+    if (next_ids_.count(duplicate_id) == 0) {
+        return false;
+    }
+
+    auto previous_id = duplicate_id;
+    while (next_ids_[previous_id] != duplicate_id) {
+        previous_id = next_ids_[previous_id];
+    }
+
+    const auto next_id = next_ids_[duplicate_id];
+    next_ids_.erase(duplicate_id);
+    if (previous_id == next_id) {
+        next_ids_.erase(previous_id);
+        duplicate_count_--;
+        return true;
+    }
+
+    next_ids_[previous_id] = next_id;
+    return true;
+}
+
 auto
 SparseDuplicateTracker::GetDuplicateIds(InnerIdType id) const -> std::vector<InnerIdType> {
     std::shared_lock lock(mutex_);
--- a/src/datacell/sparse_duplicate_tracker.cpp
+++ b/src/datacell/sparse_duplicate_tracker.cpp
@@ -128,8 +128,13 @@ SparseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
     }
 
     auto previous_id = duplicate_id;
+    uint64_t count = 0;
+    const auto max_count = next_ids_.size();
     while (next_ids_[previous_id] != duplicate_id) {
-        previous_id = next_ids_[previous_id];
+        previous_id = next_ids_.at(previous_id);
+        if (++count > max_count) {
+            return false;
+        }
     }
 
     const auto next_id = next_ids_[duplicate_id];
--- a/tests/test_hgraph.cpp
+++ b/tests/test_hgraph.cpp
@@ -386,6 +386,17 @@ RunWithGeneratedBlockSizeLimit(Fn&& fn) {
     vsag::Options::Instance().set_block_size_limit(origin_size);
 }
 
+bool
+FindDistanceById(const vsag::DatasetPtr& result, int64_t id, float& distance) {
+    for (int64_t i = 0; i < result->GetDim(); ++i) {
+        if (result->GetIds()[i] == id) {
+            distance = result->GetDistances()[i];
+            return true;
+        }
+    }
+    return false;
+}
+
 template <typename Cases, typename Fn>
 void
 ForEachHGraphCase(const fixtures::HGraphResourcePtr& resource, const Cases& test_cases, Fn&& fn) {
@@ -2031,6 +2042,93 @@ TestHGraphDuplicateBuild(const fixtures::HGraphTestIndexPtr& test_index,
 HGRAPH_PR_DAILY_CASE("HGraph Duplicate Build",
                      "[ft][build][duplicate][hgraph]",
                      TestHGraphDuplicateBuild)
+
+TEST_CASE("HGraph UpdateVector refreshes duplicate member state",
+          "[ft][hgraph][duplicate][update]") {
+    using namespace fixtures;
+
+    constexpr int64_t kEntryId = 10;
+    constexpr int64_t kRepresentativeId = 11;
+    constexpr int64_t kDuplicateId = 12;
+    constexpr int64_t kDim = 2;
+    constexpr float kNearRadius = 0.01F;
+    const std::string search_param = R"({"hgraph":{"ef_search":32}})";
+
+    auto make_updated_index = [&]() -> TestIndex::IndexPtr {
+        HGraphTestIndex::HGraphBuildParam build_param("l2", kDim, "fp32");
+        build_param.support_duplicate = true;
+        build_param.thread_count = 1;
+
+        auto param = HGraphTestIndex::GenerateHGraphBuildParametersString(build_param);
+        auto index = TestIndex::TestFactory(HGraphTestIndex::name, param, true);
+
+        float entry_vector[kDim] = {-1.0F, 0.0F};
+        int64_t entry_id[1] = {kEntryId};
+        auto entry = vsag::Dataset::Make();
+        entry->NumElements(1)->Dim(kDim)->Ids(entry_id)->Float32Vectors(entry_vector)->Owner(false);
+        REQUIRE(index->Build(entry).has_value());
+
+        float representative_vector[kDim] = {1.0F, 0.0F};
+        int64_t representative_id[1] = {kRepresentativeId};
+        auto representative = vsag::Dataset::Make();
+        representative->NumElements(1)
+            ->Dim(kDim)
+            ->Ids(representative_id)
+            ->Float32Vectors(representative_vector)
+            ->Owner(false);
+        REQUIRE(index->Add(representative).has_value());
+
+        float duplicate_vector[kDim] = {1.0F, 0.0F};
+        int64_t duplicate_id[1] = {kDuplicateId};
+        auto duplicate = vsag::Dataset::Make();
+        duplicate->NumElements(1)->Dim(kDim)->Ids(duplicate_id)->Float32Vectors(duplicate_vector);
+        duplicate->Owner(false);
+        REQUIRE(index->Add(duplicate).has_value());
+        REQUIRE(index->CheckIdExist(kDuplicateId));
+
+        float moved_vector[kDim] = {0.0F, 1.0F};
+        auto moved = vsag::Dataset::Make();
+        moved->NumElements(1)->Dim(kDim)->Float32Vectors(moved_vector);
+        moved->Owner(false);
+        auto update = index->UpdateVector(kDuplicateId, moved, true);
+        REQUIRE(update.has_value());
+        REQUIRE(update.value());
+
+        return index;
+    };
+
+    SECTION("search expansion uses the updated duplicate member distance") {
+        auto index = make_updated_index();
+        float representative_query_vector[kDim] = {1.0F, 0.0F};
+        auto query = vsag::Dataset::Make();
+        query->NumElements(1)->Dim(kDim)->Float32Vectors(representative_query_vector);
+        query->Owner(false);
+
+        auto result = index->KnnSearch(query, 3, search_param);
+        REQUIRE(result.has_value());
+
+        float returned_distance = 0.0F;
+        REQUIRE(FindDistanceById(result.value(), kDuplicateId, returned_distance));
+
+        auto exact_distance = index->CalcDistanceById(representative_query_vector, kDuplicateId);
+        REQUIRE(exact_distance.has_value());
+        REQUIRE(std::abs(returned_distance - exact_distance.value()) < 1e-5F);
+    }
+
+    SECTION("updated duplicate-only member is searchable at its new position") {
+        auto index = make_updated_index();
+        float moved_query_vector[kDim] = {0.0F, 1.0F};
+        auto query = vsag::Dataset::Make();
+        query->NumElements(1)->Dim(kDim)->Float32Vectors(moved_query_vector);
+        query->Owner(false);
+
+        auto result = index->RangeSearch(query, kNearRadius, search_param, 1);
+        REQUIRE(result.has_value());
+        REQUIRE(result.value()->GetDim() == 1);
+        REQUIRE(result.value()->GetIds()[0] == kDuplicateId);
+        REQUIRE(std::abs(result.value()->GetDistances()[0]) < 1e-5F);
+    }
+}
 
 static void
 TestHGraphEstimateMemoryAndGetMemoryUsage(const fixtures::HGraphTestIndexPtr& test_index,
--- a/tests/test_hgraph.cpp
+++ b/tests/test_hgraph.cpp
@@ -2054,7 +2054,7 @@ TEST_CASE("HGraph UpdateVector refreshes duplicate member state",
     constexpr float kNearRadius = 0.01F;
     const std::string search_param = R"({"hgraph":{"ef_search":32}})";
 
-    auto make_updated_index = [&]() -> TestIndex::IndexPtr {
+    auto make_duplicate_index = [&]() -> TestIndex::IndexPtr {
         HGraphTestIndex::HGraphBuildParam build_param("l2", kDim, "fp32");
         build_param.support_duplicate = true;
         build_param.thread_count = 1;
@@ -2086,6 +2086,11 @@ TEST_CASE("HGraph UpdateVector refreshes duplicate member state",
         REQUIRE(index->Add(duplicate).has_value());
         REQUIRE(index->CheckIdExist(kDuplicateId));
 
+        return index;
+    };
+
+    auto make_updated_index = [&]() -> TestIndex::IndexPtr {
+        auto index = make_duplicate_index();
         float moved_vector[kDim] = {0.0F, 1.0F};
         auto moved = vsag::Dataset::Make();
         moved->NumElements(1)->Dim(kDim)->Float32Vectors(moved_vector);
@@ -2127,6 +2132,28 @@ TEST_CASE("HGraph UpdateVector refreshes duplicate member state",
         REQUIRE(result.value()->GetDim() == 1);
         REQUIRE(result.value()->GetIds()[0] == kDuplicateId);
         REQUIRE(std::abs(result.value()->GetDistances()[0]) < 1e-5F);
+    }
+
+    SECTION("updated duplicate representative detaches remaining members") {
+        auto index = make_duplicate_index();
+        float moved_vector[kDim] = {0.0F, 1.0F};
+        auto moved = vsag::Dataset::Make();
+        moved->NumElements(1)->Dim(kDim)->Float32Vectors(moved_vector);
+        moved->Owner(false);
+        auto update = index->UpdateVector(kRepresentativeId, moved, true);
+        REQUIRE(update.has_value());
+        REQUIRE(update.value());
+
+        float duplicate_query_vector[kDim] = {1.0F, 0.0F};
+        auto duplicate_query = vsag::Dataset::Make();
+        duplicate_query->NumElements(1)->Dim(kDim)->Float32Vectors(duplicate_query_vector);
+        duplicate_query->Owner(false);
+
+        auto duplicate_result = index->RangeSearch(duplicate_query, kNearRadius, search_param, 1);
+        REQUIRE(duplicate_result.has_value());
+        REQUIRE(duplicate_result.value()->GetDim() == 1);
+        REQUIRE(duplicate_result.value()->GetIds()[0] == kDuplicateId);
+        REQUIRE(std::abs(duplicate_result.value()->GetDistances()[0]) < 1e-5F);
     }
 }
 
--- a/src/datacell/dense_duplicate_tracker.h
+++ b/src/datacell/dense_duplicate_tracker.h
@@ -29,6 +29,9 @@ public:
     void
     SetDuplicateId(InnerIdType group_id, InnerIdType duplicate_id) override;
 
+    bool
+    RemoveDuplicateId(InnerIdType duplicate_id) override;
+
     auto
     GetDuplicateIds(InnerIdType id) const -> std::vector<InnerIdType> override;
 
--- a/src/datacell/dense_duplicate_tracker_test.cpp
+++ b/src/datacell/dense_duplicate_tracker_test.cpp
@@ -67,6 +67,30 @@ TEST_CASE("DenseDuplicateTracker ignores duplicate reinsertion", "[ut][DenseDupl
     REQUIRE(tracker.GetDuplicateIds(0) == std::vector<InnerIdType>{1});
 }
 
+TEST_CASE("DenseDuplicateTracker removes duplicate ids", "[ut][DenseDuplicateTracker]") {
+    auto allocator = std::make_shared<DefaultAllocator>();
+    DenseDuplicateTracker tracker(allocator.get());
+    tracker.Resize(6);
+
+    tracker.SetDuplicateId(0, 1);
+    tracker.SetDuplicateId(0, 2);
+    tracker.SetDuplicateId(3, 4);
+
+    REQUIRE(tracker.RemoveDuplicateId(1));
+    REQUIRE(sorted_duplicates(tracker.GetDuplicateIds(0)) == std::vector<InnerIdType>{2});
+    REQUIRE(tracker.GetDuplicateIds(1).empty());
+    REQUIRE(tracker.GetGroupId(1) == 1);
+    REQUIRE(tracker.GetGroupId(2) == 0);
+
+    REQUIRE(tracker.RemoveDuplicateId(2));
+    REQUIRE(tracker.GetDuplicateIds(0).empty());
+    REQUIRE(tracker.GetGroupId(2) == 2);
+
+    REQUIRE(tracker.RemoveDuplicateId(4));
+    REQUIRE(tracker.GetDuplicateIds(3).empty());
+    REQUIRE_FALSE(tracker.RemoveDuplicateId(5));
+}
+
 TEST_CASE("DenseDuplicateTracker serialize and deserialize", "[ut][DenseDuplicateTracker]") {
     auto allocator = std::make_shared<DefaultAllocator>();
     DenseDuplicateTracker tracker(allocator.get());
--- a/src/datacell/duplicate_interface.h
+++ b/src/datacell/duplicate_interface.h
@@ -33,6 +33,9 @@ public:
     virtual void
     SetDuplicateId(InnerIdType group_id, InnerIdType duplicate_id) = 0;
 
+    virtual bool
+    RemoveDuplicateId(InnerIdType duplicate_id) = 0;
+
     [[nodiscard]] virtual auto
     GetDuplicateIds(InnerIdType id) const -> std::vector<InnerIdType> = 0;
 
--- a/src/datacell/graph_interface.h
+++ b/src/datacell/graph_interface.h
@@ -217,6 +217,14 @@ public:
         }
     }
 
+    bool
+    RemoveDuplicateId(InnerIdType duplicate_id) {
+        if (duplicate_tracker_) {
+            return duplicate_tracker_->RemoveDuplicateId(duplicate_id);
+        }
+        return false;
+    }
+
     std::vector<InnerIdType>
     GetDuplicateIds(InnerIdType id) const {
         if (duplicate_tracker_) {
--- a/src/datacell/sparse_duplicate_tracker.h
+++ b/src/datacell/sparse_duplicate_tracker.h
@@ -29,6 +29,9 @@ public:
     void
     SetDuplicateId(InnerIdType group_id, InnerIdType duplicate_id) override;
 
+    bool
+    RemoveDuplicateId(InnerIdType duplicate_id) override;
+
     auto
     GetDuplicateIds(InnerIdType id) const -> std::vector<InnerIdType> override;
 
--- a/src/datacell/sparse_duplicate_tracker_test.cpp
+++ b/src/datacell/sparse_duplicate_tracker_test.cpp
@@ -79,6 +79,29 @@ TEST_CASE("SparseDuplicateTracker ignores duplicate reinsertion", "[ut][SparseDu
     REQUIRE(tracker.GetDuplicateIds(0) == std::vector<InnerIdType>{1});
 }
 
+TEST_CASE("SparseDuplicateTracker removes duplicate ids", "[ut][SparseDuplicateTracker]") {
+    auto allocator = std::make_shared<DefaultAllocator>();
+    SparseDuplicateTracker tracker(allocator.get());
+
+    tracker.SetDuplicateId(0, 1);
+    tracker.SetDuplicateId(0, 2);
+    tracker.SetDuplicateId(3, 4);
+
+    REQUIRE(tracker.RemoveDuplicateId(1));
+    REQUIRE(sorted_duplicates(tracker.GetDuplicateIds(0)) == std::vector<InnerIdType>{2});
+    REQUIRE(tracker.GetDuplicateIds(1).empty());
+    REQUIRE(tracker.GetGroupId(1) == 1);
+    REQUIRE(tracker.GetGroupId(2) == 0);
+
+    REQUIRE(tracker.RemoveDuplicateId(2));
+    REQUIRE(tracker.GetDuplicateIds(0).empty());
+    REQUIRE(tracker.GetGroupId(2) == 2);
+
+    REQUIRE(tracker.RemoveDuplicateId(4));
+    REQUIRE(tracker.GetDuplicateIds(3).empty());
+    REQUIRE_FALSE(tracker.RemoveDuplicateId(5));
+}
+
 TEST_CASE("SparseDuplicateTracker serialize and deserialize", "[ut][SparseDuplicateTracker]") {
     auto allocator = std::make_shared<DefaultAllocator>();
     SparseDuplicateTracker tracker(allocator.get());


src/algorithm/hgraph/hgraph.cpp
```diff
@@ -585,6 +585,8 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
     void* new_base_vec = nullptr;
     uint64_t data_size = 0;
     get_vectors(data_type_, dim_, new_base, &new_base_vec, &data_size);
+    const bool was_duplicate_member =
+        this->support_duplicate_ && this->bottom_graph_->GetGroupId(inner_id) != inner_id;
```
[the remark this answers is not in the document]
jac0626: Fixed in 8d62d80f. Representative updates now snapshot the old duplicate group, dissolve remaining duplicate members, and reconnect each member with its stored vector so they are no longer expanded with the updated representative distance.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +626,21 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+        update_status = update_status && raw_vector_->UpdateVector(new_base_vec, inner_id);
+    }
+    if (not update_status) {
+        return false;
+    }
+    if (was_duplicate_member) {
+        this->bottom_graph_->RemoveDuplicateId(inner_id);
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    }
```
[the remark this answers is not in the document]
jac0626: Fixed in 8d62d80f. `UpdateVector` now handles the representative case via `GetDuplicateIds(inner_id)`: remaining duplicate-only IDs are removed from the tracker and reinserted into the bottom graph with their own vectors.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -585,6 +585,8 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
     void* new_base_vec = nullptr;
     uint64_t data_size = 0;
     get_vectors(data_type_, dim_, new_base, &new_base_vec, &data_size);
+    const bool was_duplicate_member =
+        this->support_duplicate_ && this->bottom_graph_->GetGroupId(inner_id) != inner_id;
```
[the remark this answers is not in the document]
jac0626: Fixed in 8d62d80f. The cleanup path is no longer limited to `GetGroupId(inner_id) != inner_id`; canonical IDs with duplicate members now detach and reconnect those members.

src/algorithm/hgraph/hgraph.cpp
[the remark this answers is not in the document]
jac0626: Already covered in the current head. `UpdateVector` checks `RemoveDuplicateId` before every `graph_add_one` path and returns false instead of inserting a duplicate node if removal fails.

src/datacell/dense_duplicate_tracker.cpp
```diff
@@ -46,6 +46,31 @@ DenseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplicat
...
+    std::scoped_lock lock(mutex_);
+
+    if (duplicate_id >= duplicate_ids_.size() || duplicate_ids_[duplicate_id] == duplicate_id) {
+        return false;
+    }
+
+    auto previous_id = duplicate_id;
+    while (duplicate_ids_[previous_id] != duplicate_id) {
+        previous_id = duplicate_ids_[previous_id];
+    }
```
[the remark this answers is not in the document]
jac0626: Already covered in the current head. Dense `RemoveDuplicateId` bounds traversal with `max_count = duplicate_ids_.size()` and returns false on a corrupted cycle instead of looping indefinitely.

src/datacell/sparse_duplicate_tracker.cpp
```diff
@@ -119,6 +119,31 @@ SparseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplica
...
+    std::scoped_lock lock(mutex_);
+
+    if (next_ids_.count(duplicate_id) == 0) {
+        return false;
+    }
+
+    auto previous_id = duplicate_id;
+    while (next_ids_[previous_id] != duplicate_id) {
+        previous_id = next_ids_[previous_id];
+    }
```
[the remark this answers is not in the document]
jac0626: Fixed in 8d62d80f. Sparse `RemoveDuplicateId` now uses `find()` during traversal, keeps the loop bound, and returns false on missing links without inserting map entries.

src/datacell/sparse_duplicate_tracker.cpp
```diff
@@ -119,6 +119,36 @@ SparseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplica
...
+    uint64_t count = 0;
+    const auto max_count = next_ids_.size();
+    while (next_ids_[previous_id] != duplicate_id) {
+        previous_id = next_ids_.at(previous_id);
+        if (++count > max_count) {
+            return false;
+        }
+    }
+
+    const auto next_id = next_ids_[duplicate_id];
```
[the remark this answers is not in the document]
jac0626: Fixed in 8d62d80f. The sparse tracker no longer uses `next_ids_[previous_id]` while traversing; read-only lookup uses `find()` so failed removal does not mutate `next_ids_`.

tests/test_hgraph.cpp
```diff
@@ -386,6 +386,17 @@ RunWithGeneratedBlockSizeLimit(Fn&& fn) {
     vsag::Options::Instance().set_block_size_limit(origin_size);
 }
 
+bool
+FindDistanceById(const vsag::DatasetPtr& result, int64_t id, float& distance) {
+    for (int64_t i = 0; i < result->GetDim(); ++i) {
```
LHT129, line 391: [critical] `FindDistanceById` iterates over `result->GetDim()` which returns the per-vector dimensionality, not the number of results. When `topk != dim`, this will either miss matching IDs or read out of bounds on `GetIds()` / `GetDistances()`.

The loop should use `result->GetNumElements()` instead:

```cpp
for (int64_t i = 0; i < result->GetNumElements(); ++i) {
```

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+    if (was_duplicate_member) {
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
+            Vector<int8_t> duplicate_data(data_size, allocator_);
+            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
```
LHT129, line 652: [suggestion] The `duplicate_data` buffer is declared as `Vector<int8_t>` and sized by `data_size`, but then cast to `float*` for `GetVectorByInnerId`. This has two issues:

1. **Buffer size**: `data_size` is the byte size of the input dataset, but `GetVectorByInnerId` writes `dim_` floats. The correct buffer size should be `dim_ * sizeof(float)`.
2. **Strict aliasing**: Casting `int8_t*` to `float*` is technically undefined behavior in C++. While `Vector` likely provides sufficient alignment, using a `Vector<float>` would be cleaner and avoid the cast entirely.

Suggested fix:
```cpp
Vector<float> duplicate_data(dim_, allocator_);
GetVectorByInnerId(duplicate_id, duplicate_data.data());
// ...
this->graph_add_one(duplicate_data.data(), -1, duplicate_id);
```

LHT129 (commented):
Thanks for this fix. The core logic for `RemoveDuplicateId` in both trackers and the `UpdateVector` integration looks solid — the linked-list manipulation is correct and the loop guards prevent infinite loops.

Two issues found:

1. **[critical] `FindDistanceById` in `tests/test_hgraph.cpp:391`**: Uses `GetDim()` instead of `GetNumElements()` for the loop bound, which will miss results or read out of bounds when `topk != dim`.

2. **[suggestion] `Vector<int8_t>` buffer in `hgraph.cpp:652`**: The duplicate reinsertion path uses an `int8_t` vector cast to `float*` for `GetVectorByInnerId`. The buffer should be `Vector<float>` sized by `dim_` to avoid strict aliasing concerns and ensure correct buffer sizing.

Otherwise the PR looks good — the `RemoveDuplicateId` implementations are correct, the representative-update case is properly handled via the `else if` branch, and the test coverage is thorough.

LHT129 (commented):
I have completed a thorough review of this PR. The overall approach of adding `RemoveDuplicateId` support to both duplicate trackers and refreshing duplicate state during `UpdateVector()` is sound.

**Issues already identified by @LHT129 that I confirm:**

1. **[critical] `FindDistanceById` in `tests/test_hgraph.cpp:391`**: The loop iterates over `result->GetDim()` (vector dimensionality) instead of `result->GetNumElements()` (number of results). When `topk != dim`, this will either miss matching IDs or read out of bounds. Should use `GetNumElements()`.

2. **[suggestion] `Vector<int8_t> duplicate_data` in `src/algorithm/hgraph/hgraph.cpp:652`**: The buffer is sized by `data_size` (byte size of input dataset) but used to store `dim_` floats via `GetVectorByInnerId`. Using `Vector<float> duplicate_data(dim_, allocator_)` would be correct and avoid the `int8_t*` to `float*` cast.

**Additional observations:**

3. The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` are correct. The circular linked list unlinking logic properly handles all cases (middle element removal, last element removal, single-duplicate group dissolution). The loop counter guards against infinite loops on corrupted data.

4. The `UpdateVector` logic correctly handles both cases: when the updated ID is a duplicate member (removes it from the group and reconnects to graph) and when it is a representative with duplicates (removes all duplicates from the group and reconnects them individually).

5. The `raw_vector_->UpdateVector()` addition is a good catch — without it, the raw vector storage would be stale after an update, causing incorrect distance calculations.

No other blocking issues found. The two items above should be addressed before merging.

tests/test_hgraph.cpp
```diff
@@ -386,6 +386,17 @@ RunWithGeneratedBlockSizeLimit(Fn&& fn) {
...
+        if (result->GetIds()[i] == id) {
```
LHT129, line 392: **issue:** `result->GetDim()` returns the dimension count, not the number of result entries. For search results, the number of entries is given by `result->GetNumElements()`. Using `GetDim()` here will iterate only `dim` times instead of over all result entries, potentially missing the target ID if there are more results than dimensions.

```suggestion
    for (int64_t i = 0; i < result->GetNumElements(); ++i) {
```

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
 
     // note that only modify vector need to obtain unique lock
     // and the lock has been obtained inside datacell
-    auto codes = (has_precise_reorder()) ? high_precise_codes_ : basic_flatten_codes_;
     bool update_status = basic_flatten_codes_->UpdateVector(new_base_vec, inner_id);
     if (has_precise_reorder()) {
         update_status = update_status && high_precise_codes_->UpdateVector(new_base_vec, inner_id);
     }
+    if (create_new_raw_vector_ && raw_vector_ != nullptr) {
+        update_status = update_status && raw_vector_->UpdateVector(new_base_vec, inner_id);
```
LHT129, line 638: **suggestion:** `Vector<int8_t>` allocates `data_size` bytes, but `GetVectorByInnerId` writes `float` values (4 bytes each). For a vector of dimension `dim_`, the required size is `dim_ * sizeof(float)`, not `data_size` (which equals `dim_ * sizeof(float)` only for `fp32`). Consider using `Vector<float>` directly or sizing with `dim_ * sizeof(float)` to make the intent explicit and avoid subtle bugs if the data type ever changes.

```suggestion
            Vector<float> duplicate_data(dim_, allocator_);
            GetVectorByInnerId(duplicate_id, duplicate_data.data());
```

LHT129 (commented):
Thanks for this fix. The approach of removing stale duplicate membership and reconnecting updated members to the bottom graph is sound.

I have reviewed the full diff and the implementations of `RemoveDuplicateId` in both `DenseDuplicateTracker` and `SparseDuplicateTracker` — the linked-list traversal with cycle detection guards looks correct, and the `UpdateVector` logic properly handles both the "was duplicate member" and "representative with duplicates" cases.

There are two outstanding issues already flagged by @LHT129 that should be addressed before merging:

1. **`Vector<int8_t>` buffer in the representative-update path** (`src/algorithm/hgraph/hgraph.cpp:652`): The buffer is sized by `data_size` (byte size) and cast from `int8_t*` to `float*`. Using `Vector<float>` sized by `dim_` would be cleaner and avoid strict aliasing concerns.

2. **`FindDistanceById` uses `GetDim()` instead of `GetNumElements()`** (`tests/test_hgraph.cpp:392`): `GetDim()` returns the vector dimension, not the number of result entries. This should be `GetNumElements()` to correctly iterate over all search results.

Otherwise the change looks correct and the test coverage (unit tests for both trackers + integration test for UpdateVector) is thorough.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+    }
+    if (was_duplicate_member) {
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
+            Vector<int8_t> duplicate_data(data_size, allocator_);
```
LHT129, line 651: [suggestion] If `RemoveDuplicateId` fails for a duplicate in the middle of this loop, earlier duplicates have already been removed from the tracker and re-added to the graph, leaving the system in a partially-updated state. The function returns `false` but the side effects from earlier iterations are not rolled back.

Consider collecting the duplicate IDs first, validating that all can be removed, and only then applying the mutations. Alternatively, document that callers should treat a `false` return as a fatal error requiring index rebuild.

LHT129 (commented):
[critical] When the updated ID is a duplicate-group representative (i.e., GetGroupId(inner_id) == inner_id but duplicate_ids is non-empty), the else-if branch at hgraph.cpp:649 removes and re-adds only the duplicate members to the graph. The representative itself stays in the graph with its old topology connections, but its vector has already been updated in the flatten codes storage. This means graph traversal will encounter the representative at its old graph position with the new vector, leading to incorrect search results for queries near the representative new position.

The representative should also be re-added to the graph after its vector is updated, similar to how the was_duplicate_member branch handles inner_id.

The existing test "updated duplicate representative detaches remaining members" only verifies that the remaining duplicate member is searchable at its old position, but does not verify that the representative itself is searchable at its new position.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+    if (not update_status) {
+        return false;
+    }
+    if (was_duplicate_member) {
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
```
LHT129, line 649: [suggestion] The `else if (not duplicate_ids.empty())` branch handles the case where the updated ID is a duplicate-group representative. However, the representative itself is not re-added to the graph after its vector is updated — only the duplicate members are. The representative stays in the graph with stale topology connections while its vector has changed in the flatten codes storage.

Consider also calling `graph_add_one` for the representative after processing all duplicates, so that its graph connections reflect the new vector position.

LHT129 (commented):
I have reviewed this PR and confirm the overall approach is sound. The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` are correct — the circular linked-list unlinking logic properly handles all cases (middle element removal, last element removal, single-duplicate group dissolution) with appropriate loop guards against infinite loops.

The `UpdateVector` logic correctly handles both cases: when the updated ID is a duplicate member (removes it from the group and reconnects to graph) and when it is a representative with duplicates (removes all duplicates from the group and reconnects them individually).

The `raw_vector_->UpdateVector()` addition is a good catch — without it, the raw vector storage would be stale after an update.

I concur with the issues already flagged by @LHT129:

1. **[critical] `FindDistanceById` in `tests/test_hgraph.cpp:392`**: Uses `GetDim()` instead of `GetNumElements()` for the loop bound.
2. **[suggestion] `Vector<int8_t>` buffer in `src/algorithm/hgraph/hgraph.cpp:652`**: Should be `Vector<float>` sized by `dim_`.

No other blocking issues found. The test coverage (unit tests for both trackers + integration test for UpdateVector) is thorough.

src/datacell/dense_duplicate_tracker.cpp
```diff
@@ -46,6 +46,36 @@ DenseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplicat
     duplicate_ids_[group_id] = duplicate_id;
 }
 
+bool
+DenseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
+    std::scoped_lock lock(mutex_);
+
+    if (duplicate_id >= duplicate_ids_.size() || duplicate_ids_[duplicate_id] == duplicate_id) {
```
LHT129, line 53: [suggestion] `RemoveDuplicateId` does not guard against being called on a duplicate-group representative. When called on a representative of a group with 2+ duplicate members, the function creates a cycle of duplicate members with no self-pointing node (e.g., `dup1->dup2->...->dupN->dup1`). This would cause `GetGroupId` (which has no loop counter limit in the dense tracker) to infinite-loop on those IDs.

The early-return check only rejects standalone IDs (`duplicate_ids_[id] == id`), but a representative has `duplicate_ids_[rep] != rep` (it points to the first duplicate), so it passes through.

While current callers in `hgraph.cpp` only invoke this on confirmed duplicate members (not representatives), the function should either:
1. Detect and reject representative IDs explicitly, or
2. Handle the representative case by designating a new representative from the remaining duplicates.

The same latent issue applies to `SparseDuplicateTracker::RemoveDuplicateId`.

LHT129 (commented):
[note] This PR adds valuable functionality: `RemoveDuplicateId` for both dense and sparse duplicate trackers, raw vector storage update during `UpdateVector`, and comprehensive test coverage. The core logic in `RemoveDuplicateId` implementations is correct.

Several substantive issues have already been flagged by other reviewers and remain to be addressed:

1. **Representative not re-added to graph** (LHT129, `src/algorithm/hgraph/hgraph.cpp:649`): When the updated ID is a duplicate-group representative, the `else if` branch detaches and re-adds duplicate members but does not refresh the representative's own graph connections. The representative stays with stale topology while its vector has changed.

2. **Partial update in else-if loop** (LHT129, `src/algorithm/hgraph/hgraph.cpp:651`): If `RemoveDuplicateId` fails for a duplicate mid-loop, earlier duplicates have already been mutated. The function returns `false` but side effects from earlier iterations are not rolled back.

3. **`RemoveDuplicateId` on representative IDs** (LHT129, `src/datacell/dense_duplicate_tracker.cpp:53`): The function does not guard against being called on a duplicate-group representative, which would corrupt the circular list structure. Current callers are safe, but the API should either reject this case explicitly or handle it correctly.

The test coverage is thorough and the `raw_vector_` update fix is a good catch.

LHT129 (commented):
[critical] `graph_add_one` is called under `std::shared_lock` in `UpdateVector()` (hgraph.cpp lines 647, 656), but `graph_add_one` modifies graph topology (connecting new nodes, updating neighbor lists). A `shared_lock` allows concurrent readers (e.g., search operations) to access the graph simultaneously, creating a data race.

The original `UpdateVector` code did not call `graph_add_one` at all — it only updated the flatten codes storage. Now that graph insertion is added here, the lock must be upgraded to `std::unique_lock` to ensure exclusive access during the mutation.

Suggested fix: use `std::unique_lock` instead of `std::shared_lock` for the scope of `graph_add_one` calls.

LHT129 (commented):
[suggestion] In the else-if branch for duplicate_ids (hgraph.cpp line 649), the representative (inner_id) is not re-added to the graph via graph_add_one after its vector is updated. Only the duplicate members are removed from the tracker and re-inserted. The representative stays with old topology connections while its vector has changed, potentially degrading search recall. Consider also calling graph_add_one for inner_id after processing all duplicates.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 649: [suggestion] The return value of `GetVectorByInnerId` is not checked. If the call fails (e.g., the duplicate_id is invalid or the vector storage is corrupted), `duplicate_data` will contain uninitialized data that gets passed to `graph_add_one`, potentially inserting garbage into the graph topology.

Consider checking the return value and skipping the iteration on failure, or returning `false` to signal the caller.

Also, `Vector<int8_t>` is used as the buffer type while the data is cast to `float*`. Consider using `Vector<float>` directly to make the intent clearer and avoid the reinterpret cast.

LHT129 (commented):
Code review for PR #2354 (commit 8d62d80f).

src/algorithm/hgraph/hgraph.cpp
LHT129, line 649: [suggestion] `GetVectorByInnerId` is called outside the `global_mutex_` shared lock in the `else if (not duplicate_ids.empty())` branch. `GetVectorByInnerId` reads `high_precise_codes_`, `basic_flatten_codes_`, `raw_vector_`, and `create_new_raw_vector_` — all of which can be modified by `Tune()` under an exclusive `global_mutex_` lock. This is a potential data race.

Consider acquiring the shared lock before the loop so that `GetVectorByInnerId` is protected. Note: `RemoveDuplicateId` has its own internal mutex so it is safe to call under the shared lock.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+    }
+    if (not update_status) {
+        return false;
+    }
+    if (was_duplicate_member) {
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
```
LHT129, line 648: [suggestion] The return value of `graph_add_one` is not checked in the `was_duplicate_member` branch. `graph_add_one` returns `false` when a duplicate match is found (i.e., the updated vector still matches an existing duplicate group). In that case, the member has already been removed from the duplicate tracker by `RemoveDuplicateId` above, but was not re-added to the graph — leaving the member in an inconsistent state where it is neither in the graph topology nor in a duplicate group.

Consider checking the return value and handling the failure case, e.g. by re-inserting the member back into the duplicate tracker or returning an error.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
+            Vector<int8_t> duplicate_data(data_size, allocator_);
+            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
+            if (not this->bottom_graph_->RemoveDuplicateId(duplicate_id)) {
+                return false;
+            }
+            std::shared_lock rlock(this->global_mutex_);
+            this->graph_add_one(duplicate_data.data(), -1, duplicate_id);
```
LHT129, line 657: [suggestion] The return value of `graph_add_one` is not checked in the `else if` branch either. Same concern as above: if `graph_add_one` returns `false` (duplicate match found), the duplicate member has already been removed from the tracker but was not re-added to the graph. The loop also continues to process remaining duplicates rather than aborting, which could compound the inconsistency.

Consider checking the return value and returning `false` on failure.

LHT129 (commented):
Thanks for this fix. The approach of removing stale duplicate membership and reconnecting updated members to the bottom graph is sound, and the test coverage is thorough. I left a few inline comments:

1. **Data race on `GetVectorByInnerId`** in the `else if` branch — called outside `global_mutex_` while `Tune()` can modify the underlying pointers under exclusive lock.

2. **`graph_add_one` return value not checked** in both branches — if it returns `false` (duplicate match), the member has already been removed from the tracker but not re-added to the graph, leaving an inconsistent state.

These are suggestions, not blockers. The core fix logic is correct.

LHT129 (commented):
Thank you for this fix. The changes look well-structured and address the root cause described in #2322.

The PR has evolved well through review iterations — the loop-count protection in both `DenseDuplicateTracker::RemoveDuplicateId` and `SparseDuplicateTracker::RemoveDuplicateId` is a good defensive addition, and the `else if` branch correctly handles the representative-update case that was flagged in earlier reviews.

The remaining concerns (data race on `GetVectorByInnerId` outside `global_mutex_`, and unchecked return values from `graph_add_one`) are already captured in the inline comments from @LHT129. Once those are addressed, this should be ready to merge.

The test coverage is thorough — unit tests for both tracker implementations plus integration tests covering the three key scenarios (duplicate member distance, searchability at new position, and representative detaching remaining members).

LHT129 (commented):
Thank you for this fix. The changes look well-structured and address the root cause described in #2322.

The PR has evolved well through review iterations — the loop-count protection in both DenseDuplicateTracker::RemoveDuplicateId and SparseDuplicateTracker::RemoveDuplicateId is a good defensive addition, and the else-if branch correctly handles the representative-update case that was flagged in earlier reviews.

The remaining concerns (data race on GetVectorByInnerId outside global_mutex_, and unchecked return values from graph_add_one) are already captured in the inline comments from @LHT129. Once those are addressed, this should be ready to merge.

The test coverage is thorough — unit tests for both tracker implementations plus integration tests covering the three key scenarios (duplicate member distance, searchability at new position, and representative detaching remaining members).

LHT129 (commented):
Thank you for this fix. The changes look well-structured and address the root cause described in #2322.

The PR has evolved well through review iterations — the loop-count protection in both DenseDuplicateTracker::RemoveDuplicateId and SparseDuplicateTracker::RemoveDuplicateId is a good defensive addition, and the else-if branch correctly handles the representative-update case that was flagged in earlier reviews.

The remaining concerns (data race on GetVectorByInnerId outside global_mutex_, and unchecked return values from graph_add_one) are already captured in the inline comments from @LHT129. Once those are addressed, this should be ready to merge.

The test coverage is thorough — unit tests for both tracker implementations plus integration tests covering the three key scenarios (duplicate member distance, searchability at new position, and representative detaching remaining members).

LHT129 (commented):
Thank you for this PR. The fix addresses the core issue described in #2322 by properly refreshing duplicate member state during `UpdateVector()`.

I have reviewed the changes and the following points have already been raised by other reviewers (LHT129) and remain relevant:

1. **Data race in `GetVectorByInnerId`**: In the `else if (not duplicate_ids.empty())` branch, `GetVectorByInnerId` is called outside the `global_mutex_` shared lock. This reads `high_precise_codes_`, `basic_flatten_codes_`, `raw_vector_`, and `create_new_raw_vector_` — all of which can be modified by `Tune()` under an exclusive lock. Consider acquiring the shared lock before the loop.

2. **Unchecked `graph_add_one` return value (both branches)**: `graph_add_one` returns `false` when a duplicate match is found. If this happens after `RemoveDuplicateId` has already succeeded, the member ends up in an inconsistent state — removed from the duplicate tracker but not re-added to the graph topology. Consider checking the return value and handling the failure.

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` look correct, with proper cycle detection guards. The test coverage is good and covers the key scenarios (duplicate member update, representative update, search correctness).

No additional blocking issues from my side beyond what has already been noted.

LHT129 (commented):
I have completed a thorough review of this PR (commit 8d62d80f). The overall approach of adding `RemoveDuplicateId` support to both duplicate trackers and refreshing duplicate state during `UpdateVector()` is sound.

**Summary of findings:**

The core logic in `DenseDuplicateTracker::RemoveDuplicateId` and `SparseDuplicateTracker::RemoveDuplicateId` is correct — the circular linked-list manipulation properly handles removal from groups of any size, and the loop guards prevent infinite traversal on corrupted state. The test coverage for both trackers and the HGraph integration test are comprehensive.

**Key issues already identified by other reviewers that still need attention:**

1. `[critical]` **`graph_add_one` called under `shared_lock`** (hgraph.cpp:647, 656): `graph_add_one` modifies graph topology but is invoked under `std::shared_lock` on `global_mutex_`, which allows concurrent readers. This needs a unique/exclusive lock.

2. `[critical]` **`FindDistanceById` uses `GetDim()` instead of result count** (test_hgraph.cpp:391): `GetDim()` returns per-vector dimensionality (2), not the number of search results. This will cause the test to miss results beyond index 1, potentially masking real bugs.

3. `[suggestion]` **Representative not re-added to graph after vector update** (hgraph.cpp:649): When the updated ID is a duplicate-group representative, only the duplicate members are re-added via `graph_add_one`. The representative itself retains stale graph connectivity after its vector changes.

4. `[suggestion]` **No rollback on partial `RemoveDuplicateId` failure** (hgraph.cpp:651): If `RemoveDuplicateId` fails for a duplicate mid-loop, earlier duplicates have already been removed from their groups and re-added to the graph, leaving the index in an inconsistent state.

5. `[suggestion]` **`Vector<int8_t>` buffer cast to `float*`** (hgraph.cpp:652): The buffer is sized by `data_size` (bytes) but `GetVectorByInnerId` writes `float` values, which are 4 bytes each. The buffer should be sized as `data_size / sizeof(float)` or use `Vector<float>`.

The duplicate tracker implementations and the integration test design are well done. Once the above issues (particularly the locking concern) are addressed, this will be a solid fix for #2322.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
+            Vector<int8_t> duplicate_data(data_size, allocator_);
+            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
+            if (not this->bottom_graph_->RemoveDuplicateId(duplicate_id)) {
```
LHT129, line 653: [suggestion] If `RemoveDuplicateId` fails for a `duplicate_id` mid-loop (returning `false`), earlier iterations have already removed their members from the duplicate tracker and re-added them to the graph via `graph_add_one`. This leaves the index in a partially-updated state — some former duplicate members are now standalone graph nodes while others remain in the old duplicate group.

Consider either: (a) collecting all duplicate data and performing removals in a first pass, then doing graph insertion in a second pass so that a failure in one step does not leave partial state; or (b) documenting that `RemoveDuplicateId` is infallible for valid inputs and the early-return is purely a defense against internal corruption.

LHT129 (commented):
Left one additional review comment regarding partial-state handling in the duplicate member refresh loop.

LHT129 (commented):
Thanks for this fix. The overall approach — adding `RemoveDuplicateId` to both dense and sparse duplicate trackers, and refreshing HGraph duplicate-member state during `UpdateVector()` — is sound and well-tested.

I have reviewed all 10 changed files and the core logic is correct. The duplicate tracker removal correctly handles the circular linked-list unlinking for groups of size 2 and >2, and the cycle guard prevents infinite loops on corrupted data.

Most substantive issues have already been raised by other reviewers (LHT129), including:

1. **`graph_add_one` return value not checked** (both branches at `hgraph.cpp:648` and `hgraph.cpp:657`) — if `graph_add_one` returns `false` (duplicate match found), the member has already been removed from the tracker but was not re-added to the graph, leaving an inconsistent state.

2. **`GetVectorByInnerId` called outside `global_mutex_` shared lock** (`hgraph.cpp:652`) — potential data race with `Tune()` which modifies the underlying codec pointers under an exclusive lock.

3. **Representative not re-added to graph** in the `else if` branch — when the updated ID is a duplicate-group representative, only the duplicate members are detached and re-inserted; the representative itself retains stale graph topology connections.

4. **Partial failure in the `else if` loop** (`hgraph.cpp:653`) — if `RemoveDuplicateId` fails mid-loop, earlier iterations have already modified the index, leaving it in a partially-updated state.

I would recommend addressing these before merging. The tests are comprehensive and cover the key scenarios well.

LHT129 (commented):
## Review Summary

This PR fixes #2322 by refreshing duplicate state during `UpdateVector()`. The changes are well-structured across three layers: the `DuplicateInterface` abstraction, the dense/sparse tracker implementations, and the HGraph integration.

### What was reviewed

- **`hgraph.cpp`**: The `UpdateVector()` changes correctly handle both cases — when the updated ID is a duplicate member itself (`was_duplicate_member`) and when it is a representative with other duplicate members (`else if` branch). The raw vector storage update was also added, which was previously missing.
- **`DenseDuplicateTracker::RemoveDuplicateId()`**: Correctly unlinks a node from the circular linked list with proper cycle detection and loop-count protection.
- **`SparseDuplicateTracker::RemoveDuplicateId()`**: Correctly unlinks a node from the unordered_map-based circular structure with `find()`-based traversal (avoiding accidental insertions from `operator[]`) and loop-count protection.
- **`graph_interface.h`**: The `RemoveDuplicateId()` forwarding method follows the existing pattern of `SetDuplicateId()` and `GetDuplicateIds()`.
- **Tests**: Unit tests for both trackers cover removal of middle, last, and non-existent IDs. The integration test in `test_hgraph.cpp` covers three scenarios: duplicate member distance correctness after update, searchability at new position, and representative update detaching remaining members.

### Existing review comments

Several reviewers have already raised important points that are still open:

1. The return value of `graph_add_one` is not checked in either the `was_duplicate_member` or `else if` branches. If `graph_add_one` returns `false` (e.g., the updated vector still matches an existing duplicate group), the member has already been removed from the duplicate tracker but was not re-added to the graph, leaving an inconsistent state.

2. In the `else if` loop, if `RemoveDuplicateId` fails mid-loop, earlier iterations have already modified state, leaving a partially-updated index.

These are worth addressing before merge. Overall the fix is correct in its core logic and the test coverage is thorough.

LHT129 (commented):
This PR has already received extensive review coverage (33+ inline comments) from prior review passes. The core logic of `RemoveDuplicateId` in both Dense and Sparse trackers is correct, with proper cycle detection, loop bounds, and mutex protection. The `UpdateVector` changes correctly handle both the duplicate-member and representative cases.

Key remaining items already flagged by prior reviews that still apply to the current diff:
- `FindDistanceById` in `tests/test_hgraph.cpp:391` uses `GetDim()` instead of `GetNumElements()` to iterate search results
- `Vector<int8_t>` buffer cast to `float*` in `hgraph.cpp:652` — consider using `Vector<float>` directly
- `GetVectorByInnerId` is called outside `global_mutex_` shared lock in the `else if` branch
- Return values of `GetVectorByInnerId` and `graph_add_one` are not checked

No new issues identified beyond those already raised.

LHT129 (commented):
Thanks for this fix. The PR is well-structured with good test coverage covering the key scenarios (duplicate member update, duplicate-only searchability, and representative update detaching members).

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` look correct, with proper cycle detection guards and correct linked-list removal logic.

The main remaining concern (already raised by @LHT129) is that `graph_add_one` return values are not checked in either the `was_duplicate_member` branch or the `else if` loop. If `graph_add_one` returns `false` (duplicate match found), the member has already been removed from the duplicate tracker but was not re-added to the graph, leaving it in an inconsistent state. In the `else if` loop specifically, the loop continues processing remaining duplicates even after a failure, compounding the inconsistency.

Consider checking `graph_add_one` return values and returning `false` on failure to prevent silent corruption.

LHT129 (commented):
Thanks for this fix. The PR is well-structured with good test coverage covering the key scenarios (duplicate member update, duplicate-only searchability, and representative update detaching members).

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` look correct, with proper cycle detection guards and correct linked-list removal logic.

The main remaining concern (already raised by @LHT129) is that `graph_add_one` return values are not checked in either the `was_duplicate_member` branch or the `else if` loop. If `graph_add_one` returns `false` (duplicate match found), the member has already been removed from the duplicate tracker but was not re-added to the graph, leaving it in an inconsistent state. In the `else if` loop specifically, the loop continues processing remaining duplicates even after a failure, compounding the inconsistency.

Consider checking `graph_add_one` return values and returning `false` on failure to prevent silent corruption.

LHT129 (commented):
test review from LHT129

LHT129 (commented):
Thanks for this fix. The PR is well-structured with good test coverage covering the key scenarios (duplicate member update, duplicate-only searchability, and representative update detaching members).

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` look correct, with proper cycle detection guards and correct linked-list removal logic.

The main remaining concern (already raised by @LHT129) is that `graph_add_one` return values are not checked in either the `was_duplicate_member` branch or the `else if` loop. If `graph_add_one` returns `false` (duplicate match found), the member has already been removed from the duplicate tracker but was not re-added to the graph, leaving it in an inconsistent state. In the `else if` loop specifically, the loop continues processing remaining duplicates even after a failure, compounding the inconsistency.

Consider checking `graph_add_one` return values and returning `false` on failure to prevent silent corruption.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 657: [suggestion] The return value of `graph_add_one` is not checked here. If `graph_add_one` returns `false` (meaning the duplicate member's vector matches an existing duplicate group and was re-grouped rather than added as a standalone node), the function still returns `true` at line 660, misleading the caller into thinking the update succeeded fully.

This is the same concern @LHT129 raised for the `was_duplicate_member` branch above. Consider:

```cpp
if (not this->graph_add_one(duplicate_data.data(), -1, duplicate_id)) {
    return false;
}
```

src/algorithm/hgraph/hgraph.cpp
LHT129, line 648: [suggestion] The return value of `graph_add_one` is not checked in either the `was_duplicate_member` branch or the `else if` branch. `graph_add_one` returns `false` when a duplicate match is found (i.e., the updated vector still matches an existing duplicate group). In that case, the member has already been removed from the duplicate tracker by `RemoveDuplicateId` above, but was not re-added to the graph — leaving the member in an inconsistent state.

Consider checking the return value and returning `false` on failure.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 657: [suggestion] Same concern as above: the return value of `graph_add_one` is not checked in this `else if` branch. If `graph_add_one` returns `false` (duplicate match found), the duplicate member has already been removed from the tracker but was not re-added to the graph. The loop also continues to process remaining duplicates rather than aborting, which could compound the inconsistency.

Consider checking the return value and returning `false` on failure.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 653: [suggestion] `GetVectorByInnerId` can throw a `VsagException` if the inner id is not found (e.g., if the duplicate member was concurrently removed). If this happens mid-loop, earlier iterations have already removed their members from the duplicate tracker and re-added them to the graph via `graph_add_one`, leaving the index in a partially-updated state.

Consider either: (a) collecting all duplicate data in a first pass before any mutations, so a failure leaves the original state intact; or (b) wrapping the loop body in a try-catch and attempting rollback on failure.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 651: [suggestion] In the `else if` branch, `GetVectorByInnerId(duplicate_id, ...)` is called before acquiring `global_mutex_`. At this point the duplicate member has not yet been removed from the duplicate tracker, so the vector read is safe. However, the vector data is read outside the lock while `graph_add_one` is called inside the lock — if a concurrent `UpdateVector` or `Add` modifies the vector storage between these two calls, `graph_add_one` would use stale vector data.

Consider moving the `GetVectorByInnerId` call inside the locked section, or alternatively, read the vector data after `RemoveDuplicateId` but before `graph_add_one`, both under the same lock scope to ensure consistency.

src/datacell/sparse_duplicate_tracker.cpp
```diff
@@ -119,6 +119,43 @@ SparseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplica
...
+        const auto previous_iter = next_ids_.find(previous_id);
+        if (previous_iter == next_ids_.end()) {
+            return false;
+        }
+        if (previous_iter->second == duplicate_id) {
+            break;
+        }
+        previous_id = previous_iter->second;
+        if (++count > max_count) {
+            return false;
```
LHT129, line 143: [suggestion] `next_ids_[duplicate_id]` on this line uses `operator[]` on `std::unordered_map`, which will insert a default-constructed entry if the key is unexpectedly missing. The `while` loop above already uses `find()` for safe traversal, but this line falls back to `operator[]`. While `duplicate_id` is guaranteed to exist at this point (checked by the early `count()` guard), using `at()` would make the intent clearer and provide a hard error (exception) rather than silent insertion if the invariant is ever violated.

Consider: `const auto next_id = next_ids_.at(duplicate_id);`

src/algorithm/hgraph/hgraph.cpp
LHT129, line 651: [suggestion] `duplicate_data` is allocated as `Vector<int8_t>` with `data_size` bytes on every iteration of the loop. If there are many duplicate members, this repeated allocation could be moved outside the loop and reused (e.g., by calling `.data()` to get the write pointer each iteration). The `data_size` is constant across all duplicates in this group, so a single allocation is sufficient.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -585,6 +585,12 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
     void* new_base_vec = nullptr;
```
LHT129, line 585: test comment

LHT129 (commented):
Review summary

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+        return false;
+    }
+    if (was_duplicate_member) {
+        if (not this->bottom_graph_->RemoveDuplicateId(inner_id)) {
+            return false;
+        }
+        std::shared_lock rlock(this->global_mutex_);
+        this->graph_add_one(new_base_vec, -1, inner_id);
+    } else if (not duplicate_ids.empty()) {
+        for (const auto duplicate_id : duplicate_ids) {
```
LHT129, line 650: [suggestion] If `GetVectorByInnerId` throws an exception mid-loop (e.g., due to an internal error when retrieving the vector for a duplicate member), earlier iterations have already called `RemoveDuplicateId` and `graph_add_one` on previous duplicate members. This leaves the duplicate tracker and graph in a partially-updated state — some former duplicate members are now standalone graph nodes while others remain in the old duplicate group.

Consider either: (a) collecting all duplicate vector data in a first pass before any removals, so that a failure in data retrieval does not leave partial state; or (b) wrapping the loop body in a try-catch that restores removed members on failure.

src/datacell/dense_duplicate_tracker.cpp
```diff
@@ -46,6 +46,36 @@ DenseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplicat
...
+        return false;
```
LHT129, line 54: [note] `RemoveDuplicateId` does not guard against being called on a group representative (head of the duplicate chain). In the dense tracker, `duplicate_ids_[representative]` points to the first duplicate member (not itself), so the `duplicate_ids_[duplicate_id] == duplicate_id` check does not catch this case. If called on a representative, the function would remove it from the chain and leave the remaining members in an orphaned cycle, and `duplicate_count_` would not be decremented.

The current callers in `hgraph.cpp` never pass a representative (they check `was_duplicate_member` first), so this is not a live bug. Consider adding an assertion or guard to make the contract explicit and prevent future misuse.

src/datacell/sparse_duplicate_tracker.cpp
```diff
@@ -119,6 +119,43 @@ SparseDuplicateTracker::SetDuplicateId(InnerIdType group_id, InnerIdType duplica
     next_ids_[group_id] = duplicate_id;
 }
 
+bool
+SparseDuplicateTracker::RemoveDuplicateId(InnerIdType duplicate_id) {
+    std::scoped_lock lock(mutex_);
+
+    if (next_ids_.count(duplicate_id) == 0) {
+        return false;
```
LHT129, line 127: [note] Same concern as the dense tracker: `RemoveDuplicateId` does not guard against being called on a group representative. In the sparse tracker, `next_ids_.count(representative) > 0` (the representative has an entry pointing to the first duplicate), so the early-return guard does not catch this case. If called on a representative, the group would be corrupted.

The current callers never pass a representative, but consider adding a guard for defensive correctness.

LHT129 (commented):
Thanks for this fix. The core logic of removing stale duplicate membership and reconnecting updated members to the bottom graph looks correct, and the test coverage for the three key scenarios (duplicate member update, searchability at new position, representative update detaching members) is solid.

The remaining concerns around unchecked `graph_add_one` return values and partial-update consistency in the loop have already been raised by @LHT129 in the inline comments. Those are the main items I would suggest addressing before merging.

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` are well-structured with proper cycle-detection guards and correct linked-list removal logic.

LHT129 (commented):
The fix correctly addresses the duplicate state refresh issue described in #2322. The implementation adds `RemoveDuplicateId` to both dense and sparse duplicate trackers with proper cycle detection guards, and handles both the duplicate-member-update and representative-update cases in `UpdateVector()`.

The existing inline comments from LHT129 raise valid points about checking the return value of `graph_add_one` and handling partial state on mid-loop failure in the `else if` branch. These are worth addressing before merging.

LHT129 (commented):
I have reviewed this PR. The changes look good overall — the `RemoveDuplicateId` logic in both `DenseDuplicateTracker` and `SparseDuplicateTracker` is correct, and the `UpdateVector` duplicate-state refresh in `hgraph.cpp` properly handles both the duplicate-member and representative-update cases. The test coverage is thorough with three distinct scenarios.

The remaining concern (return value of `graph_add_one` not being checked) has already been raised by @LHT129 in earlier review comments. Once that is addressed, this PR should be ready to merge.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
     if (has_precise_reorder()) {
         update_status = update_status && high_precise_codes_->UpdateVector(new_base_vec, inner_id);
     }
+    if (create_new_raw_vector_ && raw_vector_ != nullptr) {
+        update_status = update_status && raw_vector_->UpdateVector(new_base_vec, inner_id);
+    }
+    if (not update_status) {
+        return false;
+    }
+    if (was_duplicate_member) {
```
LHT129, line 643: [suggestion] `duplicate_id` is copied on every iteration. Since `InnerIdType` is likely an integer type the overhead is negligible, but for consistency with the project style and to avoid any potential copy, consider using `const auto&`:

```suggestion
    } else if (not duplicate_ids.empty()) {
        for (const auto& duplicate_id : duplicate_ids) {
```

src/algorithm/hgraph/hgraph.cpp
LHT129, line 648: [suggestion] `Vector<int8_t> duplicate_data(data_size, allocator_)` allocates heap memory on every iteration of the loop. Since `data_size` is constant across iterations, consider allocating the buffer once outside the loop and reusing it:

```cpp
    } else if (not duplicate_ids.empty()) {
        Vector<int8_t> duplicate_data(data_size, allocator_);
        for (const auto& duplicate_id : duplicate_ids) {
            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
            ...
        }
    }
```

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
-    auto codes = (has_precise_reorder()) ? high_precise_codes_ : basic_flatten_codes_;
     bool update_status = basic_flatten_codes_->UpdateVector(new_base_vec, inner_id);
     if (has_precise_reorder()) {
         update_status = update_status && high_precise_codes_->UpdateVector(new_base_vec, inner_id);
     }
+    if (create_new_raw_vector_ && raw_vector_ != nullptr) {
+        update_status = update_status && raw_vector_->UpdateVector(new_base_vec, inner_id);
+    }
+    if (not update_status) {
+        return false;
```
LHT129, line 641: [suggestion] In the `else if (not duplicate_ids.empty())` branch, the duplicate members are detached from the tracker and re-added to the graph, but the representative (`inner_id`) itself is not re-added to the graph after its vector was updated. Since the representative&apos;s vector changed, its existing graph connectivity may no longer be correct for the new position.

Consider also re-adding the representative to the graph after updating its vector, similar to what the `was_duplicate_member` branch does:

```cpp
    } else if (not duplicate_ids.empty()) {
        for (const auto& duplicate_id : duplicate_ids) {
            ...
        }
        std::shared_lock rlock(this->global_mutex_);
        this->graph_add_one(new_base_vec, -1, inner_id);
    }
```

Alternatively, if the existing graph connectivity is intentionally preserved for the representative, a brief comment explaining why would help future readers.

src/algorithm/hgraph/hgraph.cpp
LHT129, line 651: [suggestion] `Vector<int8_t> duplicate_data(data_size, allocator_)` is allocated inside the for loop, which performs a heap allocation on every iteration when there are multiple duplicate members. Since `data_size` is the same for all iterations, consider moving the allocation outside the loop and reusing the buffer.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -624,11 +630,33 @@ HGraph::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update)
...
+            Vector<int8_t> duplicate_data(data_size, allocator_);
+            GetVectorByInnerId(duplicate_id, (float*)duplicate_data.data());
+            if (not this->bottom_graph_->RemoveDuplicateId(duplicate_id)) {
+                return false;
+            }
+            std::shared_lock rlock(this->global_mutex_);
+            this->graph_add_one(duplicate_data.data(), -1, duplicate_id);
+        }
+    }
     return update_status;
```
LHT129, line 660: [note] After the new duplicate-handling code, `update_status` is guaranteed to be `true` when execution reaches `return update_status` (all failure paths return `false` early). Consider changing this to `return true` for clarity, or alternatively, the early returns in the duplicate-handling branches could set `update_status = false` and fall through to a single return point.

LHT129 (commented):
Thank you for this fix. The PR correctly addresses the root cause described in #2322 by refreshing duplicate membership state during `UpdateVector()`. The implementation is well-structured with proper lock usage and good test coverage covering all three scenarios (duplicate member update, representative update, and search correctness).

A few suggestions for consideration (also left as inline comments):

1. **[suggestion] `graph_add_one` return value**: The return value of `graph_add_one` is not checked in either the `was_duplicate_member` or `else if` branch. If `graph_add_one` returns `false` (e.g., the updated vector still matches an existing duplicate group), the member has already been removed from the duplicate tracker but was not re-added to the graph, leaving it in an inconsistent state.

2. **[suggestion] Partial state on mid-loop failure**: In the `else if` branch, if `RemoveDuplicateId` fails for a `duplicate_id` mid-loop, earlier iterations have already modified the duplicate tracker and graph, leaving the index in a partially-updated state.

3. **[suggestion] Vector allocation inside loop**: `Vector<int8_t> duplicate_data(data_size, allocator_)` is allocated on every iteration. Consider moving it outside the loop since `data_size` is invariant.

4. **[note] Return value clarity**: `return update_status` at the end of the function always returns `true` at that point since all failure paths return `false` early. Consider using `return true` for clarity.

The `RemoveDuplicateId` implementations in both `DenseDuplicateTracker` and `SparseDuplicateTracker` look correct, with proper cycle detection guards and correct linked-list removal logic. The test coverage is thorough and covers the key scenarios.

Issue #2363, opened 2026-06-29
[Bug] Multiple out-of-bounds array accesses in src/ (GraphDataCell, ODescent, SparseTermDataCell)


LHT129:
# Multiple out-of-bounds array accesses in src/

## Description

Three sites in `src/` perform array/vector access without proper bounds checking, causing undefined behavior. Because the fixes are all minimal boundary tighterings and do not change behavior on normal paths, they are combined here into a single issue/PR.

## Background

A code audit identified these potential out-of-bounds reads/writes. Triggers depend on input distribution:

- `GraphDataCell`: when an id equal to `max_capacity_` is queried for delete/recover.
- `ODescent`: when `update_neighbors` shrinks neighbor lists below `max_degree`.
- `SparseTermDataCell`: when `doc_retain_ratio_` is close to 1.0 and floating-point accumulation keeps `temp_mass` slightly below `part_mass` even after all elements are consumed.

### 1. `GraphDataCell` — off-by-one (`<=` should be `<`)

`src/datacell/graph_datacell.h` (`DeleteNeighborsById`, `RecoverDeleteNeighborsById`)

`node_versions_` has size `max_capacity_`, so the valid index range is `[0, max_capacity_ - 1]`. The current comparison `id <= max_capacity_` permits `id == max_capacity_`, which is one past the end of `node_versions_`.

```cpp
// before
if (id <= max_capacity_) {
    node_versions_[id]  // OOB when id == max_capacity_
}
// after
if (id < max_capacity_) {
    node_versions_[id]
}
```

### 2. `ODescent::repair_no_in_edge` — link size not re-checked after `update_neighbors`

`src/impl/odescent/odescent_graph_builder.cpp`

The while loop only enforces `need_replace_loc < odescent_param_->max_degree`, but `update_neighbors` may already have shrunk `graph_[i].neighbors` (i.e. `link`) below `max_degree`, so `link[need_replace_loc]` is read out of bounds. Likewise, `replace_pos[need_replace_id]` is used to index `graph_[need_replace_id].neighbors` without verifying that `neighbors.size()` is large enough.

```cpp
// while: add actual size check
while (in_edges_count[i] < min_in_degree &&
       need_replace_loc < static_cast<int>(link.size()) &&
       need_replace_loc < odescent_param_->max_degree) {
    ...
}
// if: add neighbor-list size check
if (replace_pos[need_replace_id] > 0 &&
[...]

PR #2371, opened 2026-06-29
feat: expose mrle_dim parameter to HGraph public API

### LINKED ISSUE #2372: Expose mrle_dim parameter to HGraph public API
The MRLE transformer supports mrle_dim internally but it is not exposed through the public API. Users should be able to configure this via JSON: {"mrle_dim": 127}. Need to add INDEX_MRLE_DIM constant, external param mapping in hgraph_param_mapping.cpp, validation, and unit test.

LHT129:
## Summary

Expose the internal `mrle_dim` parameter to the HGraph public API, allowing users to configure MRLE (Multi-Resolution Length Encoding) dimension truncation.

Fixes #2372

## Changes

- **`include/vsag/constants.h`**: Declare `INDEX_MRLE_DIM` external constant
- **`src/constants.cpp`**: Define `INDEX_MRLE_DIM = "mrle_dim"`
- **`src/algorithm/hgraph/hgraph_param_mapping.cpp`**:
  - Add mapping entry: `INDEX_MRLE_DIM` → `BASE_CODES_KEY/QUANTIZATION_PARAMS_KEY/MRLE_DIM_KEY`
  - Add default value `0` in JSON template
  - Add validation: `mrle_dim` must be in range [0, dim]
- **`src/algorithm/hgraph/hgraph_parameter_test.cpp`**: Add unit test verifying external param mapping

## Usage

```json
{
  "mrle_dim": 127
}
```

When `mrle_dim` is set to a non-zero value, the MRLE transformer truncates the vector dimension to that value.

### DIFF
--- a/include/vsag/constants.h
+++ b/include/vsag/constants.h
@@ -172,6 +172,7 @@ extern const char* const RABITQ_BITS_PER_DIM_FILTER;
 extern const char* const RABITQ_ERROR_RATE;
 extern const char* const RABITQ_USE_FHT;
 extern const char* const INDEX_TQ_CHAIN;
+extern const char* const INDEX_MRLE_DIM;
 
 extern const char* const HGRAPH_SUPPORT_REMOVE;
 extern const char* const HGRAPH_SUPPORT_FORCE_REMOVE;
--- a/src/algorithm/hgraph/hgraph_param_mapping.cpp
+++ b/src/algorithm/hgraph/hgraph_param_mapping.cpp
@@ -359,6 +359,14 @@ HGraph::map_hgraph_param(const JsonType& hgraph_json) {
             },
         },
         {
+            INDEX_MRLE_DIM,
+            {
+                BASE_CODES_KEY,
+                QUANTIZATION_PARAMS_KEY,
+                MRLE_DIM_KEY,
+            },
+        },
+        {
             RABITQ_BITS_PER_DIM_QUERY,
             {
                 BASE_CODES_KEY,
@@ -494,6 +502,7 @@ HGraph::map_hgraph_param(const JsonType& hgraph_json) {
                 "{RABITQ_QUANTIZATION_BITS_PER_DIM_BASE_KEY}": 1,
                 "{RABITQ_QUANTIZATION_ERROR_RATE_KEY}": 1.9,
                 "{TQ_CHAIN_KEY}": "",
+                "mrle_dim": 0,
                 "nbits": 8,
                 "{PRODUCT_QUANTIZATION_DIM_KEY}": 1,
                 "{HOLD_MOLDS}": false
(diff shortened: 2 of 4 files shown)


LHT129:
## Review fixes (ac3b551a)

- **Validation added**: Added range check for `mrle_dim` in `CheckAndMappingExternalParam` — must be in `[0, dim]`
- **PR description fixed**: Removed incorrect "Fixes #2363" reference (that issue is unrelated)
- Rebased on latest main, single commit preserved

LHT129:
## Review fixes (ca71a89b)

- **Type safety**: Added `IsInt() || IsInt64()` check before reading `mrle_dim` from JSON; switched to `GetInt64()` to avoid silent wrapping of large values
- **Grammar**: Fixed error message from `"must in range[0, {}]"` to `"must be in range [0, {}]"`
- **Tests**: Added 3 test cases for validation:
  - `mrle_dim` exceeds `dim` → rejected
  - `mrle_dim` is negative → rejected
  - `mrle_dim` omitted → defaults to 0

Also rebased on latest main (da3b8350) and ensured single commit.

LHT129:
## Fix (c1aa78c6)

Moved `mrle_dim` validation **before** `FromJson()` so that invalid values (e.g. Int64 > INT_MAX) are rejected before `VectorTransformerParameter::FromJson()` calls `GetInt()` on the internal key. Since the validated range is `[0, dim]` and `dim` fits in int, `FromJson()` will always see a safe value.

Also rebased on latest main (da3b8350), single commit, no conflicts.

LHT129:
## Fix (998f82e4)

**Compilation fix**: `JsonWrapper` does not have `IsInt()`/`IsInt64()`/`GetInt64()` — those are RapidJSON methods. Changed to:
- `IsNumberInteger()` for type check
- `GetInt()` (returns `int64_t`) for value retrieval

**Test improvements** (per review feedback):
- `REQUIRE_THROWS` → `REQUIRE_THROWS_WITH(..., ContainsSubstring("mrle_dim"))` to verify the specific validation path
- Added `Contains()` checks before accessing nested JSON in the default-value test
- Parse separate JSON strings per SECTION instead of using non-existent `AddMember()`/`GetAllocator()` (JsonWrapper has no such methods)

Rebased on latest main (da3b8350), single commit, no conflicts.

LHT129:
## Review fixes (60079ec7)

- **Unused variable**: Removed `base_json_str` that was declared but never used (each SECTION parses its own JSON with different `mrle_dim` values)
- **tq_chain/MRLE coupling**: Acknowledged. The current behavior (silently ignoring `mrle_dim` when `tq_chain` does not enable MRLE) is consistent with how other optional parameters work in the HGraph parameter mapping. Adding coupling validation is a valid enhancement but out of scope for this PR which focuses on exposing the parameter. Can be tracked as a follow-up.

Also rebased on latest main (da3b8350) and ensured single commit.

LHT129:
## Review fixes (ffb66248)

Rebased on latest `main` after SIMQ fix #2373 and addressed the latest Copilot comments:

- Made `mrle_dim` read explicit as `int64_t` after `JsonWrapper::IsNumberInteger()` validation.
- Improved the type error message to include the malformed JSON value via `Dump()`.
- Replaced test assertions that hard-coded `"mrle_dim"` with `vsag::MRLE_DIM_KEY`.
- Fixed the HGraph parameter template default: replaced unresolved `"{MRLE_DIM_KEY}"` placeholder with the real `"mrle_dim"` key.

The branch is still a single commit on top of current main.

wxyucs (approved):
lgtm

PR #2376, opened 2026-06-30
refactor(param): introduce declarative schema and migrate BruteForceParameter

LHT129:
## What

Add a small declarative schema framework under `src/param/` and use it for `BruteForceParameter` as the first migration. The framework lets a parameter class describe its own fields once and have `FromJson`, `ToJson`, `CheckCompatibility` (and unknown-key detection) derived from that single source of truth.

## Why

`BruteForceParameter` (and every other `*_parameter.{h,cpp}` that will follow) hand-writes three near-identical blocks: a `FromJson` with `CHECK_ARGUMENT(json.Contains(...))`, a `ToJson` with the matching `json[KEY].SetJson(...)`, and a `CheckCompatibility` that chains `CHECK_FIELD_EQ` / `CHECK_SUB_PARAM`. The boilerplate grows linearly with the field count and is silent on unknown keys, so typos like `{"pq_dim": 8}` on a BruteForce index parse cleanly today.

A schema-driven approach lets each class declare the fields once and produce the three behaviours from that single declaration. `Range<T>` covers numeric validation, and a `ConstantStringField` handles the common `"type": "brute_force"` tag without giving it special-case treatment.

## Schema framework (`src/param/`)

* `validators.h` provides a header-only `Range<T>` validator that callers can attach to a scalar field. It throws `VsagException` with `ErrorType::INVALID_ARGUMENT` if the value is outside the closed `[min, max]` range, mirroring the existing `CHECK_ARGUMENT` style.
* `schema.h` exposes `Schema`, `SchemaBuilder<Owner>`, and a small set of `VSAG_PARAM*` macros. Two field kinds are supported in this PR:
  * `ScalarField<T>` for `uint64_t` / `int64_t` / `std::string` / `bool` / `float` (the types BruteForce-class parameters use today).
  * `SubParamField<Sub>` for `Parameter`-derived sub-objects that are constructed via a factory (e.g. `FlattenInterfaceParameter` for `BruteForceParameter::base_codes_param`).
  * A `ConstantStringField` covers the `"type": "brute_force"` tag pattern that several index parameters share.
  * `Schema::Parse` calls each field's parser, fills the C++ member, runs the per-field validator, and refuses unknown JSON keys via `CheckUnknownKeys`. `Schema::Serialize` writes the fields back in the order they were declared so the JSON layout stays stable. `Schema::Equal` walks the fields once and stops at the first mismatch, replacing manual `CHECK_FIELD_EQ` / `CHECK_SUB_PARAM` ladders.
* `schema.cpp` implements `CheckUnknownKeys` using `JsonWrapper::GetInnerJson()->items()` so callers can register the set of keys the base class already consumes (`extra_known_keys`) without forcing every parameter to inherit a common JSON pre-filter.

The framework is intentionally narrow: it does not touch any `InnerIndexParameter` field, does not add a `Validate` hook (left for PR #2), and does not migrate any of the other parameter classes. That restraint is what keeps the BruteForce-side diff small.

## BruteForce migration

`BruteForceParameter` now declares its schema in-class via `VSAG_PARAM_SCHEMA_DECL()` and defines it out-of-line in the `.cpp`:

```cpp
VSAG_PARAM_SCHEMA(BruteForceParameter)
    VSAG_PARAM_TYPE_TAG(TYPE_KEY, INDEX_TYPE_BRUTE_FORCE)
    VSAG_PARAM_SUBPARAM_REQUIRED(FlattenInterfaceParameter, base_codes_param, BASE_CODES_KEY, CreateFlattenParam)
VSAG_PARAM_SCHEMA_END()
```

Three overrides (`FromJson` / `ToJson` / `CheckCompatibility`) delegate to `schema().Parse` / `schema().Serialize` / `schema().Equal` after forwarding to the `InnerIndexParameter` base, so inherited fields keep their existing behaviour. The hand-written `CHECK_ARGUMENT(json.Contains(BASE_CODES_KEY))`, `CHECK_SUB_PARAM(*this, *p, base_codes_param)`, and `json[BASE_CODES_KEY].SetJson(base_codes_param->ToJson())` blocks are gone.

## JSON compatibility

The schema keeps the field order stable and uses the same keys / types as the hand-written version, so `BruteForceParameter::FromJson(user_json)` followed by `ToJson()` produces byte-identical output before and after this PR. Unknown keys (`{"pq_dim": 8}` etc.) now throw `INVALID_ARGUMENT` instead of being silently accepted.

## Tests

* `src/param/schema_test.cpp` exercises the framework end-to-end with a minimal in-process `TestParam` (one scalar with a `Range`, one string, one sub-parameter) and covers four behaviours: scalar + sub-parameter parse, JSON round-trip stability, `Range` rejection, and unknown-key rejection. Both rejection paths assert `VsagException`.
* The existing `BruteForce Parameters CheckCompatibility` test in `src/algorithm/bruteforce/bruteforce_parameter_test.cpp` is unchanged and continues to exercise the public surface of `BruteForceParameter` (now backed by the schema).

## Build wiring

* `src/param/CMakeLists.txt` follows the project convention: an `OBJECT` library `param` for production code and a `STATIC` library `param_test` for the unit tests, gated on `ENABLE_TESTS`.
* `src/CMakeLists.txt` registers `param` as a subdirectory and adds it to `VSAG_DEP_LIBS` so `vsag` / `vsag_static` pick it up.
* `tests/CMakeLists.txt` links `param_test` into `unittests` with the same `--whole-archive` (Linux) / `-Wl,-force_load` (macOS) treatment used for the other `*_test` static libraries.

## Verification

* `cmake -B build -DCMAKE_BUILD_TYPE=Debug -DENABLE_TESTS=ON -G 'Unix Makefiles' -S.` succeeded.
* `cmake --build build --target unittests` built successfully (including `param`, `param_test`, `bruteforce`, `algorithm_test`).
* Ran the targeted filter `[ut][param][schema],[ut][BruteForceParameter],[brute_force_threshold]`:
  ```
  All tests passed (17 assertions in 6 test cases)
  ```
* `make fmt` (clang-format v15) clean.

Signed-off-by: claude &lt;claude@anthropic.com&gt;
Signed-off-by: LHT129 &lt;tianlan.lht@antgroup.com&gt;


src/algorithm/bruteforce/bruteforce_parameter.cpp
```diff
@@ -15,32 +15,34 @@
...
 
 void
 BruteForceParameter::FromJson(const JsonType& json) {
     InnerIndexParameter::FromJson(json);
-    CHECK_ARGUMENT(json.Contains(BASE_CODES_KEY),
-                   fmt::format("bruteforce parameters must contains {}", BASE_CODES_KEY));
-    const auto& base_codes_json = json[BASE_CODES_KEY];
-    this->base_codes_param = CreateFlattenParam(base_codes_json);
+    schema().Parse(this, json);
 }
```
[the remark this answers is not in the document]
LHT129: Fixed in c6328159. Added `schema().CheckUnknownKeys(json, base_keys)` in `BruteForceParameter::FromJson` after `schema().Parse`. The base_keys set contains all 11 keys from InnerIndexParameter. Unknown keys now throw INVALID_ARGUMENT.

src/param/validators.h
```diff
@@ -0,0 +1,66 @@
...
+    template <typename T>
+    void
+    operator()(const std::string& key, const T& value) const {
+        const T min_cast = static_cast<T>(this->min_value);
+        const T max_cast = static_cast<T>(this->max_value);
+        CHECK_ARGUMENT(
+            value >= min_cast && value <= max_cast,
+            fmt::format(
+                "parameter '{}' value {} out of range [{}, {}]", key, value, min_cast, max_cast));
+    }
```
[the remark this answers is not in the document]
LHT129: Fixed in c6328159. Restructured `Range` so the template parameter T must match the field type exactly. Comparisons are now same-type (no cross-signedness cast at comparison time). The constructor accepts arbitrary arithmetic literals via a SFINAE-constrained template constructor that static_casts to T once during construction. Usage updated to `Range<uint64_t>(1, 1000000)`. Added `#include <type_traits>` for `std::enable_if_t` and `std::is_arithmetic_v`.

LHT129 (commented):
[critical] **tests/CMakeLists.txt line 45: Generator expression typo in macOS linker flags**

`$<TARGET_FILE: param_test>` contains a space before `param_test`, which will cause a CMake configuration error on macOS.

Additionally, `param_test` and `datacell_test` are incorrectly merged into a single `-Wl,-force_load` flag. Each test library needs its own `-Wl,-force_load` prefix, matching the pattern used for every other test library on the surrounding lines (lines 39-44, 46-51).

Current (line 45):
```cmake
"-Wl,-force_load,$<TARGET_FILE: param_test> $<TARGET_FILE:datacell_test>"
```

Should be two separate flags:
```cmake
"-Wl,-force_load,$<TARGET_FILE:param_test>"
"-Wl,-force_load,$<TARGET_FILE:datacell_test>"
```

Note: line 42 already correctly adds `param_test` with its own `-Wl,-force_load`, so line 45 appears to be a duplicate/merge error. The fix should remove line 45 and ensure `datacell_test` retains its own `-Wl,-force_load` flag (which was lost when the original line 43 was replaced).

---

[suggestion] **src/algorithm/bruteforce/bruteforce_parameter.cpp: Unused `<unordered_set>` include**

The `<unordered_set>` include was added (replacing `<fmt/format.h>`) but is not used anywhere in the current version of this file. This appears to be a leftover from a `CheckUnknownKeys` fix that was partially reverted. The include should be removed to keep the code clean.

---

[suggestion] **src/algorithm/bruteforce/bruteforce_parameter.cpp: Missing `CheckUnknownKeys` call in `FromJson`**

The PR description states that unknown keys (e.g. `{"pq_dim": 8}`) now throw `INVALID_ARGUMENT` instead of being silently accepted. However, `BruteForceParameter::FromJson` only calls `schema().Parse(this, json)` and does not call `schema().CheckUnknownKeys(json, extra_known_keys)`. As a result, unknown keys are still silently accepted for `BruteForceParameter`.

The `InnerIndexParameter::FromJson` consumes keys like `USE_REORDER_KEY`, `BUILD_THREAD_COUNT_KEY`, etc. that are not declared in the `BruteForceParameter` schema. These should be passed as `extra_known_keys` to `CheckUnknownKeys`. The `<unordered_set>` include appears to be a leftover from a previous attempt to add this call.

---

[note] **src/inner_string_params.h: Inconsistent key naming in DEFAULT_MAP**

The newly added entry `{"INDEX_BRUTE_FORCE", INDEX_BRUTE_FORCE}` uses the key `"INDEX_BRUTE_FORCE"` but all other index type entries in the map use the pattern `"INDEX_TYPE_*"` (e.g. `"INDEX_TYPE_HGRAPH"`, `"INDEX_TYPE_IVF"`). For consistency, consider using `"INDEX_TYPE_BRUTE_FORCE"` as the key, or verify that `"INDEX_BRUTE_FORCE"` is intentionally different.

tests/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ target_link_libraries (unittests
         PRIVATE
         Catch2::Catch2 
         simd_test vsag_test algorithm_test factory_test attr_test
-        datacell_test quantizer_test storage_test io_test utils_test impl_test
+        datacell_test param_test quantizer_test storage_test io_test utils_test impl_test
         vsag_static
 )
```
LHT129, line 35: [critical] Generator expression typo: `$<TARGET_FILE: param_test>` contains a space before `param_test`, which will cause a CMake configuration error on macOS.

Additionally, `param_test` and `datacell_test` are incorrectly merged into a single `-Wl,-force_load` flag. Each test library needs its own `-Wl,-force_load` prefix, matching the pattern used for every other test library on the surrounding lines.

Current:
```cmake
"-Wl,-force_load,$<TARGET_FILE: param_test> $<TARGET_FILE:datacell_test>"
```

Should be two separate flags:
```cmake
"-Wl,-force_load,$<TARGET_FILE:param_test>"
"-Wl,-force_load,$<TARGET_FILE:datacell_test>"
```

Note: line 42 already correctly adds `param_test` with its own `-Wl,-force_load`, so this line appears to be a duplicate/merge error. The fix should remove this line and ensure `datacell_test` retains its own `-Wl,-force_load` flag.

tests/CMakeLists.txt
```diff
@@ -39,9 +39,10 @@ if (APPLE)
         "-Wl,-force_load,$<TARGET_FILE:simd_test>"
         "-Wl,-force_load,$<TARGET_FILE:vsag_test>"
         "-Wl,-force_load,$<TARGET_FILE:algorithm_test>"
+        "-Wl,-force_load,$<TARGET_FILE:param_test>"
         "-Wl,-force_load,$<TARGET_FILE:factory_test>"
```
LHT129, line 43: [critical] Generator expression typo: `$<TARGET_FILE: param_test>` contains a space before `param_test`, which will cause a CMake configuration error on macOS.

Additionally, `param_test` and `datacell_test` are incorrectly merged into a single `-Wl,-force_load` flag. Each test library needs its own `-Wl,-force_load` prefix, matching the pattern used for every other test library on the surrounding lines.

Current:
```cmake
"-Wl,-force_load,$<TARGET_FILE: param_test> $<TARGET_FILE:datacell_test>"
```

Should be two separate flags:
```cmake
"-Wl,-force_load,$<TARGET_FILE:param_test>"
"-Wl,-force_load,$<TARGET_FILE:datacell_test>"
```

Note: the line above (position 10) already correctly adds `param_test` with its own `-Wl,-force_load`, so this line appears to be a duplicate/merge error. The fix should remove this line and ensure `datacell_test` retains its own `-Wl,-force_load` flag (which was lost when the original line was replaced).

src/algorithm/bruteforce/bruteforce_parameter.cpp
```diff
@@ -15,7 +15,7 @@
 
 #include "bruteforce_parameter.h"
 
-#include <fmt/format.h>
```
LHT129, line 18: [suggestion] The `<unordered_set>` include is unused in the current version of this file. It appears to be a leftover from a previous iteration that called `schema().CheckUnknownKeys(json, extra_known_keys)` with a `std::unordered_set<std::string>` for base class keys. Since that call is not present in the current code, the include should be removed.

If the `CheckUnknownKeys` call is added back (see related comment on `FromJson`), then this include will be needed.

src/algorithm/bruteforce/bruteforce_parameter.cpp
```diff
@@ -24,23 +24,27 @@
 
 namespace vsag {
```
LHT129, line 26: [suggestion] The PR description states that unknown keys (e.g. `{"pq_dim": 8}`) now throw `INVALID_ARGUMENT` instead of being silently accepted. However, `FromJson` only calls `schema().Parse(this, json)` and does not call `schema().CheckUnknownKeys(json, extra_known_keys)`.

`InnerIndexParameter::FromJson` (called on the line above) consumes keys like `USE_REORDER_KEY`, `BUILD_THREAD_COUNT_KEY`, `LABEL_REMAP_TYPE_KEY`, `USE_ATTRIBUTE_FILTER_KEY`, `PRECISE_CODES_KEY`, `STORE_RAW_VECTOR_KEY`, `RAW_VECTOR_KEY`, `EXTRA_INFO_KEY`, `TRAIN_SAMPLE_COUNT_KEY`, `ATTR_PARAMS_KEY`, and `REORDER_SOURCE_KEY`. These should be passed as `extra_known_keys` so that only truly unknown keys are rejected.

Suggested fix:
```cpp
void
BruteForceParameter::FromJson(const JsonType& json) {
    InnerIndexParameter::FromJson(json);
    schema().Parse(this, json);
    schema().CheckUnknownKeys(json, {
        USE_REORDER_KEY, REORDER_SOURCE_KEY, USE_ATTRIBUTE_FILTER_KEY,
        ATTR_PARAMS_KEY, BUILD_THREAD_COUNT_KEY, LABEL_REMAP_TYPE_KEY,
        PRECISE_CODES_KEY, STORE_RAW_VECTOR_KEY, RAW_VECTOR_KEY,
        EXTRA_INFO_KEY, TRAIN_SAMPLE_COUNT_KEY
    });
}
```

This would also justify the `<unordered_set>` include that is currently unused.

src/inner_string_params.h
```diff
@@ -200,6 +200,7 @@ const std::unordered_map<std::string, std::string> DEFAULT_MAP = {
     {"INDEX_TYPE_IVF", INDEX_TYPE_IVF},
     {"INDEX_TYPE_GNO_IMI", INDEX_TYPE_GNO_IMI},
     {"INDEX_TYPE_PYRAMID", INDEX_TYPE_PYRAMID},
+    {"INDEX_BRUTE_FORCE", INDEX_BRUTE_FORCE},
```
LHT129, line 203: [note] The newly added entry uses the key `"INDEX_BRUTE_FORCE"` but all other index type entries in `DEFAULT_MAP` use the pattern `"INDEX_TYPE_*"` (e.g. `"INDEX_TYPE_HGRAPH"`, `"INDEX_TYPE_IVF"`, `"INDEX_TYPE_GNO_IMI"`, `"INDEX_TYPE_PYRAMID"`). For consistency, consider using `"INDEX_TYPE_BRUTE_FORCE"` as the key, or verify that `"INDEX_BRUTE_FORCE"` is intentionally different from the established naming convention.

tests/CMakeLists.txt
LHT129, line 43: [critical] The Apple `-force_load` line for `param_test` has a malformed generator expression: `"$<TARGET_FILE: param_test>"` contains a space between the colon and the target name. CMake will not expand this correctly, causing a link failure on macOS.

The line also incorrectly merges `param_test` and `datacell_test` into a single `-force_load` argument. Each library needs its own `-force_load` flag.

Current (broken):
```cmake
"-Wl,-force_load,$<TARGET_FILE: param_test> $<TARGET_FILE:datacell_test>"
```

Should be two separate entries:
```cmake
"-Wl,-force_load,$<TARGET_FILE:param_test>"
"-Wl,-force_load,$<TARGET_FILE:datacell_test>"
```

src/algorithm/bruteforce/bruteforce_parameter.h
```diff
@@ -34,8 +35,8 @@ class BruteForceParameter : public InnerIndexParameter {
     bool
     CheckCompatibility(const vsag::ParamPtr& other) const override;
 
-public:
     FlattenInterfaceParamPtr base_codes_param{nullptr};
```
LHT129, line 38: [suggestion] The `public:` label before `base_codes_param` was removed, making the member private by default (since the preceding `CheckCompatibility` override ends the public section). This is a behavioral change: `base_codes_param` was previously accessible from outside the class.

If this visibility change is intentional, consider adding an explicit `private:` label to make the intent clear and avoid confusion. If it is unintentional, restore the `public:` label.

Note that `DEFINE_POINTER2(FlattenDataCellParam, FlattenDataCellParameter)` was also removed from the header — verify that no other translation unit depends on this typedef being available through `bruteforce_parameter.h`.

src/algorithm/bruteforce/bruteforce_parameter.cpp
```diff
@@ -15,7 +15,7 @@
 
 #include "bruteforce_parameter.h"
 
-#include <fmt/format.h>
+#include <unordered_set>
```
LHT129, line 18: [suggestion] The `#include <unordered_set>` on line 18 is unused in the current code. `FromJson` calls `schema().Parse(this, json)` but does not call `schema().CheckUnknownKeys(json, extra_known_keys)`, so unknown keys in the JSON input are still silently accepted for `BruteForceParameter`.

As noted in the PR description, one of the benefits of the schema framework is rejecting unknown keys. To achieve this, add the `CheckUnknownKeys` call with the keys consumed by `InnerIndexParameter::FromJson`:

```cpp
void
BruteForceParameter::FromJson(const JsonType& json) {
    InnerIndexParameter::FromJson(json);
    schema().Parse(this, json);
    schema().CheckUnknownKeys(json, {
        USE_REORDER_KEY, REORDER_SOURCE_KEY, USE_ATTRIBUTE_FILTER_KEY,
        ATTR_PARAMS_KEY, BUILD_THREAD_COUNT_KEY, LABEL_REMAP_TYPE_KEY,
        PRECISE_CODES_KEY, STORE_RAW_VECTOR_KEY, RAW_VECTOR_KEY,
        EXTRA_INFO_KEY, TRAIN_SAMPLE_COUNT_KEY
    });
}
```

If the `CheckUnknownKeys` call is intentionally deferred to a follow-up PR, then the `<unordered_set>` include should be removed for now.

LHT129 (commented):
## Review Summary

This PR introduces a clean declarative schema framework for parameter classes and migrates `BruteForceParameter` as the first user. The design is well-thought-out: a single source of truth drives `FromJson`, `ToJson`, and `CheckCompatibility`, and the macro DSL is ergonomic. The test coverage for the framework itself is solid.

Three issues found:

1. **[critical] `tests/CMakeLists.txt:43`** — The Apple `-force_load` generator expression has a space in `$<TARGET_FILE: param_test>`, which will cause a CMake expansion failure on macOS. Additionally, `param_test` and `datacell_test` are incorrectly merged into a single `-force_load` argument instead of being separate entries.

2. **[suggestion] `bruteforce_parameter.h:38`** — The `public:` label was removed, changing `base_codes_param` from public to private. If intentional, an explicit `private:` label would clarify intent. Also, `DEFINE_POINTER2(FlattenDataCellParam, FlattenDataCellParameter)` was removed — verify no other TU depends on it being transitively available through this header.

3. **[suggestion] `bruteforce_parameter.cpp:18`** — `<unordered_set>` is included but unused. The `CheckUnknownKeys` call (a key benefit of the schema framework per the PR description) is missing from `BruteForceParameter::FromJson`. Either add the call with the 11 base-class keys, or remove the unused include.

The schema framework code itself (`schema.h`, `schema.cpp`, `validators.h`, `schema_test.cpp`) looks correct and well-structured. The `Range` validator fix (same-type comparisons) addresses the earlier gemini-code-assist concern.

PR #2435, opened 2026-07-13
fix: reuse visited lists across sub-pools

### LINKED ISSUE #2427: [improve](hgraph): reuse idle visited lists across sub-pools before allocating
### Kind of improvement
perf

### Affected area
hgraph

### Summary
`ResourceObjectPool<VisitedList>::TakeOne()` partitions the pool into 16 sub-pools and prefers a thread-local sub-pool. When the chosen sub-pool is empty, it allocates a new `VisitedList` immediately instead of checking whether other sub-pools already hold idle objects. In HGraph search, each request takes one visited list and returns it after route-graph and bottom-graph search (`src/algorithm/hgraph/hgraph_search.cpp:453`, `src/algorithm/hgraph/hgraph_search.cpp:544`), so peak allocation count depends heavily on this pool policy.

For large-capacity indexes, each `VisitedList` is expensive because it stores one `uint16_t` tag per inner id (`src/utils/visited_list.h:30`, `src/utils/visited_list.h:54`). Under search pressure, the current allocation policy can grow the cached object count more than necessary even when other sub-pools already have idle objects.

Suggested direction: before allocating a new object, try to reuse from another non-empty sub-pool; alternatively maintain a small global overflow free list.

### Expected benefit
Reduce the peak number of live `VisitedList` objects without changing HGraph search semantics. This should lower scratch RSS / allocator live bytes for large-capacity indexes and make memory usage less sensitive to thread scheduling.

### Benchmark evidence
Observed in a search stress test with a custom allocator instrumented by backtrace:
- call path: `HGraph::KnnSearch -> HGraph::SearchWithRequest -> ResourceObjectPool<VisitedList>::TakeOne -> VisitedList::VisitedList -> SafeAllocator::Allocate`
- the visited-list pool reached `live_count=34` and `live_bytes=680026112`
- this works out to roughly 20 MB per object, consistent with `VisitedList` using `uint16_t[max_capacity]`

`// TODO`: add before/after numbers from a patched allocator strategy.

### References
- src/utils/resource_object_pool.h:37
- src/utils/resource_object_pool.h:69
- src/utils/visited_list.h:30
- src/utils/visited_list.h:54
- src/algorithm/hgraph/hgraph_search.cpp:453
- src/algorithm/hgraph/hgraph_search.cpp:544

### Additional context
This is not a correctness bug by itself; the issue is that the runtime scratch pool keeps more large objects live than necessary.

## Related issues
- #49 fix memory leak in hnsw

---

jac0626:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

- Fixes: #2427

## What Changed

- Scan all visited-list sub-pools for an idle object before allocating a new one.
- Preserve the existing thread-local sub-pool preference and non-blocking lock strategy.
- Add regression coverage for reusing an object from another sub-pool.

## Test Evidence

- [x] Other (described below)

Test details:

```text
clang-format 15.0.7 --dry-run --Werror (changed files): passed
make test CASE='[ut][VisitedListPool]':
All tests passed (10001 assertions in 2 test cases)
```

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: pool allocation occurs only after all sub-pools are observed empty

- Performance impact: reduces unnecessary large visited-list allocations
- Concurrency/thread-safety impact: retains the existing per-sub-pool mutexes and `try_lock` behavior

## Documentation Impact

- [x] No docs update needed

## Risk and Rollback

- Risk level: low
- Rollback plan: revert the commit to restore the previous allocation policy

## Checklist

---

### DIFF
--- a/src/utils/resource_object_pool.h
+++ b/src/utils/resource_object_pool.h
@@ -72,23 +72,31 @@ public:
             std::hash<std::thread::id>()(std::this_thread::get_id()) % kSubPoolCount};
         auto pool_id = prefer_pool_id_;
         while (true) {
-            if (sub_pool_mutexes_[pool_id].try_lock()) {
-                prefer_pool_id_ = pool_id;
-                if (pool_[pool_id]->empty()) {
-                    sub_pool_mutexes_[pool_id].unlock();
-                    auto obj = this->constructor_();
-                    obj->source_pool_id_ = pool_id;
-                    return obj;
+            uint64_t empty_pool_count = 0;
+            for (uint64_t i = 0; i < kSubPoolCount; ++i) {
+                if (sub_pool_mutexes_[pool_id].try_lock()) {
+                    if (pool_[pool_id]->empty()) {
+                        sub_pool_mutexes_[pool_id].unlock();
+                        ++empty_pool_count;
+                    } else {
+                        std::shared_ptr<T> obj = pool_[pool_id]->front();
+                        pool_[pool_id]->pop_front();
+                        sub_pool_mutexes_[pool_id].unlock();
+                        prefer_pool_id_ = pool_id;
+                        obj->source_pool_id_ = pool_id;
+                        obj->Reset();
+                        return obj;
+                    }
                 }
-                std::shared_ptr<T> obj = pool_[pool_id]->front();
-                pool_[pool_id]->pop_front();
-                sub_pool_mutexes_[pool_id].unlock();
-                obj->source_pool_id_ = pool_id;
-                obj->Reset();
+
+                ++pool_id;
+                pool_id %= kSubPoolCount;
+            }
+            if (empty_pool_count == kSubPoolCount) {
+                auto obj = this->constructor_();
+                obj->source_pool_id_ = prefer_pool_id_;
                 return obj;
             }
-            ++pool_id;
-            pool_id %= kSubPoolCount;
         }
     }
 
--- a/src/utils/visited_list_test.cpp
+++ b/src/utils/visited_list_test.cpp
@@ -121,3 +121,15 @@ TEST_CASE("VisitedListPool Basic Test", "[ut][VisitedListPool]") {
         }
     }
 }
+
+TEST_CASE("VisitedListPool reuses objects across sub-pools", "[ut][VisitedListPool]") {
+    auto allocator = std::make_shared<DefaultAllocator>();
+    auto pool = std::make_shared<VisitedListPool>(
+        VisitedListPool::kSubPoolCount, allocator.get(), 1000, allocator.get());
+    auto initial_memory_usage = pool->GetMemoryUsage();
+
+    auto first = pool->TakeOne();
+    auto second = pool->TakeOne();
+
+    REQUIRE(pool->GetMemoryUsage() == initial_memory_usage);
+}


LHT129 (commented):
[suggestion] The new test verifies that memory usage stays constant after taking two objects, which is good for a basic sanity check. However, it does not verify that the second `TakeOne()` actually returned an object from a *different* sub-pool than the first one. Consider adding an assertion that the two returned objects have different `source_pool_id_` values to confirm cross-pool reuse is actually happening.

Additionally, there is no concurrent test coverage for the new cross-pool scanning logic. The existing `test concurrency` section exercises `TakeOne`/`ReturnOne` under multiple threads but does not specifically verify that objects are reused across sub-pools under contention. Consider adding a concurrent scenario where multiple threads take and return objects, then verify that no extra allocations occur beyond the initial pool size.

src/utils/resource_object_pool.h
```diff
@@ -72,23 +72,31 @@ class ResourceObjectPool {
             std::hash<std::thread::id>()(std::this_thread::get_id()) % kSubPoolCount};
```
LHT129, line 72: [note] The `empty_pool_count` variable is declared inside the `while(true)` loop (line 75) and re-initialized to 0 on each iteration. This is correct for the current logic, but it is worth noting that if `try_lock()` fails on a sub-pool because another thread holds the lock, that sub-pool is silently skipped and not counted as empty. This means the busy-spin `while(true)` loop will keep retrying until all sub-pools can be locked and are observed empty. Under high contention, this could cause a brief CPU spike. The gemini-code-assist bot already suggested a single-pass approach that avoids this entirely.

src/utils/visited_list_test.cpp
```diff
@@ -121,3 +121,15 @@ TEST_CASE("VisitedListPool Basic Test", "[ut][VisitedListPool]") {
...
+    auto allocator = std::make_shared<DefaultAllocator>();
+    auto pool = std::make_shared<VisitedListPool>(
+        VisitedListPool::kSubPoolCount, allocator.get(), 1000, allocator.get());
+    auto initial_memory_usage = pool->GetMemoryUsage();
+
+    auto first = pool->TakeOne();
+    auto second = pool->TakeOne();
+
+    REQUIRE(pool->GetMemoryUsage() == initial_memory_usage);
+}
```
LHT129, line 135: [note] The test name says "reuses objects across sub-pools" but the test does not actually verify cross-pool reuse behavior. The pool is initialized with `kSubPoolCount` (16) objects, and `TakeOne()` is called only twice — both objects likely come from the same preferred sub-pool. The assertion `GetMemoryUsage() == initial_memory_usage` would pass even with the old code, since pre-filled objects are being consumed without new allocations.

To meaningfully test cross-pool reuse, consider a scenario where:
1. Take objects from the preferred sub-pool until it is empty
2. Return those objects (which sets their `source_pool_id_` to the preferred pool)
3. Take objects again — the new code should find them in the preferred sub-pool, but the old code would have allocated new ones if the preferred sub-pool was empty

Alternatively, verify that after exhausting all sub-pools and returning objects, subsequent `TakeOne()` calls reuse returned objects rather than allocating new ones.

src/utils/resource_object_pool.h
```diff
@@ -72,23 +72,31 @@ class ResourceObjectPool {
...
-                obj->Reset();
+
+                ++pool_id;
+                pool_id %= kSubPoolCount;
+            }
+            if (empty_pool_count == kSubPoolCount) {
+                auto obj = this->constructor_();
+                obj->source_pool_id_ = prefer_pool_id_;
                 return obj;
             }
```
LHT129, line 99: [note] In the allocation fallback path (line 99), `prefer_pool_id_` is not updated before being used as `source_pool_id_`. In the old code, `prefer_pool_id_` was set to the pool that was found empty (the pool that triggered the allocation). This behavioral difference means newly allocated objects are always assigned to the thread's last successful pool rather than the pool that was observed empty. While this is unlikely to cause correctness issues, it subtly changes the distribution pattern of allocated objects across sub-pools compared to the previous behavior.

If this change is intentional, it would be helpful to document the rationale. Otherwise, consider updating `prefer_pool_id_` to the current `pool_id` (or the first pool observed empty) before the allocation, consistent with the old behavior.

PR #2436, opened 2026-07-13
fix: bound hgraph visited-list cache

### LINKED ISSUE #2428: [improve](hgraph): support trimming visited-list pool memory after load drops
### Kind of improvement
perf

### Affected area
hgraph

### Summary
The visited-list pool keeps runtime scratch memory at its peak high-water mark after search pressure drops. `ReturnOne()` always puts a `VisitedList` back into the pool cache (`src/utils/resource_object_pool.h:96`), but there is no trimming policy to release idle objects. On the HGraph side, `resize()` rebuilds the pool only when capacity grows (`src/algorithm/hgraph/hgraph_build.cpp:549`), and `shrink_to_fit()` shrinks index data structures but does not rebuild or shrink the visited-list pool (`src/algorithm/hgraph/hgraph_modify.cpp:195`).

As a result, once a large number of `VisitedList` objects have been allocated during a search spike, allocator live bytes stay high even after concurrency drops or after force-remove reduces the effective index size.

Suggested direction: support bounded caching or trimming for idle visited lists, and consider rebuilding or shrinking the pool when `shrink_to_fit()` reduces effective capacity.

### Expected benefit
Allow runtime scratch memory to fall after traffic drops or after force-remove / shrink-to-fit. This would make HGraph memory usage better match current demand instead of permanent historical peak pressure.

### Benchmark evidence
Observed in a search stress test with a custom allocator instrumented by backtrace:
- call path: `HGraph::KnnSearch -> HGraph::SearchWithRequest -> ResourceObjectPool<VisitedList>::TakeOne -> VisitedList::VisitedList -> SafeAllocator::Allocate`
- the visited-list pool reached `live_count=34` and `live_bytes=680026112`
- returning those objects to the pool did not reduce allocator live bytes because the objects remained cached rather than freed

`// TODO`: add allocator snapshots before and after load drop, and after an explicit trim or shrink implementation.

### References
- src/utils/resource_object_pool.h:96
- src/algorithm/hgraph/hgraph_build.cpp:549
- src/algorithm/hgraph/hgraph_build.cpp:560
- src/algorithm/hgraph/hgraph_modify.cpp:195
- src/utils/visited_list.h:54

### Additional context
An explicit runtime API such as `TrimRuntimeScratch()` would be acceptable if fully automatic trimming is too disruptive for hot paths.

## Related issues
_None found._

---

jac0626:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

- Fixes: #2428

## What Changed

- Add an optional per-sub-pool cache limit to `ResourceObjectPool`; existing callers remain unlimited by default.
- Configure HGraph to retain at most one idle visited list in each of its 16 sub-pools.
- Release excess returned objects and subtract their memory from pool statistics.
- Add regression coverage for memory usage falling after excess objects are returned.

## Test Evidence

- [x] Other (described below)

Test details:

```text
clang-format 15.0.7 --dry-run --Werror (changed files): passed
make test CASE='[ut][VisitedListPool]':
All tests passed (10002 assertions in 2 test cases)
```

## Compatibility Impact

- API/ABI compatibility: no public VSAG API changes
- Behavior changes: HGraph retains at most 16 idle visited lists instead of retaining the historical peak

- Performance impact: bounds idle scratch memory; workloads above 16 concurrent searches may reallocate excess lists on a later burst
- Concurrency/thread-safety impact: uses the existing per-sub-pool mutex and adds no global lock or background work

## Documentation Impact

- [x] No docs update needed

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert the commit to restore unlimited HGraph visited-list caching

## Checklist

---

### DIFF
--- a/src/utils/visited_list_test.cpp
+++ b/src/utils/visited_list_test.cpp
@@ -121,3 +121,22 @@ TEST_CASE("VisitedListPool Basic Test", "[ut][VisitedListPool]") {
         }
     }
 }
+
+TEST_CASE("VisitedListPool releases objects above the cache limit", "[ut][VisitedListPool]") {
+    auto allocator = std::make_shared<DefaultAllocator>();
+    auto pool = std::make_shared<VisitedListPool>(0, allocator.get(), 1000, allocator.get());
+    pool->SetMaxCachedObjectCountPerSubPool(1);
+    auto initial_memory_usage = pool->GetMemoryUsage();
+
+    std::vector<std::shared_ptr<VisitedList>> visited_lists;
+    for (uint64_t i = 0; i < 3; ++i) {
+        visited_lists.emplace_back(pool->TakeOne());
+    }
+    REQUIRE(pool->GetMemoryUsage() > initial_memory_usage);
+    auto object_memory_usage = visited_lists.front()->GetMemoryUsage();
+
+    for (auto& visited_list : visited_lists) {
+        pool->ReturnOne(visited_list);
+    }
+    REQUIRE(pool->GetMemoryUsage() == initial_memory_usage + object_memory_usage);
+}
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -597,6 +597,7 @@ private:
     float alpha_{1.0};            // Relative Neighborhood Graph pruning coefficient
 
     std::shared_ptr<VisitedListPool> pool_{nullptr};  // pool of visited-lists for search
+    static constexpr uint64_t MAX_CACHED_VISITED_LIST_COUNT_PER_SUB_POOL = 1;
 
     mutable std::shared_mutex global_mutex_;        // guards total_count_, entry_point_id_
     mutable MutexArrayPtr neighbors_mutex_;         // per-node locks for neighbor lists
--- a/src/algorithm/hgraph/hgraph_build.cpp
+++ b/src/algorithm/hgraph/hgraph_build.cpp
@@ -558,6 +558,7 @@ HGraph::resize(uint64_t new_size) {
     if (cur_size < new_size_power_2) {
         this->neighbors_mutex_->Resize(new_size_power_2);
         pool_ = std::make_shared<VisitedListPool>(1, allocator_, new_size_power_2, allocator_);
+        pool_->SetMaxCachedObjectCountPerSubPool(MAX_CACHED_VISITED_LIST_COUNT_PER_SUB_POOL);
         this->label_table_->Resize(new_size_power_2);
         bottom_graph_->Resize(new_size_power_2);
         this->basic_flatten_codes_->Resize(new_size_power_2);
--- a/src/algorithm/hgraph/hgraph_serialize.cpp
+++ b/src/algorithm/hgraph/hgraph_serialize.cpp
@@ -752,6 +752,7 @@ HGraph::read_streaming_body(StreamReader& reader,
     auto new_size = max_capacity_.load();
     this->neighbors_mutex_->Resize(new_size);
     pool_ = std::make_shared<VisitedListPool>(1, allocator_, new_size, allocator_);
+    pool_->SetMaxCachedObjectCountPerSubPool(MAX_CACHED_VISITED_LIST_COUNT_PER_SUB_POOL);
     this->total_count_ = this->basic_flatten_codes_->TotalCount();
     if (this->raw_vector_ != nullptr) {
         this->has_raw_vector_ = true;
@@ -786,6 +787,7 @@ HGraph::Deserialize(StreamReader& reader) {
         this->neighbors_mutex_->Resize(new_size);
 
         pool_ = std::make_shared<VisitedListPool>(1, allocator_, new_size, allocator_);
+        pool_->SetMaxCachedObjectCountPerSubPool(MAX_CACHED_VISITED_LIST_COUNT_PER_SUB_POOL);
 
         if (this->extra_info_size_ > 0 && this->extra_infos_ != nullptr) {
             this->extra_infos_->Deserialize(reader);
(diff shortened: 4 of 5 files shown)


src/utils/visited_list_test.cpp
```diff
@@ -121,3 +121,22 @@ TEST_CASE("VisitedListPool Basic Test", "[ut][VisitedListPool]") {
...
+    auto pool = std::make_shared<VisitedListPool>(0, allocator.get(), 1000, allocator.get());
+    pool->SetMaxCachedObjectCountPerSubPool(1);
+    auto initial_memory_usage = pool->GetMemoryUsage();
+
+    std::vector<std::shared_ptr<VisitedList>> visited_lists;
+    for (uint64_t i = 0; i < 3; ++i) {
+        visited_lists.emplace_back(pool->TakeOne());
+    }
+    REQUIRE(pool->GetMemoryUsage() > initial_memory_usage);
+    auto object_memory_usage = visited_lists.front()->GetMemoryUsage();
```
LHT129, line 136: [suggestion] The test assumes all VisitedList objects returned by TakeOne() have identical GetMemoryUsage() values. While this holds for VisitedList with the same size parameter, the final assertion would give a misleading failure if per-object memory ever diverged. Consider adding an explicit check that all taken objects report the same memory usage.

LHT129 (commented):
Overall this is a clean, well-scoped change that addresses the memory issue described in #2428. The approach of adding an optional cache limit to `ResourceObjectPool` and configuring HGraph to retain at most 1 idle visited list per sub-pool is sound.

**What works well:**
- The default (`uint64_t::max()`) preserves backward compatibility for all existing callers.
- All 4 HGraph pool-creation sites (`resize`, `read_streaming_body`, and both `Deserialize` paths) consistently apply the limit.
- The test covers the core behavior: excess objects are released and memory is correctly accounted.
- The lock is released before `obj.reset()` and `memory_usage_.fetch_sub()`, avoiding contention on the sub-pool mutex during destruction.

**Items already raised by other reviewers that I agree with:**
- `max_cached_object_count_per_sub_pool_` should be `std::atomic<uint64_t>` to prevent a data race if `SetMaxCachedObjectCountPerSubPool` is ever called concurrently with `ReturnOne`.
- `ReturnOne` should use `std::move(obj)` when emplacing into the pool so the caller always relinquishes ownership, making the side effects consistent between the cached and dropped paths.
- A null-pointer guard at the top of `ReturnOne` would be a reasonable defensive addition.

src/utils/visited_list_test.cpp
LHT129, line 136: [note] This test implicitly relies on the fact that in a single-threaded execution, `TakeOne()` always assigns objects to the same sub-pool (via `thread_local prefer_pool_id_`). With `max_cached_object_count_per_sub_pool = 1` and 16 sub-pools, if the 3 objects were distributed across different sub-pools, all 3 would be cached (1 per sub-pool) and the final assertion would fail. The test passes today because the pool is empty on creation and `try_lock` always succeeds on the first sub-pool, but this assumption is fragile — a future change to `TakeOne()` or a multi-threaded variant of this test could break silently.

Consider either:
- Explicitly verifying that all taken objects share the same `source_pool_id_`, or
- Using a pool with a single sub-pool (e.g., by testing `ResourceObjectPool` directly with `kSubPoolCount = 1`) to make the cache-limit semantics unambiguous.

LHT129 (commented):
Thanks for this PR. The change is well-scoped and addresses the memory bounding issue described in #2428.

I have reviewed the diff and the existing inline comments already cover the key points:

1. **`max_cached_object_count_per_sub_pool_` thread safety** — The plain `uint64_t` is read in `ReturnOne()` and written in `SetMaxCachedObjectCountPerSubPool()`. In the current usage, `SetMaxCachedObjectCountPerSubPool` is only called during initialization (under `global_mutex_` in `resize()`, or during deserialization before the pool is exposed to concurrent access), so there is no actual data race in practice. However, making it `std::atomic<uint64_t>` would be a defensive improvement if the setter is ever called at runtime in the future.

2. **`ReturnOne` ownership semantics** — When the object is cached, `emplace_back(obj)` copies the `shared_ptr`, leaving the caller with a live reference. Using `std::move(obj)` would make the ownership transfer consistent in both the cached and dropped paths.

3. **Test sub-pool distribution assumption** — The new test relies on all 3 objects landing in the same sub-pool, which holds in single-threaded execution due to `thread_local prefer_pool_id_`. This is fragile; an explicit check or a single-sub-pool test would be more robust.

The 4 call sites in `hgraph_build.cpp` and `hgraph_serialize.cpp` all correctly set the cache limit. The `#include <limits>` addition is appropriate. The `static constexpr` naming follows the existing `DEFAULT_RESIZE_BIT` convention.

No blocking issues from my side — the existing comments cover the areas worth addressing.

LHT129 (commented):
Thanks for this PR. The change to bound HGraph visited-list cache is well-scoped and addresses the memory issue described in #2428.

The existing inline comments from other reviewers have already covered the key concerns:

1. **Ownership consistency in `ReturnOne`**: Using `std::move` when emplacing into the sub-pool would make ownership transfer explicit and consistent between the cached and released paths.
2. **Thread safety of `max_cached_object_count_per_sub_pool_`**: Making this field `std::atomic<uint64_t>` would prevent a data race if `SetMaxCachedObjectCountPerSubPool` is ever called concurrently with `ReturnOne`.
3. **Test robustness**: The new test relies on `thread_local prefer_pool_id_` routing all objects to the same sub-pool, which is fragile. Consider verifying `source_pool_id_` explicitly or testing `ResourceObjectPool` directly with a single sub-pool.

The core logic — checking the sub-pool size against a configurable limit before caching, and releasing excess objects while tracking memory — looks correct. The 4 call sites in HGraph consistently apply the limit after pool creation.

No additional blocking issues from my side.

PR #2483, opened 2026-07-20
fix(ci): prevent PR lint from being silently skipped

### LINKED ISSUE #2482: PR CI can silently skip changed-file clang-tidy lint
The PR CI changed-file detection step can produce an empty lint pattern when its shell escaping command fails. Because the lint step is conditional on the pattern being non-empty, the job succeeds while changed C++ files are not linted. This was observed in PR #2437.

wxyucs:
## Summary

Fix changed-source detection in PR CI so clang-tidy cannot be silently skipped when the lint filter generation fails.

## Root cause

The previous `sed` escaping expression was malformed. Its failure was not propagated, leaving the output pattern empty and causing the conditional `Run Lint` step to be skipped.

## Changes

- Use Python `re.escape` to build the changed-file regex safely.
- Fail the job when `git diff` or filter generation fails.
- Fail the job if changed source files exist but the resulting filter is empty.

Validation: YAML parsing, shell syntax, regex generation, and `git diff --check` passed.

Fixes: #2482

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -96,13 +96,27 @@ jobs:
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
           git fetch origin ${{ github.base_ref }} --depth=1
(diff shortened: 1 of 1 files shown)


LHT129 (approved):
LGTM

PR #2486, opened 2026-07-20
Revert filter visit stage changes on v0.18

LHT129:
Revert #2394 (d9679079) and #2369 (feb5e4a5) to restore the original filter->CheckValid() visit condition on v0.18 branch.

These two commits replaced the OR-condition (ShouldSkipFilterCheck || CheckValid) with a single probability gate (ShouldVisit), causing valid neighbors to be skipped and recall to drop in high-filter scenarios.


wxyucs (approved):
lgtm

PR #2488, opened 2026-07-20
fix(ci): prevent PR lint from being silently skipped on 1.0

wxyucs:
## Summary

Backport the changed-source lint detection fix to the `1.0` branch.

## Root cause

The previous `sed` escaping expression was malformed. Its failure was not propagated, leaving the output pattern empty and causing the conditional `Run Lint` step to be skipped.

## Changes

- Use Python `re.escape` to build the changed-file regex safely.
- Fail the job when `git diff` or filter generation fails.
- Fail the job if changed source files exist but the resulting filter is empty.

Validation: YAML parsing, shell syntax, and `git diff --check` passed.

Fixes: #2482


LHT129 (approved):
LGTM

Issue #1087, opened 2025-08-26
HGraph RangeSearch support extra_info_filter


yulijunzj:

wxyucs:
Version-to-PR mapping for the `use_extra_info_filter` support in HGraph RangeSearch, for cross-checking this issue's version labels:

| Version label | PR that introduces the behavior |
| --- | --- |
| `version/0.16` | #2489 |
| `version/0.18` | #2490 |
| `version/1.0` | #2139 |
| `version/1.1` | Inherited from main through #2139; there is no separate 1.1 introduction PR. |

The 0.16 and 0.18 PRs are minimal branch-compatible backports. PR #2139 introduced the behavior on main through the shared HGraph search helper refactor.

wxyucs:
done

PR #2490, opened 2026-07-20
[0.18] fix(hgraph): support extra info filter in RangeSearch

wxyucs:
### Description

Backport the HGraph RangeSearch extra-info filter fix to the 0.18 branch. When `use_extra_info_filter` is enabled and extra-info storage is available, RangeSearch now wraps the caller filter with `ExtraInfoWrapperFilter` instead of always treating it as an ID filter.

This is the minimal branch-compatible equivalent of the behavior introduced on main by #2139.

### Changes

- Parse HGraph search parameters before constructing the filter wrapper.
- Select `ExtraInfoWrapperFilter` for extra-info filtering and keep `InnerIdWrapperFilter` as the fallback.
- Add a regression case whose ID predicate accepts all records while its extra-info predicate rejects all records.

### Testing

- Added focused regression coverage for the affected HGraph RangeSearch path.
- Ran `clang-format 15.0.7` on the changed C++ files.
- Ran `git diff --check`.
- Build and tests were not run: the current host is macOS, and the requested Linux container/network environment is not ready.

Fixes: #1087

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -1103,9 +1103,14 @@ HGraph::RangeSearch(const DatasetPtr& query,
     SearchStatistics stats;
     QueryContext ctx{.stats = &stats};
 
-    std::shared_ptr<InnerIdWrapperFilter> ft = nullptr;
+    auto params = HGraphSearchParameters::FromJson(parameters);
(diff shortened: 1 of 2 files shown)


LHT129 (approved):
LGTM

Issue #2319, opened 2026-06-23
Fix HGraph Range Search recall issue in two-stage search mode


LHT129:
## Problem

HGraph index supports two-stage search mode:
- First stage: uses low-precision quantizer for coarse filtering
- Second stage: uses high-precision quantizer for reordering

This design works well for KNN search, but has recall issues in Range Search scenarios.

### Root Cause

In `src/algorithm/hgraph/hgraph_search.cpp`, the `HGraph::RangeSearch` function:

1. **First stage filtering**: Uses low-precision quantizer with radius filtering in `basic_searcher.cpp`
2. **Second stage reordering**: Uses high-precision quantizer to recalculate distances
3. **Missing step**: After reordering, there is NO radius filtering with high-precision distances!

### Impact

Due to distance errors between low-precision and high-precision quantizers:
- **False positives**: Points with true distance > radius but low-precision distance <= radius are incorrectly kept
- **False negatives**: Points with true distance <= radius but low-precision distance > radius are incorrectly filtered

### Comparison with Other Indexes

`Pyramid::RangeSearch` correctly handles radius filtering after reordering by checking both topk and radius constraints.

## Proposed Solution

Add radius filtering after reordering in `HGraph::RangeSearch` to ensure only points within the radius (based on high-precision distances) are returned.

## Additional Context

This issue affects range search accuracy when using two-stage search with quantization. The fix is minimal and aligns with the existing implementation in Pyramid.

wxyucs:
Version-to-PR mapping for the HGraph RangeSearch rerank-radius fix, for cross-checking this issue's version labels:

| Version label | PR that introduces the fix |
| --- | --- |
| `version/0.16` | #2491 |
| `version/0.18` | #2492 |
| `version/1.0` | #2321 |
| `version/1.1` | Inherited from main through #2321; there is no separate 1.1 introduction PR. |

The 0.16 and 0.18 PRs are minimal branch-compatible backports. PR #2321 introduced the post-rerank radius filtering on main.

wxyucs:
done

PR #2492, opened 2026-07-20
[0.18] fix(hgraph): filter reranked RangeSearch results by radius

wxyucs:
### Description

Backport the HGraph RangeSearch rerank-radius fix from #2321 to the 0.18 branch. After high-precision reranking, results whose final distance exceeds `radius + THRESHOLD_ERROR` are now removed before applying `limited_size`.

### Changes

- Reapply the RangeSearch radius constraint after high-precision reorder.
- Preserve the existing threshold tolerance used by range search.
- Add a quantization-plus-FP32-rerank regression case that expects only the exact-radius result.

### Testing

- Added focused regression coverage for the affected HGraph RangeSearch path.
- Ran `clang-format 15.0.7` on the changed C++ files.
- Ran `git diff --check`.
- Build and tests were not run: the current host is macOS, and the requested Linux container/network environment is not ready.

Fixes: #2319

### DIFF
--- a/src/algorithm/hgraph.cpp
+++ b/src/algorithm/hgraph.cpp
@@ -1172,6 +1172,10 @@ HGraph::RangeSearch(const DatasetPtr& query,
             raw_query, this->high_precise_codes_, search_result, limited_size, nullptr, ctx);
     }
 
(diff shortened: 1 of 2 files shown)


LHT129 (approved):
LGTM

PR #2498, opened 2026-07-21
[bug](hgraph): include reader caches in memory usage

### LINKED ISSUE #2497: [bug](hgraph): include reader caches in memory usage
### Affected area

hgraph

### Interface

cpp

### VSAG version

efdaf17a10e96cdb5222baf558d50dfacbdc672e

### Summary

`HGraph::GetMemoryUsage()` significantly under-reports memory when an HGraph index is loaded
through `ReaderIO` and the retained `Reader` owns a resident page cache.

`ReaderIO` is provided by VSAG, while the concrete `Reader` is supplied through `ReaderSet`. The
index retains that reader, but the public `Reader` interface has no way to report reader-owned
memory. Memory released by `SetImmutable()` and memory added by the visited-list pool can also
leave the cached total stale.

### Minimal reproducer

```cpp
constexpr uint64_t cache_size = 64ULL * 1024ULL * 1024ULL;

auto allocator = std::make_shared<RecordingAllocator>();
auto resource = std::make_shared<vsag::Resource>(allocator, nullptr);
vsag::Engine engine(resource);

auto index = engine.CreateIndex("hgraph", hgraph_params).value();

vsag::ReaderSet readers;
// CachingReader reads the serialized index and owns a 64 MiB cache allocated
// through RecordingAllocator.
readers.Set(
    "hgraph",
    std::make_shared<CachingReader>(serialized_index, allocator, cache_size));

index->Deserialize(readers).value();
index->SetImmutable().value();

for (uint64_t i = 0; i < 3; ++i) {
    index->KnnSearch(query, 10, search_params).value();
}

std::cout << allocator->CurrentBytes() << '\n';
std::cout << index->GetMemoryUsage() << '\n';
```

Configuration:

- 1,000,000 vectors
- HGraph
- dimensions: 128
- base quantization: RaBitQ
- precise quantization: FP32
- `use_reorder=true`
- `SetImmutable()`
- three queries after loading
- reader, mixed, and block-memory IO configurations

### Environment

```shell
- OS: Ubuntu 22.04
- vsag version: v1.0.0-dirty
```

### Expected behavior

`GetMemoryUsage()` should include memory owned by a retained Reader, such as an in-memory page
cache, and should reflect memory changes caused by `SetImmutable()` and visited-list pool growth.

The underlying file or remote data source should not be counted as reader-owned memory.

### Actual behavior

For the 1M x 128 test:

| IO mode | Allocator current | GetMemoryUsage | Difference |
| --- | ---: | ---: | ---: |
| reader | 89.01 MiB | 23.47 MiB | 65.54 MiB |
| mixed | 251.02 MiB | 184.25 MiB | 66.77 MiB |
| block | 743.02 MiB | 674.25 MiB | 68.77 MiB |

The stable difference is dominated by the retained Reader's 64 MiB cache.

### References

- `include/vsag/readerset.h:132`
- `src/io/reader_io/reader_io.cpp`
- `src/algorithm/hgraph/hgraph.cpp:452`
- `src/algorithm/hgraph/hgraph.cpp:660`
- `src/algorithm/hgraph/hgraph_serialize.cpp:854`
- `docs/docs/en/src/advanced/memory.md`
- `docs/docs/zh/src/advanced/memory.md`

### Additional context

A local prototype adds an optional `Reader::GetMemoryUsage()` method whose default implementation
returns zero. HGraph dynamically adds the retained reader and visited-list pool usage, avoids
double-counting a shared precise reader, and refreshes its static memory snapshot after
`SetImmutable()`.

With a 64 MiB reader cache, the remaining allocator-versus-interface difference is 14.21-17.78
MiB across the three IO modes. This residual comes from allocation capacity, alignment,
hash-table buckets, and block slack rather than another large omitted component.

## Related issues

_None found._

---

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2497

## What Changed

- Add an optional, ABI-safe `MemoryUsageReader` subtype for resident reader-owned caches while
  keeping the existing `Reader` vtable unchanged.
- Detect the optional capability at runtime; existing readers report zero without requiring a
  rebuild.
- Include retained base and precise readers in HGraph total and detailed memory reporting without
  double-counting a shared reader.
- Read Reader and visited-list pool usage dynamically so cache and query-time pool growth are
  reflected immediately.
- Refresh HGraph's static memory snapshot after `SetImmutable()` releases neighbor mutex storage.
- Add regression coverage for both cache-reporting and ordinary readers, and update the English and
  Chinese memory documentation.

## Test Evidence

- [x] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build --target vsag functests -j2
# [100%] Built target vsag
# [100%] Built target functests

build/tests/functests "HGraph Reader Memory Usage"
# All tests passed (16 assertions in 1 test case)

clang-tidy-15 -p build src/reader.cpp \
  src/algorithm/hgraph/hgraph.cpp \
  src/algorithm/hgraph/hgraph_serialize.cpp --quiet
# No diagnostics

ABI checks:
- The `Reader` class definition hash matches `upstream/main` exactly.
- The built `Reader` vtable remains 0x40 bytes, matching the pre-change library.
- `MemoryUsageReader` exports separate vtable/typeinfo symbols.

1M x 128, RaBitQ2, FP32 reorder, 64 MiB reader cache, three queries:
- reader: allocator 114.92 MiB, GetMemoryUsage 100.71 MiB, difference 14.21 MiB
- mixed:  allocator 286.93 MiB, GetMemoryUsage 269.15 MiB, difference 17.77 MiB
- block:  allocator 776.93 MiB, GetMemoryUsage 759.15 MiB, difference 17.78 MiB

The reader cache and the additional 1.91 MiB query visited-list are included in each total.
```

## Compatibility Impact

- API/ABI compatibility: The existing public `Reader` interface and vtable are unchanged. Existing
  custom Reader binaries remain compatible and report zero reader-owned memory. Cache-aware readers
  can opt in by deriving from `MemoryUsageReader` and implementing `GetMemoryUsage()`.
- Behavior changes: HGraph memory usage now includes memory explicitly reported by retained
  `MemoryUsageReader` implementations and tracks query-time pool growth.

- Performance impact: Constant-time capability checks, visited-list pool access, and at most two
  reader calls are added to `GetMemoryUsage()`; search paths are unchanged.
- Concurrency/thread-safety impact: `MemoryUsageReader` implementations must make
  `GetMemoryUsage()` thread-safe. This requirement is documented.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/advanced/memory.md`

## Risk and Rollback

- Risk level: low
- Rollback plan: Revert the commits to restore cached HGraph-only accounting and remove the optional
  reader memory capability.

## Checklist

### DIFF
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -538,6 +538,7 @@ HGraph::SetImmutable() {
     this->parallel_searcher_->SetMutexArray(empty_mutex);
     this->neighbors_mutex_ = empty_mutex;
     this->immutable_.store(true, std::memory_order_release);
+    this->cal_memory_usage();
 }
 
 void
@@ -549,6 +550,8 @@ HGraph::SetIO(const std::shared_ptr<Reader> reader) {
     }
     basic_flatten_codes_->InitIO(reader_param);
     bottom_graph_->InitIO(reader_param);
+    this->reader_ = reader;
+    this->precise_reader_.reset();
 }
 
 void
@@ -556,6 +559,7 @@ HGraph::SetPreciseCodesIO(const std::shared_ptr<Reader>& reader) {
     auto reader_param = std::make_shared<ReaderIOParameter>();
     reader_param->reader = reader;
     high_precise_codes_->InitIO(reader_param);
+    this->precise_reader_ = reader;
 }
 
 const static uint64_t QUERY_SAMPLE_SIZE = 10;
@@ -745,11 +749,25 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
     this->attr_filter_index_->GetAttribute(0, inner_id, attr);
 }
 
+uint64_t
+HGraph::GetMemoryUsage() const {
+    auto memory = InnerIndexInterface::GetMemoryUsage();
+    if (this->pool_ != nullptr) {
+        memory += this->pool_->GetMemoryUsage();
+    }
+    if (this->reader_ != nullptr) {
+        memory += this->reader_->GetMemoryUsage();
+    }
+    if (this->precise_reader_ != nullptr && this->precise_reader_.get() != this->reader_.get()) {
+        memory += this->precise_reader_->GetMemoryUsage();
+    }
+    return memory;
+}
+
 void
 HGraph::cal_memory_usage() {
     auto memory = sizeof(HGraph);
     memory += this->neighbors_mutex_->GetMemoryUsage();
-    memory += this->pool_->GetMemoryUsage();
     memory += this->label_table_->GetMemoryUsage();
     memory += this->basic_flatten_codes_->GetMemoryUsage();
     if (this->code_slot_map_ != nullptr) {
@@ -770,7 +788,6 @@ HGraph::cal_memory_usage() {
     if (this->create_new_raw_vector_ and this->raw_vector_ != nullptr) {
         memory += raw_vector_->GetMemoryUsage();
     }
-
     std::unique_lock lock(this->memory_usage_mutex_);
     this->current_memory_usage_.store(memory);
 }
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -816,16 +816,25 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
 }
 
 uint64_t
+HGraph::get_reader_memory_usage(const ReaderPtr& reader) {
+    const auto* memory_usage_reader = dynamic_cast<const MemoryUsageReader*>(reader.get());
+    if (memory_usage_reader == nullptr) {
+        return 0;
+    }
+    return memory_usage_reader->GetMemoryUsage();
+}
+
+uint64_t
 HGraph::GetMemoryUsage() const {
     auto memory = InnerIndexInterface::GetMemoryUsage();
     if (this->pool_ != nullptr) {
         memory += this->pool_->GetMemoryUsage();
     }
     if (this->reader_ != nullptr) {
-        memory += this->reader_->GetMemoryUsage();
+        memory += get_reader_memory_usage(this->reader_);
     }
     if (this->precise_reader_ != nullptr && this->precise_reader_.get() != this->reader_.get()) {
-        memory += this->precise_reader_->GetMemoryUsage();
+        memory += get_reader_memory_usage(this->precise_reader_);
     }
     return memory;
 }
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -590,6 +590,7 @@ HGraph::SetImmutable() {
     this->parallel_searcher_->SetMutexArray(empty_mutex);
     this->neighbors_mutex_ = empty_mutex;
     this->immutable_.store(true, std::memory_order_release);
+    this->cal_memory_usage();
 }
 
 void
@@ -601,6 +602,8 @@ HGraph::SetIO(const std::shared_ptr<Reader> reader) {
     }
     basic_flatten_codes_->InitIO(reader_param);
     bottom_graph_->InitIO(reader_param);
+    this->reader_ = reader;
+    this->precise_reader_.reset();
 }
 
 void
@@ -608,6 +611,7 @@ HGraph::SetPreciseCodesIO(const std::shared_ptr<Reader>& reader) {
     auto reader_param = std::make_shared<ReaderIOParameter>();
     reader_param->reader = reader;
     high_precise_codes_->InitIO(reader_param);
+    this->precise_reader_ = reader;
 }
 
 const static uint64_t QUERY_SAMPLE_SIZE = 10;
@@ -811,11 +815,25 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
     this->attr_filter_index_->GetAttribute(0, inner_id, attr);
 }
 
+uint64_t
+HGraph::GetMemoryUsage() const {
+    auto memory = InnerIndexInterface::GetMemoryUsage();
+    if (this->pool_ != nullptr) {
+        memory += this->pool_->GetMemoryUsage();
+    }
+    if (this->reader_ != nullptr) {
+        memory += this->reader_->GetMemoryUsage();
+    }
+    if (this->precise_reader_ != nullptr && this->precise_reader_.get() != this->reader_.get()) {
+        memory += this->precise_reader_->GetMemoryUsage();
+    }
+    return memory;
+}
+
 void
 HGraph::cal_memory_usage() {
     auto memory = sizeof(HGraph);
     memory += this->neighbors_mutex_->GetMemoryUsage();
-    memory += this->pool_->GetMemoryUsage();
     memory += this->label_table_->GetMemoryUsage();
     memory += this->basic_flatten_codes_->GetMemoryUsage();
     if (this->code_slot_map_ != nullptr) {
--- a/tests/test_hgraph.cpp
+++ b/tests/test_hgraph.cpp
@@ -2312,6 +2312,67 @@ TestHGraphReaderIO(const fixtures::HGraphTestIndexPtr& test_index,
 
 HGRAPH_PR_DAILY_CASE("HGraph Reader IO", "[ft][serialize][hgraph]", TestHGraphReaderIO)
 
+TEST_CASE("HGraph Reader Memory Usage", "[ft][memory][hgraph]") {
+    constexpr uint64_t reader_memory = 64ULL * 1024ULL * 1024ULL;
+    constexpr int64_t dim = 16;
+    constexpr int64_t count = 200;
+    const auto build_param = R"({
+        "dtype": "float32",
+        "metric_type": "l2",
+        "dim": 16,
+        "index_param": {
+            "base_quantization_type": "sq8",
+            "precise_quantization_type": "fp32",
+            "use_reorder": true,
+            "max_degree": 16,
+            "ef_construction": 100
+        }
+    })";
+    const auto reader_param = R"({
+        "dtype": "float32",
+        "metric_type": "l2",
+        "dim": 16,
+        "index_param": {
+            "base_quantization_type": "sq8",
+            "precise_quantization_type": "fp32",
+            "use_reorder": true,
+            "max_degree": 16,
+            "ef_construction": 100,
+            "base_io_type": "reader_io",
+            "precise_io_type": "reader_io",
+            "graph_io_type": "reader_io"
+        }
+    })";
+
+    auto dataset = fixtures::HGraphTestIndex::pool.GetDatasetAndCreate(dim, count, "l2");
+    auto source = fixtures::TestIndex::TestFactory("hgraph", build_param, true);
+    fixtures::TestIndex::TestBuildIndex(source, dataset, true);
+    auto binary_set = source->Serialize().value();
+
+    auto restored = fixtures::TestIndex::TestFactory("hgraph", reader_param, true);
+    vsag::ReaderSet readers;
+    readers.Set("hgraph",
+                std::make_shared<fixtures::TestReader>(binary_set.Get("hgraph"), reader_memory));
+    REQUIRE(restored->Deserialize(readers).has_value());
+
+    auto detail = restored->GetMemoryUsageDetail();
+    REQUIRE(detail.at("reader") == reader_memory);
+    auto mutable_memory = restored->GetMemoryUsage();
+    REQUIRE(restored->SetImmutable().has_value());
+    auto immutable_memory = restored->GetMemoryUsage();
+    REQUIRE(immutable_memory < mutable_memory);
+
+    detail = restored->GetMemoryUsageDetail();
+    REQUIRE(detail.at("neighbors_mutex") == 0);
+    REQUIRE(detail.at("reader") == reader_memory);
+    uint64_t detail_sum = 0;
+    for (const auto& [_, memory] : detail) {
+        detail_sum += memory;
+    }
+    REQUIRE(immutable_memory >= detail_sum);
+    REQUIRE(immutable_memory - detail_sum < 4096);
+}
+
 static void
 TestHGraphClone(const fixtures::HGraphTestIndexPtr& test_index,
                 const fixtures::HGraphResourcePtr& resource) {
--- a/tests/test_hgraph.cpp
+++ b/tests/test_hgraph.cpp
@@ -2415,6 +2415,67 @@ TestHGraphReaderIO(const fixtures::HGraphTestIndexPtr& test_index,
 
 HGRAPH_PR_DAILY_CASE("HGraph Reader IO", "[ft][serialize][hgraph]", TestHGraphReaderIO)
 
+TEST_CASE("HGraph Reader Memory Usage", "[ft][memory][hgraph]") {
+    constexpr uint64_t reader_memory = 64ULL * 1024ULL * 1024ULL;
+    constexpr int64_t dim = 16;
+    constexpr int64_t count = 200;
+    const auto build_param = R"({
+        "dtype": "float32",
+        "metric_type": "l2",
+        "dim": 16,
+        "index_param": {
+            "base_quantization_type": "sq8",
+            "precise_quantization_type": "fp32",
+            "use_reorder": true,
+            "max_degree": 16,
+            "ef_construction": 100
+        }
+    })";
+    const auto reader_param = R"({
+        "dtype": "float32",
+        "metric_type": "l2",
+        "dim": 16,
+        "index_param": {
+            "base_quantization_type": "sq8",
+            "precise_quantization_type": "fp32",
+            "use_reorder": true,
+            "max_degree": 16,
+            "ef_construction": 100,
+            "base_io_type": "reader_io",
+            "precise_io_type": "reader_io",
+            "graph_io_type": "reader_io"
+        }
+    })";
+
+    auto dataset = fixtures::HGraphTestIndex::pool.GetDatasetAndCreate(dim, count, "l2");
+    auto source = fixtures::TestIndex::TestFactory("hgraph", build_param, true);
+    fixtures::TestIndex::TestBuildIndex(source, dataset, true);
+    auto binary_set = source->Serialize().value();
+
+    auto restored = fixtures::TestIndex::TestFactory("hgraph", reader_param, true);
+    vsag::ReaderSet readers;
+    readers.Set("hgraph",
+                std::make_shared<fixtures::TestReader>(binary_set.Get("hgraph"), reader_memory));
+    REQUIRE(restored->Deserialize(readers).has_value());
+
+    auto detail = restored->GetMemoryUsageDetail();
+    REQUIRE(detail.at("reader") == reader_memory);
+    auto mutable_memory = restored->GetMemoryUsage();
+    REQUIRE(restored->SetImmutable().has_value());
+    auto immutable_memory = restored->GetMemoryUsage();
+    REQUIRE(immutable_memory < mutable_memory);
+
+    detail = restored->GetMemoryUsageDetail();
+    REQUIRE(detail.at("neighbors_mutex") == 0);
+    REQUIRE(detail.at("reader") == reader_memory);
+    uint64_t detail_sum = 0;
+    for (const auto& [_, memory] : detail) {
+        detail_sum += memory;
+    }
+    REQUIRE(immutable_memory >= detail_sum);
+    REQUIRE(immutable_memory - detail_sum < 4096);
+}
+
 static void
 TestHGraphClone(const fixtures::HGraphTestIndexPtr& test_index,
                 const fixtures::HGraphResourcePtr& resource) {
--- a/tests/test_hgraph.cpp
+++ b/tests/test_hgraph.cpp
@@ -31,6 +31,7 @@
 #include "storage/streaming_serialization_test_utils.h"
 #include "test_index.h"
 #include "typing.h"
+#include "vsag/factory.h"
 #include "vsag/filter.h"
 #include "vsag/options.h"
 #include "vsag/search_request.h"
@@ -2451,11 +2452,11 @@ TEST_CASE("HGraph Reader Memory Usage", "[ft][memory][hgraph]") {
     auto source = fixtures::TestIndex::TestFactory("hgraph", build_param, true);
     fixtures::TestIndex::TestBuildIndex(source, dataset, true);
     auto binary_set = source->Serialize().value();
+    auto binary = binary_set.Get("hgraph");
 
     auto restored = fixtures::TestIndex::TestFactory("hgraph", reader_param, true);
     vsag::ReaderSet readers;
-    readers.Set("hgraph",
-                std::make_shared<fixtures::TestReader>(binary_set.Get("hgraph"), reader_memory));
+    readers.Set("hgraph", std::make_shared<fixtures::TestReader>(binary, reader_memory));
     REQUIRE(restored->Deserialize(readers).has_value());
 
     auto detail = restored->GetMemoryUsageDetail();
@@ -2474,6 +2475,18 @@ TEST_CASE("HGraph Reader Memory Usage", "[ft][memory][hgraph]") {
     }
     REQUIRE(immutable_memory >= detail_sum);
     REQUIRE(immutable_memory - detail_sum < 4096);
+
+    auto plain_restored = fixtures::TestIndex::TestFactory("hgraph", reader_param, true);
+    auto plain_reader = vsag::Factory::CreateReadFuncReader(
+        [binary](uint64_t offset, uint64_t len, void* dest) {
+            std::memcpy(dest, binary.data.get() + offset, len);
+        },
+        binary.size);
+    vsag::ReaderSet plain_readers;
+    plain_readers.Set("hgraph", plain_reader);
+    REQUIRE(plain_restored->Deserialize(plain_readers).has_value());
+    REQUIRE(plain_restored->GetMemoryUsage() > 0);
+    REQUIRE(plain_restored->GetMemoryUsageDetail().at("reader") == 0);
 }
 
 static void
--- a/docs/docs/en/src/advanced/memory.md
+++ b/docs/docs/en/src/advanced/memory.md
@@ -97,6 +97,10 @@ Properties:
   allocator overhead, scratch buffers, and any data held outside the index (e.g. user-owned input
   vectors). For SINDI in particular, call `GetMemoryUsage()` **after** the build completes to get
   a representative value.
+- For an HGraph loaded through `ReaderIO`, resident memory owned by a retained `Reader` is included
+  when the reader overrides `Reader::GetMemoryUsage()`. The default implementation returns zero;
+  custom readers with an in-memory page cache should override it. The underlying file or remote
+  data source is not part of the reported memory.
 
 See `examples/cpp/319_feature_get_memory_usage.cpp` for a runnable example, including a helper
 that compares the interface value with the process resident size.
@@ -118,8 +122,9 @@ understanding *where* the memory is going inside an index.
 
 Currently only HGraph provides a meaningful implementation, returning components such as
 `basic_flatten_codes`, `bottom_graph`, `route_graph`, `neighbors_mutex`, `pool`,
-`label_table`, `high_precise_codes`, `extra_infos`, and `raw_vector`. SINDI returns an empty
-map. Other index types throw an exception by default.
+`label_table`, `high_precise_codes`, `extra_infos`, `raw_vector`, `reader`, and
+`precise_reader`. Reader keys are present only when the corresponding retained reader exists.
+SINDI returns an empty map. Other index types throw an exception by default.
 
 ### Capability Flags
 
@@ -140,5 +145,6 @@ full control over parallelism and resource ownership. See
 ## Notes
 
 - A custom allocator must be thread-safe.
+- A custom `Reader::GetMemoryUsage()` override must be thread-safe.
 - The allocator's lifetime must outlive any index and result object referencing it.
 - If nothing is configured, VSAG falls back to a default `malloc`-based allocator.
--- a/docs/docs/en/src/advanced/memory.md
+++ b/docs/docs/en/src/advanced/memory.md
@@ -98,9 +98,9 @@ Properties:
   vectors). For SINDI in particular, call `GetMemoryUsage()` **after** the build completes to get
   a representative value.
 - For an HGraph loaded through `ReaderIO`, resident memory owned by a retained `Reader` is included
-  when the reader overrides `Reader::GetMemoryUsage()`. The default implementation returns zero;
-  custom readers with an in-memory page cache should override it. The underlying file or remote
-  data source is not part of the reported memory.
+  when the reader derives from `MemoryUsageReader` and implements `GetMemoryUsage()`. Readers that
+  do not derive from `MemoryUsageReader` report zero. The underlying file or remote data source is
+  not part of the reported memory.
 
 See `examples/cpp/319_feature_get_memory_usage.cpp` for a runnable example, including a helper
 that compares the interface value with the process resident size.
@@ -145,6 +145,6 @@ full control over parallelism and resource ownership. See
 ## Notes
 
 - A custom allocator must be thread-safe.
-- A custom `Reader::GetMemoryUsage()` override must be thread-safe.
+- A custom `MemoryUsageReader::GetMemoryUsage()` implementation must be thread-safe.
 - The allocator's lifetime must outlive any index and result object referencing it.
 - If nothing is configured, VSAG falls back to a default `malloc`-based allocator.
--- a/docs/docs/zh/src/advanced/memory.md
+++ b/docs/docs/zh/src/advanced/memory.md
@@ -88,6 +88,9 @@ uint64_t bytes = index->GetMemoryUsage();
   RSS 还包含 allocator 的开销、临时 scratch buffer、以及索引外部持有的数据（例如用户自有的输入
   向量缓冲）。SINDI 索引尤其建议在构建完成**之后**调用 `GetMemoryUsage()` 才能拿到具有代表性的
   数值。
+- HGraph 通过 `ReaderIO` 加载时，如果被索引持有的 `Reader` 重写了
+  `Reader::GetMemoryUsage()`，其常驻内存会计入统计。该方法默认返回 0；带内存页缓存的自定义
+  Reader 应重写此方法。底层文件或远端数据源本身不计入内存统计。
 
 可运行示例：`examples/cpp/319_feature_get_memory_usage.cpp`，其中包含一个辅助函数将接口值与进程
 驻留内存进行对照。
@@ -107,7 +110,8 @@ for (const auto& [component, bytes] : detail) {
 
 目前仅 HGraph 提供了有效实现，返回的组件包括 `basic_flatten_codes`、`bottom_graph`、
 `route_graph`、`neighbors_mutex`、`pool`、`label_table`、`high_precise_codes`、
-`extra_infos` 和 `raw_vector`。SINDI 返回空 map，其他索引类型默认抛出异常。
+`extra_infos`、`raw_vector`、`reader` 和 `precise_reader`。仅在索引持有对应 Reader 时才会
+返回 Reader 相关项。SINDI 返回空 map，其他索引类型默认抛出异常。
 
 ### 能力标志
 
@@ -127,5 +131,6 @@ for (const auto& [component, bytes] : detail) {
 ## 注意事项
 
 - 自定义 Allocator 必须是线程安全的。
+- 自定义 `Reader::GetMemoryUsage()` 实现必须是线程安全的。
 - `Allocator` 生命周期必须覆盖所有引用它的索引与结果对象。
 - 若未显式指定，VSAG 会创建一个默认的基于 `malloc` 的 allocator。
--- a/docs/docs/zh/src/advanced/memory.md
+++ b/docs/docs/zh/src/advanced/memory.md
@@ -88,9 +88,9 @@ uint64_t bytes = index->GetMemoryUsage();
   RSS 还包含 allocator 的开销、临时 scratch buffer、以及索引外部持有的数据（例如用户自有的输入
   向量缓冲）。SINDI 索引尤其建议在构建完成**之后**调用 `GetMemoryUsage()` 才能拿到具有代表性的
   数值。
-- HGraph 通过 `ReaderIO` 加载时，如果被索引持有的 `Reader` 重写了
-  `Reader::GetMemoryUsage()`，其常驻内存会计入统计。该方法默认返回 0；带内存页缓存的自定义
-  Reader 应重写此方法。底层文件或远端数据源本身不计入内存统计。
+- HGraph 通过 `ReaderIO` 加载时，如果被索引持有的 Reader 继承 `MemoryUsageReader` 并实现
+  `GetMemoryUsage()`，其常驻内存会计入统计。未继承 `MemoryUsageReader` 的 Reader 实现按 0
+  统计；底层文件或远端数据源本身不计入内存统计。
 
 可运行示例：`examples/cpp/319_feature_get_memory_usage.cpp`，其中包含一个辅助函数将接口值与进程
 驻留内存进行对照。
@@ -131,6 +131,6 @@ for (const auto& [component, bytes] : detail) {
 ## 注意事项
 
 - 自定义 Allocator 必须是线程安全的。
-- 自定义 `Reader::GetMemoryUsage()` 实现必须是线程安全的。
+- 自定义 `MemoryUsageReader::GetMemoryUsage()` 实现必须是线程安全的。
 - `Allocator` 生命周期必须覆盖所有引用它的索引与结果对象。
 - 若未显式指定，VSAG 会创建一个默认的基于 `malloc` 的 allocator。
--- a/include/vsag/readerset.h
+++ b/include/vsag/readerset.h
@@ -131,6 +131,20 @@ public:
      */
     [[nodiscard]] virtual uint64_t
     Size() const = 0;
+
+    /**
+     * @brief Returns memory owned by the reader, excluding the underlying data source.
+     *
+     * Reader implementations with resident caches should override this method so indexes that
+     * retain the reader can include that memory in their usage statistics. Overrides must be
+     * thread-safe because the method may be called concurrently with reads.
+     *
+     * @return Resident memory usage in bytes. The default implementation returns zero.
+     */
+    [[nodiscard]] virtual uint64_t
+    GetMemoryUsage() const {
+        return 0;
+    }
 };
 
 /**
--- a/include/vsag/readerset.h
+++ b/include/vsag/readerset.h
@@ -131,20 +131,27 @@ public:
      */
     [[nodiscard]] virtual uint64_t
     Size() const = 0;
+};
+
+/**
+ * @class MemoryUsageReader
+ * @brief Optional Reader extension for reporting reader-owned resident memory.
+ *
+ * Reader implementations with resident caches can derive from this class so indexes that retain
+ * the reader can include that memory in their usage statistics. Implementations must be
+ * thread-safe because GetMemoryUsage() may be called concurrently with reads.
+ */
+class MemoryUsageReader : public Reader {
+public:
+    ~MemoryUsageReader() override;
 
     /**
      * @brief Returns memory owned by the reader, excluding the underlying data source.
      *
-     * Reader implementations with resident caches should override this method so indexes that
-     * retain the reader can include that memory in their usage statistics. Overrides must be
-     * thread-safe because the method may be called concurrently with reads.
-     *
-     * @return Resident memory usage in bytes. The default implementation returns zero.
+     * @return Resident memory usage in bytes.
      */
     [[nodiscard]] virtual uint64_t
-    GetMemoryUsage() const {
-        return 0;
-    }
+    GetMemoryUsage() const = 0;
 };
 
 /**
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -109,6 +109,9 @@ public:
     uint64_t
     EstimateMemory(uint64_t num_elements) const override;
 
+    [[nodiscard]] uint64_t
+    GetMemoryUsage() const override;
+
     void
     GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const override;
 
@@ -791,6 +794,9 @@ private:
     std::atomic<int64_t> delete_count_{0};  // number of force-removed vectors
 
     std::shared_ptr<Optimizer<BasicSearcher>> optimizer_;  // search parameter optimizer
+
+    ReaderPtr reader_{nullptr};          // shared reader used by base, graph, and precise codes
+    ReaderPtr precise_reader_{nullptr};  // optional dedicated reader for precise codes
 
     bool create_new_raw_vector_{false};        // whether a separate raw vector exists
     FlattenInterfacePtr raw_vector_{nullptr};  // raw float vectors (for distance calc)
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -659,6 +659,10 @@ private:
     void
     cal_memory_usage();
 
+    /// Return resident memory reported by an optional MemoryUsageReader implementation.
+    [[nodiscard]] static uint64_t
+    get_reader_memory_usage(const ReaderPtr& reader);
+
     /// True when reorder uses a separate high-precision flatten (not base codes).
     [[nodiscard]] bool
     has_precise_reorder() const {
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -122,6 +122,9 @@ public:
     uint64_t
     EstimateMemory(uint64_t num_elements) const override;
 
+    [[nodiscard]] uint64_t
+    GetMemoryUsage() const override;
+
     void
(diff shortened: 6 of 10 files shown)


LHT129 (commented):
Thanks for this PR. The change is well-structured and correctly addresses the memory accounting gap for reader-owned caches.

**Summary of the approach:**
- Adds a default-zero `Reader::GetMemoryUsage()` virtual method so custom readers with resident caches can report their memory.
- HGraph stores `reader_` and `precise_reader_` pointers during `SetIO`/`SetPreciseCodesIO` and includes them in both `GetMemoryUsage()` (total) and `GetMemoryUsageDetail()` (breakdown).
- Avoids double-counting when the same reader instance is shared via raw-pointer comparison.
- Moves `pool_->GetMemoryUsage()` from the static `cal_memory_usage()` snapshot into the dynamic `GetMemoryUsage()` override so query-time pool growth is reflected immediately.
- Refreshes the static snapshot after `SetImmutable()` releases neighbor mutex storage.
- Includes regression tests and documentation updates in both English and Chinese.

**Existing review comments on this PR already cover the key points:** null-safety in `SetPreciseCodesIO`, ABI compatibility concerns with the new virtual method, documentation accuracy, and test improvements (magic numbers, duplicated constants, missing `@param`). I will not duplicate those here.

The core logic is sound. The test coverage for the distinct-reader path (where `SetPreciseCodesIO` is called with a different reader than `SetIO`) could be expanded, but that path is exercised in production through the streaming deserialization code path.

LHT129 (commented):
Overall this is a well-structured PR that correctly adds reader cache memory to HGraph memory usage reporting. The changes are focused and the test coverage is good.

A few observations beyond what other reviewers have already noted:

1. **Lock consistency in `GetMemoryUsage()`**: The override calls `InnerIndexInterface::GetMemoryUsage()` (which internally acquires `memory_usage_mutex_`), then adds dynamic pool and reader values outside the lock. This means the returned total is a non-atomic sum sampled at different times. This is acceptable for approximate memory accounting, but worth being aware of — the cached `current_memory_usage_` could reflect a pre-`SetImmutable()` snapshot while the pool already shows post-immutable shrinkage.

2. **`SetIO()` resets `precise_reader_`**: This is intentional (shared reader covers all components), but callers should be aware that calling `SetIO()` after `SetPreciseCodesIO()` silently drops the dedicated precise reader reference. Consider whether this warrants a warning or documentation.

3. The double-counting guard (`precise_reader_.get() != reader_.get()`) is correct and consistent between `GetMemoryUsage()` and `GetMemoryUsageDetail()`.

The existing bot comments about ABI compatibility, null checks, documentation accuracy, and test magic numbers are all valid and should be addressed.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -538,6 +538,7 @@ HGraph::SetImmutable() {
     this->parallel_searcher_->SetMutexArray(empty_mutex);
```
LHT129, line 590: [note] When `SetIO()` is called after `SetPreciseCodesIO()`, `this->precise_reader_.reset()` silently drops the dedicated precise reader reference.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -538,6 +538,7 @@ HGraph::SetImmutable() {
...
     this->neighbors_mutex_ = empty_mutex;
     this->immutable_.store(true, std::memory_order_release);
+    this->cal_memory_usage();
 }
 
 void
```
LHT129, line 596: [note] When `SetIO()` is called after `SetPreciseCodesIO()`, `this->precise_reader_.reset()` silently drops the dedicated precise reader reference. This is intentional since the shared reader now covers all components, but the silent discard could surprise callers who expect the precise reader to remain accessible for memory accounting. Consider documenting this interaction or adding a debug log.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -549,13 +550,16 @@ HGraph::SetIO(const std::shared_ptr<Reader> reader) {
     }
     basic_flatten_codes_->InitIO(reader_param);
     bottom_graph_->InitIO(reader_param);
+    this->reader_ = reader;
+    this->precise_reader_.reset();
```
LHT129, line 606: [note] When `SetIO()` is called after `SetPreciseCodesIO()`, `this->precise_reader_.reset()` silently drops the dedicated precise reader reference. This is intentional since the shared reader now covers all components, but the silent discard could surprise callers who expect the precise reader to remain accessible for memory accounting. Consider documenting this interaction or adding a debug log.

LHT129 (commented):
## Review Summary

This PR adds `Reader::GetMemoryUsage()` to include reader-owned caches in HGraph memory reporting. Overall the implementation is clean and well-tested. The key changes are:

1. **`Reader::GetMemoryUsage()`** (`include/vsag/readerset.h`): A new virtual method with a default zero-returning implementation. Reader implementations with resident caches should override this.

2. **`HGraph::GetMemoryUsage()` override** (`src/algorithm/hgraph/hgraph.cpp`): Dynamically adds pool, reader, and precise_reader memory on top of the cached `current_memory_usage_`. The pool was intentionally removed from `cal_memory_usage()` so it is tracked dynamically, reflecting query-time growth.

3. **`HGraph::GetMemoryUsageDetail()`** (`src/algorithm/hgraph/hgraph_serialize.cpp`): Includes `reader` and `precise_reader` entries when the corresponding retained readers exist, with proper deduplication when both point to the same instance.

4. **`SetImmutable()` refresh** (`src/algorithm/hgraph/hgraph.cpp`): Calls `cal_memory_usage()` after releasing neighbor mutex storage so the snapshot reflects the post-immutable state.

5. **Tests** (`tests/test_hgraph.cpp`): A new `HGraph Reader Memory Usage` test case validates reader memory inclusion, immutable memory reduction, and detail-vs-total consistency.

The existing inline comments from other reviewers cover the notable items (ABI/vtable concern, documentation precision, magic number in tests, null-check in `SetPreciseCodesIO`, and the `precise_reader_.reset()` behavior in `SetIO`). I have no additional blocking concerns to add.

LHT129 (commented):
I have reviewed this PR and the changes look good overall. The existing inline comments from other reviewers already cover the notable items:

- ABI/vtable compatibility concern with the new virtual method on `Reader`
- Null-safety in `SetPreciseCodesIO` for `high_precise_codes_`
- Documentation precision around `precise_reader` double-counting avoidance
- Test improvements (magic number `4096`, duplicated `dim` constant, missing `@param` doc)
- `precise_reader_.reset()` behavior in `SetIO()`

The core logic is correct: `GetMemoryUsage()` dynamically adds pool and reader memory on top of the cached snapshot, `GetMemoryUsageDetail()` includes reader entries with proper deduplication, and `SetImmutable()` refreshes the snapshot after releasing mutex storage. The test coverage is adequate for the happy path.

No additional blocking concerns from my side.

LHT129 (commented):
## Code Review for PR #2498

### Overall Assessment

This PR adds an optional `MemoryUsageReader` subtype to include reader-owned cache memory in HGraph memory usage reporting, with ABI-safe design and good test coverage. The approach is well thought out. A few suggestions below.

---

**[suggestion] `GetMemoryUsage()` reads `pool_` and `reader_` without synchronization**

In `src/algorithm/hgraph/hgraph.cpp`, the new `GetMemoryUsage()` override reads `pool_->GetMemoryUsage()` and `reader_`/`precise_reader_` without any lock, while the base class `InnerIndexInterface::GetMemoryUsage()` is protected by `memory_usage_mutex_`. The `pool_` is accessed concurrently by search threads (visited-list allocation/release), and `GetMemoryUsage()` is documented as callable concurrently with reads. The `VisitedListPool::GetMemoryUsage()` implementation may not be thread-safe when called concurrently with allocations.

Consider either documenting that results may be approximate under concurrency, or using `memory_usage_mutex_` to guard the dynamic reads.

**[suggestion] `cal_memory_usage()` snapshot no longer includes pool memory, but `GetMemoryUsage()` does — potential confusion**

`cal_memory_usage()` was changed to remove `pool_->GetMemoryUsage()` from the static snapshot stored in `current_memory_usage_`. Since HGraph now overrides `GetMemoryUsage()` to add pool dynamically, this is functionally correct. However, any code path that reads `current_memory_usage_` directly (e.g. through the base class accessor if called without virtual dispatch) would now miss pool memory. This is unlikely in practice but worth noting.

**[suggestion] `SetIO` resets `precise_reader_` but does not reset `high_precise_codes_` IO state**

When `SetIO()` is called after `SetPreciseCodesIO()`, `precise_reader_` is reset but `high_precise_codes_` still holds the old reader-based IO. If the old precise reader is destroyed, subsequent precise code reads would access freed memory. This is only reachable through unusual API usage, but a defensive reset of `high_precise_codes_` IO or at least a warning log would make the contract clearer.

LHT129 (commented):
Thanks for this PR. The `MemoryUsageReader` approach is a clean way to add reader memory accounting while preserving ABI compatibility of the existing `Reader` vtable — the separate subclass with `dynamic_cast` detection at runtime is the right pattern.

The implementation correctly:
- Avoids double-counting when the same reader is shared between base and precise codes (via pointer comparison)
- Refreshes the static memory snapshot after `SetImmutable()` releases neighbor mutex storage
- Tracks pool and reader memory dynamically so cache/query-time growth is reflected immediately
- Includes comprehensive test coverage for both `MemoryUsageReader` and plain `Reader` paths

No new blocking issues found beyond what has already been flagged by other reviewers.

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured and the approach of using an optional `MemoryUsageReader` subtype with `dynamic_cast` detection is a clean way to add reader memory accounting without breaking the existing `Reader` ABI.

1. **`include/vsag/readerset.h`**: New `MemoryUsageReader` class with out-of-line destructor (vtable stays in `.cpp`, preserving ABI compatibility for existing `Reader` subclasses). Clean design.

2. **`src/algorithm/hgraph/hgraph.cpp`**: 
   - `SetIO` / `SetPreciseCodesIO` now retain reader pointers for later memory queries.
   - `SetImmutable` refreshes the static memory snapshot after releasing neighbor mutex storage.
   - `GetMemoryUsage()` override dynamically includes pool and reader memory on top of the cached static snapshot — correct approach since both can grow at query time.
   - `cal_memory_usage()` correctly drops the static `pool_` accounting since it is now handled dynamically.
   - `get_reader_memory_usage()` uses `dynamic_cast` for zero-overhead capability detection on non-`MemoryUsageReader` readers.

3. **`src/algorithm/hgraph/hgraph_serialize.cpp`**: `GetMemoryUsageDetail()` includes `reader` and `precise_reader` entries with proper dedup when both point to the same instance.

4. **Tests**: Good coverage — verifies memory reporting with a `MemoryUsageReader`, verifies zero reporting with a plain `CreateReadFuncReader`, and validates the `SetImmutable` transition reduces memory.

5. **Docs**: English and Chinese memory docs updated consistently.

**No blocking issues found.** The existing inline comments from other reviewers have already addressed the minor documentation and defensive-coding suggestions. The core logic is correct, the ABI compatibility approach is sound, and the test coverage is adequate.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
...
+    if (this->precise_reader_ != nullptr && this->precise_reader_.get() != this->reader_.get()) {
+        memory += get_reader_memory_usage(this->precise_reader_);
+    }
+    return memory;
+}
+
 void
 HGraph::cal_memory_usage() {
     auto memory = sizeof(HGraph);
     memory += this->neighbors_mutex_->GetMemoryUsage();
```
LHT129, line 845: [note] `cal_memory_usage()` no longer includes `pool_->GetMemoryUsage()` (moved to the overridden `GetMemoryUsage()`), but `cal_memory_usage()` is still called from `SetImmutable()` to refresh the cached `current_memory_usage_` snapshot. This means the cached snapshot stored in the base class (`InnerIndexInterface::current_memory_usage_`) will be smaller than what `HGraph::GetMemoryUsage()` returns, since it lacks the pool and reader contributions.

This is not a correctness issue for external callers (they always go through the overridden `GetMemoryUsage()`), but it creates an internal inconsistency: if any future code path reads `current_memory_usage_` directly, it will get a stale/partial value. Consider either:
- Adding `pool_->GetMemoryUsage()` back to `cal_memory_usage()` so the cached snapshot stays consistent, or
- Removing the `cal_memory_usage()` call from `SetImmutable()` since the cached value is no longer authoritative.

LHT129 (commented):
Thank you for this well-designed PR. The approach of adding an optional `MemoryUsageReader` subtype that preserves ABI compatibility with existing `Reader` implementations is clean and well-executed.

**Summary of the change:**
- Introduces `MemoryUsageReader` as an optional extension of `Reader` for reporting reader-owned resident memory
- HGraph now tracks retained `reader_` and `precise_reader_` references and includes their reported memory in both `GetMemoryUsage()` and `GetMemoryUsageDetail()`
- Properly avoids double-counting when the same reader instance is shared
- Moves `pool_->GetMemoryUsage()` from the cached `cal_memory_usage()` to the dynamic `GetMemoryUsage()` override so query-time pool growth is reflected immediately
- Refreshes the static memory snapshot after `SetImmutable()` releases neighbor mutex storage
- Includes regression tests for both `MemoryUsageReader` and plain `Reader` paths

**Existing review comments already cover the key points I would raise:**
- The `cal_memory_usage()` / `current_memory_usage_` cache inconsistency (LHT129)
- Silent `precise_reader_.reset()` in `SetIO()` (LHT129)
- Missing `@param memory_usage` documentation in `TestReader` constructor (Copilot)
- Documentation should clarify `precise_reader` only appears when distinct from `reader` (Copilot)

The ABI compatibility approach (unchanged `Reader` vtable, runtime `dynamic_cast` detection) is well-considered. The double-counting guard (`precise_reader_.get() != reader_.get()`) is correct and properly handles the shared-reader case.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
...
+        return 0;
+    }
+    return memory_usage_reader->GetMemoryUsage();
+}
+
+uint64_t
+HGraph::GetMemoryUsage() const {
+    auto memory = InnerIndexInterface::GetMemoryUsage();
+    if (this->pool_ != nullptr) {
+        memory += this->pool_->GetMemoryUsage();
```
LHT129, line 831: [suggestion] `GetMemoryUsage()` guards `pool_->GetMemoryUsage()` with a null check (`if (this->pool_ != nullptr)`), but `GetMemoryUsageDetail()` in `hgraph_serialize.cpp:1084` calls `this->pool_->GetMemoryUsage()` unconditionally. If `pool_` can be null (as the new `GetMemoryUsage()` code implies), `GetMemoryUsageDetail()` will dereference a null pointer.

Consider adding the same null guard in `GetMemoryUsageDetail()` for consistency, or removing the null check in `GetMemoryUsage()` if `pool_` is guaranteed non-null in practice (the old `cal_memory_usage()` also called it unconditionally).

LHT129 (commented):
I completed an independent review of this PR. The existing review comments from LHT129, Copilot, and gemini-code-assist already cover all substantive issues I identified. The code changes are well-structured and the ABI-compatible approach with `MemoryUsageReader` as an opt-in subclass of `Reader` is a clean design.

Key observations already raised by other reviewers that I agree with:
- The `cal_memory_usage()` / `current_memory_usage_` inconsistency after moving `pool_` out (LHT129)
- The missing null guard for `pool_` in `GetMemoryUsageDetail()` (LHT129)
- The silent `precise_reader_.reset()` in `SetIO()` (LHT129)
- Documentation clarification for `precise_reader` dedup behavior (Copilot)

No additional issues found beyond what has already been noted.

LHT129 (commented):
Thanks for this PR. The design of an optional `MemoryUsageReader` subtype that preserves ABI compatibility with existing `Reader` implementations is clean and well-thought-out.

I have reviewed the full diff and the surrounding context. The existing inline comments from other reviewers already cover the substantive points I would raise. In summary, the key items worth addressing before merge:

1. **`GetMemoryUsageDetail()` null safety** (LHT129's comment): `pool_->GetMemoryUsage()` is called unconditionally in `GetMemoryUsageDetail()` while the new `GetMemoryUsage()` guards it with a null check. This inconsistency should be resolved — either add the guard to `GetMemoryUsageDetail()` or confirm `pool_` is never null and remove the guard from `GetMemoryUsage()`.

2. **`cal_memory_usage()` cached snapshot** (LHT129's comment): The cached `current_memory_usage_` no longer includes pool/reader contributions after `pool_->GetMemoryUsage()` was moved out of `cal_memory_usage()`. This is worth addressing for internal consistency.

3. **Test magic number**: The `4096` tolerance in the test assertion should be named and documented.

Otherwise the implementation looks correct — the double-counting guard (`precise_reader_.get() != reader_.get()`) is appropriate, the `dynamic_cast` capability detection is sound, and the test coverage is thorough.

tests/test_hgraph.cpp
```diff
@@ -2415,6 +2416,79 @@ TestHGraphReaderIO(const fixtures::HGraphTestIndexPtr& test_index,
...
+            std::memcpy(dest, binary.data.get() + offset, len);
+        },
+        binary.size);
+    vsag::ReaderSet plain_readers;
+    plain_readers.Set("hgraph", plain_reader);
+    REQUIRE(plain_restored->Deserialize(plain_readers).has_value());
+    REQUIRE(plain_restored->GetMemoryUsage() > 0);
+    REQUIRE(plain_restored->GetMemoryUsageDetail().at("reader") == 0);
+}
+
```
LHT129, line 2491: [suggestion] The test covers the case where a single shared reader is used for both base and precise codes (via `Deserialize(ReaderSet&)` -> `SetIO`), but the `precise_reader_` distinct-from-`reader_` code path (the double-counting prevention at `hgraph.cpp:840` and `hgraph_serialize.cpp:1112`) is not exercised. This path is only reachable via the streaming deserialization path (`load_streaming_body`) when a separate `"precise_reader"` key is provided in `LoadParameters`.

Consider adding a test case that uses the streaming deserialization path with a distinct precise reader to verify:
1. `GetMemoryUsageDetail()` includes both `"reader"` and `"precise_reader"` keys
2. The double-counting prevention works correctly when the same reader instance is used for both

This would also exercise the `SetPreciseCodesIO` -> `precise_reader_` assignment path.

LHT129 (commented):
## Review Summary

This PR adds an optional `MemoryUsageReader` subtype to include reader-owned resident caches in HGraph memory usage reporting. The design is clean and ABI-safe: the existing `Reader` vtable is unchanged, and the new capability is detected at runtime via `dynamic_cast`.

### What I reviewed
- `include/vsag/readerset.h` — new `MemoryUsageReader` class
- `src/algorithm/hgraph/hgraph.cpp` — `SetIO`, `SetPreciseCodesIO`, `GetMemoryUsage`, `get_reader_memory_usage`, `cal_memory_usage`
- `src/algorithm/hgraph/hgraph.h` — new members `reader_`, `precise_reader_`, method declarations
- `src/algorithm/hgraph/hgraph_serialize.cpp` — `GetMemoryUsageDetail` additions
- `src/reader.cpp` — destructor definition
- `tests/fixtures/framework/test_reader.{h,cpp}` — `TestReader` now derives from `MemoryUsageReader`
- `tests/test_hgraph.cpp` — new `"HGraph Reader Memory Usage"` test case
- `docs/docs/{en,zh}/src/advanced/memory.md` — documentation updates

### Key observations
Most substantive issues have already been raised by earlier reviewers (LHT129, Copilot, gemini-code-assist). The ones I want to highlight:

1. **`cal_memory_usage()` cached snapshot inconsistency** (LHT129): The base class cached `current_memory_usage_` no longer includes pool/reader memory, while `HGraph::GetMemoryUsage()` does. This is not a correctness issue for external callers but creates internal inconsistency.

2. **`GetMemoryUsageDetail()` null-safety inconsistency** (LHT129): `pool_->GetMemoryUsage()` is called unconditionally in `GetMemoryUsageDetail()` but guarded by a null check in `GetMemoryUsage()`.

3. **Test coverage gap for distinct `precise_reader_`** (my comment): The streaming deserialization path that sets a separate `precise_reader_` is not tested.

Overall the PR is well-structured with good ABI compatibility considerations, clear documentation, and reasonable test coverage for the primary code paths.

LHT129 (commented):
This is a well-structured PR that adds reader cache memory reporting to HGraph in an ABI-safe manner. The approach of using a separate `MemoryUsageReader` subclass (rather than modifying the `Reader` vtable) is sound.

After reviewing the diff and full file context, I concur with the existing review comments from @LHT129:

1. **`cal_memory_usage()` / `GetMemoryUsage()` inconsistency**: `cal_memory_usage()` (called from `SetImmutable()`) no longer includes `pool_->GetMemoryUsage()`, making the cached `current_memory_usage_` snapshot inconsistent with what `GetMemoryUsage()` returns. Consider either restoring pool accounting in `cal_memory_usage()` or removing the `cal_memory_usage()` call from `SetImmutable()`.

2. **`pool_` null check inconsistency**: `GetMemoryUsage()` guards `pool_->GetMemoryUsage()` with a null check, but `GetMemoryUsageDetail()` calls it unconditionally. These should be consistent.

3. **Missing test coverage for distinct precise reader**: The double-counting prevention path (`precise_reader_.get() != reader_.get()`) is not exercised by the current test. A test using the streaming deserialization path with a separate precise reader would close this gap.

The core logic (reader tracking in `SetIO`/`SetPreciseCodesIO`, `dynamic_cast` dispatch in `get_reader_memory_usage`, double-counting prevention, and `GetMemoryUsage`/`GetMemoryUsageDetail` integration) is correct and cleanly implemented.

LHT129 (commented):
I have completed a thorough review of this PR. The overall design is sound: introducing `MemoryUsageReader` as an optional ABI-safe subtype is a clean approach that preserves backward compatibility while enabling reader memory accounting.

**Summary of findings:**

The existing inline comments from other reviewers (LHT129, Copilot, gemini-code-assist) already cover the main issues I identified during my review:

1. **`pool_->GetMemoryUsage()` null-check inconsistency** (LHT129): `GetMemoryUsage()` guards with a null check but `GetMemoryUsageDetail()` calls it unconditionally. This should be made consistent.

2. **Missing test coverage for distinct precise reader** (LHT129): The streaming deserialization path where `precise_reader_` differs from `reader_` is not exercised by the new test. The double-counting prevention logic at `hgraph.cpp:840` and `hgraph_serialize.cpp:1112` would benefit from explicit test coverage.

3. **Documentation and minor issues** (Copilot/gemini-code-assist): Missing `@param` doc, magic number `4096`, documentation clarification about `precise_reader` double-counting behavior.

**Additional observations (no blocking issues):**

- The removal of `pool_->GetMemoryUsage()` from `cal_memory_usage()` and its addition to the dynamic `GetMemoryUsage()` override is correct — pool memory can grow at query time and should be read dynamically.
- The `SetIO` → `precise_reader_.reset()` pattern correctly handles the case where a single shared reader replaces previously separate readers.
- The `dynamic_cast` in `get_reader_memory_usage()` is the right tool for optional capability detection without vtable changes.
- The `SetImmutable()` → `cal_memory_usage()` call correctly refreshes the static snapshot after releasing neighbor mutex storage.

The PR is well-structured with good test coverage and documentation. I recommend addressing the two items from LHT129 before merging.

LHT129 (commented):
Thanks for this PR. The approach of introducing an optional `MemoryUsageReader` subtype with runtime detection via `dynamic_cast` is a clean way to add reader memory accounting while preserving ABI compatibility with existing `Reader` implementations.

I have reviewed the full diff and the key source files. The existing review comments from other reviewers have already covered the substantive issues:

1. **Null-safety in `SetPreciseCodesIO`** — the `high_precise_codes_` null check suggestion is worth considering for defensive robustness.
2. **`pool_` null guard inconsistency** — `GetMemoryUsage()` guards `pool_` with a null check, but `GetMemoryUsageDetail()` dereferences it unconditionally. These should be consistent.
3. **Test coverage for distinct `precise_reader_` path** — the double-counting prevention logic (`precise_reader_.get() != reader_.get()`) is not exercised by the current test, which only covers the shared-reader path via `Deserialize(ReaderSet&)`.
4. **Documentation** — the `precise_reader` key description in both EN/ZH docs should clarify it only appears when a distinct precise reader instance is retained.
5. **Test constants** — the magic `4096` tolerance and duplicated `dim` in JSON strings are minor maintainability concerns.

The overall design is sound: the static snapshot (`cal_memory_usage`) captures immutable components, while the `GetMemoryUsage()` override dynamically adds runtime-varying components (pool, reader caches). The double-counting prevention when `reader_` and `precise_reader_` point to the same instance is correctly handled.

LHT129 (commented):
Thank you for this PR. The approach of using an optional `MemoryUsageReader` subtype with runtime `dynamic_cast` detection is clean and maintains ABI compatibility with existing `Reader` implementations.

I have reviewed the changes and the existing inline comments from other reviewers (gemini-code-assist, Copilot, LHT129) have already covered the key points:

1. **`pool_` null-check inconsistency** between `GetMemoryUsage()` and `GetMemoryUsageDetail()` — the former guards with `!= nullptr` while the latter dereferences unconditionally. This should be reconciled.

2. **Missing test coverage for the distinct `precise_reader_` path** — the streaming deserialization path where `SetPreciseCodesIO` sets a reader different from `reader_` is not exercised by the current test.

3. **Documentation clarifications** — the `@param memory_usage` docstring and the double-counting prevention behavior for `precise_reader` should be documented.

The core design (optional subtype, runtime detection, double-counting prevention when readers are shared) is sound. The `cal_memory_usage()` change to exclude dynamic `pool_` from the static snapshot and instead read it fresh in `GetMemoryUsage()` is correct — this ensures query-time pool growth is reflected immediately.

Please address the items above, particularly the `pool_` null-check consistency and the streaming-path test coverage.

LHT129 (commented):
Reviewed the changes for PR #2498 (include reader caches in memory usage). 

**Summary**: This PR introduces a well-designed `MemoryUsageReader` subtype that extends the existing `Reader` interface without breaking ABI compatibility. The implementation correctly handles the shared-reader case (avoiding double-counting when `reader_` and `precise_reader_` point to the same instance) and dynamically reports pool memory in the overridden `GetMemoryUsage()`.

**Key observations** (most already noted by other reviewers):

1. `GetMemoryUsageDetail()` in `hgraph_serialize.cpp` calls `this->pool_->GetMemoryUsage()` unconditionally, while the new `GetMemoryUsage()` guards it with a null check. If `pool_` can be null, `GetMemoryUsageDetail()` will dereference a null pointer. (Noted by @LHT129)

2. The test covers the shared-reader path but not the distinct-`precise_reader_` path reachable via streaming deserialization with a separate `"precise_reader"` key. (Noted by @LHT129)

3. The `4096` tolerance in the test is a magic number that may be brittle across platforms. (Noted by Copilot)

4. Documentation for `precise_reader` could clarify the double-counting prevention behavior. (Noted by Copilot)

The overall approach is sound and the ABI compatibility strategy (separate vtable via subclass) is clean. No blocking issues found beyond what has already been raised.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
...
+HGraph::get_reader_memory_usage(const ReaderPtr& reader) {
+    const auto* memory_usage_reader = dynamic_cast<const MemoryUsageReader*>(reader.get());
+    if (memory_usage_reader == nullptr) {
+        return 0;
+    }
+    return memory_usage_reader->GetMemoryUsage();
+}
+
+uint64_t
+HGraph::GetMemoryUsage() const {
```
LHT129, line 828: [suggestion] `HGraph::GetMemoryUsage()` now bypasses the cached `current_memory_usage_` (set by `cal_memory_usage()`) and computes dynamically. However, `cal_memory_usage()` is still called in `SetImmutable()` and after deserialization — but its result is no longer used by `GetMemoryUsage()`. This means the `memory_usage_mutex_`-protected cache is dead code for HGraph.

Consider either:
1. Removing the `cal_memory_usage()` call from `SetImmutable()` (and deserialization paths) since it no longer affects `GetMemoryUsage()`, or
2. Having `GetMemoryUsage()` add the dynamic portions (pool + reader) on top of the cached value instead of calling `InnerIndexInterface::GetMemoryUsage()` directly.

Option 2 would keep the cache relevant and avoid re-summing all the static components on every call.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -601,13 +602,16 @@ HGraph::SetIO(const std::shared_ptr<Reader> reader) {
     }
     basic_flatten_codes_->InitIO(reader_param);
     bottom_graph_->InitIO(reader_param);
+    this->reader_ = reader;
```
LHT129, line 605: [suggestion] In `SetIO()`, `this->precise_reader_.reset()` unconditionally clears the precise reader. If `SetIO` is called with the same reader instance that was previously set as the precise reader via `SetPreciseCodesIO`, the precise reader reference is lost. This means subsequent `GetMemoryUsage()` and `GetMemoryUsageDetail()` calls will not report the precise reader memory even though it is still in use by `high_precise_codes_`.

Consider only resetting `precise_reader_` when it differs from the incoming reader:

```cpp
if (this->precise_reader_.get() != reader.get()) {
    this->precise_reader_.reset();
}
```

Or alternatively, if `SetIO` is always called before `SetPreciseCodesIO` and this is guaranteed by the callers, add a comment documenting that invariant.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
     this->attr_filter_index_->GetAttribute(0, inner_id, attr);
 }
 
+uint64_t
+HGraph::get_reader_memory_usage(const ReaderPtr& reader) {
+    const auto* memory_usage_reader = dynamic_cast<const MemoryUsageReader*>(reader.get());
```
LHT129, line 820: [note] `get_reader_memory_usage` uses `dynamic_cast` for the runtime capability check, which requires RTTI to be enabled. If VSAG targets any RTTI-disabled build configurations (common in some embedded/HPC environments), this would fail to compile or silently return 0 (UB with `-fno-rtti`).

This is likely fine for the current build configuration, but worth being aware of. An alternative would be a virtual method on `Reader` itself (e.g., a default `GetMemoryUsage()` returning 0), which would avoid RTTI dependency at the cost of a vtable slot on the base class.

LHT129 (commented):
## Review Summary

This PR adds support for including reader-owned resident memory in HGraph memory usage reporting. The approach is well-designed: a new `MemoryUsageReader` subclass of `Reader` provides an optional `GetMemoryUsage()` interface, and HGraph detects it at runtime via `dynamic_cast` without changing the base `Reader` vtable — preserving ABI compatibility.

### What I reviewed

- **`include/vsag/readerset.h`**: The new `MemoryUsageReader` class is cleanly separated from `Reader`, with clear documentation about thread-safety requirements. ABI-compatible approach.
- **`src/algorithm/hgraph/hgraph.cpp`**: `GetMemoryUsage()` override correctly adds dynamic pool and reader memory on top of the cached static snapshot. `SetIO`/`SetPreciseCodesIO` now retain reader references. `cal_memory_usage()` still refreshes the static snapshot in `SetImmutable()`.
- **`src/algorithm/hgraph/hgraph_serialize.cpp`**: `GetMemoryUsageDetail()` correctly reports `reader` and `precise_reader` entries, with proper deduplication when both point to the same instance.
- **`src/reader.cpp`**: Minimal change — just the destructor definition for `MemoryUsageReader`.
- **Tests**: Good coverage — verifies memory reporting with a `MemoryUsageReader`, verifies `SetImmutable()` reduces memory (mutex released), verifies plain `Reader` reports zero, and verifies detail sum consistency.
- **Docs**: English and Chinese memory docs updated with the new `reader`/`precise_reader` keys.

### Existing review comments

Several collaborators and bots have already raised valuable points on this PR. I agree with the substantive ones and won't repeat them. The key items for the author to address are:

1. **`SetIO()` unconditionally clears `precise_reader_`** — if `SetIO` is called after `SetPreciseCodesIO` with a different reader, the precise reader reference is lost. Consider only resetting when the readers differ, or document the calling order invariant.
2. **`GetMemoryUsage()` vs `cal_memory_usage()` relationship** — the current design is actually correct: `GetMemoryUsage()` calls the base class (which returns the cached static snapshot from `cal_memory_usage()`), then adds dynamic pool + reader memory. The `cal_memory_usage()` call in `SetImmutable()` is still needed to refresh the snapshot after mutex release.
3. **`dynamic_cast` RTTI dependency** — worth documenting that RTTI must be enabled for this feature.

### Overall assessment

The implementation is correct, well-tested, and ABI-safe. The design choice of a separate `MemoryUsageReader` subclass (rather than adding a virtual method to `Reader`) is the right call for maintaining backward compatibility. Once the existing review comments are addressed, this looks ready to merge.

LHT129 (commented):
Thanks for this PR. The design of an optional `MemoryUsageReader` subtype that keeps the existing `Reader` vtable unchanged is a clean approach to ABI compatibility.

I have reviewed the changes across all 10 files. The existing review comments from LHT129, Copilot, and gemini-code-assist have already covered the key points:

1. **`SetIO` unconditionally resetting `precise_reader_`** (LHT129): If `SetIO` is called after `SetPreciseCodesIO` with the same reader, the precise reader reference is lost. Consider only resetting when the readers differ.

2. **`GetMemoryUsage()` caching strategy** (LHT129): The `cal_memory_usage()` snapshot in `SetImmutable()` is still used as the base value by `GetMemoryUsage()` (via `InnerIndexInterface::GetMemoryUsage()`), with pool and reader added dynamically on top. This is a reasonable design — the cache is not dead code.

3. **`dynamic_cast` RTTI dependency** (LHT129): Worth being aware of for `-fno-rtti` builds, though likely fine for current configurations.

4. **Documentation and test nits** (Copilot): Missing `@param` doc, magic number 4096, duplicate `dim` constant, and docs clarification about `precise_reader` double-counting avoidance.

The core logic — detecting `MemoryUsageReader` at runtime, avoiding double-counting when readers are shared, and dynamically tracking pool/reader memory — looks correct. The test coverage is thorough, covering both cache-reporting and plain readers.

No additional blocking issues found from my side.

LHT129 (commented):
## Code Review

### [suggestion] `GetMemoryUsage()` returns incomplete value during mutable phase

`HGraph::GetMemoryUsage()` relies on `InnerIndexInterface::GetMemoryUsage()` which returns `current_memory_usage_` — a cached value set only by `cal_memory_usage()`. However, `cal_memory_usage()` is only called in `SetImmutable()` and during deserialization. It is never called during `Build()` or `Add()`.

This means that after `Build()` completes but before `SetImmutable()` is called, `GetMemoryUsage()` returns only `pool_` + reader memory, missing all static components (flatten codes, graphs, label table, etc.). The returned value would be significantly under-reported during the mutable phase.

Consider either:
1. Calling `cal_memory_usage()` at the end of `Build()` and after significant `Add()` batches, or
2. Having `GetMemoryUsage()` compute the static portion directly (like `cal_memory_usage()` does) and only add the dynamic portions (pool + reader) on top, rather than delegating to the cached `InnerIndexInterface::GetMemoryUsage()`.

Option 2 would also address the dead-cache concern raised in the earlier review comment about `cal_memory_usage()` results no longer being consumed by `GetMemoryUsage()`.

**File:** `src/algorithm/hgraph/hgraph.cpp:828`

LHT129 (commented):
I have reviewed this PR and the changes look good overall. The approach of introducing an optional `MemoryUsageReader` subtype to avoid breaking the existing `Reader` ABI is clean and well-designed.

The existing inline comments from other reviewers (gemini-code-assist, Copilot, LHT129) have already covered the notable points:

1. **`SetIO` unconditionally resetting `precise_reader_`** — if `SetIO` is called with the same reader instance that was previously set as the precise reader, the precise reader reference is lost. Consider only resetting when the pointers differ, or document the caller ordering invariant.

2. **`GetMemoryUsage()` bypassing the cached `current_memory_usage_`** — `cal_memory_usage()` is still called in `SetImmutable()` and after deserialization, but its result is only used as a partial sum by `GetMemoryUsage()`. The cache is still meaningful (it holds the static component sum), but the comment about it being "dead code" is worth addressing — either clarify the design or restructure.

3. **`dynamic_cast` requiring RTTI** — this is a design tradeoff worth being aware of. The current approach preserves the base `Reader` vtable at the cost of RTTI dependency. This is acceptable for the current build configuration.

4. **Documentation precision** — the docs should clarify that `precise_reader` appears in the detail map only when it is a distinct instance from `reader`.

5. **Test magic number** — the `4096` tolerance in the test should be named and documented.

The test coverage is solid, the ABI compatibility claim is verified, and the documentation updates are consistent. No blocking issues from my side.

LHT129 (commented):
This PR adds `MemoryUsageReader` support to HGraph, allowing reader-owned caches to be included in memory usage reporting. The approach is well-designed: a new `MemoryUsageReader` subclass preserves ABI compatibility with the existing `Reader` interface, and runtime `dynamic_cast` detection avoids breaking existing readers.

**Summary of changes:**
- `include/vsag/readerset.h`: New `MemoryUsageReader` class with `GetMemoryUsage()` pure virtual
- `src/algorithm/hgraph/hgraph.cpp`: `GetMemoryUsage()` override that dynamically includes pool + reader memory; `SetIO`/`SetPreciseCodesIO` now retain reader pointers; `SetImmutable()` refreshes the memory snapshot
- `src/algorithm/hgraph/hgraph.h`: New `reader_`/`precise_reader_` members, `get_reader_memory_usage()` helper
- `src/algorithm/hgraph/hgraph_serialize.cpp`: `GetMemoryUsageDetail()` now reports `reader`/`precise_reader` entries
- `src/reader.cpp`: `MemoryUsageReader` destructor out-of-line definition
- `tests/`: `TestReader` upgraded to `MemoryUsageReader`; new `HGraph Reader Memory Usage` test case covering cache-reporting and plain readers
- `docs/`: English and Chinese memory documentation updated

**Key observations already raised by other reviewers:**
1. `SetIO()` unconditionally resets `precise_reader_` — could lose the precise reader reference if the same instance is reused
2. `GetMemoryUsage()` bypasses the `cal_memory_usage()` cache, making the cached `current_memory_usage_` dead code for HGraph
3. `dynamic_cast` requires RTTI — worth documenting if no-RTTI builds are ever targeted
4. Several documentation and test polish items (magic number constant, JSON/C++ constant dedup, docstring updates)

The core logic is correct, the test coverage is solid (including the plain-reader-zero-memory path), and the ABI compatibility claim is well-verified. No blocking issues from my review beyond what has already been noted.

LHT129 (commented):
Overall this is a well-designed PR that adds reader memory accounting to HGraph while maintaining ABI compatibility through the `MemoryUsageReader` subtype pattern.

**Summary of the approach:**
- Introduces `MemoryUsageReader` as an optional subclass of `Reader` with a `GetMemoryUsage()` method
- Uses `dynamic_cast` at runtime to detect the capability without changing the `Reader` vtable
- HGraph tracks `reader_` and `precise_reader_` separately, with dedup logic to avoid double-counting when they share the same instance
- `GetMemoryUsage()` now overrides the base class to dynamically include pool and reader memory
- `GetMemoryUsageDetail()` includes `reader` and `precise_reader` keys
- `cal_memory_usage()` no longer includes pool (moved to dynamic query in `GetMemoryUsage()`)
- Good test coverage with both `MemoryUsageReader` and plain `Reader` paths

**Key observations already noted by other reviewers:**
1. `SetIO()` unconditionally resets `precise_reader_` — consider guarding this when the reader is shared
2. `GetMemoryUsage()` bypasses the `cal_memory_usage()` cache, making it dead code for HGraph
3. `dynamic_cast` requires RTTI — worth documenting if VSAG targets any `-fno-rtti` builds
4. Test uses magic number 4096 for accounting tolerance
5. Documentation should clarify `precise_reader` dedup behavior

No blocking issues found. The ABI compatibility approach is clean and the test coverage is thorough.

LHT129 (commented):
Overall this is a well-designed PR that adds reader memory accounting to HGraph while maintaining ABI compatibility through the MemoryUsageReader subtype pattern.

Summary of the approach:
- Introduces MemoryUsageReader as an optional subclass of Reader with a GetMemoryUsage() method
- Uses dynamic_cast at runtime to detect the capability without changing the Reader vtable
- HGraph tracks reader_ and precise_reader_ separately, with dedup logic to avoid double-counting
- GetMemoryUsage() now overrides the base class to dynamically include pool and reader memory
- GetMemoryUsageDetail() includes reader and precise_reader keys
- cal_memory_usage() no longer includes pool (moved to dynamic query)
- Good test coverage with both MemoryUsageReader and plain Reader paths

Key observations already noted by other reviewers:
1. SetIO() unconditionally resets precise_reader_ — consider guarding when the reader is shared
2. GetMemoryUsage() bypasses the cal_memory_usage() cache, making it dead code for HGraph
3. dynamic_cast requires RTTI — worth documenting if VSAG targets -fno-rtti builds
4. Test uses magic number 4096 for accounting tolerance
5. Documentation should clarify precise_reader dedup behavior

No blocking issues found. The ABI compatibility approach is clean and the test coverage is thorough.

LHT129 (commented):
Overall this is a well-designed PR that adds reader memory accounting to HGraph while maintaining ABI compatibility through the MemoryUsageReader subtype pattern.

Summary of the approach:
- Introduces MemoryUsageReader as an optional subclass of Reader with a GetMemoryUsage() method
- Uses dynamic_cast at runtime to detect the capability without changing the Reader vtable
- HGraph tracks reader_ and precise_reader_ separately, with dedup logic to avoid double-counting
- GetMemoryUsage() now overrides the base class to dynamically include pool and reader memory
- GetMemoryUsageDetail() includes reader and precise_reader keys
- cal_memory_usage() no longer includes pool (moved to dynamic query)
- Good test coverage with both MemoryUsageReader and plain Reader paths

Key observations already noted by other reviewers:
1. SetIO() unconditionally resets precise_reader_ — consider guarding when the reader is shared
2. GetMemoryUsage() bypasses the cal_memory_usage() cache, making it dead code for HGraph
3. dynamic_cast requires RTTI — worth documenting if VSAG targets -fno-rtti builds
4. Test uses magic number 4096 for accounting tolerance
5. Documentation should clarify precise_reader dedup behavior

No blocking issues found. The ABI compatibility approach is clean and the test coverage is thorough.

LHT129 (commented):
This is a well-structured PR that adds reader cache memory reporting to HGraph while maintaining full ABI compatibility. The approach of introducing a separate `MemoryUsageReader` subclass (rather than modifying the base `Reader` vtable) is the right call for backward compatibility.

**Summary of what was reviewed:**

1. **ABI safety** — The `Reader` vtable is untouched; `MemoryUsageReader` is an opt-in extension. The `dynamic_cast` capability check at runtime is clean and avoids breaking existing `Reader` binaries.

2. **Correctness** — The double-counting guard (`precise_reader_.get() != reader_.get()`) in both `GetMemoryUsage()` and `GetMemoryUsageDetail()` is correct. The `pool_` usage is now dynamic, which correctly reflects query-time growth.

3. **Test coverage** — The new `HGraph Reader Memory Usage` test case covers both `MemoryUsageReader` and plain `Reader` paths, validates `SetImmutable()` memory reduction, and checks detail-sum consistency. Good coverage.

4. **Documentation** — Both English and Chinese memory docs are updated with the new `reader`/`precise_reader` keys and thread-safety requirement.

A few existing review comments from other reviewers have already been raised (cache bypass in `GetMemoryUsage()`, `SetIO` precise_reader reset behavior, RTTI dependency of `dynamic_cast`, etc.) — those are worth addressing but none are blocking for this review.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
...
+    return memory_usage_reader->GetMemoryUsage();
+}
+
+uint64_t
+HGraph::GetMemoryUsage() const {
+    auto memory = InnerIndexInterface::GetMemoryUsage();
+    if (this->pool_ != nullptr) {
+        memory += this->pool_->GetMemoryUsage();
+    }
+    if (this->reader_ != nullptr) {
```
LHT129, line 833: [note] `GetMemoryUsage()` reads `reader_` and `precise_reader_` shared pointers without any synchronization. These members are written in `SetIO()` and `SetPreciseCodesIO()` during deserialization, and `GetMemoryUsage()` may be called concurrently from query threads afterward. While in practice deserialization completes before any query, there is no formal happens-before edge (e.g., no acquire/release on the pointer writes vs reads).

If this is guaranteed by the caller (i.e., `GetMemoryUsage()` is only called after deserialization fully completes), consider adding a comment documenting that invariant. Otherwise, the `reader_` and `precise_reader_` reads should be protected by the same synchronization that guards their writes.

LHT129 (commented):
This PR has already received extensive review from multiple reviewers covering all major concerns. I have reviewed the remaining changes and have no additional substantive issues to raise beyond what has already been flagged.

Key areas already covered by existing reviews:
- ABI compatibility of the new `MemoryUsageReader` subclass
- Thread safety of `reader_`/`precise_reader_` reads in `GetMemoryUsage()`
- `dynamic_cast` RTTI dependency in `get_reader_memory_usage()`
- `SetIO()` unconditionally resetting `precise_reader_`
- Cache inconsistency between `cal_memory_usage()` and `GetMemoryUsage()`
- `pool_` null check inconsistency between `GetMemoryUsage()` and `GetMemoryUsageDetail()`
- Missing test coverage for the distinct `precise_reader_` path
- Documentation clarifications for double-counting prevention

The overall approach of using an opt-in `MemoryUsageReader` subclass to avoid changing the existing `Reader` vtable is a reasonable design choice that preserves backward compatibility for existing Reader implementations.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -811,11 +815,34 @@ HGraph::GetAttributeSetByInnerId(InnerIdType inner_id, AttributeSet* attr) const
...
+    }
+    if (this->precise_reader_ != nullptr && this->precise_reader_.get() != this->reader_.get()) {
+        memory += get_reader_memory_usage(this->precise_reader_);
+    }
+    return memory;
+}
+
 void
 HGraph::cal_memory_usage() {
     auto memory = sizeof(HGraph);
```
LHT129, line 844: [note] `cal_memory_usage()` no longer includes `pool_->GetMemoryUsage()` — this was moved to the dynamic `GetMemoryUsage()` override. The function's semantics have shifted from "total memory snapshot" to "static structural memory snapshot." This is the correct design since pool usage fluctuates at query time, but the function name no longer fully reflects its scope.

Consider renaming to `cal_static_memory_usage()` or adding a brief comment above the method declaration in `hgraph.h` to clarify that it only snapshots structural (non-pool, non-reader) memory, while the dynamic portions are handled by `GetMemoryUsage()`.

LHT129 (commented):
Overall this is a well-designed PR that cleanly adds reader cache memory tracking to HGraph without breaking ABI compatibility. The use of `MemoryUsageReader` as an optional subtype with `dynamic_cast` runtime detection is a pragmatic approach that preserves backward compatibility.

**Summary of the approach:**
- Introduces `MemoryUsageReader` as an opt-in subclass of `Reader` with a `GetMemoryUsage()` method
- HGraph stores `reader_` and `precise_reader_` pointers during deserialization
- `GetMemoryUsage()` dynamically queries reader memory and visited-list pool usage
- `GetMemoryUsageDetail()` reports reader components separately
- `cal_memory_usage()` is refreshed after `SetImmutable()` to reflect released mutex storage
- Good test coverage including both `MemoryUsageReader` and plain `Reader` paths

**Key design decisions that look correct:**
- Pool memory moved from static snapshot (`cal_memory_usage`) to dynamic query (`GetMemoryUsage`) since it fluctuates at query time
- Double-counting guard when `reader_` and `precise_reader_` point to the same instance
- Out-of-line destructor for `MemoryUsageReader` ensures proper vtable emission
- Existing `Reader` vtable is unchanged, preserving ABI compatibility

The existing inline comments from other reviewers cover the remaining considerations (thread safety documentation, `precise_reader_.reset()` edge case, RTTI dependency, doc updates).

LHT129 (commented):
Reviewed the changes. The PR is well-structured with good test coverage and documentation updates. The existing inline comments from other reviewers already cover the key concerns (thread safety of reader_/precise_reader_ reads, dynamic_cast/RTTI dependency, cal_memory_usage naming clarity, null-check in SetPreciseCodesIO, magic number 4096 in tests, and doc clarifications). No additional blocking issues found.

LHT129 (commented):
I have reviewed this PR and the implementation looks solid. The existing review comments from other reviewers have already covered the key concerns:

- **RTTI dependency**: `dynamic_cast` in `get_reader_memory_usage` requires RTTI (noted by @LHT129)
- **Thread safety**: `reader_`/`precise_reader_` reads in `GetMemoryUsage()` lack formal synchronization with writes in `SetIO`/`SetPreciseCodesIO` (noted by @LHT129)
- **`cal_memory_usage()` semantics**: The function now snapshots only static structural memory, with pool and reader memory handled dynamically in `GetMemoryUsage()` (noted by @LHT129)
- **Documentation**: `precise_reader` double-counting avoidance should be clarified in docs (noted by Copilot)
- **Test**: Magic number 4096 and duplicated `dim` constant (noted by Copilot)

The core design — adding an optional `MemoryUsageReader` subtype without changing the existing `Reader` vtable — is ABI-safe and well-executed. The double-counting guard (`precise_reader_.get() != reader_.get()`) in both `GetMemoryUsage()` and `GetMemoryUsageDetail()` is correct. Moving pool memory from the static snapshot to dynamic reporting in `GetMemoryUsage()` is the right approach since pool usage fluctuates at query time.

No new blocking issues found beyond what has already been raised.

PR #2503, opened 2026-07-22
feat: support pyramid raw vector retrieval

### LINKED ISSUE #2502: [feat](other): support raw vector retrieval in pyramid
### Affected area

Pyramid public raw-vector retrieval.

### Motivation

Pyramid on current `main` already owns an internal FP32 `raw_vector_` data cell for configurations that require a decodable source, but it does not expose that storage through `Index::GetRawVectorByIds`. Explicit `store_raw_vector` in the public Pyramid parameters is also not mapped to the existing internal parameter.

Applications using quantized Pyramid base codes therefore cannot request and retrieve the original FP32 vectors through the public API. ODescent builds also need a populated external-label reverse map for lookup by non-contiguous IDs.

### Proposed solution

Reuse the implementation already on `main` and add only the public API wiring:

- Map the external `store_raw_vector` parameter.
- Set `has_raw_vector_` when the existing raw-vector data cell is created.
- Advertise `SUPPORT_GET_RAW_VECTOR_BY_IDS` when raw storage exists.
- Populate the ODescent label reverse map during Build.
- Use raw FP32 vectors for precise distance-by-id calculations.
- Add focused NSW and ODescent public API regression coverage and parameter documentation.

No new raw-vector lifecycle, FP32 aliasing, serialization format, IO override, or memory-accounting implementation is required in this issue.

Additive behavior only; no public signature or serialization-format change.

### References

- Existing internal raw-vector support landed through #2560.
- Implemented by #2503.

jac0626:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2502

## What Changed

- Map Pyramid `store_raw_vector` to the raw-vector storage already present on `main`.
- Expose that storage through public `GetRawVectorByIds` when enabled.
- Use the stored FP32 vectors for precise single and batch distance-by-id calculations.
- Build the ODescent label reverse map so non-contiguous external IDs work through the public API.
- Document `store_raw_vector` in the English and Chinese Pyramid parameter references.

This PR deliberately leaves the existing Pyramid raw-vector Train, Build, Add, Resize, serialization, and `GetVectorByInnerId` implementation unchanged. It does not add FP32 aliasing, a new serialization format, IO overrides, or memory accounting.

## Test Evidence

- [x] clang-format 15 dry-run on changed C++ files
- [x] C++ unit-test target builds
- [x] Full unit suite
- [ ] `make lint`
- [ ] Coverage suite

```text
./build/tests/unittests "Pyramid exposes stored raw vectors"
  All tests passed (14 assertions in 1 test case; NSW and ODescent)

./build/tests/unittests
  All tests passed (62,903,974 assertions in 714 test cases)
```

## Compatibility Impact

- API/ABI compatibility: additive behavior; no public signature or serialization-format changes.
- Behavior change: `store_raw_vector=true` now enables the existing public raw-vector API for Pyramid. Precise distance-by-id uses the stored FP32 vectors.

- Raw-vector storage allocation and updates retain the behavior already on `main`.
- ODescent now populates the existing label reverse map during Build; NSW already does this through its Add path.
- No new locking or concurrency model is introduced.

## Documentation Impact

- [x] Updated English and Chinese Pyramid parameter documentation.

## Risk and Rollback

- Risk level: low.
- Rollback plan: revert this PR to disable the public Pyramid raw-vector feature mapping.

## Checklist

- [x] Linked the relevant issue.
- [x] Added focused regression coverage.
- [x] Considered API compatibility.
- [x] Updated user-facing documentation.
- [x] Commit messages follow project conventions.


src/algorithm/pyramid/pyramid.h
```diff
@@ -371,16 +392,17 @@ class Pyramid : public InnerIndexInterface {
...
 
     mutable std::shared_mutex resize_mutex_;        // guards resize operations
-    std::mutex cur_element_count_mutex_;            // guards cur_element_count_ updates
+    mutable std::mutex cur_element_count_mutex_;    // guards cur_element_count_ updates
     std::string graph_type_{GRAPH_TYPE_VALUE_NSW};  // graph algorithm type
 
     std::mutex entry_point_mutex_;  // guards entry-point selection
     std::default_random_engine level_generator_{
         2021};                              // random number generator for level promotion
     ReorderInterfacePtr reorder_{nullptr};  // reorder helper (if use_reorder_)
```
LHT129, line 440: [critical] `has_raw_vector_` is used in `pyramid.cpp` (`check_and_init_raw_vector` and `InitFeatures`) but is never declared as a member variable in `pyramid.h`. This will cause a compilation error.

The variable is assigned in `check_and_init_raw_vector`:
```cpp
has_raw_vector_ = raw_vector_ != nullptr;
```

And read in `InitFeatures`:
```cpp
if (has_raw_vector_) {
    this->index_feature_list_->SetFeature(IndexFeature::SUPPORT_GET_RAW_VECTOR_BY_IDS);
}
```

But `pyramid.h` only declares `create_new_raw_vector_` and `raw_vector_`. Please add `bool has_raw_vector_{false};` to the member declarations in `pyramid.h`.

src/algorithm/pyramid/pyramid_zparameters.cpp
```diff
@@ -180,6 +180,29 @@ PyramidParameters::FromJson(const JsonType& json) {
...
+        if (not flatten_json.Contains(QUANTIZATION_PARAMS_KEY)) {
+            return;
+        }
+        const auto& quantization_json = flatten_json[QUANTIZATION_PARAMS_KEY];
+        if (quantization_json.Contains(HOLD_MOLDS) ||
+            param->quantizer_parameter->GetTypeName() != QUANTIZATION_TYPE_VALUE_FP32) {
+            return;
+        }
+        std::dynamic_pointer_cast<FP32QuantizerParameter>(param->quantizer_parameter)->hold_molds =
+            true;
```
LHT129: [suggestion] `std::dynamic_pointer_cast<FP32QuantizerParameter>(param->quantizer_parameter)->hold_molds = true;` dereferences the result without a null check. Although `GetTypeName()` returned `"fp32"`, the underlying `quantizer_parameter` may not actually be an `FP32QuantizerParameter` instance (e.g. if a subclass also reports the same type name). A null dereference here would crash during parameter parsing.

Consider:
```cpp
auto fp32_param = std::dynamic_pointer_cast<FP32QuantizerParameter>(param->quantizer_parameter);
if (fp32_param) {
    fp32_param->hold_molds = true;
}
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -228,6 +247,9 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     if (use_reorder_) {
         precise_codes_->BatchInsertVector(data_vectors, data_num);
     }
+    if (create_new_raw_vector_) {
+        raw_vector_->BatchInsertVector(data_vectors, data_num);
```
LHT129, line 237: [suggestion] The ODescent build path (`build_by_odescent`) inserts raw vectors via `raw_vector_->BatchInsertVector` but does not populate `label_table_` through `Insert` calls. The preceding `memcpy` only initializes the raw label buffer; `label_remap_` and the label-table count are never set. Consequently, vectors inserted through ODescent cannot be reached via the newly advertised `GetRawVectorByIds`/`GetDataByIds` APIs — querying by label will report them as missing, even though `GetVectorByInnerId` works correctly.

Consider building the `LabelTable` through `Insert` or explicitly rebuilding its remap/count after the raw vector batch insert, and exercise the public label-based API in the ODescent test case.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -787,6 +909,10 @@ Pyramid::Add(const DatasetPtr& base) {
                     precise_codes_->InsertVector(data_vectors + dim_ * i,
                                                  valid_id_count + local_cur_element_count);
                 }
+                if (create_new_raw_vector_) {
```
LHT129: [suggestion] In `Pyramid::Add`, `label_table_->Insert` publishes the label (line 783 in main) before `raw_vector_->InsertVector` executes (new line ~912). This means a concurrent `GetRawVectorByIds` call can resolve the label via `LabelTable::GetIdByLabel` and then attempt to read the raw vector before it has been written, leading to reading uninitialized or stale data.

The same issue exists for `base_codes_` and `precise_codes_` insertion, but the new `raw_vector_` insertion makes the window wider. Consider reordering so that all code cell writes complete before the label becomes visible, or holding the resize lock during the entire Add critical section.

LHT129 (commented):
[suggestion] `ExportModel` does not export `raw_vector_` data. When `create_new_raw_vector_` is true, the exported model will lose raw vector storage, causing `GetRawVectorByIds` to fail on the exported index. Consider adding:
```cpp
if (create_new_raw_vector_) {
    this->raw_vector_->ExportModel(index->raw_vector_);
}
```
in `Pyramid::ExportModel` (pyramid.cpp, around line 877).

src/algorithm/pyramid/pyramid.cpp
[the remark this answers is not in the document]
jac0626: Fixed in 44608a10. The ODescent build now populates LabelTable through Insert, which builds the reverse map and count. The regression test now retrieves both vectors through the label-based GetDataByIds API.

src/algorithm/pyramid/pyramid.h
jac0626: No additional member is needed: has_raw_vector_ is the protected member inherited from InnerIndexInterface (src/algorithm/inner_index_interface.h). The rebased Debug build and Pyramid tests compile and pass with that inherited field.

src/algorithm/pyramid/pyramid_zparameters.cpp
jac0626: Fixed in 44608a10. The cast result is now checked with CHECK_ARGUMENT before use; a mismatched concrete parameter fails parsing instead of crashing or silently skipping the required legacy restoration.

src/algorithm/pyramid/pyramid.cpp
jac0626: Fixed in 44608a10. This duplicates the ODescent label-remap finding: Build now uses LabelTable::Insert and the test exercises label-based raw retrieval.

src/algorithm/pyramid/pyramid.cpp
jac0626: Fixed in 44608a10. Base, precise, and dedicated raw cells are now written before LabelTable::Insert publishes the label, so label-based readers cannot observe an unwritten vector.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -652,6 +735,15 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
                     loaded_precise_codes = true;
                 }
                 break;
+            case StreamSerializationTag::RAW_VECTOR:
+                loaded_raw_vector = true;
+                if (this->create_new_raw_vector_) {
```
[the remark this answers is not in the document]
jac0626: Fixed in 44608a10. The factory load override helper is now shared by HGraph and Pyramid and applies base, precise, and raw IO placement before construction. The regression removes the producer directory and loads successfully with memory IO overrides.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -787,6 +909,10 @@ Pyramid::Add(const DatasetPtr& base) {
                     precise_codes_->InsertVector(data_vectors + dim_ * i,
                                                  valid_id_count + local_cur_element_count);
                 }
+                if (create_new_raw_vector_) {
+                    raw_vector_->InsertVector(data_vectors + dim_ * i,
```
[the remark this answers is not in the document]
jac0626: Fixed in 44608a10 by staging every code-cell write before publishing the label. If a backend write throws, the label, logical count, and hierarchy remain unpublished; retry uses the same inner ID and overwrites any unreachable partial cell state.

jac0626:
Review follow-up: I also evaluated the non-inline ExportModel suggestion and intentionally made no change. Pyramid::ExportModel exports reusable quantizer/model state, not indexed vector data. A dedicated Pyramid raw cell is constrained to FP32, which has no trained model state to transfer; the target is already constructed with an empty raw cell from the same parameters. Copying raw vectors would violate the existing empty-model contract rather than fix retrieval.

jac0626:
Final rebase note: upstream/main advanced again while validation was running. The review-fix commit referenced in the thread replies was rewritten from 44608a10 to d8dc386a; the fixes and tests are unchanged.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -420,6 +444,45 @@ Pyramid::GetNumberRemoved() const {
...
+        (void)name;
+        memory += usage;
+    }
+    return memory;
+}
+
+std::unordered_map<std::string, uint64_t>
+Pyramid::GetMemoryUsageDetail() const {
+    std::lock_guard count_lock(cur_element_count_mutex_);
+    std::unique_lock resize_lock(resize_mutex_);
```
LHT129: [suggestion] `GetMemoryUsageDetail()` acquires `unique_lock` on `resize_mutex_`, which blocks concurrent search, `GetVectorByInnerId`, and distance-by-id calls. Since this method only reads state, a `shared_lock` would be sufficient and would avoid serializing all read operations against memory-usage queries.

Consider changing `std::unique_lock resize_lock(resize_mutex_)` to `std::shared_lock resize_lock(resize_mutex_)`.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -168,6 +170,23 @@ IndexNode::Serialize(StreamWriter& writer) const {
...
+uint64_t
+IndexNode::GetMemoryUsage() const {
+    std::shared_lock lock(mutex_);
+    uint64_t memory = sizeof(IndexNode) + ids_.capacity() * sizeof(InnerIdType);
+    memory +=
+        children_.bucket_count() * (sizeof(decltype(children_)::value_type) + sizeof(uint32_t));
+    for (const auto& [key, child] : children_) {
+        memory += key.capacity() + 1;
+        memory += child->GetMemoryUsage();
+    }
```
LHT129: [suggestion] `IndexNode::GetMemoryUsage()` multiplies `children_.bucket_count()` by `value_type` size, which overestimates memory because `bucket_count()` is the hash-table bucket count, not the number of stored entries. Use `children_.size()` instead to get the actual number of child nodes.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -929,6 +1061,18 @@ static const std::string HGRAPH_PARAMS_TEMPLATE =
                 "{HOLD_MOLDS}": false
             }
         },
+        "{STORE_RAW_VECTOR_KEY}": false,
+        "{RAW_VECTOR_KEY}": {
+            "{IO_PARAMS_KEY}": {
+                "{TYPE_KEY}": "{IO_TYPE_VALUE_BLOCK_MEMORY_IO}",
+                "{IO_FILE_PATH_KEY}": "{DEFAULT_FILE_PATH_VALUE}"
```
LHT129, line 1165: [suggestion] The `raw_vector` cell defaults to the same `./default_file_path` used by base and precise codes. With a documented configuration such as SQ8 `base_io_type=buffer_io` and `raw_vector_io_type=buffer_io`, both cells would open the same file path, leading to corruption. Consider using a distinct default path (e.g. `./default_raw_vector_path`) or requiring an explicit path when the raw-vector backend is file-backed.

LHT129 (commented):
Review of the latest push (44608a10). This push addresses several prior review comments (label publication order, null-pointer checks, factory load path). A few remaining observations below.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1362,14 +1520,119 @@ void
...
+    }
     codes->Decode(buffer, data);
     if (release) {
         codes->Release(buffer);
     }
 }
 
+void
+Pyramid::check_and_init_raw_vector(const FlattenInterfaceParamPtr& raw_vector_param,
+                                   const IndexCommonParam& common_param) {
```
LHT129: [suggestion] `check_and_init_raw_vector` checks `raw_vector_param->io_parameter->GetTypeName()` to decide whether to attempt aliasing, but the aliasing decision should be based on the *candidate* cell's IO type (i.e. whether `base_codes_` or `precise_codes_` is in-memory FP32), not the requested raw-vector backend. If the user requests `raw_vector_io_type=buffer_io` but `base_codes_` happens to be in-memory FP32, aliasing is still desirable and safe.

Consider calling `find_raw_vector_source(true)` unconditionally (regardless of `raw_vector_param` IO type) and only falling back to `MakeInstance` when no alias is found.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1362,14 +1520,119 @@ void
...
+    if (raw_vector_param == nullptr) {
+        return;
+    }
+
+    auto io_type_name = raw_vector_param->io_parameter->GetTypeName();
+    if (io_type_name == IO_TYPE_VALUE_BLOCK_MEMORY_IO || io_type_name == IO_TYPE_VALUE_MEMORY_IO) {
+        raw_vector_ = find_raw_vector_source(true);
+    }
```
LHT129: [note] `check_and_init_raw_vector` sets `has_raw_vector_ = raw_vector_ != nullptr` at line 1546, but `raw_vector_` is guaranteed non-null at this point (either from `find_raw_vector_source` or `MakeInstance`). The null check is redundant; `has_raw_vector_` will always be `true` when this code is reached. Consider setting `has_raw_vector_ = true` directly.

LHT129 (commented):
Additional observations on the latest push.

LHT129 (commented):
## Code Review for PR #2503

### [critical] Missing `has_raw_vector_` declaration in `pyramid.h`

`has_raw_vector_` is assigned in `check_and_init_raw_vector()` (pyramid.cpp line ~1558) and read in `InitFeatures()` (pyramid.cpp line ~999), but this member variable is never declared in `pyramid.h`. The header only declares `create_new_raw_vector_`. This will cause a compilation error.

**Fix**: Add `bool has_raw_vector_{false};` to the member variables section of `pyramid.h`, alongside `create_new_raw_vector_`:
```cpp
bool create_new_raw_vector_{false};  // whether raw_vector_ owns separate storage
bool has_raw_vector_{false};         // whether raw vector retrieval is available
```

### [suggestion] Unnecessary `FlattenInterface` allocation in `check_and_init_raw_vector`

In `check_and_init_raw_vector()`, `FlattenInterface::MakeInstance(raw_vector_param, common_param)` is always called first (line ~1543), even when `raw_vector_` will ultimately alias `base_codes_` or `precise_codes_` via `find_raw_vector_source()`. When aliasing succeeds, the newly constructed instance is leaked (overwritten by the aliased pointer). Consider checking for an aliasable source first, and only constructing a new instance when necessary.

### [note] `CalDistanceById` precedence order

Both `CalcDistanceById` overloads override `flat` with `raw_vector_` when `raw_vector_ != nullptr && calculate_precise_distance`. This means `raw_vector_` takes precedence over `precise_codes_` which takes precedence over `base_codes_`. This is correct behavior (raw vectors are always the most precise), but the precedence chain is implicit. A brief comment noting the priority would help future readers.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks for calling this out. We traced the actual trigger path: DEFAULT_FILE_PATH_VALUE is a placeholder ignored by the default in-memory backends, store_raw_vector is false by default, and disk-backed cells are documented to require an explicit *_file_path. A collision therefore requires an incomplete configuration that enables at least two file-backed cells while omitting or reusing their paths. HGraph already has the same template behavior for graph/base/precise/raw, so assigning only Pyramid raw a special default would be an incomplete PR-specific workaround. We are leaving this PR unchanged; if stricter validation is desired, it should be handled generically for all disk-backed cells.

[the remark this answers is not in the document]
jac0626: Thanks for calling this out. We traced the actual trigger path: DEFAULT_FILE_PATH_VALUE is a placeholder ignored by the default in-memory backends, store_raw_vector is false by default, and disk-backed cells are documented to require an explicit *_file_path. A collision therefore requires an incomplete configuration that enables at least two file-backed cells while omitting or reusing their paths. HGraph already has the same template behavior for graph/base/precise/raw, so assigning only Pyramid raw a special default would be an incomplete PR-specific workaround. We are leaving this PR unchanged; if stricter validation is desired, it should be handled generically for all disk-backed cells.

LHT129 (commented):
This commit (19a65344) looks good overall. The changes are well-scoped:

1. **`build_by_odescent` label table fix**: Replacing `std::memcpy` with `LabelTable::Insert` is the correct fix — the old code bypassed `label_remap_` construction, which would break `GetIdByLabel` and other remap-dependent operations.

2. **`pyramid_zparameters.cpp` null-safety**: Adding `CHECK_ARGUMENT` after `dynamic_pointer_cast` is a good defensive improvement.

3. **`factory.cpp` refactor**: Renaming to `apply_flatten_streaming_load_parameters` and wiring it for Pyramid is clean and enables streaming load IO overrides for Pyramid.

4. **Test coverage**: The new `GetDataByIds` assertion and streaming load storage overrides test provide good coverage.

No blocking issues found.

LHT129 (commented):
## Code Review Summary

I have performed a comprehensive review of this PR (commit `19a65344`), which adds raw vector storage and retrieval support to the Pyramid index. The feature is well-implemented with thorough test coverage (9 test cases covering storage, restoration, aliasing, streaming, legacy compatibility, memory reporting, and IO changes).

### Areas already covered by existing review comments

The 30+ existing inline comments on this PR already provide thorough coverage of the key concerns:

- **Memory accounting**: `IndexNode::GetMemoryUsage()` overestimates due to `bucket_count()` usage; `GetMemoryUsageDetail()` takes an exclusive lock that blocks readers
- **Thread safety**: Label publication ordering in `Add`, synchronization between `GetRawVectorByIds` and mutations
- **Configuration**: Default file path collision for raw_vector cell, aliasing decision logic in `check_and_init_raw_vector`
- **Serialization**: Raw-vector placement policies during streaming load, raw payload presence detection

### What looks good

- The alias-vs-dedicated storage design (`create_new_raw_vector_` flag) is clean and avoids unnecessary duplication when base/precise codes are already FP32
- `restore_raw_vector_from_source()` correctly handles the case where raw vectors must be recovered from base/precise codes during deserialization
- The `restore_legacy_raw_fp32_param` backward-compatibility shim for cosine metric is properly scoped
- Test coverage is comprehensive across build, add, serialize/deserialize, streaming, and memory reporting paths
- The factory integration correctly applies `raw_vector_io_type` and `raw_vector_file_path` overrides for Pyramid streaming loads

### Verdict

No new critical or suggestion-level issues found beyond what has already been raised in existing review comments. The existing comments should be addressed before merging.

LHT129 (commented):
[suggestion] `ExportModel` (pyramid.cpp:1034-1051) does not propagate `raw_vector_` to the exported index. When `create_new_raw_vector_` is true, the new Pyramid instance created by `ExportModel` will have an empty `raw_vector_` (only initialized by `check_and_init_raw_vector` in the constructor, but its content is never copied from the source). This means `GetVectorByInnerId` and `GetRawVectorByIds` on the exported model will return incorrect data or crash.

Consider adding raw vector export similar to how `base_codes_` and `precise_codes_` are exported:

```cpp
if (create_new_raw_vector_) {
    if (index->raw_vector_ == nullptr) {
        throw VsagException(ErrorType::INTERNAL_ERROR,
                            "Export model's pyramid raw vector is empty");
    }
    this->raw_vector_->ExportModel(index->raw_vector_);
}
```

LHT129 (commented):
test

LHT129 (commented):
test2

LHT129 (commented):
test4

LHT129 (commented):
## Code Review for commit f1555ce6 ("fix: reject unsupported pyramid reader io")

This commit adds a safety gate to reject `reader_io` for Pyramid streaming load, which is a sensible defensive measure. The implementation is clean and well-tested.

### What changed

**`src/factory/factory.cpp`**: Added `allow_reader_io` parameter to `set_streaming_io_override()` and `apply_flatten_streaming_load_parameters()`. HGraph passes `true` (reader_io allowed), Pyramid passes `false` (reader_io rejected). The rejection is enforced via `CHECK_ARGUMENT` in `set_streaming_io_override`.

**`src/algorithm/pyramid/pyramid.cpp`**: Whitespace-only change — re-indented the `raw_storage_resized` assignment for consistency with the surrounding style.

**`src/algorithm/pyramid/pyramid_test.cpp`**: Added a `"streaming load rejects reader io"` test case that serializes a Pyramid index, then attempts to load it with `raw_vector_io_type=reader_io` and verifies the load fails with `INVALID_ARGUMENT` containing "reader_io".

### Review

**[suggestion] `src/factory/factory.cpp:259`** — The `allow_reader_io=false` for Pyramid vs `true` for HGraph is correct, but a brief comment explaining *why* reader_io is unsafe for Pyramid would help future maintainers. Is it because Pyramids raw vector block depends on random-access reads that reader_io cannot support? Or is this a temporary limitation pending implementation? Without context, the asymmetry looks arbitrary.

**[note] `src/algorithm/pyramid/pyramid_test.cpp:652-663`** — The test correctly validates rejection behavior. Consider also testing that `reader_io` is rejected for *all three* Pyramid blocks (base_codes, precise_codes, raw_vector), not just raw_vector, to ensure the gate covers the full surface area.

**[note] `src/factory/factory.cpp:118-119`** — The error message `"reader_io is not supported for this streaming index"` is clear. Consider making it more specific (e.g. naming "Pyramid") so that if this gate is extended to other index types in the future, the error message remains diagnostic.

Overall this is a straightforward, well-scoped fix. No blocking issues.

src/factory/factory.cpp
LHT129, line 1: [suggestion] The `allow_reader_io=false` for Pyramid vs `true` for HGraph is correct, but a brief comment explaining *why* reader_io is unsafe for Pyramid would help future maintainers.

src/algorithm/pyramid/pyramid_test.cpp
LHT129, line 1: [note] The test correctly validates rejection behavior. Consider also testing that `reader_io` is rejected for base_codes and precise_codes blocks (not just raw_vector), to ensure the gate covers the full surface area for Pyramid.

src/factory/factory.cpp
LHT129, line 1: [note] The error message `"reader_io is not supported for this streaming index"` is clear. Consider making it more specific (e.g. naming "Pyramid") so that if this gate is extended to other index types in the future, the error message remains diagnostic about which index rejected it.

LHT129 (commented):
I have reviewed this PR which adds raw vector retrieval support to the Pyramid index. The implementation is well-structured with proper error handling, backward compatibility, and comprehensive test coverage. Below are my findings:

**[note]** In `GetMemoryUsageDetail()` (`pyramid.cpp`), the function acquires `cur_element_count_mutex_` (a `std::mutex`) first, then `resize_mutex_` (a `std::shared_mutex`) exclusively. In other code paths (e.g., `Add()`, `resize()`), `resize_mutex_` is acquired before `cur_element_count_mutex_`. While `GetMemoryUsageDetail()` is a const read-only method and unlikely to interleave with `Add()`/`resize()` in practice, the lock ordering inversion could theoretically lead to deadlock. Consider using a `std::shared_lock` on `resize_mutex_` instead of `std::unique_lock` to reduce contention and eliminate the ordering concern.

**[note]** In `check_and_init_raw_vector()`, the assignment `has_raw_vector_ = raw_vector_ != nullptr;` is always `true` when reached because the function either returns early (when `raw_vector_param == nullptr`) or guarantees `raw_vector_` is non-null. This is harmless but redundant — consider simplifying to `has_raw_vector_ = true;` or removing the null check.

**[note]** The `build_by_odescent()` function changed from `std::memcpy` to `label_table_->Insert()` in a loop for label insertion. This is a behavioral change that may have performance implications for large datasets. If `label_table_->Insert()` has per-entry overhead (e.g., locking, validation), consider whether a batch insert API would be more appropriate.

Overall the PR looks solid. The test coverage is thorough, covering dedicated storage, FP32 code reuse, streaming serialization, legacy deserialization, cosine hold_molds, and memory usage reporting.

LHT129 (commented):
## Code Review Summary

This PR adds raw vector storage and retrieval support to the Pyramid index (869 additions, 11 deletions across 9 files). Overall the implementation follows the established HGraph raw-vector pattern and includes good test coverage (6 new test cases). The latest commit (f1555ce6) correctly rejects unsupported `reader_io` for Pyramid streaming loads.

### Issues already addressed in this review round (from prior comments):
- ✅ ODescent build now uses `LabelTable::Insert` instead of raw `memcpy`
- ✅ `restore_legacy_raw_fp32_param` now has null-check via `CHECK_ARGUMENT`
- ✅ Factory now applies streaming load parameters for Pyramid (base, precise, raw_vector IO overrides)
- ✅ `reader_io` is explicitly rejected for Pyramid streaming loads with a clear error message

### Remaining open items (noted in existing review comments):
1. **`IndexNode::GetMemoryUsage()` uses `bucket_count()`** (line 268) — should use `size()` for accurate memory accounting
2. **`GetMemoryUsageDetail()` uses exclusive `unique_lock`** (line 630) — `shared_lock` would avoid blocking concurrent reads
3. **Raw vector default file path conflicts** (line 1312) — same `./default_file_path` as base/precise codes
4. **`check_and_init_raw_vector` IO type check** (line 1805) — should check candidate cell IO type, not requested param IO type
5. **Redundant `has_raw_vector_` null check** (line 1808) — `raw_vector_` is guaranteed non-null at that point

### Positive observations:
- Good test coverage including edge cases: cosine mold handling, ODescent path, streaming serialization, concurrent memory usage, and reader_io rejection
- Clean separation of `find_raw_vector_source` / `restore_raw_vector_from_source` / `check_and_init_raw_vector`
- Proper handling of `create_new_raw_vector_` flag to distinguish aliased vs owned storage
- Documentation updated in both Chinese and English

No blocking issues found. The remaining suggestions are non-critical improvements for memory accounting accuracy, lock contention, and code clarity.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -826,8 +906,17 @@ Pyramid::Deserialize(StreamReader& reader) {
...
         raw_vector_->Deserialize(buffer_reader);
+    } else if (raw_vector_size > 0) {
+        auto raw_vector_begin = buffer_reader.GetCursor();
+        if (create_new_raw_vector_) {
+            raw_vector_->Deserialize(buffer_reader);
+            CHECK_ARGUMENT(buffer_reader.GetCursor() - raw_vector_begin == raw_vector_size,
+                           "serialized Pyramid raw vector size mismatch");
+        } else {
+            buffer_reader.Seek(raw_vector_begin + raw_vector_size);
+        }
```
LHT129: [critical] When `raw_vector_` aliases `base_codes_` or `precise_codes_` (FP32 reuse path), the legacy deserialization branch (`has_legacy_raw_vector == true`) unconditionally calls `raw_vector_->Deserialize(buffer_reader)`, which writes the serialized raw-vector block into the aliased storage — corrupting the already-deserialized base or precise codes.

This happens because `check_and_init_raw_vector` may set `raw_vector_ = base_codes_` (or `precise_codes_`) when the requested raw_vector IO is memory-backed and an in-memory FP32 source exists, leaving `create_new_raw_vector_ == false`. The legacy path then deserializes into `raw_vector_` without checking whether it owns separate storage.

Suggested fix: guard the legacy deserialization with `create_new_raw_vector_`:

```cpp
if (has_legacy_raw_vector) {
    if (create_new_raw_vector_) {
        raw_vector_->Deserialize(buffer_reader);
    } else {
        // raw_vector_ aliases base/precise; the block payload must be skipped.
        // Since the legacy format has no raw_vector_size in the footer,
        // the block size is unknown here — consider rejecting this combination
        // or computing the remaining payload size from the stream.
    }
}
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -757,12 +821,12 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
                 }
                 break;
             case StreamSerializationTag::RAW_VECTOR:
```
LHT129, line 774: [suggestion] In `read_streaming_body`, when `create_new_raw_vector_` is false and a `RAW_VECTOR` block is encountered, `loaded_raw_vector` is set to `true` but `ReadSeekableBlockPayload` is never called. This means the stream position is not advanced past the block payload, which will cause subsequent block parsing to read garbage.

The new-format path (`else if (raw_vector_size > 0)`) in `Deserialize` correctly handles this by calling `buffer_reader.Seek()` when `create_new_raw_vector_` is false. The streaming path should similarly consume (skip) the block payload.

Suggested fix: always call `ReadSeekableBlockPayload` to advance the stream, but only deserialize when `create_new_raw_vector_` is true:

```cpp
case StreamSerializationTag::RAW_VECTOR:
    loaded_raw_vector = true;
    ReadSeekableBlockPayload(
        block_reader, block_header, [this](StreamReader& block) {
            if (this->create_new_raw_vector_) {
                this->raw_vector_->Deserialize(block);
            }
        });
    break;
```

LHT129 (commented):
Test review comment

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -519,6 +540,45 @@ Pyramid::GetNumberRemoved() const {
...
+        (void)name;
+        memory += usage;
+    }
+    return memory;
+}
+
+std::unordered_map<std::string, uint64_t>
+Pyramid::GetMemoryUsageDetail() const {
+    std::lock_guard count_lock(cur_element_count_mutex_);
+    std::unique_lock resize_lock(resize_mutex_);
```
LHT129: [suggestion] `GetMemoryUsageDetail` acquires an exclusive `resize_mutex_` via `std::unique_lock`, which blocks all concurrent search operations (they require `shared_lock`). While memory reporting is not a hot path, frequent calls could impact search latency. Consider using `std::shared_lock` instead, since the method only reads internal state without modifying it.

```cpp
// current:
std::unique_lock resize_lock(resize_mutex_);
// suggestion:
std::shared_lock resize_lock(resize_mutex_);
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1647,6 +1757,106 @@ Pyramid::GetVectorByInnerId(InnerIdType inner_id, float* data) const {
...
+    }
+    if (raw_vector_ == nullptr) {
+        raw_vector_ = FlattenInterface::MakeInstance(raw_vector_param, common_param);
+        create_new_raw_vector_ = true;
+    }
+
+    CHECK_ARGUMENT(raw_vector_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_FP32,
+                   "Pyramid raw vector storage must use fp32 quantization");
+    const bool valid_cosine_raw_vector =
+        metric_ != MetricType::METRIC_TYPE_COSINE || raw_vector_->HoldMolds();
```
LHT129: [note] The assignment `has_raw_vector_ = raw_vector_ != nullptr;` in `check_and_init_raw_vector` is redundant. The method already sets `create_new_raw_vector_ = true` in the branches that create a new instance, and `has_raw_vector_` is always derivable from `raw_vector_ != nullptr`. Consider removing `has_raw_vector_` entirely and using `raw_vector_ != nullptr` directly, or at minimum removing this redundant assignment since it duplicates the same check already performed by callers.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -519,6 +540,45 @@ Pyramid::GetNumberRemoved() const {
...
+    for (const auto& [name, usage] : detail) {
+        (void)name;
+        memory += usage;
+    }
+    return memory;
+}
+
+std::unordered_map<std::string, uint64_t>
+Pyramid::GetMemoryUsageDetail() const {
+    std::lock_guard count_lock(cur_element_count_mutex_);
```
LHT129: [suggestion] `GetMemoryUsageDetail()` acquires `cur_element_count_mutex_` but never accesses `cur_element_count_` within the function body. The lock appears unnecessary here. If the intent is to ensure a consistent snapshot alongside other operations that modify `cur_element_count_`, consider adding a comment explaining the purpose, or remove the lock to avoid confusion.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -839,8 +919,17 @@ Pyramid::Deserialize(StreamReader& reader) {
     if (has_precise_reorder()) {
         precise_codes_->Deserialize(buffer_reader);
     }
-    if (raw_vector_ != nullptr) {
+    if (has_legacy_raw_vector) {
```
LHT129: [note] The `Deserialize` legacy path at line 922 calls `raw_vector_->Deserialize(buffer_reader)` without checking `create_new_raw_vector_`. This is safe because `CheckCompatibility` (line 902) already ensures `store_raw_vector` matches between the stored and loading parameters, so `raw_vector_` is always initialized when `has_legacy_raw_vector` is true. However, the asymmetry with the `raw_vector_size > 0` path (which does check `create_new_raw_vector_`) may confuse future readers. Consider adding a brief comment explaining why the check is unnecessary here.

LHT129 (commented):
## Review Summary

This PR adds `store_raw_vector` support to Pyramid, enabling `GetRawVectorByIds` and precise distance-by-id calculations. The implementation is thorough and well-tested, covering construction, serialization (both legacy and streaming), deserialization, memory reporting, and export.

### What was reviewed

The latest commit (259e04bc) fixes legacy serialization compatibility by detecting older Pyramid snapshots that lack the `raw_vector_size` footer field and handling their inline raw vector blocks correctly.

### Key observations

1. **Legacy compatibility**: The `Deserialize` path now correctly distinguishes three cases: legacy inline raw vectors (no `raw_vector_size` field), new format with explicit size, and no raw vector block at all. The `CheckCompatibility` guard ensures parameter mismatches are caught early.

2. **Memory reporting**: `GetMemoryUsageDetail()` correctly avoids double-counting by only reporting `raw_vector` when `create_new_raw_vector_` is true (dedicated storage). When aliased to base/precise codes, the memory is already accounted for under those entries.

3. **Cosine safety**: `find_raw_vector_source()` and `check_and_init_raw_vector()` both validate `HoldMolds()` for cosine metric, preventing accidental aliasing of normalized vectors as raw vectors. The `restore_legacy_raw_fp32_param` in `pyramid_zparameters.cpp` ensures backward compatibility for older FP32 parameter snapshots.

4. **Test coverage**: The test suite covers build, ODescent build, FP32 reuse, cosine preservation, legacy deserialization validation, memory reporting, IO storage round-trips (6 scenarios), and concurrent Add with memory queries. The new `legacy upstream raw block without size` test specifically validates the latest fix.

### Minor suggestions

Two inline comments were posted regarding code clarity (unnecessary mutex acquisition and legacy path asymmetry). Neither is blocking.

Overall, the implementation is solid and ready to merge after addressing the existing review feedback from other reviewers.

src/algorithm/pyramid/pyramid.cpp
LHT129: [suggestion] `GetMemoryUsageDetail()` acquires `cur_element_count_mutex_` and a unique `resize_mutex_`, which blocks concurrent `Add` and search operations for the duration of the memory walk. Since this is a public API (`GetMemoryUsage()` → `GetMemoryUsageDetail()`), frequent calls could introduce latency spikes.

Consider using `std::shared_lock` on `resize_mutex_` instead of `std::unique_lock`, since the function only reads internal state. The `cur_element_count_mutex_` lock is fine as-is since `Add` also acquires it first.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -225,13 +244,15 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     const auto* data_ids = base->GetIds();
 
     resize(data_num);
-    std::memcpy(label_table_->label_table_.data(), data_ids, sizeof(LabelType) * data_num);
+    for (InnerIdType inner_id = 0; inner_id < data_num; ++inner_id) {
```
LHT129, line 228: [note] The `build_by_odescent` change from `std::memcpy(label_table_->label_table_.data(), data_ids, ...)` to a loop calling `label_table_->Insert(inner_id, data_ids[inner_id])` appears to be a correctness fix (the original `memcpy` likely bypassed proper label-to-inner-id mapping initialization). This change is not directly related to raw vector support and could have been a separate commit. Please confirm this is intentional and not an accidental inclusion.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -519,6 +540,45 @@ Pyramid::GetNumberRemoved() const {
...
+    uint64_t hierarchy_memory = hierarchies_.bucket_count() *
+                                (sizeof(decltype(hierarchies_)::value_type) + sizeof(uint32_t));
+    for (const auto& [name, hierarchy] : hierarchies_) {
+        hierarchy_memory += name.capacity() + 1;
+        hierarchy_memory += sizeof(Hierarchy) + hierarchy->name.capacity() + 1;
+        hierarchy_memory += hierarchy->no_build_levels.capacity() * sizeof(int32_t);
+        hierarchy_memory += hierarchy->root->GetMemoryUsage();
+    }
+    memory_usage["hierarchies"] = hierarchy_memory;
+    return memory_usage;
```
LHT129: [suggestion] `GetMemoryUsageDetail()` reports `base_codes`, `precise_codes`, and `raw_vector` memory, but does not include the `reorder_` object's memory usage. When `use_reorder_` is true, the `FlattenReorder` instance at `reorder_` also consumes memory that should be accounted for in the detail report. Consider adding `reorder_->GetMemoryUsage()` (or equivalent) to the detail map.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1647,6 +1757,106 @@ Pyramid::GetVectorByInnerId(InnerIdType inner_id, float* data) const {
...
+            if (release) {
+                source->Release(codes);
+            }
+            if (not decoded) {
+                throw VsagException(
+                    ErrorType::READ_ERROR,
+                    fmt::format("failed to decode Pyramid raw vector {}", begin + offset));
+            }
+        }
+        raw_vector_->BatchInsertVector(vectors.data(),
```
LHT129: [suggestion] `restore_raw_vector_from_source()` calls `raw_vector_->BatchInsertVector(vectors.data(), current_batch_size)` without passing explicit inner IDs. This relies on `BatchInsertVector` assigning sequential inner IDs starting from the current `TotalCount()`. If `raw_vector_` was pre-sized via `Resize()` but `TotalCount()` is 0 (as checked at line 1813), this should work correctly. However, the implicit ID assignment is fragile — consider explicitly passing the inner IDs (e.g. `begin + offset`) to make the intent clear and avoid breakage if `BatchInsertVector` semantics change.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -828,6 +890,24 @@ Pyramid::Deserialize(StreamReader& reader) {
...
+    const bool has_raw_vector_size = basic_info.Contains(PYRAMID_RAW_VECTOR_SIZE);
+    uint64_t raw_vector_size = 0;
+    if (has_raw_vector_size) {
+        raw_vector_size = basic_info[PYRAMID_RAW_VECTOR_SIZE].GetUint64();
+    }
+    bool has_legacy_raw_vector = false;
+    if (basic_info.Contains(INDEX_PARAM)) {
+        auto index_param = std::make_shared<PyramidParameters>();
+        index_param->FromString(basic_info[INDEX_PARAM].GetString());
+        if (not this->create_param_ptr_->CheckCompatibility(index_param)) {
```
LHT129: [suggestion] The `Deserialize` path now validates parameter compatibility via `CheckCompatibility` (line 902). This is a good safety check, but it may reject valid deserialization when the stored index was created with different IO types (e.g. `block_memory_io` vs `buffer_io`) that are logically equivalent for the same quantizer. The streaming load path in `factory.cpp` handles this by overriding IO types before construction, but the non-streaming `Deserialize` path does not. Consider whether IO type differences should be excluded from the compatibility check, or document that non-streaming deserialization requires exact IO type matching.

src/algorithm/pyramid/pyramid.cpp
LHT129: [suggestion] `GetMemoryUsageDetail` acquires `cur_element_count_mutex_` but never reads `cur_element_count_` within the locked scope. The `resize_mutex_` acquired immediately after already serializes access to all the members being read (`base_codes_`, `precise_codes_`, `raw_vector_`, `hierarchies_`). Holding both locks adds unnecessary contention with `Add()` which also acquires `cur_element_count_mutex_` first.

Consider dropping the `cur_element_count_mutex_` lock here, or adding a comment explaining why both locks are needed.

src/algorithm/pyramid/pyramid.cpp
LHT129, line 228: [note] The `build_by_odescent` change from `std::memcpy(label_table_->label_table_.data(), ...)` to `label_table_->Insert(inner_id, ...)` is a good fix that uses the proper public API instead of directly accessing a private member. However, this change is unrelated to the raw vector feature and is not mentioned in the PR description. Consider calling it out explicitly or splitting into a separate commit for cleaner history.

src/factory/factory.cpp
```diff
@@ -122,14 +125,17 @@ set_streaming_io_override(JsonType& index_param,
...
-apply_hgraph_streaming_load_parameters(JsonType& index_param, const std::string& parameters) {
+apply_flatten_streaming_load_parameters(JsonType& index_param,
+                                        const std::string& parameters,
+                                        bool allow_reader_io) {
     auto load_json = JsonType::Parse(parameters.empty() ? "{}" : parameters);
     if (load_json.Contains(HGRAPH_BASE_IO_TYPE)) {
         require_string_load_parameter(load_json, HGRAPH_BASE_IO_TYPE);
         set_streaming_io_override(index_param,
                                   BASE_CODES_KEY,
                                   load_json[HGRAPH_BASE_IO_TYPE].GetString().c_str(),
```
LHT129, line 131: [note] Renaming `apply_hgraph_streaming_load_parameters` to `apply_flatten_streaming_load_parameters` with the new `allow_reader_io` parameter is a reasonable generalization, but this function is now shared between HGraph and Pyramid. If more index types adopt streaming load in the future, consider extracting this into a common utility rather than keeping it in `factory.cpp`.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1036,7 +1128,7 @@ Pyramid::resize(int64_t new_max_capacity) {
     if (has_precise_reorder()) {
         precise_codes_->Resize(new_max_capacity);
     }
-    if (raw_vector_ != nullptr) {
+    if (create_new_raw_vector_) {
         raw_vector_->Resize(new_max_capacity);
```
LHT129, line 1042: [suggestion] In `ExportModel`, the raw_vector_ is serialized only when `create_new_raw_vector_` is true, but `has_raw_vector_` can also be true when aliasing base_codes_ or precise_codes_. If ExportModel is called in the alias case, the raw vector data would be silently omitted. Consider adding a comment or assertion to clarify this is intentional (since the data is already exported via the aliased codes).

LHT129 (commented):
Thanks for this well-structured PR. The implementation handles the key concerns well:

- **Aliasing logic**: `check_and_init_raw_vector` + `find_raw_vector_source` correctly reuses existing in-memory FP32 base/precise codes when possible, and falls back to dedicated storage otherwise.
- **Cosine safety**: `HoldMolds()` checks prevent aliasing normalized FP32 codes as raw vectors for cosine metric.
- **Backward compatibility**: Legacy deserialization path (`has_legacy_raw_vector`) and `restore_legacy_raw_fp32_param` in `FromJson` handle older snapshots correctly.
- **Serialization robustness**: `raw_vector_size` footer field enables safe skip/validation of raw blocks, and `CheckCompatibility` is called before reading the body.
- **Memory reporting**: `GetMemoryUsageDetail` properly reports raw_vector only when `create_new_raw_vector_` is true, avoiding double-counting aliases.
- **Test coverage**: Comprehensive tests cover SQ8+raw, FP32 alias, ODescent build, cosine, legacy deserialize, IO storage changes, streaming load, and concurrent Add memory reporting.

No blocking issues found. The existing review comments from earlier revisions appear to have been addressed in this latest commit.

LHT129 (commented):
This PR maps the existing `store_raw_vector` infrastructure to Pyramid with minimal, focused changes. The implementation is clean and well-scoped:

- **`build_by_odescent`**: Replacing `std::memcpy` with `label_table_->Insert` is necessary for ODescent to build the label reverse map, enabling non-contiguous external IDs through `GetRawVectorByIds`.
- **`CalcDistanceById` / `CalDistanceById`**: Correctly prioritizes `raw_vector_` over `precise_codes_` when `calculate_precise_distance` is true, since FP32 raw vectors are always the most precise source.
- **`InitFeatures`**: Conditionally sets `SUPPORT_GET_RAW_VECTOR_BY_IDS` based on `has_raw_vector_`, consistent with how other indexes expose this feature.
- **Test coverage**: The new test case covers both NSW and ODescent graph types, verifying `GetRawVectorByIds`, `CalcDistanceById`, and `CalcDistancesById` with non-contiguous IDs.

No issues found. The changes are additive, backward-compatible, and well-tested.

src/algorithm/pyramid/pyramid.h
```diff
@@ -176,6 +176,7 @@ class Pyramid : public InnerIndexInterface {
         if (pyramid_param->store_raw_vector) {
             raw_vector_ =
                 FlattenInterface::MakeInstance(pyramid_param->raw_vector_param, common_param);
+            has_raw_vector_ = true;
         }
```
LHT129, line 180: [critical] `has_raw_vector_` is set in the constructor (pyramid.h:179) and checked in `InitFeatures()` (pyramid.cpp:1092), but it is never declared as a member variable in `pyramid.h`. The header only declares `raw_vector_` (line 422) but has no `bool has_raw_vector_{false};` declaration. This will cause a compile error. Add the missing member declaration in the private section of `pyramid.h`.

LHT129 (commented):
Code review for commit 08ef22d4.

LHT129 (approved):
LGTM

PR #2516, opened 2026-07-23
fix(eval): correct knn latency and QPS measurements

### LINKED ISSUE #2470: [bug](eval_performance): measure knn latency and qps at search boundaries
### Affected area

eval_performance

### Interface

cpp

### VSAG version

v1.0.0 (`efdaf17a`)

### Summary

`eval_performance` measures intervals between adjacent monitor callbacks rather
than timing each `KnnSearch` call. QPS is then derived from these intervals
instead of measured from parallel-region wall time.

Since commit `cabba248a368d21d43e26e77d385de03a1bca2b8`, per-query statistics
parsing also runs before the latency record point. JSON parsing, string
conversion, shared atomic updates, query preparation, scheduling, and monitor
lock contention therefore affect reported latency and QPS.

### Minimal reproducer

The current execution order is:

```cpp
auto result = this->index_->KnnSearch(query, topk, config_.search_param);

if (collect_statistics) {
    this->record_statistics(result.value());
}

monitor->Record(&record);
```

The QPS formula is:

```text
QPS = retained_samples * active_threads * 1000
      / sum(sample_intervals)
```

Therefore:

```text
QPS * latency_avg = active_threads * 1000
```

For example, with two active threads, one thread can complete only one 100 ms
query while the other completes about one hundred 1 ms queries. The slow
thread's first and only query creates no latency sample but still contributes
to `active_threads`. The reported QPS is approximately 2000 while actual
throughput is approximately 1000 QPS.

### Environment

```shell
cat: /etc/*-release: No such file or directory
cat: /etc/*-release: No such file or directory
- OS:
- vsag version: v1.0.0
- compiler version: Apple clang version 21.0.0 (clang-2100.1.1.101)

OS: Darwin 25.5.0 arm64
CPU: Apple M5 Pro
Commit: efdaf17a
```

### Expected behavior

- Measure each query immediately before and after `KnnSearch`.
- Compute QPS as successful query count divided by parallel-region wall time.
- Keep statistics extraction outside the performance timing pass.
- Use thread-local latency samples.
- Include first-query latency, or define an explicit warm-up phase.

### Actual behavior

- Each thread's first query is discarded.
- Later samples measure time between adjacent `Record()` calls.
- Samples include work outside `KnnSearch` and global monitor-lock waiting.
- Statistics collection parses JSON and updates shared atomics before recording.
- QPS assumes balanced, continuously saturated threads.
- Empty sample sets can cause division by zero or invalid percentile access.

### References

- `tools/eval/case/search_eval_case.cpp:76-96`
- `tools/eval/case/search_eval_case.cpp:154-198`
- `tools/eval/case/search_eval_case.cpp:303-346`
- `tools/eval/monitor/latency_monitor.cpp:28-55`
- `tools/eval/monitor/latency_monitor.cpp:79-98`
- `src/dataset_impl.cpp:602-612`
- https://github.com/antgroup/vsag/commit/cabba248a368d21d43e26e77d385de03a1bca2b8

### Additional context

Results before and after `cabba248` are not directly comparable. The impact is
expected to be largest for fast indexes and high-concurrency runs.

## Related issues

- [#408 — eval the toolkit calculates the recall incorrectly](https://github.com/antgroup/vsag/issues/408)
- [#1109 — eval_performance reports incorrect memory usage](https://github.com/antgroup/vsag/issues/1109)

---

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [x] Documentation
- [x] CI/Build/Infra

## Linked Issue

- Fixes: #2470

## What Changed

- Root cause: `LatencyMonitor` inferred latency from gaps between per-thread
  `Record()` callbacks. This dropped each worker's first query and included
  query preparation, statistics extraction, scheduling, and monitor-lock
  contention. QPS was then estimated from those intervals and active-thread
  count instead of measured wall time, producing biased results for imbalanced
  workloads.
- Measure every KNN and filtered-KNN search directly around
  `Index::KnnSearch` with `std::chrono::steady_clock`, including the first
  query.
- Compute QPS as successful query count divided by the measured search-pass
  wall time.
- Store latency samples in preallocated per-iteration slots and publish the
  batch after workers join, avoiding a lock in the timed path.
- Keep statistics extraction, recall calculation, and memory sampling outside
  the latency/QPS pass.
- Preserve existing metric keys and add measurement method, sample count,
  successful query count, and duration metadata.
- Handle empty or invalid timing batches safely.
- Add deterministic regression coverage, wire it into the default test target
  and PR CI, and document the new semantics in English and Chinese.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
eval_monitor_test: 5 test cases, 33 assertions passed
ASan target with ENABLE_TOOLS=OFF: eval_monitor_test passed
clang 15 + libomp compile check for search_eval_case.cpp: passed
clang-format 15 validation: passed
git diff --check: passed
GitHub workflow YAML parse: passed

A complete tools build with the default Apple clang remains blocked by the
repository's existing raw -fopenmp flag:
clang++: error: unsupported option '-fopenmp'
The changed search source was independently validated with clang 15 + libomp.
```

## Compatibility Impact

- API/ABI compatibility: None. The change is internal to `eval_performance`;
  existing metric keys are preserved and JSON output is extended with metadata.
- Behavior changes: Latency and QPS now use explicit search-boundary
  measurements. Results from the previous callback-interval implementation are
  not directly comparable.

- Performance impact: Measurement accuracy is improved. The timed path adds
  steady-clock reads but removes callback locking; configurations that need a
  separate statistics pass may take longer overall.
- Concurrency/thread-safety impact: Each OpenMP iteration writes to a unique
  preallocated sample slot, and the monitor receives the completed batch after
  the parallel region.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/en/src/resources/eval.md`,
    `docs/docs/zh/src/resources/eval.md`

## Risk and Rollback

- Risk level: Medium, because benchmark semantics intentionally change.
- Rollback plan: Revert this PR to restore the previous callback-interval
  implementation and output semantics.

## Checklist

- [x] I have linked the relevant issue (required for `kind/bug` and
  `kind/feature`; see "Linked Issue" above)
- [x] My commit messages follow project conventions (Conventional Commits,
  optional `[skip ci]` prefix)

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -27,6 +27,7 @@ jobs:
               - 'src/**'
               - 'include/**'
               - 'tests/**'
+              - 'tools/**'
               - 'CMakeLists.txt'
               - 'Makefile'
               - 'extern/**'
@@ -263,6 +264,8 @@ jobs:
           export LSAN_OPTIONS=suppressions=omp.supp
           chmod +x ./build/tests/functests
           ./scripts/testing/test_parallel_by_name.sh functests
+          chmod +x ./build/tests/eval_monitor_test
+          ./build/tests/eval_monitor_test -d yes
       - name: Upload Log
         uses: actions/upload-artifact@v5
         with:
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ test:                    ## Build and run unit tests.
 	cmake --build ${DEBUG_BUILD_DIR} --parallel ${COMPILE_JOBS}
 	./build/tests/unittests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
 	./build/tests/functests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
+	./build/tests/eval_monitor_test -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
 	./build/mockimpl/tests_mockimpl -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
 
 .PHONY: asan
--- a/docs/docs/en/src/resources/eval.md
+++ b/docs/docs/en/src/resources/eval.md
@@ -87,6 +87,24 @@ Note: under `global.exporters`, each entry is a **named** exporter (a YAML map),
 - **Latency**: average, P50/P95/P99
 - **Resource**: peak memory usage
 
+### Search Measurement Semantics
+
+- **Latency** is the elapsed wall duration of each measured `Index::KnnSearch` call, measured
+  with the monotonic `std::chrono::steady_clock`.
+- **QPS** is the number of successful queries divided by the wall duration, in seconds, of the
+  measured search pass.
+- Statistics extraction, recall calculation, and memory sampling run outside the performance
+  pass and do not contribute to latency or QPS.
+- Every measured query is included, including the first query on each worker. The tool does not
+  apply an implicit warm-up; run any desired warm-up explicitly before the measured pass.
+
+JSON results include `measurement_method`, `measurement_sample_count`,
+`measurement_successful_query_count`, and `measurement_duration(s)` so the measurement population
+and method can be audited.
+
+Results produced by older versions that measured intervals between monitor callbacks are not
+directly comparable with results that use these semantics.
+
 ## Search Modes
 
 `search_mode` accepts `knn`, `range`, `knn_filter`, and `range_filter`.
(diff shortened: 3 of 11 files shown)


wxyucs (approved):
lgtm

PR #2522, opened 2026-07-23
fix(eval): correct peak memory measurement

### LINKED ISSUE #1109: `eval_performance` Reports Incorrect Memory Usage (16777216.00 TB)
#### **1. Describe the bug**

When running the `eval_performance` script, the reported `memory_usage` metric shows an impossibly large value: `16777216.00 TB`. This is clearly a bug in how memory is being calculated or displayed.

Image

#### **2. To Reproduce**

1.  Run the `eval_performance` script with following configuration.

```bash
# PQ Performance Test Configuration

global:
  exporters:
    e1:
      format: "text"
      to: "stdout"
    e2:
      format: "json"
      to: "file://tmp/github-perf.json"

SIFT-pq:
    datapath: "/tmp/data/sift-128-euclidean.hdf5"
    type: "build"
    index_name: "ivf"
    create_params: '{"dim":128,"dtype":"float32","metric_type":"l2","index_param": {
            "buckets_count": 50,
            "ivf_train_type": "kmeans",
            "base_quantization_type": "pqfs",
            "precise_quantization_type": "fp32",
            "use_reorder": true,
            "base_pq_dim": 32,
            "precise_io_type": "async_io",
            "precise_file_path": "./precise_codes"
        }}'
    search_params: '{"ivf": {
            "scan_buckets_count": 10
        }}'
    index_path: "/tmp/sift-128-euclidean/index/hgraph_index"
    topk: 10
    search_mode: "knn"
    range: 0.5
    delete_index_after_search: false

### PR DESCRIPTION
**Pull request by jac0626**
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [x] CI/Build/Infra

## Linked Issue

- Fixes: #1109

## What Changed

- Replace the underflow-prone, point-in-time RSS calculation with a restartable background sampler that reads process RSS every 5 ms on Linux and macOS.
- Preserve `memory_peak(build/search)` while adding raw baseline, absolute peak, delta, availability, sample-count, and error fields; clamp negative deltas to zero.
- Measure only `Index::Build()` during build and the latency/QPS query pass during search; run KNN statistics outside monitored intervals.
- Reuse one evaluation dataset in `build,search` mode and release the build index before creating/deserializing the search index.
- Report index-owned memory through `index_memory(B)` in addition to process RSS and existing memory details.
- Add regression tests, macOS eval-test coverage, and synchronized English/Chinese documentation.

## Root Cause

The previous monitor subtracted two `uint64_t` RSS page counts without validating reads or ordering. If the final RSS was below the constructor-time baseline, subtraction wrapped near `UINT64_MAX`, producing `16777216.00 TB`. Build also sampled only after completion, search used a later intrusive query pass, and `build,search` retained duplicate datasets and index objects.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (described below)

Test details:

```text
cmake --build build --target unittests eval_performance --parallel 6
./build/tests/unittests '[eval]' --allow-running-no-tests
# All tests passed: 64 assertions in 11 test cases

# Ubuntu clang-format version 15.0.7

python3 -c "import yaml; yaml.safe_load(open('.github/workflows/pr-ci.yml'))"
git diff --check
```

The eval tests also passed 50 consecutive repetitions. A temporary 20,000-vector HDF5 `build,search` end-to-end run completed successfully and produced sane byte/RSS fields without the oversized TB value.

## Compatibility Impact

- API/ABI compatibility: none; changes are confined to the eval tool, tests, CI, and documentation.
- Behavior changes: memory output now represents sampled process-RSS peak delta for the measured phase. Existing human-readable `memory_peak(build/search)` keys remain available; additional diagnostics are emitted in JSON.

- Performance impact: no change to VSAG's `Index::KnnSearch()` implementation. When memory and latency are enabled together, one 5 ms RSS sampling thread runs concurrently and may add small benchmark noise; KNN statistics no longer contaminate latency/QPS measurements.
- Concurrency/thread-safety impact: adds a phase-scoped sampler thread guarded by private synchronization; it does not take index or per-query locks.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/en/src/resources/eval.md`, `docs/docs/zh/src/resources/eval.md`

## Risk and Rollback

- Risk level: medium; the change updates eval measurement boundaries and output diagnostics but does not modify the library API or index algorithms.
- Rollback plan: revert commit `af9b6b1e` to restore the previous eval monitor and pass lifecycle.

## Checklist

- [x] I have added/updated tests for the bug fix.
- [x] I have updated docs for the behavior change.

### DIFF
--- a/tools/eval/case/build_eval_case.cpp
+++ b/tools/eval/case/build_eval_case.cpp
@@ -16,6 +16,7 @@
 #include "./build_eval_case.h"
 
 #include <algorithm>
+#include <exception>
 #include <filesystem>
 #include <utility>
 
@@ -25,11 +26,54 @@
 
 namespace vsag::eval {
 
+namespace {
+
+class StartedMonitorGuard {
+public:
+    explicit StartedMonitorGuard(std::vector<MonitorPtr>& monitors) : monitors_(monitors) {
+    }
+
+    ~StartedMonitorGuard() {
+        while (stopped_monitor_count_ < started_monitor_count_) {
+            const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
+            ++stopped_monitor_count_;
+            try {
+                monitors_[monitor_id]->Stop();
+            } catch (...) {
+            }
+        }
+    }
+
+    void
+    StartAll() {
+        for (auto& monitor : monitors_) {
+            monitor->Start();
+            ++started_monitor_count_;
+        }
+    }
+
+    void
+    StopNext() {
+        const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
+        monitors_[monitor_id]->Stop();
+        ++stopped_monitor_count_;
+    }
+
+private:
+    std::vector<MonitorPtr>& monitors_;
+    uint64_t started_monitor_count_{0};
+    uint64_t stopped_monitor_count_{0};
+};
+
+}  // namespace
+
 BuildEvalCase::BuildEvalCase(const std::string& dataset_path,
                              const std::string& index_path,
                              vsag::IndexPtr index,
-                             EvalConfig config)
-    : EvalCase(dataset_path, index_path, index), config_(std::move(config)) {
+                             EvalConfig config,
+                             EvalDatasetPtr dataset)
+    : EvalCase(dataset_path, index_path, std::move(index), std::move(dataset)),
+      config_(std::move(config)) {
     this->init_monitors();
 }
 
@@ -68,16 +112,15 @@ BuildEvalCase::do_build() {
     } else {
         base->SparseVectors((const SparseVector*)this->dataset_ptr_->GetTrain());
     }
-    for (auto& monitor : monitors_) {
-        monitor->Start();
-    }
+    StartedMonitorGuard monitor_guard(monitors_);
+    monitor_guard.StartAll();
     auto build_index = index_->Build(base);
     if (not build_index.has_value()) {
         throw std::runtime_error(build_index.error().message);
     }
-    for (auto& monitor : monitors_) {
-        monitor->Record();
-        monitor->Stop();
+    for (auto monitor = monitors_.rbegin(); monitor != monitors_.rend(); ++monitor) {
+        (*monitor)->Record();
+        monitor_guard.StopNext();
     }
 }
 void
@@ -105,6 +148,11 @@ BuildEvalCase::process_result() {
     result["index_info"] = JsonType::parse(config_.build_param);
     result["action"] = "build";
     result["index"] = config_.index_name;
+    try {
+        result["index_memory(B)"] = this->index_->GetMemoryUsage();
+    } catch (const std::exception& e) {
+        logger_->Error(e.what());
+    }
     try {
         auto detail = this->index_->GetMemoryUsageDetail();
         for (const auto& [name, size] : detail) {
--- a/tools/eval/case/build_eval_case.cpp
+++ b/tools/eval/case/build_eval_case.cpp
@@ -26,6 +26,47 @@
 
 namespace vsag::eval {
 
+namespace {
+
+class StartedMonitorGuard {
+public:
+    explicit StartedMonitorGuard(std::vector<MonitorPtr>& monitors) : monitors_(monitors) {
+    }
+
+    ~StartedMonitorGuard() {
+        while (stopped_monitor_count_ < started_monitor_count_) {
+            const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
+            ++stopped_monitor_count_;
+            try {
+                monitors_[monitor_id]->Stop();
+            } catch (...) {
+            }
+        }
+    }
+
+    void
+    StartAll() {
+        for (auto& monitor : monitors_) {
+            monitor->Start();
+            ++started_monitor_count_;
+        }
+    }
+
+    void
+    StopNext() {
+        const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
+        monitors_[monitor_id]->Stop();
+        ++stopped_monitor_count_;
+    }
+
+private:
+    std::vector<MonitorPtr>& monitors_;
+    uint64_t started_monitor_count_{0};
+    uint64_t stopped_monitor_count_{0};
+};
+
+}  // namespace
+
 BuildEvalCase::BuildEvalCase(const std::string& dataset_path,
                              const std::string& index_path,
                              vsag::IndexPtr index,
@@ -83,16 +124,15 @@ BuildEvalCase::do_build() {
     if (this->dataset_ptr_->GetTrainPaths() != nullptr) {
         base->Paths(this->dataset_ptr_->GetTrainPaths());
     }
-    for (auto& monitor : monitors_) {
-        monitor->Start();
-    }
+    StartedMonitorGuard monitor_guard(monitors_);
+    monitor_guard.StartAll();
     auto build_index = index_->Build(base);
     if (not build_index.has_value()) {
         throw std::runtime_error(build_index.error().message);
     }
-    for (auto& monitor : monitors_) {
-        monitor->Record();
-        monitor->Stop();
+    for (auto monitor = monitors_.rbegin(); monitor != monitors_.rend(); ++monitor) {
+        (*monitor)->Record();
+        monitor_guard.StopNext();
     }
 }
 void
--- a/tools/eval/case/build_search_eval_case.h
+++ b/tools/eval/case/build_search_eval_case.h
@@ -25,19 +25,27 @@ public:
     BuildSearchEvalCase(const std::string& dataset_path,
                         const std::string& index_path,
                         vsag::IndexPtr index,
-                        EvalConfig config)
-        : EvalCase(dataset_path, index_path, index) {
-        build_ = EvalCase::MakeInstance(config, "build");
-        search_ = EvalCase::MakeInstance(config, "search");
-        config_ = std::move(config);
+                        EvalConfig config,
+                        EvalDatasetPtr dataset = nullptr)
+        : EvalCase(dataset_path, index_path, nullptr, std::move(dataset)),
+          config_(std::move(config)) {
+        build_ = std::make_shared<BuildEvalCase>(
+            dataset_path_, index_path_, std::move(index), config_, dataset_ptr_);
     }
 
     ~BuildSearchEvalCase() override = default;
 
     JsonType
     Run() override {
-        auto build_result = build_->Run();
-        auto search_result = search_->Run();
+        auto build = std::move(build_);
+        if (build == nullptr) {
+            build = EvalCase::MakeInstance(config_, "build", dataset_ptr_);
+        }
+        auto build_result = build->Run();
+        build.reset();
+
+        auto search = EvalCase::MakeInstance(config_, "search", dataset_ptr_);
+        auto search_result = search->Run();
         return merge_results(build_result, search_result);
     }
 
@@ -50,9 +58,8 @@ private:
     }
 
 private:
-    EvalCasePtr build_;
-    EvalCasePtr search_;
     EvalConfig config_;
+    EvalCasePtr build_;
 };
 
 }  // namespace vsag::eval
--- a/tools/eval/case/build_search_eval_case.h
+++ b/tools/eval/case/build_search_eval_case.h
@@ -27,18 +27,25 @@ public:
                         vsag::IndexPtr index,
                         EvalConfig config,
                         EvalDatasetPtr dataset = nullptr)
-        : EvalCase(dataset_path, index_path, std::move(index), dataset) {
-        build_ = EvalCase::MakeInstance(config, "build", dataset);
-        search_ = EvalCase::MakeInstance(config, "search", dataset);
-        config_ = std::move(config);
+        : EvalCase(dataset_path, index_path, nullptr, std::move(dataset)),
+          config_(std::move(config)) {
+        build_ = std::make_shared<BuildEvalCase>(
+            dataset_path_, index_path_, std::move(index), config_, dataset_ptr_);
     }
 
     ~BuildSearchEvalCase() override = default;
 
     JsonType
     Run() override {
-        auto build_result = build_->Run();
-        auto search_result = search_->Run();
+        auto build = std::move(build_);
+        if (build == nullptr) {
+            build = EvalCase::MakeInstance(config_, "build", dataset_ptr_);
+        }
+        auto build_result = build->Run();
+        build.reset();
+
+        auto search = EvalCase::MakeInstance(config_, "search", dataset_ptr_);
+        auto search_result = search->Run();
         return merge_results(build_result, search_result);
     }
 
@@ -51,9 +58,8 @@ private:
     }
 
 private:
-    EvalCasePtr build_;
-    EvalCasePtr search_;
     EvalConfig config_;
+    EvalCasePtr build_;
 };
 
 }  // namespace vsag::eval
--- a/tools/eval/case/search_eval_case.cpp
+++ b/tools/eval/case/search_eval_case.cpp
@@ -26,6 +26,7 @@
 #include "../monitor/latency_monitor.h"
 #include "../monitor/memory_peak_monitor.h"
 #include "../monitor/recall_monitor.h"
+#include "search_pass_runner.h"
 #include "typing.h"
 #include "vsag/filter.h"
 #include "vsag_exception.h"
@@ -57,8 +58,10 @@ private:
 SearchEvalCase::SearchEvalCase(const std::string& dataset_path,
                                const std::string& index_path,
                                vsag::IndexPtr index,
-                               EvalConfig config)
-    : EvalCase(dataset_path, index_path, index), config_(std::move(config)) {
+                               EvalConfig config,
+                               EvalDatasetPtr dataset)
+    : EvalCase(dataset_path, index_path, std::move(index), std::move(dataset)),
+      config_(std::move(config)) {
     auto search_mode = config_.search_mode;
     if (search_mode == "knn") {
         this->search_type_ = SearchType::KNN;
@@ -93,7 +96,8 @@ SearchEvalCase::init_latency_monitor() {
         if (config_.enable_percent_latency) {
             latency_monitor->SetMetrics("percent_latency");
         }
-        this->monitors_.emplace_back(std::move(latency_monitor));
+        this->latency_monitor_ = latency_monitor;
+        this->monitors_.emplace_back(this->latency_monitor_);
     }
 }
 
@@ -115,8 +119,7 @@ SearchEvalCase::init_recall_monitor() {
 void
 SearchEvalCase::init_memory_monitor() {
     if (config_.enable_memory) {
-        auto memory_peak_monitor = std::make_shared<MemoryPeakMonitor>("search");
-        this->monitors_.emplace_back(std::move(memory_peak_monitor));
+        this->memory_monitor_ = std::make_shared<MemoryPeakMonitor>("search");
     }
 }
 
@@ -152,39 +155,46 @@ SearchEvalCase::deserialize(std::ifstream& infile) {
 
 void
 SearchEvalCase::do_knn_search() {
+    SearchPassRunner::Run(
+        this->monitors_,
+        this->latency_monitor_,
+        this->memory_monitor_,
+        [this](const MonitorPtr& monitor) { this->run_knn_search_pass(monitor.get(), false); },
+        [this]() { this->run_knn_search_pass(nullptr, true); });
+}
+
+void
+SearchEvalCase::run_knn_search_pass(Monitor* monitor, bool collect_statistics) {
     uint64_t topk = config_.top_k;
     auto query_count = this->dataset_ptr_->GetNumberOfQuery();
     this->logger_->Debug("query count is " + std::to_string(query_count));
     auto min_query = std::max(static_cast<uint64_t>(query_count), config_.search_query_count);
-    for (uint64_t monitor_id = 0; monitor_id < this->monitors_.size(); ++monitor_id) {
-        auto& monitor = this->monitors_[monitor_id];
-        const bool collect_statistics = monitor_id == 0;
-        monitor->Start();
 
-        omp_set_num_threads(config_.num_threads_searching);
+    omp_set_num_threads(config_.num_threads_searching);
 #pragma omp parallel for schedule(dynamic)
-        for (int64_t id = 0; id < min_query; ++id) {
-            auto i = id % query_count;
-            auto query = vsag::Dataset::Make();
-            query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
-            const void* query_vector = this->dataset_ptr_->GetOneTest(i);
-            if (this->dataset_ptr_->GetVectorType() == DENSE_VECTORS) {
-                if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
-                    query->Float32Vectors((const float*)query_vector);
-                } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
-                    query->Int8Vectors((const int8_t*)query_vector);
-                }
-            } else {
-                query->SparseVectors((const SparseVector*)query_vector);
+    for (int64_t id = 0; id < min_query; ++id) {
+        auto i = id % query_count;
+        auto query = vsag::Dataset::Make();
+        query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
+        const void* query_vector = this->dataset_ptr_->GetOneTest(i);
+        if (this->dataset_ptr_->GetVectorType() == DENSE_VECTORS) {
+            if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
+                query->Float32Vectors((const float*)query_vector);
+            } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
+                query->Int8Vectors((const int8_t*)query_vector);
             }
-            auto result = this->index_->KnnSearch(query, topk, config_.search_param);
-            if (not result.has_value()) {
-                std::cerr << "query error: " << result.error().message << std::endl;
-                exit(-1);
-            }
-            if (collect_statistics) {
-                this->record_statistics(result.value());
-            }
+        } else {
+            query->SparseVectors((const SparseVector*)query_vector);
+        }
+        auto result = this->index_->KnnSearch(query, topk, config_.search_param);
+        if (not result.has_value()) {
+            std::cerr << "query error: " << result.error().message << std::endl;
+            exit(-1);
+        }
+        if (collect_statistics) {
+            this->record_statistics(result.value());
+        }
+        if (monitor != nullptr) {
             const int64_t* neighbors = result.value()->GetIds();
             int64_t* ground_truth_neighbors = dataset_ptr_->GetNeighbors(i);
             auto record = std::make_tuple(neighbors,
@@ -194,7 +204,6 @@ SearchEvalCase::do_knn_search() {
                                           result.value()->GetDim());
             monitor->Record(&record);
         }
-        monitor->Stop();
     }
 }
 
@@ -204,6 +213,16 @@ SearchEvalCase::do_range_search() {
 
 void
 SearchEvalCase::do_knn_filter_search() {
+    SearchPassRunner::Run(
+        this->monitors_,
+        this->latency_monitor_,
+        this->memory_monitor_,
+        [this](const MonitorPtr& monitor) { this->run_knn_filter_search_pass(monitor.get()); },
+        []() {});
+}
+
+void
+SearchEvalCase::run_knn_filter_search_pass(Monitor* monitor) {
     uint64_t topk = config_.top_k;
     auto query_count = this->dataset_ptr_->GetNumberOfQuery();
     auto train_labels = this->dataset_ptr_->GetTrainLabels();
@@ -216,33 +235,32 @@ SearchEvalCase::do_knn_filter_search() {
     }
     this->logger_->Debug("query count is " + std::to_string(query_count));
     auto min_query = std::max<int64_t>(query_count, 10000);
-    for (auto& monitor : this->monitors_) {
-        monitor->Start();
-        for (int64_t id = 0; id < min_query; ++id) {
-            auto i = id % query_count;
-            auto query = vsag::Dataset::Make();
-            query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
-            const void* query_vector = this->dataset_ptr_->GetOneTest(i);
-            if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
-                query->Float32Vectors((const float*)query_vector);
-            } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
-                query->Int8Vectors((const int8_t*)query_vector);
-            }
-            auto test_label = test_labels[i];
-            auto filter = std::make_shared<FilterObj>(
-                train_labels, test_label, this->dataset_ptr_->GetValidRatio(test_label));
-            auto result = this->index_->KnnSearch(query, topk, config_.search_param, filter);
-            if (not result.has_value()) {
-                std::cerr << "query error: " << result.error().message << std::endl;
-                exit(-1);
-            }
+
+    for (int64_t id = 0; id < min_query; ++id) {
+        auto i = id % query_count;
+        auto query = vsag::Dataset::Make();
+        query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
+        const void* query_vector = this->dataset_ptr_->GetOneTest(i);
+        if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
+            query->Float32Vectors((const float*)query_vector);
+        } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
+            query->Int8Vectors((const int8_t*)query_vector);
+        }
+        auto test_label = test_labels[i];
+        auto filter = std::make_shared<FilterObj>(
+            train_labels, test_label, this->dataset_ptr_->GetValidRatio(test_label));
+        auto result = this->index_->KnnSearch(query, topk, config_.search_param, filter);
+        if (not result.has_value()) {
+            std::cerr << "query error: " << result.error().message << std::endl;
+            exit(-1);
+        }
+        if (monitor != nullptr) {
             const int64_t* neighbors = result.value()->GetIds();
             int64_t* ground_truth_neighbors = dataset_ptr_->GetNeighbors(i);
             auto record = std::make_tuple(
                 neighbors, ground_truth_neighbors, dataset_ptr_.get(), query_vector, topk);
             monitor->Record(&record);
         }
-        monitor->Stop();
     }
 }
 
@@ -257,11 +275,19 @@ SearchEvalCase::process_result() {
         const auto& one_result = monitor->GetResult();
         EvalCase::MergeJsonType(one_result, result);
     }
+    if (this->memory_monitor_ != nullptr) {
+        EvalCase::MergeJsonType(this->memory_monitor_->GetResult(), result);
+    }
     result["action"] = "search";
     result["search_mode"] = config_.search_mode;
     result["index_info"] = JsonType::parse(config_.build_param);
     result["search_param"] = config_.search_param;
     result["index"] = config_.index_name;
+    try {
+        result["index_memory(B)"] = this->index_->GetMemoryUsage();
+    } catch (const std::exception& e) {
+        logger_->Error(e.what());
+    }
     try {
         auto detail = this->index_->GetMemoryUsageDetail();
         for (const auto& [name, size] : detail) {
--- a/tools/eval/case/search_eval_case.cpp
+++ b/tools/eval/case/search_eval_case.cpp
@@ -30,6 +30,7 @@
 #include "../monitor/latency_monitor.h"
 #include "../monitor/memory_peak_monitor.h"
 #include "../monitor/recall_monitor.h"
+#include "search_pass_runner.h"
 #include "search_timing.h"
 #include "typing.h"
 #include "vsag/filter.h"
@@ -156,8 +157,7 @@ SearchEvalCase::init_recall_monitor() {
 void
 SearchEvalCase::init_memory_monitor() {
     if (config_.enable_memory) {
-        auto memory_peak_monitor = std::make_shared<MemoryPeakMonitor>("search");
-        this->monitors_.emplace_back(std::move(memory_peak_monitor));
+        this->memory_monitor_ = std::make_shared<MemoryPeakMonitor>("search");
     }
 }
 
@@ -198,11 +198,27 @@ SearchEvalCase::deserialize(std::ifstream& infile) {
 
 void
 SearchEvalCase::do_knn_search() {
+    bool statistics_collected = false;
+    SearchPassRunner::Run(
+        this->monitors_,
+        this->latency_monitor_,
+        this->memory_monitor_,
+        [this](const MonitorPtr& monitor) { this->run_knn_search_pass(monitor.get(), false); },
+        [this, &statistics_collected]() {
+            this->run_knn_search_pass(nullptr, true);
+            statistics_collected = true;
+        });
+    if (not statistics_collected) {
+        this->run_knn_search_pass(nullptr, true);
+    }
+}
+
+void
+SearchEvalCase::run_knn_search_pass(Monitor* monitor, bool collect_statistics) {
     uint64_t topk = config_.top_k;
     auto query_count = this->dataset_ptr_->GetNumberOfQuery();
     this->logger_->Debug("query count is " + std::to_string(query_count));
     auto min_query = std::max(static_cast<uint64_t>(query_count), config_.search_query_count);
-
     auto prepare_query = [this](uint64_t query_id) {
         auto query = vsag::Dataset::Make();
         query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
@@ -222,49 +238,14 @@ SearchEvalCase::do_knn_search() {
         return std::make_pair(std::move(query), query_vector);
     };
 
-    bool statistics_collected = false;
-    for (auto& monitor : this->monitors_) {
-        const bool is_latency_monitor =
-            this->latency_monitor_ != nullptr and monitor.get() == this->latency_monitor_.get();
-        // Statistics share the first non-performance pass, or use the fallback pass below.
-        const bool collect_statistics = not is_latency_monitor and not statistics_collected;
-        monitor->Start();
-
-        omp_set_num_threads(config_.num_threads_searching);
-        if (is_latency_monitor) {
-            using Clock = std::chrono::steady_clock;
-            std::vector<double> latency_records(min_query);
-            SearchFailure search_failure;
-            const auto wall_start = Clock::now();
-#pragma omp parallel for schedule(dynamic)
-            for (int64_t id = 0; id < min_query; ++id) {
-                if (search_failure.Failed()) {
-                    continue;
-                }
-                auto i = static_cast<uint64_t>(id) % query_count;
-                auto query_and_vector = prepare_query(i);
-                auto& query = query_and_vector.first;
-                auto [result, latency_ms] = MeasureSearch(
-                    [&]() { return this->index_->KnnSearch(query, topk, config_.search_param); });
-                if (not result.has_value()) {
-                    search_failure.Record(result.error().message);
-                    continue;
-                }
-                latency_records[static_cast<uint64_t>(id)] = latency_ms;
-            }
-            search_failure.ThrowIfFailed();
-            const auto wall_end = Clock::now();
-            auto timing_batch = LatencyTimingBatch{
-                std::move(latency_records),
-                min_query,
-                std::chrono::duration<double>(wall_end - wall_start).count(),
-            };
-            this->latency_monitor_->SetTimingBatch(std::move(timing_batch));
-            monitor->Stop();
-            continue;
-        }
-
+    omp_set_num_threads(config_.num_threads_searching);
+    const bool is_latency_monitor =
+        this->latency_monitor_ != nullptr and monitor == this->latency_monitor_.get();
+    if (is_latency_monitor) {
+        using Clock = std::chrono::steady_clock;
+        std::vector<double> latency_records(min_query);
         SearchFailure search_failure;
+        const auto wall_start = Clock::now();
 #pragma omp parallel for schedule(dynamic)
         for (int64_t id = 0; id < min_query; ++id) {
             if (search_failure.Failed()) {
@@ -273,15 +254,44 @@ SearchEvalCase::do_knn_search() {
             auto i = static_cast<uint64_t>(id) % query_count;
             auto query_and_vector = prepare_query(i);
             auto& query = query_and_vector.first;
-            const void* query_vector = query_and_vector.second;
-            auto result = this->index_->KnnSearch(query, topk, config_.search_param);
+            auto [result, latency_ms] = MeasureSearch(
+                [&]() { return this->index_->KnnSearch(query, topk, config_.search_param); });
             if (not result.has_value()) {
                 search_failure.Record(result.error().message);
                 continue;
             }
-            if (collect_statistics) {
-                this->record_statistics(result.value());
-            }
+            latency_records[static_cast<uint64_t>(id)] = latency_ms;
+        }
+        search_failure.ThrowIfFailed();
+        const auto wall_end = Clock::now();
+        auto timing_batch = LatencyTimingBatch{
+            std::move(latency_records),
+            min_query,
+            std::chrono::duration<double>(wall_end - wall_start).count(),
+        };
+        this->latency_monitor_->SetTimingBatch(std::move(timing_batch));
+        return;
+    }
+
+    SearchFailure search_failure;
+#pragma omp parallel for schedule(dynamic)
+    for (int64_t id = 0; id < min_query; ++id) {
+        if (search_failure.Failed()) {
+            continue;
+        }
+        auto i = static_cast<uint64_t>(id) % query_count;
+        auto query_and_vector = prepare_query(i);
+        auto& query = query_and_vector.first;
+        const void* query_vector = query_and_vector.second;
+        auto result = this->index_->KnnSearch(query, topk, config_.search_param);
+        if (not result.has_value()) {
+            search_failure.Record(result.error().message);
+            continue;
+        }
+        if (collect_statistics) {
+            this->record_statistics(result.value());
+        }
+        if (monitor != nullptr) {
             SearchRecord record{result.value()->GetIds(),
                                 dataset_ptr_->GetNeighbors(i),
                                 dataset_ptr_.get(),
@@ -290,31 +300,8 @@ SearchEvalCase::do_knn_search() {
                                 topk};
             monitor->Record(&record);
         }
-        search_failure.ThrowIfFailed();
-        monitor->Stop();
-        statistics_collected = statistics_collected or collect_statistics;
     }
-
-    if (not statistics_collected) {
-        omp_set_num_threads(config_.num_threads_searching);
-        SearchFailure search_failure;
-#pragma omp parallel for schedule(dynamic)
-        for (int64_t id = 0; id < min_query; ++id) {
-            if (search_failure.Failed()) {
-                continue;
-            }
-            auto i = static_cast<uint64_t>(id) % query_count;
-            auto query_and_vector = prepare_query(i);
-            auto& query = query_and_vector.first;
-            auto result = this->index_->KnnSearch(query, topk, config_.search_param);
-            if (not result.has_value()) {
-                search_failure.Record(result.error().message);
-                continue;
-            }
-            this->record_statistics(result.value());
-        }
-        search_failure.ThrowIfFailed();
-    }
+    search_failure.ThrowIfFailed();
 }
 
 void
@@ -323,6 +310,16 @@ SearchEvalCase::do_range_search() {
 
 void
 SearchEvalCase::do_knn_filter_search() {
+    SearchPassRunner::Run(
+        this->monitors_,
+        this->latency_monitor_,
+        this->memory_monitor_,
+        [this](const MonitorPtr& monitor) { this->run_knn_filter_search_pass(monitor.get()); },
+        []() {});
+}
+
+void
+SearchEvalCase::run_knn_filter_search_pass(Monitor* monitor) {
     uint64_t topk = config_.top_k;
     auto query_count = this->dataset_ptr_->GetNumberOfQuery();
     auto train_labels = this->dataset_ptr_->GetTrainLabels();
@@ -335,48 +332,14 @@ SearchEvalCase::do_knn_filter_search() {
     }
     this->logger_->Debug("query count is " + std::to_string(query_count));
     auto min_query = std::max<int64_t>(query_count, 10000);
-    for (auto& monitor : this->monitors_) {
-        const bool is_latency_monitor =
-            this->latency_monitor_ != nullptr and monitor.get() == this->latency_monitor_.get();
-        monitor->Start();
-        if (is_latency_monitor) {
-            using Clock = std::chrono::steady_clock;
-            std::vector<double> latency_records(min_query);
-            const auto wall_start = Clock::now();
-            for (int64_t id = 0; id < min_query; ++id) {
-                auto i = id % query_count;
-                auto query = vsag::Dataset::Make();
-                query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
-                const void* query_vector = this->dataset_ptr_->GetOneTest(i);
-                if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
-                    query->Float32Vectors((const float*)query_vector);
-                } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
-                    query->Int8Vectors((const int8_t*)query_vector);
-                }
-                auto test_label = test_labels[i];
-                auto filter = std::make_shared<FilterObj>(
-                    train_labels, test_label, this->dataset_ptr_->GetValidRatio(test_label));
-                auto [result, latency_ms] = MeasureSearch([&]() {
-                    return this->index_->KnnSearch(query, topk, config_.search_param, filter);
-                });
-                if (not result.has_value()) {
-                    throw std::runtime_error("query error: " + result.error().message);
-                }
-                latency_records[static_cast<uint64_t>(id)] = latency_ms;
-            }
-            const auto wall_end = Clock::now();
-            auto timing_batch = LatencyTimingBatch{
-                std::move(latency_records),
-                static_cast<uint64_t>(min_query),
-                std::chrono::duration<double>(wall_end - wall_start).count(),
-            };
-            this->latency_monitor_->SetTimingBatch(std::move(timing_batch));
-            monitor->Stop();
-            continue;
-        }
-
+    const bool is_latency_monitor =
+        this->latency_monitor_ != nullptr and monitor == this->latency_monitor_.get();
+    if (is_latency_monitor) {
+        using Clock = std::chrono::steady_clock;
+        std::vector<double> latency_records(min_query);
+        const auto wall_start = Clock::now();
         for (int64_t id = 0; id < min_query; ++id) {
-            auto i = id % query_count;
+            auto i = static_cast<uint64_t>(id) % query_count;
             auto query = vsag::Dataset::Make();
             query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
             const void* query_vector = this->dataset_ptr_->GetOneTest(i);
@@ -388,10 +351,42 @@ SearchEvalCase::do_knn_filter_search() {
             auto test_label = test_labels[i];
             auto filter = std::make_shared<FilterObj>(
                 train_labels, test_label, this->dataset_ptr_->GetValidRatio(test_label));
-            auto result = this->index_->KnnSearch(query, topk, config_.search_param, filter);
+            auto [result, latency_ms] = MeasureSearch([&]() {
+                return this->index_->KnnSearch(query, topk, config_.search_param, filter);
+            });
             if (not result.has_value()) {
                 throw std::runtime_error("query error: " + result.error().message);
             }
+            latency_records[static_cast<uint64_t>(id)] = latency_ms;
+        }
+        const auto wall_end = Clock::now();
+        auto timing_batch = LatencyTimingBatch{
+            std::move(latency_records),
+            static_cast<uint64_t>(min_query),
+            std::chrono::duration<double>(wall_end - wall_start).count(),
+        };
+        this->latency_monitor_->SetTimingBatch(std::move(timing_batch));
+        return;
+    }
+
+    for (int64_t id = 0; id < min_query; ++id) {
+        auto i = static_cast<uint64_t>(id) % query_count;
+        auto query = vsag::Dataset::Make();
+        query->NumElements(1)->Dim(this->dataset_ptr_->GetDim())->Owner(false);
+        const void* query_vector = this->dataset_ptr_->GetOneTest(i);
+        if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_FLOAT32) {
+            query->Float32Vectors((const float*)query_vector);
+        } else if (this->dataset_ptr_->GetTestDataType() == vsag::DATATYPE_INT8) {
+            query->Int8Vectors((const int8_t*)query_vector);
+        }
+        auto test_label = test_labels[i];
+        auto filter = std::make_shared<FilterObj>(
+            train_labels, test_label, this->dataset_ptr_->GetValidRatio(test_label));
+        auto result = this->index_->KnnSearch(query, topk, config_.search_param, filter);
+        if (not result.has_value()) {
+            throw std::runtime_error("query error: " + result.error().message);
+        }
+        if (monitor != nullptr) {
             SearchRecord record{result.value()->GetIds(),
                                 dataset_ptr_->GetNeighbors(i),
                                 dataset_ptr_.get(),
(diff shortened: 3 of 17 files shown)


LHT129 (commented):
Code review for PR #2522: [WIP]fix(eval): correct peak memory measurement

---

**[suggestion] `tools/eval/monitor/memory_peak_monitor.cpp:1949` — `last_error_` not cleared on successful samples**

In `sample_memory()`, `last_error_` is never cleared when a sample succeeds. A transient failure leaves a stale error string even when subsequent samples succeed, making `memory_peak_error(<phase>)` misleading — it can report an old error despite a valid baseline and peak.

The failure count already captures transient issues; the error field should reflect the last sampling outcome. Suggested fix: add `last_error_.clear();` in the success branch of `sample_memory()`.

---

**[suggestion] `tools/eval/case/search_pass_runner.cpp:1316-1323` — `MonitorPassGuard` destruction order is implicit**

In `run_monitored_pass()`, when `memory_monitor` is not null, `memory_guard` is declared after `monitor_guard`, so `memory_guard` destructs first (memory stops before the ordinary monitor). This is correct behavior, but it relies on C++ local variable declaration order which is easy to break during refactoring. Consider making the order explicit, e.g. by calling `memory_guard.Stop()` before the end of scope, or adding a comment noting the intentional ordering.

---

**[suggestion] `tools/eval/case/build_search_eval_case.h:87-89` — `build_` null fallback creates ambiguity**

In `Run()`, if `build_` is null after `std::move`, the code falls back to `EvalCase::MakeInstance(config_, "build", dataset_ptr_)`. This creates a new index via `Factory::CreateIndex`, not the original index passed to the constructor. Since the constructor always sets `build_`, this fallback path is unreachable in practice. Consider removing the dead code or adding a comment explaining when this path could be triggered.

---

**[note] `tools/eval/case/search_eval_case.cpp:553` — `do_knn_filter_search()` passes empty statistics callback**

`do_knn_filter_search()` passes `[]() {}` as the `statistics_pass`, meaning KNN result statistics are not collected in filter search mode. This differs from `do_knn_search()` which collects statistics after all monitor passes. This appears intentional (filter search does not collect recall statistics), but is worth confirming.

LHT129 (commented):
[suggestion] Review of `codex/fix-eval-memory-1109` (af9b6b1e). This is a WIP PR so these are non-blocking suggestions.

### 1. `sample_memory()` does not clear `last_error_` on successful reads

In `tools/eval/monitor/memory_peak_monitor.cpp`, the `sample_memory()` method sets `last_error_` when a sample fails but never clears it when a subsequent sample succeeds. This means once any transient failure occurs, `memory_peak_error(<phase>)` will report a stale error for the remainder of the measurement phase, even though the baseline and peak are valid.

Suggested fix: add `last_error_.clear();` in the success branch of `sample_memory()`.

### 2. `StartedMonitorGuard::StopNext()` naming is misleading

In `tools/eval/case/build_eval_case.cpp`, `StopNext()` always stops the *last* started monitor (reverse-start order), but the name suggests it stops the "next" one in forward order. The destructor already handles reverse-order unwind correctly. Consider renaming to `StopLast()` to document the contract, or removing `StopNext()` entirely and letting the destructor handle all stops.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
-        max_memory_ = val2;
+void
+MemoryPeakMonitor::sample_memory() {
+    const auto sample = read_memory();
+    std::lock_guard<std::mutex> lock(state_mutex_);
+    if (sample.available) {
+        absolute_peak_bytes_ = std::max(absolute_peak_bytes_, sample.bytes);
+        ++sample_count_;
+        return;
     }
```
LHT129, line 254: [suggestion] `last_error_` is never cleared after a successful sample in `sample_memory()`. If a transient failure occurs (e.g., a single `/proc/self/statm` read fails) and subsequent samples succeed, `last_error_` retains the stale error string. This causes `GetResult()` to report a non-empty `memory_peak_error` even though the overall measurement succeeded.

The `failure_count_` field already captures transient failures. Consider clearing `last_error_` on successful samples so the error field reflects only the most recent outcome:

```cpp
if (sample.available) {
    absolute_peak_bytes_ = std::max(absolute_peak_bytes_, sample.bytes);
    ++sample_count_;
    last_error_.clear();  // clear stale error on success
    return;
}
```

tools/eval/case/build_search_eval_case.h
```diff
@@ -25,19 +25,27 @@ class BuildSearchEvalCase : public EvalCase {
...
-        build_ = EvalCase::MakeInstance(config, "build");
-        search_ = EvalCase::MakeInstance(config, "search");
-        config_ = std::move(config);
+                        EvalConfig config,
+                        EvalDatasetPtr dataset = nullptr)
+        : EvalCase(dataset_path, index_path, nullptr, std::move(dataset)),
+          config_(std::move(config)) {
+        build_ = std::make_shared<BuildEvalCase>(
+            dataset_path_, index_path_, std::move(index), config_, dataset_ptr_);
     }
```
LHT129, line 35: [note] `Run()` now creates a new `SearchEvalCase` on every call via `MakeInstance`, whereas previously `search_` was constructed once in the constructor. This means each `Run()` call will deserialize the index from disk again. While this is intentional (per the PR description, to release the build index before creating the search index), it is worth noting that `Run()` is no longer safe to call multiple times on the same `BuildSearchEvalCase` instance without re-incurring the full deserialization cost. If callers previously relied on `Run()` being repeatable without re-deserialization, this is a behavioral change.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 254: [suggestion] `last_error_` is never cleared after a successful sample. Once any transient failure occurs, `last_error_` remains non-empty even if all subsequent samples succeed, causing `memory_peak_error(<phase>)` to report a stale error. The `memory_peak_failed_sample_count` already captures failure history, so `last_error_` should reflect the most recent sampling outcome. In `sample_memory()`, add `last_error_.clear();` in the success path before the `return`.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
+    running_ = baseline.available;
+    worker_ready_ = true;
+    stop_condition_.notify_all();
+    if (not running_) {
+        return;
+    }
+    while (
+        not stop_condition_.wait_for(lock, sample_interval_, [this]() { return not running_; })) {
+        lock.unlock();
+        sample_memory();
```
LHT129, line 279: [note] In `sampling_loop()`, the final `lock.unlock(); sample_memory();` call runs `sample_memory()` without the mutex held. While `sample_memory()` internally acquires `state_mutex_`, the `running_` flag is already `false` at this point, so the `wait_for` predicate is satisfied and the loop has exited. This final sample is intentional (to capture one last reading after the stop signal), but consider documenting this with a brief comment since the unlock-before-call pattern is unusual.

tools/eval/case/build_eval_case.cpp
```diff
@@ -68,16 +112,15 @@ BuildEvalCase::do_build() {
...
+    monitor_guard.StartAll();
     auto build_index = index_->Build(base);
     if (not build_index.has_value()) {
         throw std::runtime_error(build_index.error().message);
     }
-    for (auto& monitor : monitors_) {
-        monitor->Record();
-        monitor->Stop();
+    for (auto monitor = monitors_.rbegin(); monitor != monitors_.rend(); ++monitor) {
+        (*monitor)->Record();
```
LHT129, line 134: [note] `MemoryPeakMonitor::Record()` is now a no-op since sampling moved to the background thread. The call to `monitor->Record()` in `do_build()` (line 122) still executes but has no effect for `MemoryPeakMonitor`. This is harmless but may confuse future readers. Consider either removing the `Record()` call or adding a comment noting that `MemoryPeakMonitor` samples continuously in the background and ignores explicit `Record()` calls.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
+MemoryPeakMonitor::sampling_loop() {
+    const auto baseline = read_memory();
+    std::unique_lock<std::mutex> lock(state_mutex_);
+    available_ = baseline.available;
+    baseline_bytes_ = baseline.available ? baseline.bytes : 0;
+    absolute_peak_bytes_ = baseline_bytes_;
+    sample_count_ = baseline.available ? 1 : 0;
+    failure_count_ = baseline.available ? 0 : 1;
+    last_error_ = baseline.error;
+    running_ = baseline.available;
```
LHT129, line 270: [suggestion] In `sampling_loop()`, the baseline read calls `read_memory()` directly (the private method that invokes `reader_()`), which bypasses the exception-safe wrapper. If the reader callback throws, the exception propagates out of `sampling_loop()` and will call `std::terminate` since uncaught exceptions in a thread terminate the process. The `MemoryPeakMonitor::read_memory()` wrapper already catches `std::exception` and unknown exceptions, converting them to `MemoryReadResult` with an error message. Consider using the wrapper for the baseline read so that a throwing reader is reported as a baseline failure rather than crashing the process.

Current:
```cpp
const auto baseline = read_memory();
```

Suggested:
```cpp
const auto baseline = this->read_memory();
```

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 254: [suggestion] `sample_memory()` sets `last_error_` on failure but never clears it on success. Once a transient sampling failure occurs, `last_error_` retains the error message permanently even if all subsequent samples succeed. This means `memory_peak_error(<phase>)` in the JSON output will report a stale error despite successful peak measurement. The `failure_count_` already captures transient issues, so `last_error_` should reflect the most recent sampling outcome.

Consider clearing `last_error_` on successful samples:
```cpp
if (sample.available) {
    absolute_peak_bytes_ = std::max(absolute_peak_bytes_, sample.bytes);
    ++sample_count_;
    last_error_.clear();
    return;
}
```

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
 
 namespace vsag::eval {
 
+namespace {
+
+class StartedMonitorGuard {
+public:
+    explicit StartedMonitorGuard(std::vector<MonitorPtr>& monitors) : monitors_(monitors) {
+    }
```
LHT129, line 34: [suggestion] When `index_->Build(base)` throws at line 128, the `StartedMonitorGuard` destructor will call `Stop()` on all started monitors in reverse order via the `while` loop in the destructor. However, the `StopNext()` method is designed to be called in LIFO order (matching the reverse iteration at line 133). The destructor's cleanup loop also uses LIFO order, which is correct. But there is a subtle issue: if `StartAll()` was called but `index_->Build(base)` throws before any `StopNext()` calls, the destructor will stop monitors in the order `monitors_[started-1]`, `monitors_[started-2]`, ..., `monitors_[0]`. This is the reverse of the start order, which is correct for LIFO cleanup. No issue here, just noting for review awareness.

However, there is a more concrete concern: the destructor catches all exceptions from `monitor->Stop()` silently. If a monitor's `Stop()` throws, the exception is swallowed and the remaining monitors are still stopped. This is the right behavior for a destructor, but consider whether a log message would help diagnose silent Stop failures during error recovery.

tools/eval/case/build_search_eval_case.h
```diff
@@ -27,18 +27,25 @@ class BuildSearchEvalCase : public EvalCase {
...
 
     ~BuildSearchEvalCase() override = default;
 
     JsonType
     Run() override {
-        auto build_result = build_->Run();
-        auto search_result = search_->Run();
+        auto build = std::move(build_);
+        if (build == nullptr) {
+            build = EvalCase::MakeInstance(config_, "build", dataset_ptr_);
```
LHT129, line 42: [note] In `Run()`, `build.reset()` is called before creating the search `EvalCase`. However, the `build` index was created with `std::move(index)` in the constructor, so the original `index` passed to `BuildSearchEvalCase` is now owned by the build case. After `build.reset()`, the build index is destroyed, which is correct for releasing memory before the search phase. But `EvalCase::MakeInstance(config_, "search", dataset_ptr_)` at line 43 creates a new `SearchEvalCase` that will deserialize the index from disk. This means the search phase uses a freshly deserialized index, not the in-memory build result. This is intentional per the PR description ("release the build index before creating/deserializing the search index"), but worth confirming that the index was serialized to disk by the build phase before being destroyed.

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace vsag::eval {
+
+namespace {
+
+class MonitorPassGuard {
+public:
+    explicit MonitorPassGuard(const MonitorPtr& monitor) : monitor_(monitor) {
+    }
+
```
LHT129, line 28: [suggestion] `MonitorPassGuard` stores `const MonitorPtr& monitor_` as a reference member. This is safe as long as the referenced `MonitorPtr` outlives the guard, which is true for all current call sites (the `monitor` parameter in `run_monitored_pass`, and `memory_monitor` in `SearchPassRunner::Run`). However, if someone later creates a `MonitorPassGuard` from a temporary `MonitorPtr`, the reference would dangle. Consider storing by value (`MonitorPtr monitor_`) to make the guard self-contained and eliminate this footgun. The cost is one additional `shared_ptr` copy, which is negligible for a scope guard used once per pass.

tools/eval/case/search_eval_case.cpp
```diff
@@ -198,11 +198,27 @@ SearchEvalCase::deserialize(std::ifstream& infile) {
...
+    SearchPassRunner::Run(
+        this->monitors_,
+        this->latency_monitor_,
+        this->memory_monitor_,
+        [this](const MonitorPtr& monitor) { this->run_knn_search_pass(monitor.get(), false); },
+        [this, &statistics_collected]() {
+            this->run_knn_search_pass(nullptr, true);
+            statistics_collected = true;
+        });
+    if (not statistics_collected) {
```
LHT129, line 211: [note] In `do_knn_search()`, `statistics_collected` is captured by reference in the lambda passed to `SearchPassRunner::Run`. The fallback `if (not statistics_collected)` at line 211 runs when no query pass executed (i.e., when `monitors_` is empty and there is no latency or memory monitor). In that case, `SearchPassRunner::Run` skips the statistics pass (since `ran_query_pass` is false), and the fallback here correctly runs the statistics pass. This is correct behavior, but the dual statistics-collection paths (inside `SearchPassRunner::Run` and the fallback here) could be simplified by having `SearchPassRunner::Run` always run statistics when at least one pass executed. Consider whether the fallback is still needed after the refactoring.

tools/eval/monitor/memory_peak_monitor.cpp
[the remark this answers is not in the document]
jac0626: Thanks for flagging this. We intentionally keep `last_error_` as the most recent failure reason for the phase, paired with a nonzero `memory_peak_failed_sample_count`. Clearing it after a later successful sample would preserve the failure count but discard the only diagnostic for that failure. It remains empty when the phase has no failed samples, so no code change is needed.

jac0626: Thanks. The intended semantics are that `last_error_` retains the most recent failed-sample reason for the phase, while `failure_count_` records how many samples failed. Clearing the error after a later success would leave a nonzero failure count without its diagnostic reason. The error remains empty when the phase has no failures, so we will keep the current behavior.

tools/eval/case/build_search_eval_case.h
jac0626: Confirmed that each outer `Run()` deserializes for the search phase, but this is not a new behavioral cost. The previous implementation also called `SearchEvalCase::Run()` on every outer run, and `SearchEvalCase::Run()` always deserializes the index. Creating a fresh search case now ensures that the build index has been released before the search index is created, so no change is needed.

tools/eval/monitor/memory_peak_monitor.cpp
jac0626: This field intentionally means the most recent failure reason, not the status of the most recent sample. Keeping it allows a nonzero `memory_peak_failed_sample_count` to retain its diagnostic context even if later samples succeed. It is empty when no sample failed during the phase, so we will not clear it on success.

tools/eval/monitor/memory_peak_monitor.cpp
jac0626: Confirmed; the final sample is intentional. The explicit unlock is required because `sample_memory()` acquires `state_mutex_` internally, so calling it while this lock is held would deadlock. The current sequence releases the loop lock and then captures the final reading after the stop signal. The code already expresses that ownership transition directly, so no additional change is needed.

tools/eval/case/build_eval_case.cpp
jac0626: Confirmed that `MemoryPeakMonitor::Record()` is a no-op. `do_build()` intentionally applies the generic monitor lifecycle to every monitor; individual monitor types may ignore `Record()`. Removing the call or branching on concrete monitor types would couple the build path to monitor implementations without changing behavior, so we will keep the generic call.

tools/eval/monitor/memory_peak_monitor.cpp
jac0626: The baseline already calls the exception-safe member wrapper. In this member function, unqualified `read_memory()` and `this->read_memory()` resolve to the same `MemoryPeakMonitor::read_memory()` method, which catches both `std::exception` and unknown exceptions and converts them to `MemoryReadResult`. Adding `this->` would not change behavior, so no code change is needed.

tools/eval/monitor/memory_peak_monitor.cpp
jac0626: We intentionally retain the most recent failure reason for the measurement phase. `failure_count_` records that failures occurred, and `last_error_` preserves the diagnostic for the latest one; clearing it after a successful sample would lose that diagnostic while the failure count remains nonzero. The field is empty for a phase with no failed samples, so the current behavior is retained.

tools/eval/case/build_eval_case.cpp
jac0626: Agreed that the reverse-order cleanup is correct. Swallowing `Stop()` exceptions in the destructor is deliberate: during stack unwinding it preserves the original build exception and still attempts to stop the remaining monitors. On the normal path, explicit `StopNext()` calls are outside the destructor and their exceptions still propagate. Adding logging would introduce a logging dependency into failure cleanup, so no change is needed here.

tools/eval/case/build_search_eval_case.h
jac0626: Confirmed. `BuildEvalCase::Run()` completes `do_build()`, serialization, and result processing before it returns. `build.reset()` executes only after that return, so the built index has already been serialized before its in-memory instance is destroyed. The subsequent search case intentionally deserializes a fresh index after build memory is released.

tools/eval/case/search_pass_runner.cpp
jac0626: All current construction sites pass an lvalue `MonitorPtr` whose scope outlives the guard, and `MonitorPassGuard` is private to this `.cpp` file in an anonymous namespace. Therefore the current reference member cannot dangle in any existing path. Storing by value would only guard against a hypothetical future misuse, so we are keeping the smaller current scope unchanged.

tools/eval/case/search_eval_case.cpp
jac0626: `SearchPassRunner::Run()` already invokes the statistics callback once whenever at least one query pass ran. This fallback covers the distinct zero-pass case where all monitors are disabled; without it, ordinary KNN evaluation would collect no index statistics. Moving that behavior into the generic runner would change its contract and contradict its existing zero-pass test, so the fallback remains.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
+    worker_ready_ = true;
+    stop_condition_.notify_all();
+    if (not running_) {
+        return;
+    }
+    while (
+        not stop_condition_.wait_for(lock, sample_interval_, [this]() { return not running_; })) {
```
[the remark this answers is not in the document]
jac0626: The described race exists if `Start()` and `Stop()` are called concurrently. Current production call sites use a sequential lifecycle, and the `Monitor` interface does not define concurrent lifecycle calls as supported. The suggested one-line assignment change is also insufficient to provide thread safety because access to `sampling_thread_` would remain unsynchronized. Supporting concurrent `Start()`/`Stop()` requires a separate lifecycle contract, mutex, and tests; we will not add a partial fix in this PR.

jac0626:
Follow-up on the non-inline review notes:

- **Monitor stop order:** the normal path already calls `memory_guard.Stop()` before `monitor_guard.Stop()`; exception unwinding preserves the same reverse-start order, and the ordering is covered by tests.
- **`build_` fallback:** this path is reachable on the second `BuildSearchEvalCase::Run()` because the first run moves `build_` out. It intentionally recreates the build case for repeat runs.
- **Filter statistics callback:** the empty internal-statistics callback preserves the pre-existing filter behavior. Configured monitors, including recall, still run; adding an index-statistics pass would execute another full filtered query batch and change evaluation cost.
- **`StopNext()` naming:** this is a private scope-guard helper whose LIFO behavior is local and tested. Renaming it is cosmetic and outside this fix.
- **`stop_sampling()` join behavior:** under the supported sequential lifecycle, a baseline failure is joined immediately in `Start()`, and a successful worker exits only after `Stop()` signals and joins it. Concurrent `Start()`/`Stop()` is not part of the monitor contract and requires full lifecycle synchronization rather than a local join change.
- **Opening `/proc/self/statm` per sample:** this is a measurable syscall cost, but it is an optimization topic rather than a correctness defect in this change. Caching an fd adds lifecycle and error-recovery behavior and should be evaluated separately with a benchmark. The deterministic fallback message remains available even when `errno` has no useful value.
- **CMake source factoring:** the current test target has no duplicate-definition or compile-configuration issue, and the repository already uses this direct-source test pattern for eval monitors. Introducing a shared eval library would broaden this bug-fix PR and can be handled as a separate build cleanup.

The repeated `last_error_` note is addressed in the corresponding inline threads: it intentionally retains the most recent failure reason so a nonzero failed-sample count keeps its diagnostic context. No code changes are needed for these non-inline notes.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 270: [suggestion] In `sampling_loop()`, the `running_` flag is overwritten with `baseline.available` after the mutex is acquired but before `worker_ready_` is signaled. If the baseline read succeeds, `running_` is set to `true`. However, `Start()` first calls `Stop()`, which sets `running_ = false` under the mutex. Since `Start()` releases the mutex before creating the thread, there is a window where `Stop()` could be called concurrently (from another thread) between the mutex release in `Start()` and the mutex acquisition in `sampling_loop()`. If `Stop()` sets `running_ = false` during this window, the subsequent `running_ = baseline.available` in `sampling_loop()` would flip it back to `true`, causing `Stop()` to hang in `join()`.

While the current call sites use a sequential lifecycle (as noted in earlier discussion), this implicit contract is fragile. Consider computing `running_` as `running_ && baseline.available` so that a stop request is never silently overwritten.

LHT129 (commented):
After a thorough review of all changed files in this PR, I find the implementation to be well-structured and comprehensive. The existing review comments (from Copilot, LHT129, and others) have already covered the substantive issues thoroughly, including:

- The `last_error_` clearing semantics in `sample_memory()` (author confirmed intentional)
- The `running_` flag race condition in `sampling_loop()` (author confirmed sequential lifecycle)
- The `MonitorPassGuard` reference member safety
- The `Record()` no-op design
- Various documentation and test coverage points

The PR adds robust background RSS sampling with proper thread synchronization, comprehensive unit tests (64 assertions across 11 test cases per the PR description), clear phase boundaries for memory measurement, and synchronized English/Chinese documentation. The `SearchPassRunner` abstraction cleanly separates pass orchestration from the eval cases.

I have no additional concerns to raise beyond what has already been discussed in the existing review threads.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 270: [suggestion] The `running_ = baseline.available` assignment on this line can silently overwrite a concurrent `Stop()` request.

When `Start()` releases `state_mutex_` before creating the sampling thread, another thread may call `Stop()`, which acquires the mutex, sets `running_ = false`, releases the mutex, and notifies `stop_condition_`. The sampling thread then acquires the mutex here and unconditionally sets `running_ = baseline.available` (typically `true`), discarding the stop signal. This causes `Stop()` to block indefinitely in `join()`.

While current call sites use a sequential lifecycle, this implicit contract is fragile and could break under future refactoring. Consider preserving a prior stop request:

```cpp
running_ = running_ && baseline.available;
```

This ensures that once `Stop()` has set `running_ = false`, the sampling loop will not restart regardless of the baseline result.

LHT129 (commented):
Thank you for this well-structured PR. The migration from a point-in-time RSS calculation to a background sampler is a solid fix for the underflow bug, and the new `SearchPassRunner` abstraction cleanly separates pass orchestration from the eval cases.

I have reviewed the full diff and the two substantive issues I identified are already covered by existing inline comments:

1. **`last_error_` not cleared on successful samples** (by @Copilot) — once a transient sampling failure occurs, `last_error_` stays non-empty even if later samples succeed, which can be misleading in the JSON output.

2. **`running_ = baseline.available` race with concurrent `Stop()`** (by @LHT129) — while current call sites use a sequential lifecycle, the unconditional assignment can discard a stop signal, making the contract fragile under future refactoring.

The rest of the code looks correct: the RAII guards in `StartedMonitorGuard` and `MonitorPassGuard` properly handle exception safety, the phase boundaries are well-defined, the test coverage is thorough (including platform-availability, restart/idempotency, and pass-ordering tests), and the `build_search_eval_case.h` correctly releases the build index before creating the search index.

One minor observation: in `search_pass_runner.cpp`, `run_monitored_pass` explicitly calls `Stop()` on the guards and then the destructor also calls `Stop()`. This is harmless since `MonitorPassGuard::Stop()` sets `started_ = false`, making the destructor a no-op, but the explicit calls are redundant with the RAII design.

LHT129 (commented):
## Review Summary

This PR replaces the underflow-prone RSS calculation with a restartable background sampler, adds comprehensive memory metrics, refactors the search pass orchestration via `SearchPassRunner`, and includes thorough test coverage (64 assertions across 11 test cases, plus macOS CI coverage).

### What was reviewed

- `tools/eval/monitor/memory_peak_monitor.cpp/.h` — Core memory sampling with background thread, Linux `/proc/self/statm` and macOS `task_info` support
- `tools/eval/case/search_pass_runner.cpp/.h` — New pass orchestration separating memory, latency, and statistics phases
- `tools/eval/case/search_eval_case.cpp/.h` — Refactored to use `SearchPassRunner`, memory monitor extracted from `monitors_`
- `tools/eval/case/build_eval_case.cpp` — Added `StartedMonitorGuard` for RAII monitor lifecycle
- `tools/eval/case/build_search_eval_case.h` — Release build index before creating search index
- `tests/CMakeLists.txt`, `tools/eval/CMakeLists.txt` — Build system changes
- `.github/workflows/pr-ci.yml` — macOS CI now runs eval tests
- Test files: `search_pass_runner_test.cpp`, `memory_peak_monitor_test.cpp`
- Documentation: `eval.md` (English and Chinese)

### Key findings already discussed

The existing review comments (28 total) have thoroughly covered the main design considerations:

1. **`last_error_` semantics** — Multiple reviewers noted that `last_error_` is not cleared on successful samples. The author confirmed this is intentional: the field retains the most recent failure reason for the phase, paired with `failure_count_`.

2. **`running_ = baseline.available` race** — Multiple reviewers flagged the potential for `Stop()` to be overwritten. The author confirmed the `Monitor` interface assumes a sequential lifecycle, and a partial fix without full synchronization would be misleading.

3. **Various design notes** — `MonitorPassGuard` reference member, `Record()` no-op, `build.reset()` timing, statistics fallback path, etc. — all have been discussed and resolved with author responses.

### Assessment

The implementation is well-structured with clear separation of concerns. The test coverage is thorough, covering normal operation, error paths, restart/idempotency, and platform-specific behavior. The existing review discussion has addressed the substantive design questions. No blocking issues remain.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
+        sample_memory();
+        lock.lock();
+    }
+    lock.unlock();
+    sample_memory();
+}
+
+std::string
+MemoryPeakMonitor::metric_name(const std::string& metric) const {
+    return metric + "(" + process_name_ + ")";
```
LHT129, line 288: [suggestion] The final `sample_memory()` call after the while-loop exits executes after the stop signal has been received and `running_` is already `false`. This means the last sample is taken outside the measured phase window and may capture RSS changes that occurred after `Stop()` was called (e.g., during teardown of other components).

If the intent is to capture a final sample at the moment the stop is requested, consider sampling *before* releasing the lock. Alternatively, if this post-stop sample is intentional for completeness, consider adding a brief comment explaining the rationale.

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+run_monitored_pass(const MonitorPtr& monitor,
+                   const MonitorPtr& memory_monitor,
+                   const SearchPassRunner::SearchPass& search_pass) {
+    MonitorPassGuard monitor_guard(monitor);
+    monitor_guard.Start();
+
+    if (memory_monitor != nullptr) {
+        MonitorPassGuard memory_guard(memory_monitor);
+        memory_guard.Start();
+        search_pass(monitor);
```
LHT129, line 65: [note] The memory monitor baseline is taken after the main monitor (e.g., latency) has already been started via `monitor_guard.Start()` on line 1116. This means the RSS baseline includes any allocations performed by the main monitor's `Start()` method. While `LatencyMonitor::Start()` is likely lightweight, `RecallMonitor::Start()` or future monitors could allocate non-trivial memory, inflating the baseline and reducing the reported delta.

Consider starting the memory monitor first, before the main monitor, so the baseline reflects the pre-pass state:
```cpp
MonitorPassGuard monitor_guard(monitor);
if (memory_monitor != nullptr) {
    MonitorPassGuard memory_guard(memory_monitor);
    memory_guard.Start();
    monitor_guard.Start();
    search_pass(monitor);
    memory_guard.Stop();
} else {
    monitor_guard.Start();
    search_pass(monitor);
}
monitor_guard.Stop();
```

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
...
+            try {
+                monitors_[monitor_id]->Stop();
+            } catch (...) {
+            }
+        }
+    }
+
+    void
+    StartAll() {
+        for (auto& monitor : monitors_) {
```
LHT129, line 49: [suggestion] If `StartAll()` throws partway through (e.g., the 3rd monitor's `Start()` throws), the `StartedMonitorGuard` destructor will correctly stop only the monitors that were started (tracked by `started_monitor_count_`). However, the exception propagates out of `do_build()` at line 224, and `monitor_guard`'s destructor runs during stack unwinding. This is correct behavior, but the `build_index` result (which may be a successful build) is lost because the exception from the monitor prevents the build result from being used.

Consider whether a monitor `Start()` failure should prevent the build result from being recorded. If the build itself succeeded, it may be preferable to log the monitor failure and continue rather than throwing away the build result.

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured with good test coverage (unit tests for both `MemoryPeakMonitor` and `SearchPassRunner`), clear documentation updates in both English and Chinese, and proper CI integration.

The core fix — replacing the underflow-prone RSS calculation with a background sampling thread — is correctly implemented. The `SearchPassRunner` abstraction cleanly separates the pass scheduling logic from the search execution.

The existing review comments from LHT129 and Copilot have already covered the main substantive points:
- `last_error_` retention semantics (author confirmed intentional)
- Post-stop `sample_memory()` call (author confirmed intentional)
- Memory monitor start ordering relative to main monitor (LHT129 suggestion)
- `MonitorPassGuard` reference member lifetime (author confirmed safe for current usage)

One additional observation:

**[note]** In `memory_peak_monitor.cpp`, the `sampling_loop()` function body is not wrapped in a top-level try-catch. While `read_memory()` (called internally) catches all exceptions, if an unexpected exception were to propagate from mutex/lock operations or other standard library calls, the thread would call `std::terminate`. Consider adding a top-level catch-all in the thread function as a safety net, consistent with the defensive style used elsewhere in this PR.

```cpp
void sampling_loop() {
    try {
        // ... existing body ...
    } catch (const std::exception& e) {
        std::lock_guard<std::mutex> lock(state_mutex_);
        running_ = false;
        last_error_ = std::string("sampling loop failed: ") + e.what();
        worker_ready_ = true;
        stop_condition_.notify_all();
    } catch (...) {
        std::lock_guard<std::mutex> lock(state_mutex_);
        running_ = false;
        last_error_ = "sampling loop failed with unknown error";
        worker_ready_ = true;
        stop_condition_.notify_all();
    }
}
```

This is a defensive measure and not blocking — the current code is safe given that `read_memory()` never throws.

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    void
+    Start() {
+        monitor_->Start();
+        started_ = true;
+    }
+
+    void
```
LHT129, line 44: [suggestion] `MonitorPassGuard::Stop()` is not exception-safe: if `monitor_->Stop()` throws, `started_` remains `true`, causing the destructor to call `monitor_->Stop()` a second time.

While `MemoryPeakMonitor::Stop()` is idempotent (it checks `running_` before proceeding), other `Monitor` implementations may not be. Consider setting `started_ = false` before calling `monitor_->Stop()`:

```cpp
void
Stop() {
    started_ = false;
    monitor_->Stop();
}
```

This way the destructor will not retry even if `Stop()` throws.

tools/eval/case/build_eval_case.cpp
```diff
@@ -83,16 +124,15 @@ BuildEvalCase::do_build() {
...
+    monitor_guard.StartAll();
     auto build_index = index_->Build(base);
     if (not build_index.has_value()) {
         throw std::runtime_error(build_index.error().message);
     }
-    for (auto& monitor : monitors_) {
-        monitor->Record();
-        monitor->Stop();
+    for (auto monitor = monitors_.rbegin(); monitor != monitors_.rend(); ++monitor) {
+        (*monitor)->Record();
```
LHT129, line 134: [suggestion] In `do_build()`, `monitor_guard.StopNext()` at line 134 is called without a try-catch inside the Record loop. If any monitor's `Stop()` throws, subsequent monitors in the reverse iteration will be skipped — they won't get their `Record()` or `Stop()` calls.

The destructor of `StartedMonitorGuard` already handles this correctly with try-catch, so the explicit `StopNext()` path should match that behavior. Consider wrapping `StopNext()` in a try-catch, or simply removing the explicit `StopNext()` calls and letting the destructor handle all cleanup:

```cpp
for (auto monitor = monitors_.rbegin(); monitor != monitors_.rend(); ++monitor) {
    (*monitor)->Record();
}
// Destructor will stop all started monitors in reverse order with exception safety.
```

LHT129 (commented):
I have reviewed this PR thoroughly. The changes are well-structured with good test coverage, proper error handling, and cross-platform support (Linux/macOS).

The existing review comments from @LHT129 and @Copilot have already flagged the main issues I would have raised:

1. **`last_error_` not cleared on successful samples** (by @Copilot) — `sample_memory()` sets `last_error_` on failure but never clears it on subsequent success, so the error field can become stale.

2. **Final `sample_memory()` after stop signal** (by @LHT129) — The post-loop sample in `sampling_loop()` captures RSS after `Stop()` was requested, potentially including teardown allocations.

3. **Memory monitor baseline ordering** (by @LHT129) — In `run_monitored_pass`, the memory monitor starts after the main monitor, so the RSS baseline includes the main monitor's `Start()` overhead.

Beyond these already-flagged items, the PR is solid. The `SearchPassRunner` abstraction cleanly separates pass scheduling from the search logic, the `StartedMonitorGuard` provides proper RAII cleanup, and the test coverage (64 assertions across 11 test cases) is thorough.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 270: [note] In `sampling_loop()`, `running_` is set to `baseline.available`. If the baseline read succeeds but the condition variable wait is prematurely notified (spurious wakeup or a concurrent `Stop()` before the while-loop starts), the predicate `not running_` would be true and the loop would exit immediately, calling `sample_memory()` once and returning. This would result in only 2 samples (baseline + one post-"stop") instead of continuous sampling.

Consider restructuring so that `running_` is not overloaded to mean both "baseline was successfully read" and "stop has been requested". A separate flag like `baseline_ok_` could decouple these two concerns, making the state machine easier to reason about.

tools/eval/case/search_pass_runner.cpp
LHT129, line 65: [note] In `SearchPassRunner::Run()`, when `memory_monitor != nullptr && not has_latency_pass`, the memory-only pass runs `search_pass(nullptr)` (line 878). The callback in `do_knn_search` forwards to `run_knn_search_pass(nullptr, false)`, which correctly skips `monitor->Record()`. However, the `do_knn_filter_search` callback does not distinguish between a null monitor and a regular monitor — it unconditionally calls `monitor->Record()` when `monitor != nullptr` (line 714). If a future code path passes a null monitor to the filter search callback, this would crash. The current call sites are safe because the memory-only path is only used for KNN search, but consider adding a null check in `run_knn_filter_search_pass` for robustness.

tools/eval/case/build_eval_case.cpp
LHT129, line 49: [suggestion] `StartedMonitorGuard::StartAll()` increments `started_monitor_count_` before `monitor->Start()` completes. If `monitor->Start()` throws, the count has already been incremented for that monitor, and the destructor will attempt to call `Stop()` on a monitor whose `Start()` failed, leaving it in an unknown state.

Consider swapping the order so the counter is only incremented after a successful `Start()`:
```cpp
void StartAll() {
    for (auto& monitor : monitors_) {
        monitor->Start();
        ++started_monitor_count_;
    }
}
```

In practice `MemoryPeakMonitor::Start()` and `DurationMonitor::Start()` are unlikely to throw, so this is a defensive improvement rather than a bug fix.

LHT129 (commented):
Thanks for this PR. The changes are well-structured and the test coverage is thorough. I have a few additional observations beyond the existing review comments:

1. The `StartedMonitorGuard` in `build_eval_case.cpp` increments `started_monitor_count_` before `monitor->Start()` completes. If `Start()` throws, the destructor will call `Stop()` on a partially-initialized monitor.

2. The `sampling_loop()` in `memory_peak_monitor.cpp` takes a final sample after the stop signal (already noted by @LHT129). This is a design choice worth documenting.

3. Overall the refactoring correctly separates concerns: `SearchPassRunner` handles pass orchestration, `MemoryPeakMonitor` handles sampling, and the eval cases are simplified. The memory-only pass when latency is disabled is a good touch.

No blocking issues from my side. The existing inline comments from @Copilot and @LHT129 cover the actionable items.

LHT129 (commented):
## Code Review Summary

This is a well-structured PR that correctly addresses the root cause (underflow-prone RSS calculation) with a robust background sampling approach. The test coverage is thorough, and the documentation updates are comprehensive.

### Issues already raised by other reviewers (no need to repeat):

1. **`last_error_` not cleared on successful samples** (by Copilot) — `memory_peak_monitor.cpp:254`: After a transient failure, subsequent successful samples do not clear `last_error_`, so the error field may be stale.

2. **Post-stop `sample_memory()` call** (by @LHT129) — `memory_peak_monitor.cpp:288`: The final sample after the while-loop exits captures RSS after `Stop()` was called, which may include teardown allocations.

3. **Memory monitor baseline ordering** (by @LHT129) — `search_pass_runner.cpp:65`: The memory monitor starts after the main monitor, so the baseline includes the main monitor's `Start()` allocations.

### Additional observation:

**[note]** `search_pass_runner.cpp:31` — `MonitorPassGuard` stores a `const MonitorPtr&` reference. While all current call sites are safe (named variables with sufficient lifetime), storing a reference rather than copying the `shared_ptr` is fragile against future misuse (e.g., passing a temporary). Consider storing `MonitorPtr` by value; the cost is one `shared_ptr` copy per guard, negligible relative to the search pass.

---

Overall the implementation is solid. The three existing inline comments cover the substantive issues; the additional note above is a defensive design suggestion. No blocking concerns.

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <stdexcept>
+
+namespace vsag::eval {
+
+namespace {
+
+class MonitorPassGuard {
+public:
+    explicit MonitorPassGuard(const MonitorPtr& monitor) : monitor_(monitor) {
+    }
```
LHT129, line 27: [note] MonitorPassGuard stores a const reference to a MonitorPtr (shared_ptr). If a caller ever passes a temporary shared_ptr, the reference dangles. All current call sites pass lvalues that outlive the guard, so this is safe today, but the design is fragile.

Consider storing a copy of the shared_ptr instead (adds one ref-count operation per guard instance but eliminates the lifetime dependency on the caller's variable):

    class MonitorPassGuard {
    public:
        explicit MonitorPassGuard(MonitorPtr monitor) : monitor_(std::move(monitor)) {}
        // ...
    private:
        MonitorPtr monitor_;
        bool started_{false};
    };

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 254: [suggestion] `last_error_` is never cleared when a sample succeeds after a previous transient failure. If a single sample fails (e.g., `/proc/self/statm` is temporarily unreadable), `last_error_` retains that error message permanently, even though subsequent samples succeed and the peak measurement is valid. This causes `memory_peak_error(<phase>)` to report a stale error despite a successful overall measurement.

Consider clearing `last_error_` on successful samples so the error field reflects only the most recent sampling outcome, while `memory_peak_failed_sample_count` preserves the failure history:

```cpp
if (sample.available) {
    absolute_peak_bytes_ = std::max(absolute_peak_bytes_, sample.bytes);
    ++sample_count_;
    last_error_.clear();
    return;
}
```

LHT129 (commented):
Review summary for PR #2522

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
...
+            const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
+            ++stopped_monitor_count_;
+            try {
+                monitors_[monitor_id]->Stop();
+            } catch (...) {
+            }
+        }
+    }
+
+    void
```
LHT129, line 47: [suggestion] `StopNext()` uses `started_monitor_count_ - stopped_monitor_count_ - 1` to compute the monitor index. Both counters are `uint64_t`. If `StopNext()` were ever called before `StartAll()`, the subtraction would underflow and produce a very large index, leading to an out-of-bounds access on `monitors_`. Consider adding an assertion or guard so the invariant is explicit and protects against future misuse.

LHT129 (commented):
test

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    ~MonitorPassGuard() {
```
LHT129, line 29: [suggestion] `MonitorPassGuard` stores `const MonitorPtr&` — a reference to an external `shared_ptr`. In `run_monitored_pass` this is safe because both `monitor` and `memory_monitor` are const-ref parameters whose referents outlive the guard. However, the reference semantics make the guard fragile: if it were ever used with a temporary `shared_ptr`, the reference would dangle after the full expression.

Consider storing `MonitorPtr` by value instead. The `shared_ptr` copy is cheap (one atomic increment) and makes the guard self-contained:
```cpp
class MonitorPassGuard {
public:
    explicit MonitorPassGuard(MonitorPtr monitor) : monitor_(std::move(monitor)) {}
    // ...
private:
    MonitorPtr monitor_;
    bool started_{false};
};
```
This would require updating call sites to pass by value or `std::move`.

tools/eval/case/build_search_eval_case.h
LHT129, line 42: [suggestion] `BuildSearchEvalCase::Run()` moves `build_` into a local `shared_ptr`, then checks for `nullptr` to decide whether to create a new instance. After the first call, `build_` is null, so a second call would create a fresh `BuildEvalCase` via `EvalCase::MakeInstance`.

However, the newly created `BuildEvalCase` reuses `dataset_ptr_` and `config_`, which means the second call would re-run the build phase on the same dataset and config. If `Run()` is intended to be single-use, consider adding an assertion or throwing if `build_` is already null and the intent is not to support re-execution:
```cpp
if (build == nullptr) {
    throw std::logic_error("BuildSearchEvalCase::Run() can only be called once");
}
```
If re-execution is intentional, the current fallback is fine.

LHT129 (commented):
Overall this is a well-structured PR that correctly addresses the root cause of the memory measurement bug. The switch from a point-in-time RSS read to a background sampling thread is the right approach, and the test coverage is thorough.

A few observations beyond the inline comments already posted:

1. The `StartedMonitorGuard` in `build_eval_case.cpp` correctly handles both normal and exceptional paths with its RAII-style destructor — the reverse-order Stop in the destructor mirrors the normal path.

2. The separation of `memory_monitor_` from the main `monitors_` list in `SearchEvalCase` is clean and avoids the previous problem where memory/latency/recall monitors were conflated in a single loop.

3. The `BuildSearchEvalCase` refactor to release the build index before creating the search index is a good improvement that prevents double-counting memory.

4. The `SearchPassRunner` abstraction cleanly separates the pass orchestration logic from the individual search implementations, making the code easier to follow and test.

No blocking issues found beyond what has already been noted in the existing inline comments.

LHT129 (commented):
test

tools/eval/case/build_search_eval_case.h
```diff
@@ -27,18 +27,25 @@ class BuildSearchEvalCase : public EvalCase {
...
+        }
+        auto build_result = build->Run();
+        build.reset();
+
+        auto search = EvalCase::MakeInstance(config_, "search", dataset_ptr_);
```
LHT129, line 47: [suggestion] The `if (build == nullptr)` check after `auto build = std::move(build_);` is unreachable dead code. `build_` is always initialized in the constructor via `std::make_shared<BuildEvalCase>(...)`. If the fallback path is intentional as a safety net, consider adding a comment explaining when it could be triggered. Otherwise it can be removed.

LHT129 (commented):
Code review for PR #2522

LHT129 (commented):
Additional code review comments for PR #2522

tools/eval/case/build_eval_case.cpp
LHT129, line 134: [suggestion] `do_build()` calls `Record()` and `StopNext()` in reverse iteration order (`rbegin`/`rend`), while `StartAll()` starts monitors in forward order. `StopNext()` uses LIFO semantics (`started_monitor_count_ - stopped_monitor_count_ - 1`), so the stop order is consistent (last-started stops first). However, `Record()` is also called in reverse order, which means the first-started monitor gets its `Record()` called last. For the current set of monitors (memory_peak_monitor whose `Record()` is a no-op, and duration_monitor) this is harmless, but if a future monitor relies on `Record()` being called in the same order as `Start()`, this could be surprising.

Consider iterating forward for `Record()` calls to match the `Start()` order, or document that reverse iteration is intentional.

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
...
+    StartAll() {
+        for (auto& monitor : monitors_) {
+            monitor->Start();
+            ++started_monitor_count_;
+        }
+    }
+
+    void
```
LHT129, line 55: [suggestion] If `monitors_[monitor_id]->Stop()` throws in `StopNext()`, `stopped_monitor_count_` is not incremented. The destructor will then attempt to call `Stop()` on the same monitor again during unwind. While the destructor swallows exceptions from `Stop()`, calling `Stop()` twice on a monitor whose `Stop()` is not idempotent could cause unexpected behavior.

Consider incrementing `stopped_monitor_count_` before calling `Stop()`, so the destructor does not retry a monitor whose `Stop()` already failed:

```cpp
void
StopNext() {
    const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
    ++stopped_monitor_count_;
    monitors_[monitor_id]->Stop();
}
```

This matches the pattern used in the destructor, where `stopped_monitor_count_` is incremented before the `Stop()` call.

LHT129 (commented):
Overall this is a well-structured PR that replaces the fragile point-in-time RSS calculation with a robust background sampling approach. The separation of concerns via `SearchPassRunner` and `StartedMonitorGuard` improves code organization, and the comprehensive test coverage (64 assertions across 11 test cases) is commendable.

One suggestion: in `StartedMonitorGuard::StopNext()`, consider incrementing `stopped_monitor_count_` before calling `Stop()` to prevent the destructor from retrying a monitor whose `Stop()` threw an exception (see inline comment).

The existing review discussions have already covered the key design decisions (error retention semantics, post-stop sampling, race condition on `running_`, memory baseline ordering), and the author has provided clear rationale for each.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 288: [suggestion] `sample_memory()` is called after the lock is released at the end of `sampling_loop()`, which means the final sample is taken outside the critical section while `running_` is already `false`. This is a benign data race on `absolute_peak_bytes_`, `sample_count_`, `failure_count_`, and `last_error_` since `Stop()` may concurrently call `GetResult()` or a new `Start()` may reset these fields. Consider either keeping the lock held through the final `sample_memory()` call, or documenting that the post-loop sample is intentionally lock-free and that `GetResult()`/`Start()` synchronization already prevents concurrent access at a higher level.

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
 
 namespace vsag::eval {
 
+namespace {
+
+class StartedMonitorGuard {
+public:
+    explicit StartedMonitorGuard(std::vector<MonitorPtr>& monitors) : monitors_(monitors) {
```
LHT129, line 33: [note] The `StartedMonitorGuard` destructor stops monitors in reverse order of `StartAll()`, but `do_build()` also calls `Record()` and `StopNext()` in reverse order via `rbegin()/rend()`. If an exception is thrown by `index_->Build()` between `StartAll()` and the manual `Record()`/`StopNext()` loop, the destructor will call `Stop()` on monitors that were started but never had `Record()` called. This is correct for cleanup but means the memory monitor may report zero samples for a failed build. The behavior is reasonable, but worth noting that a build failure will produce a `memory_peak_sample_count` of 1 (baseline only) rather than 0.

tools/eval/case/build_search_eval_case.h
```diff
@@ -27,18 +27,25 @@ class BuildSearchEvalCase : public EvalCase {
...
+            dataset_path_, index_path_, std::move(index), config_, dataset_ptr_);
     }
 
     ~BuildSearchEvalCase() override = default;
 
     JsonType
     Run() override {
-        auto build_result = build_->Run();
-        auto search_result = search_->Run();
+        auto build = std::move(build_);
```
LHT129, line 40: [suggestion] The `Run()` method moves `build_` out via `std::move(build_)`, leaving the member in a moved-from state. If `Run()` is called a second time on the same `BuildSearchEvalCase` instance, `build` will be null and the fallback `EvalCase::MakeInstance(config_, "build", dataset_ptr_)` will create a new `BuildEvalCase` — but the original `index` was already moved into the first `build_` in the constructor. The fallback `MakeInstance` will create a new index internally, which may produce different results than the original. This is unlikely in practice since `Run()` is typically called once, but consider adding a guard or documenting the single-use semantics.

tools/eval/monitor/memory_peak_monitor.cpp
LHT129, line 254: [suggestion] `last_error_` is set on every failed sample but never cleared on successful samples. Once a transient failure occurs (e.g., a single `/proc/self/statm` read fails), `last_error_` will permanently retain that error message even if all subsequent samples succeed. This means `memory_peak_error(<phase>)` in the JSON output will report a stale error despite the measurement being valid overall. Consider clearing `last_error_` on successful samples so the error field reflects only the most recent outcome, while `memory_peak_failed_sample_count` already preserves the failure history.

LHT129 (commented):
Thank you for this well-structured PR. The migration from a point-in-time RSS read to a background sampling thread with proper baseline/delta tracking is a solid fix for the underflow issue described in #1109.

I have reviewed the core changes across `memory_peak_monitor.cpp/.h`, `search_pass_runner.cpp/.h`, `search_eval_case.cpp`, `build_eval_case.cpp`, and `build_search_eval_case.h`. The existing inline comments from Copilot and LHT129 already cover the main areas worth attention:

1. **`last_error_` not cleared on successful samples** (Copilot) — transient failures leave a stale error string even when subsequent samples succeed.
2. **Final `sample_memory()` after stop signal** (LHT129) — the post-loop sample executes after `running_` is already `false`, which may capture teardown RSS.
3. **Memory monitor baseline ordering** (LHT129) — the memory baseline is taken after the main monitor has already started, which could inflate the baseline slightly.

Beyond those, the code looks clean: the `MonitorPassGuard` / `StartedMonitorGuard` RAII wrappers are well-designed, the `SearchPassRunner` abstraction cleanly separates pass scheduling from the eval cases, the test coverage is thorough (64 assertions across 11 test cases plus the new unit tests), and the `build,search` lifecycle now correctly releases the build index before creating the search index.

No additional blocking issues from my side.

LHT129 (commented):
Thank you for this well-structured PR. The migration from a point-in-time RSS read to a background sampling thread with proper baseline/delta tracking is a solid fix for the underflow issue described in #1109.

I have reviewed the core changes across memory_peak_monitor.cpp/.h, search_pass_runner.cpp/.h, search_eval_case.cpp, build_eval_case.cpp, and build_search_eval_case.h. The existing inline comments from Copilot and LHT129 already cover the main areas worth attention:

1. last_error_ not cleared on successful samples (Copilot) — transient failures leave a stale error string even when subsequent samples succeed.
2. Final sample_memory() after stop signal (LHT129) — the post-loop sample executes after running_ is already false, which may capture teardown RSS.
3. Memory monitor baseline ordering (LHT129) — the memory baseline is taken after the main monitor has already started, which could inflate the baseline slightly.

Beyond those, the code looks clean: the MonitorPassGuard / StartedMonitorGuard RAII wrappers are well-designed, the SearchPassRunner abstraction cleanly separates pass scheduling from the eval cases, the test coverage is thorough (64 assertions across 11 test cases plus the new unit tests), and the build,search lifecycle now correctly releases the build index before creating the search index.

No additional blocking issues from my side.

LHT129 (commented):
Thank you for this well-structured PR. The migration from a point-in-time RSS read to a background sampling thread with proper baseline/delta tracking is a solid fix for the underflow issue described in #1109.

I have reviewed the core changes across memory_peak_monitor.cpp/.h, search_pass_runner.cpp/.h, search_eval_case.cpp, build_eval_case.cpp, and build_search_eval_case.h. The existing inline comments from Copilot and LHT129 already cover the main areas worth attention:

1. last_error_ not cleared on successful samples (Copilot) — transient failures leave a stale error string even when subsequent samples succeed.
2. Final sample_memory() after stop signal (LHT129) — the post-loop sample executes after running_ is already false, which may capture teardown RSS.
3. Memory monitor baseline ordering (LHT129) — the memory baseline is taken after the main monitor has already started, which could inflate the baseline slightly.

Beyond those, the code looks clean: the MonitorPassGuard / StartedMonitorGuard RAII wrappers are well-designed, the SearchPassRunner abstraction cleanly separates pass scheduling from the eval cases, the test coverage is thorough (64 assertions across 11 test cases plus the new unit tests), and the build,search lifecycle now correctly releases the build index before creating the search index.

No additional blocking issues from my side.

LHT129 (commented):
I have reviewed this PR and the code quality is solid. The changes correctly address the root cause (underflow-prone RSS calculation) by replacing it with a restartable background sampler. The test coverage is thorough with 7 test cases covering baseline failure, clamping, transient peaks, restart/idempotency, format boundaries, and platform availability.

The existing inline comments from LHT129 cover the remaining areas worth attention:
- The final `sample_memory()` call after the while-loop in `sampling_loop()` executes after `running_` is already `false`
- The memory monitor baseline is taken after the main monitor has already started in `run_monitored_pass`

These are non-blocking suggestions that would further tighten the measurement boundaries. The core fix is correct and the refactoring of search pass orchestration via `SearchPassRunner` is clean and well-tested.

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
...
+    }
+
+    ~StartedMonitorGuard() {
+        while (stopped_monitor_count_ < started_monitor_count_) {
```
LHT129, line 37: [note] In `build_eval_case.cpp`, the `StartedMonitorGuard::StopNext()` method computes the monitor index as `started_monitor_count_ - stopped_monitor_count_ - 1`, which assumes the caller iterates monitors in reverse order (as `do_build()` does with `monitors_.rbegin()`). This implicit coupling between the guard and the call site is fragile — if someone changes the iteration direction in `do_build()` without updating `StopNext()`, monitors would be stopped in the wrong order.

Consider making `StopNext()` take the monitor index explicitly, or document the reverse-order requirement in a comment on `StopNext()`.

LHT129 (commented):
[note] `read_process_memory()` on Linux opens `/proc/self/statm` from scratch on every 5 ms sample tick. While this is an eval tool and not a core library hot path, the repeated `std::ifstream` construction, file open, and close per sample adds unnecessary syscall overhead. Consider keeping the file stream open across samples and using `seekg(0)` + `clear()` to re-read, or reading via a lower-level mechanism such as `pread` on an already-open fd. This would reduce noise in the memory measurement itself.

`tools/eval/monitor/memory_peak_monitor.cpp`, around line 38 (the `read_process_memory()` function).

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured with comprehensive test coverage and documentation updates.

**Summary of findings:**

The existing review comments from @Copilot and @LHT129 already cover the main areas of concern:

1. `last_error_` should be cleared on successful samples in `sample_memory()` to avoid stale error reporting
2. The final `sample_memory()` call after the while-loop exits in `sampling_loop()` may capture RSS outside the measured phase window
3. Memory monitor baseline ordering in `run_monitored_pass()` — starting memory before the main monitor would give a cleaner baseline

Beyond these already-flagged items, I found the code quality to be high:

- The `SearchPassRunner` abstraction cleanly separates pass orchestration from the individual search implementations
- The `StartedMonitorGuard` and `MonitorPassGuard` RAII wrappers correctly handle exception safety
- The `MemoryPeakMonitor` redesign with background sampling thread and proper synchronization eliminates the original underflow bug
- Test coverage is thorough (64 assertions across 11 test cases), covering normal operation, edge cases, error paths, restart/idempotency, and platform-specific behavior
- Documentation is synchronized between English and Chinese

No additional blocking issues found.

tools/eval/case/build_eval_case.cpp
```diff
@@ -26,6 +26,47 @@
...
+    StartAll() {
+        for (auto& monitor : monitors_) {
+            monitor->Start();
+            ++started_monitor_count_;
+        }
+    }
+
+    void
+    StopNext() {
+        const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
```
LHT129, line 57: [suggestion] `StartedMonitorGuard::StopNext()` computes `monitor_id` as `started_monitor_count_ - stopped_monitor_count_ - 1`. If `StopNext()` is called more times than `StartAll()` (e.g., due to a future code change or exception path), the unsigned subtraction wraps to a very large index, causing an out-of-bounds access on `monitors_`. Consider adding a bounds check.

```cpp
void
StopNext() {
    if (stopped_monitor_count_ >= started_monitor_count_) {
        return;
    }
    const auto monitor_id = started_monitor_count_ - stopped_monitor_count_ - 1;
    monitors_[monitor_id]->Stop();
    ++stopped_monitor_count_;
}
```

LHT129 (commented):
test

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
+    absolute_peak_bytes_ = baseline_bytes_;
+    sample_count_ = baseline.available ? 1 : 0;
+    failure_count_ = baseline.available ? 0 : 1;
+    last_error_ = baseline.error;
+    running_ = baseline.available;
+    worker_ready_ = true;
+    stop_condition_.notify_all();
+    if (not running_) {
+        return;
+    }
```
LHT129, line 275: [suggestion] In `sampling_loop`, `read_memory()` (which invokes the user-provided `reader_` callback, potentially performing I/O like reading `/proc/self/statm`) is called while holding `state_mutex_`. Although this only happens once for the baseline read, calling an external callback under a lock is an anti-pattern that could cause unexpected blocking or deadlocks if a custom reader implementation tries to acquire the same mutex.

Consider reading the baseline before acquiring the lock, then only updating the shared state under the lock:

```cpp
void
MemoryPeakMonitor::sampling_loop() {
    const auto baseline = read_memory();  // no lock held
    std::unique_lock<std::mutex> lock(state_mutex_);
    available_ = baseline.available;
    // ... rest of state updates
}
```

tools/eval/case/search_pass_runner.cpp
```diff
@@ -0,0 +1,122 @@
...
+    if (memory_monitor != nullptr) {
+        MonitorPassGuard memory_guard(memory_monitor);
+        memory_guard.Start();
+        search_pass(monitor);
+        memory_guard.Stop();
+    } else {
+        search_pass(monitor);
+    }
+
+    monitor_guard.Stop();
```
LHT129, line 71: [note] In `run_monitored_pass`, when `memory_monitor` is present, `memory_guard.Stop()` is called explicitly on line 71, and then the `MonitorPassGuard` destructor will call `monitor_->Stop()` again if `started_` is still true. This means `memory_monitor->Stop()` is called twice — once explicitly and once via the destructor.

While `MemoryPeakMonitor::Stop()` is currently idempotent (it checks `running_`), this double-stop relies on implementation-specific behavior. If a future monitor implementation is not idempotent, this could cause issues.

Consider removing the explicit `memory_guard.Stop()` call and letting the destructor handle cleanup naturally, since the guard goes out of scope immediately after the explicit call anyway:

```cpp
if (memory_monitor != nullptr) {
    MonitorPassGuard memory_guard(memory_monitor);
    memory_guard.Start();
    search_pass(monitor);
    // memory_guard.Stop();  // remove — destructor handles this
} else {
    search_pass(monitor);
}
```

tools/eval/case/build_search_eval_case.h
```diff
@@ -27,18 +27,25 @@ class BuildSearchEvalCase : public EvalCase {
...
+        if (build == nullptr) {
+            build = EvalCase::MakeInstance(config_, "build", dataset_ptr_);
+        }
+        auto build_result = build->Run();
+        build.reset();
```
LHT129, line 45: [note] `Run()` moves `build_` out on the first call, so calling `Run()` a second time would hit the `build == nullptr` branch and create a new `BuildEvalCase` via `EvalCase::MakeInstance`. However, `EvalCase::MakeInstance` is a factory that likely creates a fresh instance — this means the second `Run()` would rebuild from scratch without the original `index` pointer that was moved into the constructor. If this class is intended to be single-use (which seems to be the case given the `std::move`), consider adding a guard or documenting that `Run()` is not re-entrant.

Current code:
```cpp
auto build = std::move(build_);
if (build == nullptr) {
    build = EvalCase::MakeInstance(config_, "build", dataset_ptr_);
}
```

If re-entrancy is not supported by design, a simple assertion or early return would make the contract explicit.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -15,60 +15,277 @@
...
 Monitor::JsonType
 MemoryPeakMonitor::GetResult() {
+    bool available = false;
+    uint64_t baseline_bytes = 0;
+    uint64_t absolute_peak_bytes = 0;
+    uint64_t sample_count = 0;
+    uint64_t failure_count = 0;
+    std::string last_error;
+    {
+        std::lock_guard<std::mutex> lock(state_mutex_);
```
LHT129, line 218: [suggestion] `GetResult()` can be called while the sampling thread is still active, returning a mid-sampling snapshot that appears valid (`available_` is true, `last_error_` is empty). The eval framework calls `Stop()` before `GetResult()`, so this is not a practical issue today, but the API does not enforce the correct calling order.

Consider either:
1. Documenting that `GetResult()` must only be called after `Stop()`, or
2. Adding a `running_` check in `GetResult()` and returning `available=false` with an appropriate error message when the sampler is still active.

This would make the API self-documenting and prevent misuse if the calling code is refactored in the future.

LHT129 (commented):
Thank you for this well-structured PR. The migration from the underflow-prone point-in-time RSS calculation to a background sampler with proper baseline/peak/delta tracking is a significant improvement. The test coverage is thorough, and the documentation updates are clear.

I have left one inline comment regarding `GetResult()` API safety. Additionally, there are three existing review comments from other reviewers that are worth addressing:

1. **Copilot**: `last_error_` is never cleared after a successful sample in `sample_memory()`, which can leave stale error messages.
2. **LHT129**: The final `sample_memory()` call in `sampling_loop()` executes after the stop signal, potentially capturing post-teardown RSS.
3. **LHT129**: The memory monitor baseline is taken after the main monitor `Start()` in `run_monitored_pass()`, which may inflate the baseline.

Overall the code quality is high — the RAII guards, thread synchronization, and platform abstraction are well done. The eval tests passing 50 consecutive repetitions gives good confidence in correctness.

LHT129 (approved):
All substantive issues have been thoroughly covered by existing reviews from LHT129 and Copilot. The PR correctly fixes the memory peak measurement bug by replacing point-in-time RSS reading with a background sampler thread. The implementation is well-tested with unit tests covering normal operation, edge cases (baseline failure, below-baseline samples, restart/idempotent stop), and platform availability. The SearchPassRunner orchestration cleanly separates memory monitoring from latency/recall passes. No additional issues to raise.

PR #2523, opened 2026-07-24
perf(sindi): optimize DMQ codebooks and packed ID decoding

Roxanne0321:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## What Changed

- Add `dmq_shared_codebook_threshold` so low-frequency terms share a DMQ codebook while frequent terms keep independent codebooks.
- Remap DMQ term IDs to compact IDs and bit-pack them in encoded sparse vectors.
- Replace per-bit packed-ID extraction with a sequential bit-buffer reader in decode and distance paths.
- Document the new SINDI parameter in English and Chinese and add parameter/quantizer coverage.

## Why

The compact-ID implementation originally decoded every ID one bit at a time. During reranking this repeated shifts, masks, and byte loads for every term of every candidate, making DMQ decoding dominate search latency. Sequential byte buffering removes that per-bit hot loop.

Sharing codebooks for infrequent terms also reduces DMQ model overhead while retaining dedicated codebooks for terms with enough training samples.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (described below)

Test details:

```text
git diff --check
make release

Sparse4M, topk=10, query_prune_ratio=0.0:
n_candidate  old QPS  new QPS  old DMQ ms  new DMQ ms
100          156.15   427.13   4.299       0.257
200           90.24   373.38   8.871       0.518
500           40.03   266.58  22.481       1.311
1000          20.89   186.07  44.902       2.589
```

## Compatibility Impact

- API/ABI compatibility: no public API change; adds one optional SINDI JSON parameter.
- Behavior changes: `dmq_shared_codebook_threshold` defaults to `1024`; set it to `0` to disable sharing.
- Serialized-index compatibility: the DMQ quantizer metadata and compact-ID encoding change, so existing DMQ indexes must be rebuilt.

- Performance impact: DMQ packed-ID decode/rerank time is substantially reduced in the measured Sparse4M workload.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/indexes/sindi.md`

## Risk and Rollback

- Risk level: medium, because the DMQ serialized representation changes.
- Rollback plan: revert the two commits and rebuild affected DMQ indexes with the prior format.

## Checklist

- [x] Linked issue is not required for `kind/improvement`
- [x] I have added/updated tests for the new behavior
- [x] I have considered API and serialized-index compatibility impact
- [x] I have updated docs for the new parameter

### DIFF
--- a/src/algorithm/sindi/sindi_parameter.cpp
+++ b/src/algorithm/sindi/sindi_parameter.cpp
@@ -15,6 +15,8 @@
 
 #include "sindi_parameter.h"
 
+#include <limits>
+
 #include "impl/logger/logger.h"
 #include "inner_string_params.h"
 #include "utils/param_compat_macros.h"
@@ -137,6 +139,33 @@ SINDIParameter::FromJson(const JsonType& json) {
     CHECK_ARGUMENT(use_reorder || rerank_type == SPARSE_RERANK_TYPE_FP32,
                    "rerank_type=dmq8 requires use_reorder=true");
 
+    if (json.Contains(SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD)) {
+        const auto threshold_json = json[SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD];
+        CHECK_ARGUMENT(threshold_json.IsNumberInteger(),
+                       "dmq_shared_codebook_threshold must be an integer");
+        if (threshold_json.IsNumberUnsigned()) {
+            const auto threshold = threshold_json.GetUint64();
+            CHECK_ARGUMENT(threshold <= std::numeric_limits<uint32_t>::max(),
+                           fmt::format("dmq_shared_codebook_threshold must be in [0, {}], got {}",
+                                       std::numeric_limits<uint32_t>::max(),
+                                       threshold));
+            dmq_shared_codebook_threshold = static_cast<uint32_t>(threshold);
+        } else {
+            const auto threshold = threshold_json.GetInt();
+            CHECK_ARGUMENT(threshold >= 0,
+                           fmt::format("dmq_shared_codebook_threshold must be in [0, {}], got {}",
+                                       std::numeric_limits<uint32_t>::max(),
+                                       threshold));
+            CHECK_ARGUMENT(static_cast<uint64_t>(threshold) <= std::numeric_limits<uint32_t>::max(),
+                           fmt::format("dmq_shared_codebook_threshold must be in [0, {}], got {}",
+                                       std::numeric_limits<uint32_t>::max(),
+                                       threshold));
+            dmq_shared_codebook_threshold = static_cast<uint32_t>(threshold);
+        }
+    } else {
+        dmq_shared_codebook_threshold = DEFAULT_SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD;
+    }
+
     if (json.Contains(SPARSE_IMMUTABLE)) {
         immutable = json[SPARSE_IMMUTABLE].GetBool();
     }
@@ -160,6 +189,8 @@ SINDIParameter::ToJson() const {
         json[SPARSE_IMMUTABLE].SetBool(true);
     }
     json[SPARSE_RERANK_TYPE].SetString(rerank_type);
+    json[SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD].SetInt(
+        static_cast<int64_t>(dmq_shared_codebook_threshold));
     return json;
 }
 
@@ -175,6 +206,7 @@ SINDIParameter::CheckCompatibility(const vsag::ParamPtr& other) const {
     CHECK_FIELD_EQ(*this, *p, remap_term_ids);
     CHECK_FIELD_EQ(*this, *p, immutable);
     CHECK_FIELD_EQ(*this, *p, rerank_type);
+    CHECK_FIELD_EQ(*this, *p, dmq_shared_codebook_threshold);
     return true;
 }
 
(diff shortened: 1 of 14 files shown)


src/algorithm/sindi/sindi_parameter.cpp
```diff
@@ -160,6 +189,8 @@ SINDIParameter::ToJson() const {
         json[SPARSE_IMMUTABLE].SetBool(true);
     }
     json[SPARSE_RERANK_TYPE].SetString(rerank_type);
+    json[SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD].SetInt(
```
LHT129: [suggestion] `ToJson` unconditionally writes `dmq_shared_codebook_threshold` even when `rerank_type` is `"fp32"`. This means round-tripping a non-DMQ config through `FromJson` → `ToJson` will add a DMQ-specific parameter to the output JSON. While functionally harmless (the field is ignored when `rerank_type` is not `dmq8`), it is inconsistent with the documentation which lists this parameter under the DMQ8 rerank type.

Consider guarding the output so the field only appears when relevant:

```cpp
if (rerank_type == SPARSE_RERANK_TYPE_DMQ8) {
    json[SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD].SetInt(
        static_cast<int64_t>(dmq_shared_codebook_threshold));
}
```

src/datacell/sparse_dmq_datacell_test.cpp
```diff
@@ -92,6 +94,32 @@ TEST_CASE("SparseDmqDataCell implements FlattenInterface", "[ut][SparseDmqDataCe
             restored_distances.data(), restored_computer, inner_ids.data(), inner_ids.size());
         REQUIRE(restored_distances == distances);
     }
+
+    SECTION("reject incompatible serialized version") {
+        std::stringstream stream;
+        IOStreamWriter writer(stream);
+        cell->Serialize(writer);
+        auto serialized = stream.str();
+        constexpr uint32_t old_version = 5;
```
LHT129, line 103: [note] The version-rejection test uses `serialized.data() + sizeof(uint32_t)` to locate the version field, assuming it immediately follows the magic number. If the serialization format ever adds a field between magic and version, this test would silently pass while overwriting the wrong bytes. Consider computing the offset from the struct layout or using a named constant for clarity.

```cpp
// Current:
std::memcpy(serialized.data() + sizeof(uint32_t), &old_version, sizeof(old_version));

// More robust:
constexpr uint32_t version_offset = sizeof(K_SPARSE_DMQ_DATACELL_MAGIC);
std::memcpy(serialized.data() + version_offset, &old_version, sizeof(old_version));
```

src/quantization/sparse_quantization/sparse_dmq_quantizer.cpp
```diff
@@ -66,18 +80,33 @@ store_packed(uint8_t* bytes, uint64_t index, uint32_t bits, uint32_t value) {
...
-load_packed(const uint8_t* bytes, uint64_t index, uint32_t bits) {
-    const uint64_t bit_offset = index * bits;
-    uint32_t result = 0;
-    for (uint32_t bit = 0; bit < bits; ++bit) {
-        const uint64_t target_bit = bit_offset + bit;
-        if ((bytes[target_bit / 8] & (1U << (target_bit % 8))) != 0) {
-            result |= 1U << bit;
+class PackedReader {
+public:
+    PackedReader(const uint8_t* bytes, uint32_t bits)
```
LHT129: [note] `PackedReader` computes the mask as `bits == 32 ? std::numeric_limits<uint32_t>::max() : ((1U << bits) - 1U)`. The special case for `bits == 32` is correct since `1U << 32` is undefined behavior. However, the class does not validate that `bits` is in `[1, 32]` — passing `bits == 0` would produce a mask of 0 (silently returning 0 for every read), and `bits > 32` would be caught by the ternary but still produce a wrong mask. Consider adding an assertion or a bounds check in the constructor.

LHT129 (commented):
## Code Review Summary

**PR**: perf(sindi): optimize DMQ codebooks and packed ID decoding
**Reviewed by**: LHT129

### Overall Assessment

This is a well-executed optimization PR. The compact ID remapping and `PackedReader` approach are sound, the test coverage is thorough, and the parameter plumbing through the SINDI stack is clean. No critical issues found.

### Key Findings

1. **[suggestion]** `ToJson()` unconditionally writes `dmq_shared_codebook_threshold` even for non-DMQ rerank types. Should be guarded behind `rerank_type == SPARSE_RERANK_TYPE_DMQ8`.

2. **[note]** The version-rejection test uses `sizeof(uint32_t)` as a hardcoded offset to locate the version field. A named constant would be more maintainable.

3. **[note]** `PackedReader` does not validate its `bits` parameter range `[1, 32]`. An assertion would catch misuse early.

### What Was Verified

- Compact ID packing correctness (bit width based on distinct term count, not term_id_limit)
- Shared codebook threshold logic (≤ threshold → shared codebook)
- `PackedReader` mask computation for the `bits == 32` edge case
- Merge-join algorithm with compact IDs in `ComputeImpl`
- Serialization version bump (5 → 6) and backward-compatibility rejection
- Parameter validation covering both unsigned and signed JSON integer paths
- Test coverage for packing, sharing, empty model serialization, and version rejection

wxyucs (approved):
lgtm

PR #2538, opened 2026-07-27
fix: reclaim BruteForce force-remove storage (backport #2512 to 1.0)

### LINKED ISSUE #2501: feat](flat): release physical memory after deleting vectors
### Affected area

other

### Interface

cpp

### VSAG version

main

### Summary

# Background

  For the Flat index, deleting vectors only marks/removes them logically, while the physical memory occupied by the deleted vectors is not reclaimed.

  In long-running services with frequent insertions and deletions, memory usage remains high even after a large number of vectors have been deleted. This is especially problematic in
  memory-constrained or multi-tenant environments.

  ## Expected behavior

  The Flat index should provide a mechanism to reclaim the physical memory occupied by deleted vectors.

  Possible approaches include:

  - Compacting the underlying vector storage after deletion.
  - Providing an explicit API such as `Compact()` or `ReleaseDeletedMemory()`.
  - Automatically triggering compaction when the deleted-vector ratio exceeds a configurable threshold.

  ## Requirements

  - Deleted vectors must no longer occupy vector-data memory after reclamation.
  - Existing vector IDs and search results must remain correct.
  - Concurrent search safety should be considered.
  - The API documentation should clarify whether reclamation is automatic or explicitly triggered.
  - A test or benchmark should verify that memory usage decreases after deleting and reclaiming a large number of vectors.

  ## Use case

  A service continuously adds and deletes vectors from a Flat index. Without physical memory reclamation, its RSS keeps growing and may eventually reach the container memory limit,
  even though the number of live vectors remains stable.

### Minimal reproducer

insert and delete from flat index

### Environment

```shell
x86
```

### Expected behavior

Deleted vectors must no longer occupy vector-data memory after reclamation

### Actual behavior

Deleted vectors still occupy vector-data memory after reclamation

### References

_No response_

### Additional context

_No response_

LHT129:
Backport of #2512 to 1.0.

## Summary
- compact BruteForce internal stores after every successful FORCE_REMOVE batch
- document MARK_REMOVE and configuration-free FORCE_REMOVE semantics in English and Chinese
- add coverage for memory accounting, post-compaction Add, id/vector retrieval, search correctness, and concurrent search/delete

Fixes #2501


src/io/memory_io/memory_io.cpp
```diff
@@ -33,6 +33,23 @@ MemoryIO::ResizeImpl(uint64_t size) {
     this->size_ = size;
 }
 
+void
+MemoryIO::ShrinkImpl(uint64_t size) {
+    if (size >= this->size_) {
+        return;
+    }
+
+    auto buffer_size = std::max<uint64_t>(size, 1);
```
LHT129, line 42: [note] The `buffer_size = std::max<uint64_t>(size, 1)` guard prevents passing 0 to `Reallocate`, which would be implementation-defined behavior per the C standard. However, since the early `if (size >= this->size_) return;` already handles the no-op case, `size` reaching 0 here means `this->size_` was also 0 — an edge case that should not occur in practice for a non-empty buffer. Consider adding a comment noting this is a defensive guard, or alternatively assert that `size > 0` at this point.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -322,12 +323,15 @@ BruteForce::Remove(const std::vector<int64_t>& ids, RemoveMode mode) {
         }
 
         const auto last_inner_id = static_cast<InnerIdType>(current_total - 1);
+        Vector<float> data(allocator_);
```
LHT129, line 326: [suggestion] In the original code, `Vector<float> data(dim_, allocator_)` was constructed inside the loop body for every removed id, even when `inner_id >= last_inner_id` (no swap needed). Moving the construction outside the loop and only calling `resize(dim_)` when a swap is actually needed avoids unnecessary allocation/deallocation. This is a good optimization.

One minor consideration: `data` is now default-constructed (empty) outside the `if` block, and `resize` is called only when `inner_id < last_inner_id`. If an exception were thrown between `data.resize()` and the actual use of `data.data()`, the vector would hold allocated memory that gets cleaned up by RAII at scope exit — this is safe.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1264,15 +1268,19 @@ BruteForce::resize(uint64_t new_size) {
...
-    auto current_capacity = this->max_capacity_.load();
-    if (total_count != 0 and total_count > current_capacity / 2) {
-        return;
-    }
-
-    this->inner_codes_->ShrinkToFit(total_count);
-    this->label_table_->ShrinkToFit(total_count);
-    if (this->extra_infos_ != nullptr) {
-        this->extra_infos_->ShrinkToFit(total_count);
+    try {
```
LHT129, line 1271: [note] The removal of the early-return guard `if (total_count != 0 and total_count > current_capacity / 2) return;` is intentional and correct here. Previously this guard prevented `shrink_to_fit` from doing any work unless the live count was at most half of capacity. Now each individual `ShrinkToFit` implementation has its own threshold logic (e.g., `LabelTable::ShrinkToFit` only reallocates at half capacity, `MemoryBlockIO::ShrinkImpl` only releases complete trailing blocks), so the top-level guard is no longer needed and would incorrectly skip compaction in some cases.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1264,15 +1268,19 @@ BruteForce::resize(uint64_t new_size) {
...
+        this->inner_codes_->ShrinkToFit(total_count);
+        this->label_table_->ShrinkToFit(total_count);
+        if (this->extra_infos_ != nullptr) {
+            this->extra_infos_->ShrinkToFit(total_count);
+        }
+    } catch (const VsagException& e) {
```
LHT129, line 1277: [note] The `try/catch` around `shrink_to_fit` correctly handles two exception types:
- `VsagException` with `NO_ENOUGH_MEMORY`: compaction is best-effort; the deletion has already succeeded so we swallow this.
- `std::bad_alloc`: thrown by `AllocatorWrapper::allocate` (newly added nullptr check) and `SafeAllocator`; same rationale.

Other `VsagException` types are re-thrown since they indicate non-memory errors that should propagate. This is a well-scoped exception handler.

src/impl/allocator/allocator_wrapper.h
```diff
@@ -49,7 +49,11 @@ class AllocatorWrapper {
 
     inline pointer
     allocate(uint64_type n, const_void_pointer hint = 0) {
-        return static_cast<pointer>(allocator_->Allocate(n * sizeof(value_type)));
+        auto* p = allocator_->Allocate(n * sizeof(value_type));
```
LHT129, line 52: [note] Adding the `nullptr` check and throwing `std::bad_alloc()` is a correctness fix. Previously, if `Allocate` returned `nullptr`, the code would silently cast it to a non-null pointer, leading to undefined behavior on subsequent access. This aligns `AllocatorWrapper` with the C++ standard allocator requirements, which mandate throwing `std::bad_alloc` on allocation failure.

src/impl/label_table/label_table.h
```diff
@@ -226,7 +226,7 @@ class LabelTable {
      */
     uint64_t
     GetMemoryUsage() {
-        return sizeof(LabelTable) + label_table_.size() * sizeof(LabelType) +
+        return sizeof(LabelTable) + label_table_.capacity() * sizeof(LabelType) +
```
LHT129, line 229: [note] Changing `label_table_.size()` to `label_table_.capacity()` in `GetMemoryUsage` is the correct fix. The allocated memory is determined by capacity, not logical size. Without this fix, `GetMemoryUsage` would under-report after `ShrinkToFit` calls that only `resize` (reducing logical size without freeing capacity), which is the common case when the live count is above half capacity.

src/impl/label_table/label_table.h
```diff
@@ -365,7 +365,20 @@ class LabelTable {
 
     void
     ShrinkToFit(InnerIdType capacity) {
-        label_table_.resize(capacity);
+        // Avoid a full-table copy for small removals; vector storage is still compacted by BruteForce.
+        if (capacity <= label_table_.capacity() / 2) {
+            try {
```
LHT129, line 370: [suggestion] The half-capacity threshold (`capacity <= label_table_.capacity() / 2`) for triggering actual reallocation is a good optimization that avoids full-table copies for small removals. However, the fallback path on `std::bad_alloc` calls `label_table_.resize(capacity)` which only reduces logical size without freeing memory. This means after an OOM during compaction, the label table retains its full capacity allocation. This is the correct conservative behavior (deletion already succeeded), but it means `GetMemoryUsage` will continue to report the pre-compaction capacity. Consider whether this is the desired behavior or if a comment clarifying the tradeoff would help future readers.

tests/test_brute_force.cpp
```diff
@@ -27,6 +32,107 @@
 #include "vsag/options.h"
 #include "vsag/search_request.h"
 
+namespace {
+
+class TrackingAllocator : public vsag::Allocator {
```
LHT129, line 37: [note] The `TrackingAllocator` is a well-designed test utility. It correctly:
- Tracks all allocations/deallocations in an `unordered_map` for accurate byte accounting
- Supports `SetAllocationLimit` to simulate OOM conditions
- Uses `std::scoped_lock` for thread safety (needed since the concurrent search/delete test uses multiple threads)
- Properly handles `Reallocate` with correct accounting for size changes

The `BlockSizeLimitGuard` RAII wrapper for `Options::Instance().block_size_limit()` is also clean and ensures the global option is restored even if the test aborts.

tests/test_brute_force.cpp
```diff
@@ -962,6 +1068,112 @@ TEST_CASE("(PR) BruteForce RangeSearch After MarkRemove",
...
+        ->Owner(false);
+    auto search_result = index->KnnSearch(query, 1, "");
+    REQUIRE(search_result.has_value());
+    REQUIRE(search_result.value()->GetIds()[0] == retained_id);
+
+    std::atomic<bool> start{false};
+    std::atomic<bool> search_failed{false};
+    std::thread search_thread([&] {
+        while (not start.load(std::memory_order_acquire)) {
+        }
```
LHT129, line 1156: [note] The concurrent search/delete test correctly verifies that `FORCE_REMOVE` (which takes the exclusive global lock) does not corrupt concurrent searches. The test spawns a search thread that continuously queries while the main thread performs force removal, then verifies no search failures occurred and the final element count is correct. The use of `std::atomic<bool>` with appropriate memory ordering (`acquire`/`release` for the start flag, `relaxed` for the failure flag) is correct for this synchronization pattern.

src/algorithm/hgraph/hgraph_modify.cpp
```diff
@@ -36,7 +38,16 @@ HGraph::Remove(const std::vector<int64_t>& ids, RemoveMode mode) {
             delete_count += this->force_remove_one(id);
         }
         if (delete_count != 0) {
-            this->shrink_to_fit();
+            try {
```
LHT129, line 41: [note] The same `try/catch` pattern for `shrink_to_fit` is applied consistently in `HGraph::Remove` as in `BruteForce::Remove`. This is a good cross-cutting fix — both index types now handle OOM during compaction gracefully, treating it as best-effort since the deletion has already completed.

src/impl/label_table/label_table_test.cpp
```diff
@@ -217,6 +217,15 @@ TEST_CASE("LabelTable Memory Management", "[ut][LabelTable]") {
         auto usage = label_table.GetMemoryUsage();
         REQUIRE(usage > 0);
     }
+
+    SECTION("GetMemoryUsage includes retained capacity") {
```
LHT129, line 221: [note] This test correctly validates that `GetMemoryUsage` now reflects `capacity()` rather than `size()`. After `ShrinkToFit(127)` on a table resized to 128, the capacity remains 128 (since 127 > 128/2, only `resize` is called, not reallocation), so memory usage should be unchanged. This is a good regression test for the `size()` → `capacity()` fix in `GetMemoryUsage`.

LHT129 (commented):
This is a clean backport of #2512 to the `1.0` branch. The changes are well-structured and address the core issue (BruteForce force-remove not reclaiming storage) with several layered improvements:

**Core fix**: `BruteForce::shrink_to_fit()` now always attempts compaction (removed the overly conservative half-capacity guard), with each storage layer handling its own threshold logic.

**OOM resilience**: Both `BruteForce::shrink_to_fit()` and `HGraph::Remove()` now catch `NO_ENOUGH_MEMORY` and `std::bad_alloc` during compaction, treating it as best-effort since deletion has already succeeded.

**Correctness fixes**:
- `AllocatorWrapper::allocate()` now throws `std::bad_alloc` on null return (was silently producing UB)
- `LabelTable::GetMemoryUsage()` uses `capacity()` instead of `size()` for accurate accounting
- `LabelTable::ShrinkToFit()` only reallocates at ≤ half capacity, avoiding full-table copies for small removals
- `MemoryIO::ShrinkImpl()` added to support shrinking for non-block IO backends

**Default IO change**: BruteForce default template now uses `block_memory_io` instead of `memory_io`, which enables block-granular storage reclamation.

**Tests**: Comprehensive test covers OOM during force remove, memory accounting, post-compaction Add, id/vector retrieval, search correctness, and concurrent search/delete.

**Docs**: English and Chinese documentation updated for both `index_lifecycle` and `brute_force` pages.

No blocking issues found. LGTM for backport.

LHT129 (commented):
This is a backport of #2512 to the 1.0 branch. I have reviewed the full diff and the existing inline comments from LHT129 and vsag-bot have already covered most of the substantive points.

## Key observations

**Already flagged by vsag-bot (P2):** The change of the default IO type from `memory_io` to `block_memory_io` in `BRUTE_FORCE_PARAMS_TEMPLATE` and `WARP_PARAMS_TEMPLATE` introduces two issues that should be addressed before merging:

1. **Memory under-reporting**: `MemoryBlockIO::ShrinkImpl` sets `size_` to the exact byte count while the final partial block remains allocated. `BasicIO::GetMemoryUsage()` returns `size_`, so `GetMemoryUsage()` under-reports actual allocation. This contradicts the documented allocation-accounting contract.

2. **Null dereference in DirectReadImpl**: `MemoryBlockIO::DirectReadImpl` allocates a temporary buffer via `this->allocator_->Allocate(size)` without null-checking before calling `ReadImpl` (which does `memcpy` into the buffer). When a code crosses a block boundary and the allocator rejects the allocation, this results in a null pointer dereference instead of a proper error.

**Already covered by LHT129:** The optimization in `BruteForce::Remove` (hoisting `Vector<float> data` out of the loop), the exception handling in `shrink_to_fit`, the `AllocatorWrapper::allocate` nullptr check, the `LabelTable::GetMemoryUsage` capacity fix, the `LabelTable::ShrinkToFit` half-capacity threshold, and the `MemoryIO::ShrinkImpl` guard have all been reviewed with appropriate notes and suggestions.

**CI workflow changes** (adding `"1.*"` to branch patterns) and **documentation updates** look correct and consistent with the backport.

The core deletion logic, label table compaction, and memory reclamation are correctly implemented. The test coverage in `test_brute_force.cpp` is comprehensive, covering OOM during compaction, post-compaction Add, id/vector retrieval, search correctness, and concurrent search/delete.

wxyucs (approved):
lgtm

PR #2549, opened 2026-07-28
fix(pyramid): remove immutable state shadowing in SetImmutable

### LINKED ISSUE #2548: Pyramid SetImmutable does not block write operations
## Bug Description

Pyramid declares its own bool immutable_ member that shadows the base class std::atomic<bool> immutable_ from InnerIndexInterface. This causes SetImmutable() to only set the derived member, leaving the base class state unchanged.

As a result, the public write barrier in IndexImpl (the CHECK_IMMUTABLE_INDEX macro) does not intercept write operations (Add, Build, Remove, etc.) on Pyramid indexes after SetImmutable() is called.

## Steps to Reproduce

1. Create a Pyramid index and build it with data
2. Call index->SetImmutable()
3. Call index->Add(), index->Build(), or index->Remove() - these operations succeed instead of returning UNSUPPORTED_INDEX_OPERATION

## Expected Behavior

After calling SetImmutable(), all public write operations should be rejected with ErrorType::UNSUPPORTED_INDEX_OPERATION, matching the behavior of other indexes like HGraph.

## Actual Behavior

Write operations are not blocked because IndexImpl checks the base class atomic member (which remains false), while Pyramid::SetImmutable() only sets the derived bool member.

## Root Cause

- InnerIndexInterface declares std::atomic<bool> immutable_{false} at line 671
- IndexImpl::CHECK_IMMUTABLE_INDEX macro reads this->inner_index_->immutable_.load() at line 62
- Pyramid declares its own bool immutable_{false} at line 383 (shadowing the base class member)
- Pyramid::SetImmutable() only sets the derived member at lines 1314-1321

## Fix

Remove the shadowing bool immutable_ member from Pyramid and use the inherited atomic state with proper memory ordering, matching the pattern used by HGraph.

## Impact

This is a critical bug that violates the API contract. Users expect SetImmutable() to make the index read-only, but Pyramid continues to accept writes. This can lead to data inconsistency and potential concurrency issues.

LHT129:
## Summary

Fix Pyramid immutable state shadowing bug where SetImmutable() did not actually block write operations.

## Changes

- Remove shadowing bool immutable_ member from Pyramid (was hiding base class atomic)
- Use inherited std::atomic<bool> immutable_ with proper memory ordering
- Match the pattern used by HGraph for consistency
- Add regression test to verify Add, Build, and Remove are rejected after SetImmutable()

## Testing

New test case Pyramid Set Immutable verifies:
- SetImmutable() can be called multiple times (idempotent)
- Add() returns UNSUPPORTED_INDEX_OPERATION after SetImmutable()
- Build() returns UNSUPPORTED_INDEX_OPERATION after SetImmutable()
- Remove() returns UNSUPPORTED_INDEX_OPERATION after SetImmutable()
- Search operations continue to work correctly

Fixes: #2548

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -1311,14 +1311,14 @@ Pyramid::search_node(const IndexNode* node,
 }
 void
 Pyramid::SetImmutable() {
-    if (this->immutable_) {
+    if (this->immutable_.load(std::memory_order_acquire)) {
         return;
     }
     label_table_->SetImmutable();
     this->points_mutex_.reset();
     this->points_mutex_ = std::make_shared<EmptyMutex>();
     this->searcher_->SetMutexArray(this->points_mutex_);
-    immutable_ = true;
+    this->immutable_.store(true, std::memory_order_release);
 }
 
 float
--- a/src/algorithm/pyramid/pyramid.h
+++ b/src/algorithm/pyramid/pyramid.h
@@ -380,7 +380,6 @@ private:
     ReorderInterfacePtr reorder_{nullptr};  // reorder helper (if use_reorder_)
 
     uint32_t index_min_size_{0};  // min node size before graph is built
-    bool immutable_{false};       // true after SetImmutable()
 };
 
 }  // namespace vsag
--- a/tests/test_pyramid.cpp
+++ b/tests/test_pyramid.cpp
@@ -248,6 +248,37 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::PyramidTestIndex,
 }
 
 TEST_CASE_PERSISTENT_FIXTURE(fixtures::PyramidTestIndex,
+                             "Pyramid Set Immutable",
+                             "[ft][immutable][pyramid]") {
+    const auto metric_type = "l2";
+    PyramidParam pyramid_param;
+    const auto param =
+        GeneratePyramidBuildParametersString(metric_type, dims.front(), pyramid_param);
+    auto index = TestFactory("pyramid", param, true);
+    auto dataset =
+        pool.GetDatasetAndCreate(dims.front(), base_count, metric_type, /*with_path=*/true);
+    TestBuildIndex(index, dataset, true);
+
+    REQUIRE(index->SetImmutable().has_value());
+    REQUIRE(index->SetImmutable().has_value());
+
+    auto add_result = index->Add(dataset->base_);
+    REQUIRE_FALSE(add_result.has_value());
+    REQUIRE(add_result.error().type == vsag::ErrorType::UNSUPPORTED_INDEX_OPERATION);
+
+    auto build_result = index->Build(dataset->base_);
+    REQUIRE_FALSE(build_result.has_value());
+    REQUIRE(build_result.error().type == vsag::ErrorType::UNSUPPORTED_INDEX_OPERATION);
+
+    std::vector<int64_t> ids{dataset->base_->GetIds()[0]};
+    auto remove_result = index->Remove(ids);
+    REQUIRE_FALSE(remove_result.has_value());
+    REQUIRE(remove_result.error().type == vsag::ErrorType::UNSUPPORTED_INDEX_OPERATION);
+
+    TestKnnSearch(index, dataset, GeneratePyramidSearchParametersString(100), 0.94, true);
+}
+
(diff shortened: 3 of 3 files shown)


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1311,14 +1311,14 @@ Pyramid::search_node(const IndexNode* node,
 }
 void
```
LHT129, line 1312: [note] Unlike `HGraph::SetImmutable()` which acquires `add_mutex_` and `global_mutex_` before replacing the mutex array, `Pyramid::SetImmutable()` does not acquire any locks. This is a pre-existing behavior not introduced by this PR, but worth noting: if concurrent `Add` or `Build` operations are in-flight when `SetImmutable()` is called, the mutex replacement (`points_mutex_.reset()` + reassignment) races with those operations. The `std::atomic<bool>` check in `CHECK_IMMUTABLE_INDEX` provides ordering only for subsequent operations, not for ones already past the check.

Consider whether Pyramid needs similar lock acquisition as HGraph to ensure in-flight mutations complete before the mutex is replaced. This can be addressed in a follow-up if needed.

LHT129 (commented):
The fix is correct and minimal. Here is the analysis:

**Root cause**: `Pyramid` declared `bool immutable_{false}` in `pyramid.h:383`, which shadowed the base class `std::atomic<bool> immutable_{false}` from `InnerIndexInterface`. The `CHECK_IMMUTABLE_INDEX` macro in `index_impl.h` accesses `this->inner_index_->immutable_`, which resolves to the base class member (always `false`). Meanwhile, `Pyramid::SetImmutable()` was writing to the shadowing `bool` member. So the immutable check never blocked any write operations.

**Fix**:
1. Removes the shadowing `bool immutable_` from `pyramid.h` — the inherited `std::atomic<bool>` is now the only member
2. Updates `SetImmutable()` to use `load(std::memory_order_acquire)` / `store(true, std::memory_order_release)` — matching HGraph's pattern exactly
3. Adds a thorough regression test covering idempotency, Add/Build/Remove rejection, and search-after-immutable

**Verification**: The test validates all the critical paths that were previously broken by the shadowing bug. The memory ordering is consistent with HGraph and the `CHECK_IMMUTABLE_INDEX` macro.

LHT129:
Good catch on the locking behavior note. This is indeed a pre-existing issue unrelated to the immutable state shadowing bug being fixed here. The shadowing bug was purely about the wrong member being written to (shadowing bool vs inherited atomic<bool>), which is what this PR addresses.

The race condition with in-flight mutations during mutex replacement is a separate concurrency concern that existed before. Since this PR is focused on the minimal fix for the shadowing bug, I agree this locking behavior can be addressed in a follow-up PR if needed. The regression test added here validates the core fix for the shadowing issue without introducing new concurrency patterns.

LHT129 (commented):
LGTM. The fix correctly removes the shadowing `bool immutable_` member from `Pyramid` so that `IndexImpl::CHECK_IMMUTABLE_INDEX` reads the base class `std::atomic<bool> immutable_` as intended.

**What was broken:** `Pyramid` declared `bool immutable_{false}` at line 383 of `pyramid.h`, which shadowed `InnerIndexInterface::immutable_` (`std::atomic<bool>`). The `CHECK_IMMUTABLE_INDEX` macro in `IndexImpl<T>` accesses `this->inner_index_->immutable_`, which resolves to the base class member — but `Pyramid::SetImmutable()` was writing to the shadowing member. This meant `SetImmutable()` appeared to work locally but `Add`/`Build`/`Remove` were never actually blocked.

**What this PR does:**
- Removes the shadowing `bool immutable_` from `pyramid.h`
- Updates `Pyramid::SetImmutable()` to use `this->immutable_.load(std::memory_order_acquire)` / `.store(true, std::memory_order_release)` — matching the pattern used by `HGraph`
- Adds a thorough regression test covering idempotency, write rejection (Add/Build/Remove), and search-after-immutable

The memory ordering is appropriate: acquire on the check ensures visibility of all stores before the store, and release on the store ensures the immutable flag is visible to subsequent acquire loads in `CHECK_IMMUTABLE_INDEX`. The existing `[note]` about concurrent `SetImmutable` vs in-flight mutations is a pre-existing concern not introduced by this PR and can be addressed separately.

wxyucs (approved):
lgtm

PR #2556, opened 2026-07-28
build: add experimental CRoaring amalgamation

wxyucs:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [x] CI/Build/Infra

## Linked Issue

Not applicable: this is a `kind/improvement` experimental build alternative.

## What Changed

- Keep the existing CRoaring v3.0.1 source-archive integration active and
  unchanged.
- Add `roaringbitmap_amalgamation.cmake` as an explicitly experimental,
  commented-out alternative beside the active include.
- Download the three official v3.0.1 release assets (`roaring.c`, `roaring.h`,
  and `roaring.hh`) with pinned SHA-256 verification when a developer manually
  switches the adjacent includes.
- Build the same local `roaring` target while preserving C11, include/link
  interfaces, PIC/static/shared behavior, AVX controls, warning isolation, and
  install/export behavior.

There is no option or automatic selection: normal builds retain the established
archive path. When manually enabled, the experimental implementation transfers
1,232,827 bytes instead of the 100,717,067-byte archive, a reduction of
99,484,240 bytes (98.775950%).

## Test Evidence

```text
Default active archive include, root configure and roaring build: passed
Default archive integration, CMake 3.22 URL path, linked consumer: passed
Temporary root include switch to the experimental implementation: configured
Experimental clean official download with CMake 3.22: passed
Experimental roaring target and linked bitmap consumer: passed
Experimental module included twice under CMake 3.22: passed
CMake 3.22 STREQUAL variable match and mismatch semantics: passed
Two CMake 3.22 configurations sharing the experimental cache: lock/reuse passed
Downloaded SHA-256 values and exact three-file sizes: matched pinned metadata
Experimental AVX definitions stayed private to roaring.c: verified
git diff --check: passed
git diff origin/main -- docs: empty
```

The temporary include switch used for validation was restored. The committed
state keeps `roaringbitmap.cmake` active and the experimental include commented.

## Compatibility Impact

- Default API, ABI, downloads, override behavior, and installation remain
  unchanged.
- The alternative is opt-in only through a deliberate source-level include
  switch and has no offline-build contract or user-facing documentation.

## Risk and Rollback

- Risk is isolated because the new implementation is inactive by default.
- Remove the commented include and experimental file to remove the alternative;
  the existing archive integration requires no rollback.

## Checklist

- [ ] I have linked the relevant issue (not required for `kind/improvement`)
- [x] No documentation changes are included
- [x] The commit follows the Conventional Commit and DCO requirements

### DIFF
--- a/extern/roaringbitmap/roaringbitmap_amalgamation.cmake
+++ b/extern/roaringbitmap/roaringbitmap_amalgamation.cmake
@@ -1,0 +1,101 @@
+include_guard (GLOBAL)
+
+set (roaringbitmap_version 3.0.1)
+set (roaringbitmap_release_url
+    "https://github.com/RoaringBitmap/CRoaring/releases/download/v${roaringbitmap_version}")
+set (roaringbitmap_source_dir
+    "${DOWNLOAD_DIR}/roaringbitmap-v${roaringbitmap_version}")
+file (MAKE_DIRECTORY "${roaringbitmap_source_dir}")
+
+function (vsag_download_roaringbitmap_asset asset_name asset_sha256)
+    set (asset_path "${roaringbitmap_source_dir}/${asset_name}")
+    file (LOCK "${asset_path}.lock"
+        GUARD FUNCTION
+        TIMEOUT 120
+        RESULT_VARIABLE lock_result)
+    if (NOT "${lock_result}" STREQUAL "0")
+        message (FATAL_ERROR "Failed to lock the CRoaring cache for ${asset_name}: ${lock_result}")
+    endif ()
+
+    if (EXISTS "${asset_path}")
+        file (SHA256 "${asset_path}" cached_sha256)
+        if ("${cached_sha256}" STREQUAL "${asset_sha256}")
+            message (STATUS "Using cached CRoaring asset: ${asset_path}")
+            return ()
+        endif ()
+        message (STATUS "Replacing CRoaring asset with unexpected hash: ${asset_path}")
+        file (REMOVE "${asset_path}")
+    endif ()
+
+    set (asset_url "${roaringbitmap_release_url}/${asset_name}")
+    set (temporary_path "${asset_path}.tmp")
+    file (REMOVE "${temporary_path}")
+    message (STATUS "Downloading CRoaring asset: ${asset_url}")
+    file (DOWNLOAD
+        "${asset_url}"
+        "${temporary_path}"
+        STATUS download_status
+        LOG download_log
+        SHOW_PROGRESS
+        TLS_VERIFY ON
+        INACTIVITY_TIMEOUT 15
+        TIMEOUT 90)
+    list (GET download_status 0 download_code)
+    list (GET download_status 1 download_message)
+    message (VERBOSE "CRoaring download log for ${asset_url}:\n${download_log}")
+    if (NOT download_code EQUAL 0)
+        file (REMOVE "${temporary_path}")
+        message (FATAL_ERROR
+            "Failed to download CRoaring ${asset_name} from ${asset_url}: ${download_message}")
+    endif ()
+
+    file (SHA256 "${temporary_path}" downloaded_sha256)
+    if (NOT "${downloaded_sha256}" STREQUAL "${asset_sha256}")
+        file (REMOVE "${temporary_path}")
+        message (FATAL_ERROR
+            "CRoaring ${asset_name} SHA-256 mismatch from ${asset_url}: expected "
+            "${asset_sha256}, got ${downloaded_sha256}")
+    endif ()
+    file (RENAME "${temporary_path}" "${asset_path}")
+endfunction ()
+
+# These are the official amalgamation assets attached to the pinned CRoaring release.
+vsag_download_roaringbitmap_asset (
+    roaring.c d1fe3f22b11968d6edf6648b55c96a8df57f5875453be0409c8d5b4afeec3353)
+vsag_download_roaringbitmap_asset (
+    roaring.h 57e0abc5e3f48e7c47743e49f1261747b41686993949896b3809f255ab5e282b)
+vsag_download_roaringbitmap_asset (
+    roaring.hh 81d9cfc5704ea8fddd53bfe0b20b4678c8d3143d1a47ec684c009a9100c1f530)
+
+# Preserve upstream target names for compatibility, even though roaring also exposes this include
+# path directly. Match upstream BUILD_SHARED_LIBS behavior without installing CRoaring.
+add_library (roaring-headers INTERFACE)
+add_library (roaring-headers-cpp INTERFACE)
+target_include_directories (roaring-headers INTERFACE "${roaringbitmap_source_dir}")
+target_include_directories (roaring-headers-cpp INTERFACE "${roaringbitmap_source_dir}")
+
+add_library (roaring "${roaringbitmap_source_dir}/roaring.c")
+add_library (roaring::roaring ALIAS roaring)
+set_target_properties (roaring PROPERTIES
+    C_STANDARD 11
+    C_STANDARD_REQUIRED ON
+    SOVERSION 15
+    VERSION 3.0.0
+    WINDOWS_EXPORT_ALL_SYMBOLS ON)
+target_include_directories (roaring PUBLIC "${roaringbitmap_source_dir}")
+target_link_libraries (roaring PUBLIC roaring-headers roaring-headers-cpp)
+
+if (DISABLE_AVX_FORCE OR NOT COMPILER_AVX_SUPPORTED)
+    target_compile_definitions (roaring PRIVATE ROARING_DISABLE_AVX=1)
+endif ()
+if (DISABLE_AVX512_FORCE OR NOT COMPILER_AVX512_SUPPORTED)
+    target_compile_definitions (roaring PRIVATE CROARING_COMPILER_SUPPORTS_AVX512=0)
+endif ()
+if (CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU")
+    target_compile_options (roaring PRIVATE -Wno-unused-function)
+endif ()
+
+if (NOT TARGET vsag_roaring_headers)
+    add_library (vsag_roaring_headers INTERFACE)
+endif ()
+target_include_directories (vsag_roaring_headers INTERFACE "${roaringbitmap_source_dir}")
--- a/cmake/VSAGThirdParty.cmake
+++ b/cmake/VSAGThirdParty.cmake
@@ -42,6 +42,8 @@ include (extern/cpuinfo/cpuinfo.cmake)
 include (extern/fmt/fmt.cmake)
 include (extern/thread_pool/thread_pool.cmake)
 include (extern/tsl/tsl.cmake)
+# Experimental feature; enable it by switching the commented include.
+# include (extern/roaringbitmap/roaringbitmap_amalgamation.cmake)
 include (extern/roaringbitmap/roaringbitmap.cmake)
 
 if (ENABLE_TOOLS AND ENABLE_CXX11_ABI)


extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,116 @@
...
+vsag_download_roaringbitmap_asset (
+    roaring.c d1fe3f22b11968d6edf6648b55c96a8df57f5875453be0409c8d5b4afeec3353)
+vsag_download_roaringbitmap_asset (
+    roaring.h 57e0abc5e3f48e7c47743e49f1261747b41686993949896b3809f255ab5e282b)
+vsag_download_roaringbitmap_asset (
+    roaring.hh 81d9cfc5704ea8fddd53bfe0b20b4678c8d3143d1a47ec684c009a9100c1f530)
+
+# Match the upstream target name and BUILD_SHARED_LIBS behavior without installing CRoaring.
+add_library (roaring-headers INTERFACE)
+add_library (roaring-headers-cpp INTERFACE)
```
LHT129: [note] The `roaring-headers` and `roaring-headers-cpp` INTERFACE targets are redundant here. `target_include_directories(roaring PUBLIC "${roaringbitmap_source_dir}")` already propagates the include directory to all consumers of `roaring`. Adding `target_link_libraries(roaring PUBLIC roaring-headers roaring-headers-cpp)` sets the same include paths a second time via the INTERFACE targets. This is harmless but adds unnecessary CMake targets that serve no functional purpose.

If the intent is to match upstream target names for potential future compatibility, consider adding a comment explaining that these are placeholder targets reserved for that purpose. Otherwise they can be removed.

wxyucs: Good point. These interface targets are intentionally retained to preserve the upstream CMake target names for compatibility, even though the include propagation is redundant. I added a comment documenting that purpose.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,117 @@
...
+vsag_download_roaringbitmap_asset (
+    roaring.hh 81d9cfc5704ea8fddd53bfe0b20b4678c8d3143d1a47ec684c009a9100c1f530)
+
+# Preserve upstream target names for compatibility, even though roaring also exposes this include
+# path directly. Match upstream BUILD_SHARED_LIBS behavior without installing CRoaring.
+add_library (roaring-headers INTERFACE)
+add_library (roaring-headers-cpp INTERFACE)
+target_include_directories (roaring-headers INTERFACE "${roaringbitmap_source_dir}")
+target_include_directories (roaring-headers-cpp INTERFACE "${roaringbitmap_source_dir}")
+
```
LHT129: [note] The `roaring` target does not set `POSITION_INDEPENDENT_CODE`. When `BUILD_SHARED_LIBS=ON`, CMake automatically compiles with `-fPIC`, so the shared-library path is covered. However, if someone builds `roaring` as a static library (the default when `BUILD_SHARED_LIBS=OFF`) and later links it into a shared library, the linker may fail with relocation errors on some platforms (e.g., certain Linux configurations).

Consider adding:
```cmake
set_target_properties(roaring PROPERTIES POSITION_INDEPENDENT_CODE ON)
```

This is low-risk since the PR description confirms PIC scenarios were tested, but it would make the build more robust for downstream consumers who embed VSAG as a static dependency within a shared library.

LHT129 (commented):
This is a well-executed build-system improvement that reduces CRoaring dependency download size by ~98.8% (from ~100 MB to ~1.2 MB) by switching from the full source archive to the three official amalgamation assets.

**Summary of changes:**
- `extern/roaringbitmap/roaringbitmap.cmake`: Replaced `FetchContent` with manual `file(DOWNLOAD)` for the three amalgamation files (`roaring.c`, `roaring.h`, `roaring.hh`), with SHA-256 verification and deterministic caching. Manually creates the `roaring` target with matching upstream properties (C11, SOVERSION, AVX disable, etc.).
- `cmake/VSAGOptions.cmake`: Updated `DOWNLOAD_DIR` docstring to reflect broader usage beyond `ExternalProject`.
- `docs/`: Updated both English and Chinese offline-build guides to document the CRoaring-specific override behavior (directory/base URL instead of single archive).

**What I checked:**
- Correctness: The amalgamation files are verified with pinned SHA-256 hashes. The download function handles cached, mirrored, local, and upstream fallback paths correctly. The `roaring` target properties match upstream behavior.
- Security: `TLS_VERIFY ON` is set. Hashes are pinned. No command injection vectors.
- Compatibility: `vsag_roaring_headers` now points to the amalgamation root directory (where `roaring.h` and `roaring.hh` live), which is correct for the single-file layout. The `roaring-headers`/`roaring-headers-cpp` INTERFACE targets are retained for upstream name compatibility.
- Documentation: Both EN and ZH guides accurately describe the new override behavior.

**One minor note** (posted as inline comment): The `roaring` target does not set `POSITION_INDEPENDENT_CODE`. This is fine for the shared-library path, but static-library consumers embedding VSAG in a shared library may benefit from adding it. Non-blocking.

Overall this PR is in good shape. The existing inline discussion about the INTERFACE targets has been resolved with an explanatory comment.

extern/roaringbitmap/roaringbitmap.cmake
wxyucs: This is already guaranteed by the project-wide compiler configuration: `cmake/VSAGCompilerConfig.cmake:17` sets `CMAKE_POSITION_INDEPENDENT_CODE ON`, and `CMakeLists.txt` includes that file before `VSAGThirdParty.cmake` creates `roaring`. CMake therefore initializes the `roaring` target's `POSITION_INDEPENDENT_CODE` property to ON even when it is static. The focused static target build also confirmed `-fPIC` on the `roaring.c` compile command. I am retaining the centralized setting rather than duplicating it on this target.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,117 @@
...
+    endif ()
+    list (APPEND asset_urls "${roaringbitmap_release_url}/${asset_name}")
+
+    set (download_errors)
+    foreach (asset_url IN LISTS asset_urls)
+        message (STATUS "Downloading CRoaring asset: ${asset_url}")
+        file (DOWNLOAD
+            "${asset_url}"
+            "${asset_path}"
+            EXPECTED_HASH "SHA256=${asset_sha256}"
```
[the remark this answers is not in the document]
wxyucs: Accepted. Both the active CMake and CMake 3.22 reproduce the fatal `EXPECTED_HASH` behavior before `STATUS` can be handled. Each candidate is now downloaded to a `.tmp` file without fatal hash enforcement; the code inspects the status, computes SHA-256, and renames only a verified file into the cache. A focused local HTTP test served a deliberately bad `roaring.c`: CMake reported its computed mismatch, continued to the next URL, and retained both attempts in the final diagnostic. Verified HTTP assets then configured and built the `roaring` target under CMake 3.31 and 3.22.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,117 @@
...
+                    "Ignoring CRoaring override with unexpected SHA-256: ${override_asset}")
+            else ()
+                message (WARNING "CRoaring override asset does not exist: ${override_asset}")
+            endif ()
+        else ()
+            string (REGEX REPLACE "/$" "" override_url "${roaringbitmap_override}")
+            list (APPEND asset_urls "${override_url}/${asset_name}")
+        endif ()
+    endif ()
+    list (APPEND asset_urls "${roaringbitmap_release_url}/${asset_name}")
```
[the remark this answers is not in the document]
wxyucs: The availability concern is valid, but I did not add an unverified object path: the project OSS endpoint currently returns HTTP 403 for the old archive and every plausible versioned standalone-asset path, so there is no verified project-hosted `roaring.c`/`.h`/`.hh` URL to pin. The task also requires the official versioned release assets rather than an unverified third-party source. I accepted the documentation inconsistency and now explicitly state in both languages that CRoaring uses the official release after the override, without the archive-oriented Aliyun fallback. Non-GitHub/offline operation remains supported through `VSAG_THIRDPARTY_ROARINGBITMAP`, which accepts an internal base URL or local directory and applies the same pinned SHA-256 checks.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,126 @@
...
+    set (download_errors)
+    foreach (asset_url IN LISTS asset_urls)
+        set (temporary_path "${asset_path}.tmp")
+        file (REMOVE "${temporary_path}")
+        message (STATUS "Downloading CRoaring asset: ${asset_url}")
+        file (DOWNLOAD
+            "${asset_url}"
+            "${temporary_path}"
+            STATUS download_status
+            LOG download_log
```
LHT129: [note] The `LOG download_log` variable on line 59 is captured but never read. If a download fails, the full HTTP transaction log could provide useful debugging context beyond the status message alone. Consider printing it as a verbose status message or including it in the error diagnostic when the download fails.

```cmake
# Option: print the log as a verbose status message after a failed download
if (NOT download_code EQUAL 0)
    message (VERBOSE "Download log for ${asset_url}: ${download_log}")
endif ()
```

wxyucs: Accepted. Transport failures now emit the captured `file(DOWNLOAD)` transaction log at CMake's `VERBOSE` level before the candidate loop records the concise status message. Normal successful and hash-mismatch output remains concise. The amended integration reconfigured and built the `roaring` target successfully under both CMake 3.31 and 3.22.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,129 @@
...
+            SHOW_PROGRESS
+            TLS_VERIFY ON
+            INACTIVITY_TIMEOUT 15
+            TIMEOUT 90)
+        list (GET download_status 0 download_code)
+        list (GET download_status 1 download_message)
```
LHT129: [note] When `file(DOWNLOAD)` succeeds (exit code 0) but the SHA-256 check fails, the `download_log` captured via `LOG download_log` is not emitted. This can happen when a CDN or proxy returns a 200 OK with an HTML error page instead of the expected binary. Printing the log at VERBOSE level for hash-mismatch cases as well (not just transport failures) would aid debugging in those scenarios.

Current code (line 65-67):
```cmake
if (NOT download_code EQUAL 0)
    message (VERBOSE "CRoaring download log for ${asset_url}:\n${download_log}")
endif ()
```

Suggestion: also emit the log when the hash mismatches. The simplest fix is to always emit at VERBOSE level after the download attempt regardless of outcome:
```cmake
message (VERBOSE "CRoaring download log for ${asset_url}:\n${download_log}")
```

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,129 @@
...
+target_include_directories (roaring-headers-cpp INTERFACE "${roaringbitmap_source_dir}")
+
+add_library (roaring "${roaringbitmap_source_dir}/roaring.c")
+add_library (roaring::roaring ALIAS roaring)
+set_target_properties (roaring PROPERTIES
+    C_STANDARD 11
+    C_STANDARD_REQUIRED ON
+    SOVERSION 15
+    VERSION 3.0.0
+    WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
LHT129: [note] The AVX/AVX512 disable definitions are set as `PUBLIC` on the `roaring` target, which means they propagate to all downstream consumers that link against `roaring`. Since these are implementation details of the CRoaring amalgamation source and `roaring.h`/`roaring.hh` headers do not conditionally expose API based on these macros, `PRIVATE` would be more appropriate and avoid polluting consumer compilation.

```cmake
# Current (lines 112-116):
target_compile_definitions (roaring PUBLIC ROARING_DISABLE_AVX=1)
target_compile_definitions (roaring PUBLIC CROARING_COMPILER_SUPPORTS_AVX512=0)

# Suggested:
target_compile_definitions (roaring PRIVATE ROARING_DISABLE_AVX=1)
target_compile_definitions (roaring PRIVATE CROARING_COMPILER_SUPPORTS_AVX512=0)
```

This is low-risk either way since these macro names are CRoaring-specific and unlikely to conflict with consumer code, but `PRIVATE` better expresses the intent that these are internal build configuration.

extern/roaringbitmap/roaringbitmap.cmake
wxyucs: Accepted. The captured transaction log is now emitted at `VERBOSE` after every download attempt, including HTTP-success/hash-mismatch responses. A focused local-server regression returned a 65-byte `200 OK` body for `roaring.c`; `cmake --log-level=VERBOSE` reported the request, `HTTP/1.0 200 OK`, content length, and connection details before continuing to the pinned official URL. Normal configure output remains concise because the diagnostic is still `VERBOSE` only.

extern/roaringbitmap/roaringbitmap.cmake
```diff
@@ -1,47 +1,129 @@
...
+                    message (STATUS "Copied verified CRoaring asset: ${override_asset}")
+                    return ()
+                endif ()
+                message (WARNING
+                    "Ignoring CRoaring override with unexpected SHA-256: ${override_asset}")
+            else ()
+                message (WARNING "CRoaring override asset does not exist: ${override_asset}")
+            endif ()
+        else ()
+            string (REGEX REPLACE "/$" "" override_url "${roaringbitmap_override}")
```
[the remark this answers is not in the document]
wxyucs: Accepted. A non-directory local override, or a URL ending in `.tar.gz`/`.tgz`, now takes a compatibility branch that verifies the established v3.0.1 archive MD5 and uses the original upstream subdirectory integration. A local archive configured, built `roaring`, linked a consumer, and ran with all network proxies forced to an unreachable endpoint under both current CMake and CMake 3.22; the same test passed with an HTTP archive URL on CMake 3.22. This branch is entered only for an explicit archive override, so unset and new directory/base-URL builds still download only the three pinned official amalgamation assets.

docs/docs/en/src/development/offline_build.md
```diff
@@ -41,6 +41,13 @@ The resolution order is therefore:
    **not** user-configurable — for a fully internal mirror, use the environment
    variable.
 
+CRoaring is the exception to this archive-oriented flow. VSAG downloads the
+three official v3.0.1 amalgamation assets directly, and its override names a
+base directory or URL containing `roaring.c`, `roaring.h`, and `roaring.hh`.
```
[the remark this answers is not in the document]
wxyucs: Accepted. Both language versions now scope CRoaring's directory/base-URL form to `main`, state that the `0.15`–`0.18` release lines require the single v3.0.1 source archive, and document that `main` continues to accept that legacy archive form. Focused CMake 3.22 tests cover both a base URL and the archive URL form.

extern/roaringbitmap/roaringbitmap.cmake
wxyucs: Accepted for the amalgamation target. Neither official standalone public header references `ROARING_DISABLE_AVX` or `CROARING_COMPILER_SUPPORTS_AVX512`; both affect only the compiled C implementation. The definitions are now `PRIVATE`. With both disable options forced, the recorded compile command contained both macros for `roaring.c`, while the linked C++ consumer command contained neither; that consumer compiled, linked, and exercised a bitmap successfully under current CMake and CMake 3.22.

LHT129 (commented):
This iteration looks clean. All previously raised issues have been addressed:

- The `roaring-headers`/`roaring-headers-cpp` interface targets are retained with a comment documenting the compatibility purpose.
- The `EXPECTED_HASH` fatal-error-on-first-URL issue has been resolved by switching to manual SHA-256 verification with multi-URL fallback.
- The `download_log` is now emitted at `VERBOSE` level for all download attempts.
- The AVX/AVX512 disable definitions are now `PRIVATE` on the amalgamation `roaring` target.
- Archive-form override compatibility is preserved with the `roaringbitmap_archive_override` detection branch.
- Documentation correctly scopes the new directory/base-URL form to `main` and notes the archive-only form for release lines.

No new issues found in this push.

extern/roaringbitmap/roaringbitmap.cmake
wxyucs: Scope update: the direct-asset implementation now lives only in the inactive experimental `roaringbitmap_amalgamation.cmake`. Its CMake 3.22-compatible temporary download and pinned SHA-256 validation remain, including the `VERBOSE` transaction log for successful responses that fail the hash check. The active archive integration is restored unchanged.

extern/roaringbitmap/roaringbitmap.cmake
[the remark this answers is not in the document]
wxyucs: Scope update: the requested compatibility concern is now eliminated structurally. `extern/roaringbitmap/roaringbitmap.cmake` is byte-for-byte identical to `origin/main` and remains the active default, so its established archive URLs and override contract are unchanged. The commented experimental file has no archive or offline override support.

docs/docs/en/src/development/offline_build.md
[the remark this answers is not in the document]
wxyucs: Scope update: all offline-build documentation changes have been removed, and `git diff origin/main -- docs` is empty. The experimental include is deliberately source-level and inactive, with no documented release-line or offline-build contract; the existing archive include remains active.

extern/roaringbitmap/roaringbitmap.cmake
wxyucs: Scope update: this implementation moved to the inactive experimental file, where both AVX control definitions remain `PRIVATE`. A clean CMake 3.22 compile database again showed both definitions on `roaring.c` and neither on the linked consumer. The active archive integration keeps its existing upstream behavior unchanged.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,91 @@
...
+    roaring.h 57e0abc5e3f48e7c47743e49f1261747b41686993949896b3809f255ab5e282b)
+vsag_download_roaringbitmap_asset (
+    roaring.hh 81d9cfc5704ea8fddd53bfe0b20b4678c8d3143d1a47ec684c009a9100c1f530)
+
+# Preserve upstream target names for compatibility, even though roaring also exposes this include
+# path directly. Match upstream BUILD_SHARED_LIBS behavior without installing CRoaring.
+add_library (roaring-headers INTERFACE)
+add_library (roaring-headers-cpp INTERFACE)
+target_include_directories (roaring-headers INTERFACE "${roaringbitmap_source_dir}")
+target_include_directories (roaring-headers-cpp INTERFACE "${roaringbitmap_source_dir}")
```
LHT129, line 75: [note] The `roaring`, `roaring-headers`, and `roaring-headers-cpp` targets are created without `if(NOT TARGET ...)` guards. While `VSAGThirdParty.cmake` has `include_guard(GLOBAL)` which prevents double-inclusion in normal builds, adding target-existence guards would make this module more robust if it is ever included from a different context or during incremental reconfiguration.

Suggested:
```cmake
if (NOT TARGET roaring-headers)
    add_library (roaring-headers INTERFACE)
    target_include_directories (roaring-headers INTERFACE "${roaringbitmap_source_dir}")
endif ()
```

wxyucs: Accepted the duplicate-inclusion concern, but used a module-level `include_guard(GLOBAL)` instead of per-target guards. Per-target guards could silently mask the configuration error of enabling both the archive and experimental integrations, retaining whichever `roaring` target happened to be created first. The module guard safely makes repeated inclusion of this experimental file idempotent while preserving a clear target-collision failure if both alternatives are enabled. A bounded CMake 3.22 regression included the experimental file twice, configured from the verified three-asset cache, built and linked the consumer, and ran it successfully; the unchanged archive path was separately configured, built, linked, and run within 90-second bounds.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,93 @@
+include_guard (GLOBAL)
+
+set (roaringbitmap_version 3.0.1)
+set (roaringbitmap_release_url
+    "https://github.com/RoaringBitmap/CRoaring/releases/download/v${roaringbitmap_version}")
+set (roaringbitmap_source_dir
+    "${DOWNLOAD_DIR}/roaringbitmap-v${roaringbitmap_version}")
+file (MAKE_DIRECTORY "${roaringbitmap_source_dir}")
+
```
LHT129, line 9: [note] The `file(MAKE_DIRECTORY "${roaringbitmap_source_dir}")` at module scope runs on every CMake configure, even when the experimental amalgamation path is not active (the include is commented out in `VSAGThirdParty.cmake`). This is harmless — creating an empty directory is cheap — but it is unnecessary work during normal builds.

Consider moving the `MAKE_DIRECTORY` call inside the `vsag_download_roaringbitmap_asset` function (before the first download) so it only executes when the module is actually activated. Alternatively, leave it as-is since the cost is negligible and the module is explicitly experimental.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,93 @@
...
+function (vsag_download_roaringbitmap_asset asset_name asset_sha256)
+    set (asset_path "${roaringbitmap_source_dir}/${asset_name}")
+    if (EXISTS "${asset_path}")
+        file (SHA256 "${asset_path}" cached_sha256)
+        if (cached_sha256 STREQUAL asset_sha256)
+            message (STATUS "Using cached CRoaring asset: ${asset_path}")
+            return ()
```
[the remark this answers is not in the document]
wxyucs: Declined. In CMake's `if()` grammar, these unquoted tokens are variable names and are dereferenced for `STREQUAL`; `${...}` expansion is not required here. A bounded `cmake -P` regression under CMake 3.22 set `cached_sha256`, `downloaded_sha256`, and `asset_sha256`, verified that equal values compare equal and changed values compare unequal. The integration evidence also exercises this exact cached branch: the bounded experimental reconfigure reported all three `Using cached CRoaring asset` messages, then built, linked, and ran its consumer successfully.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,93 @@
...
+            "Failed to download CRoaring ${asset_name} from ${asset_url}: ${download_message}")
+    endif ()
+
+    file (SHA256 "${temporary_path}" downloaded_sha256)
+    if (NOT downloaded_sha256 STREQUAL asset_sha256)
+        file (REMOVE "${temporary_path}")
+        message (FATAL_ERROR
+            "CRoaring ${asset_name} SHA-256 mismatch from ${asset_url}: expected "
+            "${asset_sha256}, got ${downloaded_sha256}")
+    endif ()
```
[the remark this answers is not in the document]
wxyucs: Declined independently for the post-download check. CMake 3.22 dereferences both unquoted variable operands in `if (NOT downloaded_sha256 STREQUAL asset_sha256)`. A direct bounded CMake script verified both the match and mismatch cases. Separately, the prior clean experimental configure downloaded all three official assets, promoted them past this exact check, and produced the pinned SHA-256 values; the current bounded cached configure and linked consumer also pass. If the operands were literal names as suggested, the clean configure would have failed before any asset could be promoted.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
wxyucs: Declined. The premise is incorrect: the experimental include is commented, so CMake never evaluates this module or its module-scope `file(MAKE_DIRECTORY)` during a default build. When a developer explicitly switches the includes, the directory is required before the first cache/hash/download operation. Moving it into the function would merely execute the idempotent call once per asset instead of once per enabled module. A bounded CMake 3.22 default archive configure/build linked and ran its consumer without activating this file, while the experimental configure separately passed when deliberately included.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,93 @@
...
+        if (cached_sha256 STREQUAL asset_sha256)
+            message (STATUS "Using cached CRoaring asset: ${asset_path}")
+            return ()
+        endif ()
+        message (STATUS "Replacing CRoaring asset with unexpected hash: ${asset_path}")
+        file (REMOVE "${asset_path}")
+    endif ()
+
+    set (asset_url "${roaringbitmap_release_url}/${asset_name}")
+    set (temporary_path "${asset_path}.tmp")
```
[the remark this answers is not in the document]
wxyucs: Accepted. Each asset's cache check, download, SHA-256 verification, and promotion is now serialized by a function-scoped `file(LOCK)` on `${asset_path}.lock`, with a 120-second timeout and a clear fatal diagnostic. This is supported by CMake 3.22 and keeps different assets independently lockable. Two bounded clean CMake 3.22 configurations were launched against one shared experimental cache: one transient proxy handshake failed before receiving bytes and released its lock, the other completed and built/ran the linked consumer, and retrying the failed configure reused all three verified cached assets without downloading. A separately bounded default archive configure/build/consumer also passed; the active default file is unchanged.

LHT129 (approved):
LGTM. This is a clean, well-structured experimental addition.

**Summary of changes:**
- New `extern/roaringbitmap/roaringbitmap_amalgamation.cmake`: Downloads the three CRoaring v3.0.1 amalgamation source files (~1.2MB) instead of the full 100MB source archive, with pinned SHA-256 verification and `file(LOCK)` concurrency safety.
- Two commented-out include lines in `cmake/VSAGThirdParty.cmake` showing how to opt in.

**Review notes:**
- The module is entirely opt-in (commented out by default), so there is zero risk to existing builds.
- SHA-256 verification is applied to both cached and freshly downloaded assets.
- Atomic download pattern (`.tmp` → verify → `RENAME`) is correctly implemented.
- `file(LOCK ... GUARD FUNCTION)` provides proper concurrency safety for parallel CMake invocations sharing the download cache.
- The `roaring-headers` / `roaring-headers-cpp` INTERFACE targets are intentionally retained for upstream target name compatibility (documented in the comment on line 77-78).
- AVX/AVX512 controls mirror the existing `roaringbitmap.cmake` behavior.
- `vsag_roaring_headers` is correctly guarded with `if (NOT TARGET ...)`.

No blocking issues found.

LHT129 (approved):
LGTM. The new `roaringbitmap_amalgamation.cmake` is well-structured:

- SHA-256 verification with pinned hashes for all three official release assets
- Proper `file(LOCK)` with `GUARD FUNCTION` for concurrent build safety
- Atomic promotion via `file(RENAME)` from `.tmp` to final path
- Defensive cleanup of stale cached files and temporary downloads
- AVX/AVX512 compile definitions match the existing `roaringbitmap.cmake` pattern
- `vsag_roaring_headers` target guarded with `if (NOT TARGET ...)` for compatibility
- Completely opt-in via commented include — zero risk to default builds

No issues found in this update. The prior review threads have been satisfactorily addressed.

LHT129 (approved):
LGTM. The experimental amalgamation module is clean and well-structured: SHA-256 verification, `file(LOCK)` for concurrent safety, proper cleanup of temporary files on failure, and clear error messages. The change to `VSAGThirdParty.cmake` is minimal and the new module is commented out by default, so there is no risk to existing builds. All concerns from previous review rounds have been addressed.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,101 @@
...
+        GUARD FUNCTION
+        TIMEOUT 120
+        RESULT_VARIABLE lock_result)
+    if (NOT lock_result STREQUAL "0")
+        message (FATAL_ERROR "Failed to lock the CRoaring cache for ${asset_name}: ${lock_result}")
+    endif ()
+
+    if (EXISTS "${asset_path}")
+        file (SHA256 "${asset_path}" cached_sha256)
+        if (cached_sha256 STREQUAL asset_sha256)
```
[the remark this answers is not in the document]
wxyucs: This is already comparing the two variable values. In CMake `if()`, an unquoted argument matching a defined variable is dereferenced; here both `cached_sha256` and the function argument `asset_sha256` are defined. I verified this exact form with `/usr/bin/cmake` 3.22.1 in a script covering both the equal and unequal cases. I also reconfigured the experimental path against the populated cache: it emitted `Using cached CRoaring asset` for `roaring.c`, `roaring.h`, and `roaring.hh`, then the consumer built and ran. If `asset_sha256` were literal, that cache branch could not be taken. No code change is needed.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,101 @@
...
+    list (GET download_status 1 download_message)
+    message (VERBOSE "CRoaring download log for ${asset_url}:\n${download_log}")
+    if (NOT download_code EQUAL 0)
+        file (REMOVE "${temporary_path}")
+        message (FATAL_ERROR
+            "Failed to download CRoaring ${asset_name} from ${asset_url}: ${download_message}")
+    endif ()
+
+    file (SHA256 "${temporary_path}" downloaded_sha256)
+    if (NOT downloaded_sha256 STREQUAL asset_sha256)
```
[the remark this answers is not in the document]
wxyucs: This comparison also already dereferences both defined operands under CMake `if()` semantics. Independently for this post-download branch, `/usr/bin/cmake` 3.22.1 passed the exact `if (NOT downloaded_sha256 STREQUAL asset_sha256)` form when the values matched and took the expected mismatch branch when they differed. The pinned release assets have also passed this check during clean experimental configures before being promoted from `.tmp`, and the resulting library consumer builds and runs. Quoting `${asset_sha256}` would be equivalent here, not a functional fix, so I am leaving the portable CMake 3.22-compatible code unchanged.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,101 @@
...
+        file (REMOVE "${temporary_path}")
+        message (FATAL_ERROR
+            "CRoaring ${asset_name} SHA-256 mismatch from ${asset_url}: expected "
+            "${asset_sha256}, got ${downloaded_sha256}")
+    endif ()
```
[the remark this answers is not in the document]
wxyucs: The original post-download comparison was functional, but I accepted explicit expansion as a clarity improvement. It is now `if (NOT "${downloaded_sha256}" STREQUAL "${asset_sha256}")`. A fresh CMake 3.22.1 configure downloaded the official v3.0.1 `.c`, `.h`, and `.hh` files, accepted each through this branch, and `cmake -E sha256sum` reproduced all three pinned hashes before the consumer built and ran.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,101 @@
...
+    if (NOT lock_result STREQUAL "0")
+        message (FATAL_ERROR "Failed to lock the CRoaring cache for ${asset_name}: ${lock_result}")
+    endif ()
+
+    if (EXISTS "${asset_path}")
+        file (SHA256 "${asset_path}" cached_sha256)
+        if (cached_sha256 STREQUAL asset_sha256)
+            message (STATUS "Using cached CRoaring asset: ${asset_path}")
+            return ()
+        endif ()
```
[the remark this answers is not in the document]
wxyucs: The cache comparison already evaluated variable values, but the suggested explicit expansion is useful for readability. I changed it to quote both expanded values. Focused CMake 3.22.1 evidence: the first configure downloaded and verified the three official assets, the second configure took the cached branch for all three, and the bitmap consumer built and ran successfully.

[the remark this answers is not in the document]
wxyucs: The original comparison already dereferenced both defined operands under CMake `if()` semantics, but explicit expansion is useful for clarity and avoids this recurring ambiguity. I changed the cache check to `if ("${cached_sha256}" STREQUAL "${asset_sha256}")`. With CMake 3.22.1, a clean configure downloaded and verified all three pinned assets; a second configure reported all three as cached, and the linked bitmap consumer passed.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
```diff
@@ -0,0 +1,101 @@
...
+    "${DOWNLOAD_DIR}/roaringbitmap-v${roaringbitmap_version}")
+file (MAKE_DIRECTORY "${roaringbitmap_source_dir}")
+
+function (vsag_download_roaringbitmap_asset asset_name asset_sha256)
+    set (asset_path "${roaringbitmap_source_dir}/${asset_name}")
+    file (LOCK "${asset_path}.lock"
+        GUARD FUNCTION
+        TIMEOUT 120
+        RESULT_VARIABLE lock_result)
+    if (NOT "${lock_result}" STREQUAL "0")
```
LHT129, line 16: [note] The `file(LOCK ... GUARD FUNCTION)` will release the lock when the function returns normally (including via `return()`), but if CMake terminates due to `message(FATAL_ERROR ...)` (e.g. download failure or SHA-256 mismatch), the lock file at `${asset_path}.lock` will remain on disk. This is a known limitation of CMake `file(LOCK)` — FATAL_ERROR aborts the process without running destructors.

This is harmless in practice (a stale `.lock` file does not prevent future builds), but worth noting for awareness.

wxyucs: Agreed that the filesystem entry can remain and is harmless; no source change is needed. The `.lock` file is only the path used for process-level locking—its existence does not mean the lock remains held. I verified this directly with CMake 3.22.1: all three zero-byte `.lock` files from the earlier configure were present, a new build directory acquired each lock, reused all three verified cached assets, built `roaring`, and ran the linked bitmap consumer successfully. Keeping the files also avoids unsafe deletion races between concurrent configurators.

LHT129 (approved):
LGTM. This PR adds a clean, well-structured experimental alternative for CRoaring amalgamation downloads. The implementation is properly isolated — commented out by default with no impact on existing builds. The download function includes robust SHA-256 verification, file locking for concurrent safety, and proper error handling. The target definitions mirror the existing roaringbitmap.cmake conventions (AVX controls, warning suppression, vsag_roaring_headers compatibility). No blocking issues found.

extern/roaringbitmap/roaringbitmap_amalgamation.cmake
LHT129, line 16: [note] `file(LOCK ... GUARD FUNCTION ...)` requires CMake 3.19+, but the project's `cmake_minimum_required` is 3.18. If a developer enables this experimental module on CMake 3.18, configure will fail with an error about the `GUARD` keyword not being supported.

Since this module is commented out by default and labeled experimental, this is not a blocker. Consider either:
- Adding a `cmake_minimum_required(VERSION 3.19)` guard at the top of this file, or
- Documenting the CMake 3.19+ requirement in the comment above the include in `VSAGThirdParty.cmake`.

wxyucs: This version claim is not correct, so no guard or comment change is needed. The official CMake `file()` locking documentation marks the `file(LOCK ...)` signature—including `GUARD <FUNCTION|FILE|PROCESS>`—as added in 3.2: https://cmake.org/cmake/help/v3.18/command/file.html#locking. The same syntax is also present in the versioned 3.15 documentation, below VSAG’s 3.18 minimum: https://cmake.org/cmake/help/v3.15/command/file.html#locking. The experimental module therefore remains compatible with the repository baseline; its lock/cache configure and consumer were additionally exercised under CMake 3.22.1.

PR #2557, opened 2026-07-28
ci: prefetch dependencies for host PR builds

### LINKED ISSUE #2555: [improve](build): move PR ASan builds to host and prefetch third-party sources
### Kind of improvement

perf

### Affected area

build

### Summary

Move the PR CI x86 ASan build from the `vsaglib/vsag:ci-x86-openblas`
job container to the GitHub-hosted Ubuntu 22.04 environment, while preserving
the current compiler, sanitizer, OpenBLAS, clang-tidy 15, and artifact behavior.

Accelerate third-party source preparation by:

1. Checking out FetchContent-managed dependencies at pinned revisions with
   `actions/checkout`.
2. Placing those sources under a common `FETCHCONTENT_BASE_DIR`.
3. Passing explicit `FETCHCONTENT_SOURCE_DIR_<NAME>` overrides so CMake does
   not download or update the checked-out dependencies.
4. Caching `ExternalProject_Add` source archives through the existing
   `DOWNLOAD_DIR` mechanism.

`FETCHCONTENT_BASE_DIR` alone only controls FetchContent directory layout and
does not suppress downloads. Explicit source overrides are required for
reliable pre-population.

The first implementation should keep Boost, ANTLR4, and HDF5 on their existing
`ExternalProject_Add` archive paths. Their archives can be cached separately;
they are not controlled by `FETCHCONTENT_BASE_DIR`.

### Expected benefit

- Remove the PR build job's dependency on pulling and starting the VSAG CI
  container.
- Reduce repeated third-party downloads during CMake configuration.
- Make third-party source revisions and acquisition visible in the Actions
  workflow.
- Preserve pinned and reproducible dependency inputs.
- Allow the host build path to be evaluated independently from the existing
  container test jobs.
- Retain the existing system OpenBLAS fast path while making missing host
  dependencies fail early.

The workflow should use `VSAG_USE_SYSTEM_OPENBLAS=ON` rather than the legacy
`USE_SYSTEM_OPENBLAS=ON` compatibility option.

### Benchmark evidence

// TODO: Collect step-level timing from representative PR CI runs before and
after the change.

Measure at least:

- runner and container initialization time;
- host dependency installation time;
- FetchContent checkout/population time;
- ExternalProject archive download time;
- CMake configure time;
- compilation time;
- artifact upload time;
- total `ASan Build X86` duration;
- cold-run and warm-cache behavior.

Run the host implementation in parallel with the existing container job before
making it a required check. Validate both regular and fork-originated pull
requests.

### References

- `.github/workflows/pr-ci.yml:68` — current containerized x86 ASan build
- `.github/workflows/pr-ci.yml:143` — existing host-based macOS ASan build
- `.github/workflows/pr-ci.yml:170` — existing host-based Python x86 build
- `.github/workflows/pr-ci.yml:204` — build artifact consumer jobs
- `cmake/VSAGOptions.cmake:48` — ExternalProject build metadata directory
- `cmake/VSAGOptions.cmake:50` — existing ExternalProject download cache directory
- `cmake/VSAGThirdParty.cmake:17` — third-party dependency inclusion
- `extern/json/json.cmake:14` — representative FetchContent dependency
- `extern/antlr4/antlr4.cmake:27` — representative ExternalProject dependency
- `extern/boost/boost.cmake:26` — Boost ExternalProject dependency
- `extern/hdf5/hdf5.cmake:21` — HDF5 ExternalProject dependency
- https://cmake.org/cmake/help/v3.22/module/FetchContent.html
- https://cmake.org/cmake/help/v3.22/module/ExternalProject.html
- https://github.com/actions/checkout
- https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching

### Additional context

Suggested rollout:

1. Add a non-required `build-asan-x86-host` job alongside the existing
   container build.
2. Establish a host-only timing baseline before changing dependency fetching.
3. Prefetch the largest or least reliable FetchContent dependencies first.
4. Expand checkout coverage after measuring checkout-step overhead.
5. Add an archive-only cache for ExternalProject downloads, keyed by runner OS,
   architecture, and `hashFiles('extern/**/*.cmake')`.
6. Compare cold and warm runs, then replace the original build job.
7. Audit the uploaded binaries with `ldd` because they are currently compiled
   on one environment and executed by downstream Ubuntu 22.04 container jobs.

Security requirements:

- Declare `permissions: contents: read`.
- Set `persist-credentials: false` on repository checkouts.
- Continue using the `pull_request` event rather than `pull_request_target`.
- Treat restored cache contents as untrusted and retain existing source hash
  validation.

## Related issues

_None found._

---

wxyucs:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Improvement / refactor
- [ ] Documentation
- [x] CI / build / infrastructure

## What changed

- move the PR ASan build from the build container onto the Ubuntu 22.04 runner
- pre-check out public CMake dependencies with pinned commits through a reusable composite action
- pass `FETCHCONTENT_BASE_DIR` and explicit `FETCHCONTENT_SOURCE_DIR_*` overrides into CMake
- cache third-party download archives separately from checked-out source trees
- reuse the same dependency preparation in lint and coverage jobs
- make MKL installation optional for host CI while preserving the existing default
- preserve fail-fast apt update/install behavior on x86_64 and aarch64
- expose the downloaded host build's shared library to containerized test consumers

Closes: #2555
Related: #2534, #2508

## Test evidence

- [x] `git diff --check`
- [x] YAML parsed successfully with PyYAML, including all three runtime library paths
- [x] `bash -n scripts/deps/install_deps_ubuntu.sh`
- [x] mocked failed apt updates on x86_64 and aarch64 and verified install was not called
- [x] Make dry-run verified the FetchContent arguments
- [x] verified every pinned checkout ref against GitHub
- [x] configured a real ASan/tests/tools build with strict system OpenBLAS and all source overrides
- [x] built `Catch2`, `cpuinfo`, `fmt`, `roaring`, and `thread_pool` targets successfully

Full workflow validation will run in this PR.

## Compatibility and risk

No public API or runtime behavior changes. The main risk is host-runner package/toolchain drift; the workflow pins Ubuntu 22.04 and Clang tooling versions, logs tool versions, and can be rolled back by restoring the container build and URL-based FetchContent population.

## Documentation

No user-facing documentation change is needed for this CI-only update.


scripts/deps/install_deps_ubuntu.sh
```diff
@@ -1,12 +1,41 @@
...
+    if [[ "$install_intel_mkl" == "ON" ]]; then
+        arch_pkgs+=(intel-mkl)
+    elif [[ "$install_intel_mkl" == "OFF" ]]; then
+        :
+    else
+        echo "VSAG_INSTALL_INTEL_MKL must be ON or OFF: $install_intel_mkl"
+        exit 1
+    fi
+    apt update
+    DEBIAN_FRONTEND=noninteractive apt install -y "${common_pkgs[@]}" "${arch_pkgs[@]}"
```
LHT129: [suggestion] The `&&` chaining between `apt update` and `apt install` was removed in the refactor. Previously, a failed `apt update` would stop the script early with a clear exit code. Now, if `apt update` fails (e.g. due to a transient network issue), the script will proceed to `apt install` which will fail with a less specific error, making troubleshooting harder.

Consider restoring `&&` or adding `set -e` at the top of the script:

```bash
apt update && DEBIAN_FRONTEND=noninteractive apt install -y "${common_pkgs[@]}" "${arch_pkgs[@]}"
```

or add `set -euo pipefail` at the top to fail fast on any error.

wxyucs: Implemented in 34953825. Restored apt update && apt install for both x86_64 and aarch64, so a failed index refresh cannot fall through to installation. Validation: bash -n passes, and mocked update failures on both architecture branches recorded exactly one apt call (update) with no install call.

.github/workflows/pr-ci.yml
```diff
@@ -242,6 +266,8 @@ jobs:
     needs: [changes, build-asan-x86]
     if: needs.changes.outputs.cpp == 'true'
     runs-on: ubuntu-22.04
+    env:
+      LD_LIBRARY_PATH: ${{ github.workspace }}/build/src
```
LHT129, line 270: [suggestion] The `LD_LIBRARY_PATH` is added to `test-asan-x86-functests`, `test-example-x86`, and `compatibility` jobs so the downloaded shared library can be found at runtime. However, `test-asan-x86-unittests` (line 228) does not receive the same `LD_LIBRARY_PATH` env. If the unittests binary also links against the shared VSAG library (built on the host runner and downloaded as an artifact), it may fail to locate `libvsag.so` at runtime.

Consider adding the same `env` block to `test-asan-x86-unittests` for consistency:

```yaml
  test-asan-x86-unittests:
    ...
    env:
      LD_LIBRARY_PATH: ${{ github.workspace }}/build/src
```

LHT129 (commented):
Reviewed the latest commit ba7f65ac. The changes look good:

- The `LD_LIBRARY_PATH` additions to `test-asan-x86-functests`, `test-example-x86`, and `compatibility` jobs correctly expose the host-built shared library to containerized test consumers.
- The `apt update && apt install` chain has been properly restored in `install_deps_ubuntu.sh` (addressing the earlier review comment).
- All `FETCHCONTENT_SOURCE_DIR_*` variable names in the Makefile correctly match the CMake `FetchContent_Declare` names.

One existing open item: `test-asan-x86-unittests` still lacks the `LD_LIBRARY_PATH` env (already noted in a separate inline comment). Otherwise, LGTM.

.github/workflows/pr-ci.yml
wxyucs: No change needed. Test X86 Unittests passed in the original run (job 90267449593) without LD_LIBRARY_PATH and passed again after this fix in run 30430207193 (job 90516917787). Unlike functests, examples, and compatibility, it has no runtime dependency on the downloaded shared lib; adding the path would be unnecessary and could mask a future linkage regression.

LHT129 (approved):
LGTM

wxyucs:
image

PR #2567, opened 2026-07-29
feat(io): add optional read cache for disk IO

### LINKED ISSUE #2420: Add CacheIO read-cache layer for disk IO
## Background

VSAG IO module provides multiple disk IO implementations (BufferIO, MMapIO, AsyncIO) for vector index persistence and loading. In vector search scenarios, graph traversal (e.g. HNSW/HGraph) generates many random small reads with strong locality — hot nodes are accessed repeatedly. Currently all reads go directly to disk (pread/mmap page fault) without leveraging access locality, leaving optimization opportunity.

## Proposed Solution

Add CacheIO<InnerIO> as a read-cache layer that wraps any disk-backed IO:

- Page-based cache management (default 128KB, configurable)
- Configurable total cache size (default 256MB)
- Pluggable eviction strategy (LRU as first implementation)
- Read/DirectRead/MultiRead go through cache; Write passes through to InnerIO
- CRTP architecture, zero virtual function overhead

## Related PR

https://github.com/antgroup/vsag/pull/2419

LHT129:
## Summary

Add an optional page-based read cache to concrete disk IO implementations.

Closes #2420

## Configuration

Keep the concrete IO type and set `enable_read_cache` to true. `total_cache_size` supplies the page-cache budget.

## Validation

- make fmt
- git diff --check
- source ~/.bashrc && cmake -S . -B build-release && cmake --build build-release --target datacell -j 1

Local test configuration is blocked by the environment Boost mirror HTTP 403; CI runs the full test suite.


src/io/read_cache/page.h
```diff
@@ -0,0 +1,58 @@
...
+#include "vsag/allocator.h"
+
+namespace vsag {
+
+class Page {
+public:
+    explicit Page(Allocator* allocator) : allocator_(allocator) {
+        data_ = static_cast<uint8_t*>(allocator_->Allocate(Page::DEFAULT_PAGE_SIZE));
+    }
+
```
LHT129, line 29: [suggestion] `Page` constructor does not check if `Allocator::Allocate` returns `nullptr`. If allocation fails, `data_` will be `nullptr` and subsequent access to `Data()` would dereference a null pointer.

Consider adding a null check or using an exception:

```cpp
explicit Page(Allocator* allocator) : allocator_(allocator) {
    data_ = static_cast<uint8_t*>(allocator_->Allocate(Page::DEFAULT_PAGE_SIZE));
    if (data_ == nullptr) {
        throw VsagException(ErrorType::NO_ENOUGH_MEMORY, "failed to allocate page");
    }
}
```

src/io/common/basic_io.h
```diff
@@ -232,24 +271,40 @@ class BasicIO {
...
+        ClearCache();
     }
 
     inline void
     Shrink(uint64_t size) {
         if constexpr (has_ShrinkImpl<IOTmpl>::value) {
-            return cast().ShrinkImpl(size);
+            cast().ShrinkImpl(size);
         } else {
             if (size <= this->size_) {
```
LHT129, line 324: [suggestion] `GetOrLoadPage` holds `cache_mutex_` while calling `cast().ReadImpl(...)`, which may perform blocking disk I/O (e.g., for `BufferIO`, `AsyncIO`, `MMapIO`). This means all concurrent cache lookups for other pages will be blocked during the I/O operation.

Consider a double-checked locking pattern: check the cache without the lock, and only acquire the lock for insertion after the I/O completes:

```cpp
PagePtr GetOrLoadPage(uint64_t page_id) const {
    // ... bounds checks ...
    auto page = cache_->Get(page_id);
    if (page != nullptr) {
        return page;
    }
    auto new_page = std::make_shared<Page>(allocator_);
    if (new_page->Data() == nullptr) {
        return nullptr;
    }
    uint64_t read_size = std::min(Page::DEFAULT_PAGE_SIZE, size_ - offset);
    if (not cast().ReadImpl(read_size, offset, new_page->Data())) {
        return nullptr;
    }
    std::scoped_lock<std::mutex> lock(cache_mutex_);
    return cache_->Insert(page_id, std::move(new_page));
}
```

This way I/O happens outside the lock, and `PageCache::Insert` handles the duplicate-insert case internally.

src/io/read_cache/page_cache.cpp
```diff
@@ -0,0 +1,84 @@
...
+    std::scoped_lock<std::mutex> lock(mutex_);
+    auto existing = pages_.find(page_id);
+    if (existing != pages_.end()) {
+        OnAccess(page_id);
+        return existing->second;
+    }
+    if (max_pages_ == 0) {
+        return page;
+    }
+    while (pages_.size() >= max_pages_) {
```
LHT129, line 46: [suggestion] When `PickVictim()` returns `UINT64_MAX` or an invalid page_id, the fallback picks `pages_.begin()->first`. This could evict a page that was just inserted or accessed, defeating the eviction policy. While this is a rare edge case (only when the derived class returns an invalid victim), it would be safer to log a warning here so the issue can be diagnosed in production.

src/datacell/bucket_datacell.h
```diff
@@ -23,12 +23,25 @@
...
+#include "io/read_cache/page.h"
 #include "quantization/product_quantization/pq_fastscan_quantizer.h"
 #include "simd/fp32_simd.h"
 #include "utils/byte_buffer.h"
 
 namespace vsag {
 
+inline IOParamPtr
+AdjustBucketReadCacheParam(const IOParamPtr& io_param, BucketIdType bucket_count) {
+    if (io_param == nullptr or not io_param->enable_read_cache_ or bucket_count == 0) {
```
LHT129, line 35: [note] `AdjustBucketReadCacheParam` divides the total cache pages evenly across buckets. When `total_pages < bucket_count`, `pages_per_bucket` becomes 0, which means each bucket gets a cache size of 0 and `EnableReadCache` will reset the cache (no caching). This is correct behavior, but it might be surprising to users who configure a small cache with many buckets — they would get no caching at all. Consider documenting this behavior or adding a minimum per-bucket page count.

src/io/common/basic_io.h
LHT129, line 324: [suggestion] `GetOrLoadPage` holds `cache_mutex_` while calling `cast().ReadImpl(...)`, which may perform blocking disk I/O. This means all concurrent cache lookups for other pages will be blocked during the I/O operation.

Consider a double-checked locking pattern: perform the I/O outside the lock, then only acquire the lock for insertion. Since `PageCache::Insert` already handles the duplicate-insert case (returns the existing page), this is safe:

```cpp
PagePtr GetOrLoadPage(uint64_t page_id) const {
    // ... bounds checks ...
    {
        std::scoped_lock<std::mutex> lock(cache_mutex_);
        auto page = cache_->Get(page_id);
        if (page != nullptr) return page;
    }
    auto new_page = std::make_shared<Page>(allocator_);
    if (new_page->Data() == nullptr) return nullptr;
    uint64_t read_size = std::min(Page::DEFAULT_PAGE_SIZE, size_ - offset);
    if (not cast().ReadImpl(read_size, offset, new_page->Data())) return nullptr;
    std::scoped_lock<std::mutex> lock(cache_mutex_);
    return cache_->Insert(page_id, std::move(new_page));
}
```

src/io/read_cache/page_cache.cpp
LHT129, line 46: [suggestion] When `PickVictim()` returns `UINT64_MAX` or an invalid page_id, the fallback picks `pages_.begin()->first`. This could evict a page that was just inserted or accessed, defeating the eviction policy. While this is a rare edge case, consider logging a warning here so the issue can be diagnosed in production.

src/datacell/bucket_datacell.h
LHT129, line 35: [note] `AdjustBucketReadCacheParam` divides the total cache pages evenly across buckets. When `total_pages < bucket_count`, `pages_per_bucket` becomes 0, and `EnableReadCache` will reset the cache (no caching). This is correct but may surprise users who configure a small cache with many buckets. Consider documenting this behavior.

src/io/common/basic_io.h
```diff
@@ -101,6 +110,22 @@ class BasicIO {
     [[nodiscard]] inline const uint8_t*
     Read(uint64_t size, uint64_t offset, bool& need_release) const {
         static_assert(has_DirectReadImpl<IOTmpl>::value);
+        if (cache_ != nullptr) {
+            need_release = false;
+            if (size == 0 or not IsValidRange(size, offset)) {
+                return nullptr;
+            }
+            auto* data = static_cast<uint8_t*>(allocator_->Allocate(size));
```
LHT129: [note] When the read cache is enabled, `DirectReadImpl` is bypassed entirely — the cache path allocates a new buffer, copies data into it via `ReadCached`, and the caller must call `Release` to free it. This changes the semantics from zero-copy (returning an internal pointer) to always-copy when the cache is active. Callers that rely on `need_release == false` for performance (e.g., `scan_bucket_by_id` in `bucket_datacell.h`) will now always pay the allocation + copy cost. This is expected given the cache architecture, but worth being aware of for hot paths.

src/io/common/basic_io.h
```diff
@@ -232,24 +271,40 @@ class BasicIO {
...
     Shrink(uint64_t size) {
         if constexpr (has_ShrinkImpl<IOTmpl>::value) {
-            return cast().ShrinkImpl(size);
+            cast().ShrinkImpl(size);
         } else {
             if (size <= this->size_) {
                 this->size_ = size;
             }
         }
+        ClearCache();
```
LHT129: [suggestion] The `GetMemoryUsage()` method in `BasicIO` only returns `size_`, which does not account for the memory consumed by the page cache (`cache_`). When the read cache is enabled, the actual memory footprint can be significantly larger. This could lead to inaccurate memory budgeting.

Consider including the cache memory in the usage calculation:

```cpp
inline int64_t
GetMemoryUsage() const {
    int64_t usage = this->size_;
    if (cache_ != nullptr) {
        usage += static_cast<int64_t>(cache_->Size() * Page::DEFAULT_PAGE_SIZE);
    }
    return usage;
}
```

src/io/common/basic_io.h
```diff
@@ -101,6 +110,22 @@ class BasicIO {
     [[nodiscard]] inline const uint8_t*
```
LHT129, line 122: [suggestion] The `DirectReadImpl` overload (returning `const uint8_t*` with `need_release`) allocates a temporary buffer via `allocator_->Allocate(size)` and copies data through `ReadCached`. When the cache is active, this always sets `need_release = true`, which changes the semantics for callers that previously relied on `need_release == false` for zero-copy reads (e.g., mmap-backed IO).

This is an intentional design trade-off (cache coherence requires copying), but it would be helpful to document this behavioral change in the method comment, since callers on hot paths like `scan_bucket_by_id` will now always pay the allocation + copy cost when the cache is active.

src/io/common/basic_io.h
```diff
@@ -232,24 +271,40 @@ class BasicIO {
...
     }
 
     inline void
     Shrink(uint64_t size) {
         if constexpr (has_ShrinkImpl<IOTmpl>::value) {
-            return cast().ShrinkImpl(size);
+            cast().ShrinkImpl(size);
         } else {
             if (size <= this->size_) {
                 this->size_ = size;
```
LHT129, line 325: [note] The `Shrink` method previously had `return cast().ShrinkImpl(size);` which prevented the else branch from executing when the CRTP derived class has `ShrinkImpl`. Now the `return` is removed, so `ClearCache()` always executes after `Shrink`. This is the correct behavior (shrinking changes the valid data range, so the cache should be cleared regardless of path), but it is worth confirming that no `ShrinkImpl` implementation relies on the old early-return behavior.

LHT129 (commented):
## Code Review Summary

Thanks for this PR! The page-based read cache is a well-structured addition. Here are my findings:

### Already Noted (by self-review)
- **Page constructor null check** — `Allocate` return value not checked (page.h)
- **GetOrLoadPage mutex contention** — cache mutex held during disk I/O (basic_io.h)
- **PageCache::Insert eviction logic** — fallback to `pages_.begin()->first()` (page_cache.cpp)

### Additional Findings

**[suggestion] GetMemoryUsage doesn't account for cache memory**
The `BasicIO::GetMemoryUsage()` method only returns `size_`, not including the page cache memory. When the read cache is enabled, the actual memory footprint can be significantly larger, leading to inaccurate memory budgeting. Consider adding `cache_->Size() * Page::DEFAULT_PAGE_SIZE` to the returned value.

**[suggestion] Serialize populates cache with transient reads**
`Serialize` calls `this->Read(...)` which goes through `ReadCached` when the cache is enabled. This populates the cache with every page read during serialization, potentially evicting useful pages that were already cached. Consider clearing the cache before serialization or adding a bypass flag.

**[note] Shrink return removal**
The `Shrink` method previously had `return cast().ShrinkImpl(size);` which prevented the else branch from executing. Now the `return` is removed so `ClearCache()` always executes. This is correct behavior, but worth confirming no `ShrinkImpl` implementation relied on the old early-return.

**[suggestion] DirectReadImpl always sets need_release=true with cache**
When the cache is active, the `DirectReadImpl` overload that returns `const uint8_t*` always allocates and copies, setting `need_release = true`. This changes semantics for callers that previously relied on zero-copy reads. Worth documenting.

**[note] BucketDataCell GetMemoryUsage under-reports**
`AdjustBucketReadCacheParam` creates separate IO instances per bucket, each with its own cache, but `BucketDataCell::GetMemoryUsage()` doesn't include cache memory. Consistent with existing `IOArray` behavior, but worth noting for capacity planning.

### Overall Assessment

The design is clean: pluggable eviction policy (PageCache base + LRUPageCache), `shared_ptr` ownership so evicted pages remain valid for in-flight readers, and proper cache invalidation on writes. The test coverage is good with unit tests for PageCache/LRUPageCache and integration tests for the full read/write/invalidation flow. The examples are helpful for demonstrating usage.

No blocking issues found. The suggestions above are non-blocking improvements.

src/io/buffer_io/buffer_io.cpp
```diff
@@ -36,6 +36,10 @@ BufferIO::BufferIO(std::string filename, Allocator* allocator)
         throw VsagException(ErrorType::INTERNAL_ERROR,
                             fmt::format("open file {} error {}", this->filepath_, strerror(errno)));
     }
+    // Stat file to get size
+    if (this->exist_file_) {
+        this->size_ = std::filesystem::file_size(this->filepath_);
```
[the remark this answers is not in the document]
LHT129: Fixed in a5f9bb54: Deserialize now resizes the backing IO to the serialized extent before writing, which clears stale physical and logical tails.

src/io/common/basic_io.h
```diff
@@ -210,6 +248,7 @@ class BasicIO {
      */
     inline void
     InitIO(const IOParamPtr& io_param) {
+        EnableReadCache(io_param);
```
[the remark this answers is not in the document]
LHT129: Fixed in a5f9bb54: InitIO now retains an existing configured cache when binding a ReaderIOParameter that does not explicitly enable one.

src/datacell/bucket_datacell.h
```diff
@@ -23,12 +23,25 @@
...
+        return io_param;
+    }
+    JsonType json = io_param->ToJson();
+    uint64_t total_pages = io_param->read_cache_total_size_ / Page::DEFAULT_PAGE_SIZE;
+    uint64_t pages_per_bucket = total_pages / bucket_count;
+    json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(pages_per_bucket * Page::DEFAULT_PAGE_SIZE);
+    return IOParameter::GetIOParameterByJson(json);
```
LHT129, line 42: [suggestion] `AdjustBucketReadCacheParam` round-trips through JSON serialization (`ToJson` → modify → `GetIOParameterByJson`) solely to adjust the per-bucket cache size. This is both inefficient (full serialize/deserialize cycle) and fragile — if any `ToJson` implementation omits a type-specific field, it would be silently lost during the round-trip.

Consider adding a lightweight `Clone()` or `WithReadCacheSize(uint64_t)` method to `IOParameter` that copies the parameter and only adjusts the cache size field, avoiding the JSON round-trip entirely.

src/io/noncontinuous_io/noncontinuous_io.h
```diff
@@ -91,7 +91,7 @@ class NonContinuousIO : public BasicIO<NonContinuousIO<IOTmpl>> {
         while (cur_size < size) {
             auto area = start_area->first;
             auto area_size = std::min(size - cur_size, area.size - (start_offset - area.offset));
-            inner_io_->WriteImpl(data + cur_size, area_size, start_offset);
+            inner_io_->Write(data + cur_size, area_size, start_offset);
```
LHT129, line 94: [note] The change from `inner_io_->WriteImpl(...)` to `inner_io_->Write(...)` (and similarly `MultiReadImpl` → `MultiRead`) is a good improvement: it routes writes through `BasicIO::Write`, which now calls `InvalidateCacheRange` to keep the inner IO's read cache coherent after writes. Reads through `NonContinuousIO` also benefit from the inner IO's cache via `BasicIO::MultiRead`.

This is worth calling out because it changes the call path for NonContinuousIO — previously it bypassed the `BasicIO` wrapper methods entirely.

LHT129 (commented):
## Code Review Summary

### Overview
This PR adds an optional page-based read cache for disk IO operations in VSAG. The implementation is well-structured with a pluggable eviction policy architecture, proper mutex-based thread safety, and integration across all IO backends (BufferIO, AsyncIO, MMapIO, UringIO, MemoryIO, MemoryBlockIO, ReaderIO, NonContinuousIO).

### What was reviewed
- Core cache infrastructure: `Page`, `PageCache`, `LRUPageCache` (new files)
- Cache integration in `BasicIO` template (Read, DirectRead, MultiRead, Write, InitIO, Deserialize)
- IO parameter configuration (`IOParameter::LoadReadCacheConfig`, `AppendReadCacheConfig`)
- Per-bucket cache size adjustment (`AdjustBucketReadCacheParam`)
- External parameter mapping for HGraph and IVF indexes
- All IO backend constructors (EnableReadCache call)
- NonContinuousIO routing change (WriteImpl→Write, MultiReadImpl→MultiRead)
- Test coverage (unit tests for PageCache/LRU, integration tests for HGraph/IVF with async backend)
- Example programs

### Issues previously identified and fixed in a5f9bb54
- **Reader binding disables cache**: InitIO now preserves existing cache when binding a ReaderIOParameter
- **Stale logical tails in backing files**: Deserialize now resizes the backing IO to the serialized extent

### Remaining open issues (from other reviewers)
- Page constructor null check after Allocate
- GetOrLoadPage holds cache_mutex_ during blocking I/O
- PickVictim fallback edge case
- AdjustBucketReadCacheParam division when total_pages < bucket_count
- DirectRead semantics change when cache is active
- GetMemoryUsage does not account for cache memory
- Shrink behavior change (ClearCache always called)
- P1: Preserve deserialized section length when binding reader
- P1: Clear cached pages when rebinding backing reader
- P2: Avoid resizing MMapIO to zero-length mapping
- P2: Reject non-integer cache-size values

### New findings in this review
1. **[suggestion] AdjustBucketReadCacheParam JSON round-trip**: The function serializes the entire IO parameter to JSON and deserializes it back just to adjust the cache size. This is fragile — if any `ToJson` implementation misses a field, it would be silently lost.
2. **[note] NonContinuousIO routing change**: The switch from `WriteImpl`/`MultiReadImpl` to `Write`/`MultiRead` is a positive change that ensures cache coherence for non-continuous IO.

### Overall assessment
The implementation is solid. The two P2 issues that were fixed in this commit address the most significant concerns. The remaining open issues are primarily suggestions for improvement rather than correctness bugs. The test coverage is reasonable with unit tests for the cache components and integration tests for HGraph/IVF with async backend.

LHT129:
## Code Review Summary

### Overview
This PR adds an optional page-based read cache for disk IO operations in VSAG. The implementation is well-structured with a pluggable eviction policy architecture, proper mutex-based thread safety, and integration across all IO backends (BufferIO, AsyncIO, MMapIO, UringIO, MemoryIO, MemoryBlockIO, ReaderIO, NonContinuousIO).

### What was reviewed
- Core cache infrastructure: `Page`, `PageCache`, `LRUPageCache` (new files)
- Cache integration in `BasicIO` template (Read, DirectRead, MultiRead, Write, InitIO, Deserialize)
- IO parameter configuration (`IOParameter::LoadReadCacheConfig`, `AppendReadCacheConfig`)
- Per-bucket cache size adjustment (`AdjustBucketReadCacheParam`)
- External parameter mapping for HGraph and IVF indexes
- All IO backend constructors (EnableReadCache call)
- NonContinuousIO routing change (WriteImpl→Write, MultiReadImpl→MultiRead)
- Test coverage (unit tests for PageCache/LRU, integration tests for HGraph/IVF with async backend)
- Example programs

### Issues previously identified and fixed in a5f9bb54
- **Reader binding disables cache**: InitIO now preserves existing cache when binding a ReaderIOParameter
- **Stale logical tails in backing files**: Deserialize now resizes the backing IO to the serialized extent

### Remaining open issues (from other reviewers)
- Page constructor null check after Allocate
- GetOrLoadPage holds cache_mutex_ during blocking I/O
- PickVictim fallback edge case
- AdjustBucketReadCacheParam division when total_pages < bucket_count
- DirectRead semantics change when cache is active
- GetMemoryUsage does not account for cache memory
- Shrink behavior change (ClearCache always called)
- P1: Preserve deserialized section length when binding reader
- P1: Clear cached pages when rebinding backing reader
- P2: Avoid resizing MMapIO to zero-length mapping
- P2: Reject non-integer cache-size values

### New findings in this review
1. **[suggestion] AdjustBucketReadCacheParam JSON round-trip**: The function serializes the entire IO parameter to JSON and deserializes it back just to adjust the cache size. This is fragile — if any `ToJson` implementation misses a field, it would be silently lost.
2. **[note] NonContinuousIO routing change**: The switch from `WriteImpl`/`MultiReadImpl` to `Write`/`MultiRead` is a positive change that ensures cache coherence for non-continuous IO.

### Overall assessment
The implementation is solid. The two P2 issues that were fixed in this commit address the most significant concerns. The remaining open issues are primarily suggestions for improvement rather than correctness bugs. The test coverage is reasonable with unit tests for the cache components and integration tests for HGraph/IVF with async backend.

src/io/reader_io/reader_io.cpp
```diff
@@ -46,6 +47,7 @@ ReaderIO::InitIOImpl(const vsag::IOParamPtr& io_param) {
                             "ReaderIOParam is required for ReaderIO initialization.");
     }
     reader_ = reader_param->reader;
+    this->size_ = reader_->Size();
```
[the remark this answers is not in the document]
LHT129: Fixed in ed0f8f0b: ReaderIO::InitIOImpl now only sets size_ from reader_->Size() when size_ == 0, preserving any previously deserialized section extent.

src/io/common/basic_io.h
```diff
@@ -210,6 +256,11 @@ class BasicIO {
      */
     inline void
     InitIO(const IOParamPtr& io_param) {
+        // Reader-backed IOs are initialized after construction. Keep the cache
+        // configured by the concrete IO unless InitIO explicitly enables it.
+        if (cache_ == nullptr or (io_param != nullptr and io_param->enable_read_cache_)) {
```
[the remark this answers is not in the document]
LHT129: Fixed in ed0f8f0b: BasicIO::InitIO now calls ClearCache() when rebinding without explicitly enabling cache, so stale pages from the previous backing reader are evicted while the cache structure is preserved.

src/io/common/basic_io.h
```diff
@@ -170,6 +204,14 @@ class BasicIO {
             reader.Seek(reader.GetCursor() + size);
             this->size_ = std::max(this->size_, size);
         } else {
+            // Reset the logical and physical extent so a shorter deserialization
+            // cannot retain stale bytes from a previously opened file.
+            if constexpr (has_ResizeImpl<IOTmpl>::value) {
+                Resize(size);
```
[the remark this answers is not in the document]
LHT129: Fixed in ed0f8f0b: Deserialize now special-cases size == 0 to set logical size and clear cache without invoking ResizeImpl, avoiding the zero-length mremap/mmap failure in MMapIO.

src/io/common/io_parameter.cpp
```diff
@@ -82,8 +82,31 @@ IOParameter::GetIOParameterByJson(const JsonType& json) {
...
+
+void
+IOParameter::LoadReadCacheConfig(const JsonType& json) {
+    if (json.Contains(READ_CACHE_ENABLED_KEY)) {
+        CHECK_ARGUMENT(json[READ_CACHE_ENABLED_KEY].IsBool(),
+                       "enable_read_cache must be a boolean");
+        enable_read_cache_ = json[READ_CACHE_ENABLED_KEY].GetBool();
+    }
+    if (json.Contains(READ_CACHE_TOTAL_CACHE_SIZE_KEY)) {
+        read_cache_total_size_ = json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].GetUint64();
```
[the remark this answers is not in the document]
LHT129: Fixed in ed0f8f0b: LoadReadCacheConfig now validates total_cache_size with IsNumberUnsigned() before assignment, rejecting floating-point or negative values.

src/io/common/basic_io.h
```diff
@@ -232,24 +302,46 @@ class BasicIO {
...
+        }
     }
 
     inline void
     Shrink(uint64_t size) {
         if constexpr (has_ShrinkImpl<IOTmpl>::value) {
-            return cast().ShrinkImpl(size);
+            cast().ShrinkImpl(size);
         } else {
             if (size <= this->size_) {
```
LHT129, line 324: [suggestion] `GetOrLoadPage` still holds `cache_mutex_` while calling `cast().ReadImpl(...)`, which may perform blocking disk I/O. This means all concurrent cache lookups for other pages will be blocked during the I/O operation.

The previous review suggested a double-checked locking pattern but it was not adopted. Since `PageCache::Insert` already handles the duplicate-insert case (returns the existing page), the I/O can safely happen outside the lock:

```cpp
PagePtr GetOrLoadPage(uint64_t page_id) const {
    if (page_id > UINT64_MAX / Page::DEFAULT_PAGE_SIZE) {
        return nullptr;
    }
    uint64_t offset = page_id * Page::DEFAULT_PAGE_SIZE;
    if (offset >= size_) {
        return nullptr;
    }
    {
        std::scoped_lock<std::mutex> lock(cache_mutex_);
        auto page = cache_->Get(page_id);
        if (page != nullptr) {
            return page;
        }
    }
    auto new_page = std::make_shared<Page>(allocator_);
    if (new_page->Data() == nullptr) {
        return nullptr;
    }
    uint64_t read_size = std::min(Page::DEFAULT_PAGE_SIZE, size_ - offset);
    if (not cast().ReadImpl(read_size, offset, new_page->Data())) {
        return nullptr;
    }
    std::scoped_lock<std::mutex> lock(cache_mutex_);
    return cache_->Insert(page_id, std::move(new_page));
}
```

This way I/O happens outside the lock, and concurrent readers for different pages are not serialized.

src/io/read_cache/page.h
LHT129, line 29: [suggestion] The `Page` constructor does not check if `Allocator::Allocate` returns `nullptr`. If allocation fails, `data_` will be `nullptr` and subsequent access to `Data()` would dereference a null pointer.

This was flagged in the previous review but remains unaddressed. Consider adding a null check:

```cpp
explicit Page(Allocator* allocator) : allocator_(allocator) {
    data_ = static_cast<uint8_t*>(allocator_->Allocate(Page::DEFAULT_PAGE_SIZE));
    if (data_ == nullptr) {
        throw std::bad_alloc();
    }
}
```

Alternatively, if the project prefers not to throw from constructors, the callers in `GetOrLoadPage` and tests already check `new_page->Data() == nullptr`, so the null-pointer dereference risk is partially mitigated at the call sites. However, it would be safer to enforce the invariant at construction time.

src/io/noncontinuous_io/noncontinuous_io.h
LHT129, line 94: [note] `NonContinuousIO` now calls `inner_io_->Write(...)` and `inner_io_->MultiRead(...)` instead of `WriteImpl`/`MultiReadImpl`. This correctly routes through `BasicIO`'s public interface which handles cache invalidation (for writes) and cache reads (for reads). This is a necessary change for the read cache to work correctly with non-continuous IO layouts.

This also means `NonContinuousIO` no longer bypasses the cache layer, which is the correct behavior — the cache operates at the logical offset level, and `NonContinuousIO` maps logical offsets to physical offsets transparently.

src/io/common/basic_io.h
```diff
@@ -170,6 +212,19 @@ class BasicIO {
             reader.Seek(reader.GetCursor() + size);
             this->size_ = std::max(this->size_, size);
         } else {
+            // Reset the logical and physical extent so a shorter deserialization
+            // cannot retain stale bytes from a previously opened file.
+            if constexpr (has_ResizeImpl<IOTmpl>::value) {
+                if (size > 0) {
+                    Resize(size);
+                } else {
+                    this->size_ = 0;
```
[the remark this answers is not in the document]
LHT129: Fixed in 5b2e73d6: zero-length Deserialize now invokes Resize(0) for backends that support zero-size truncation. MMapIO explicitly retains its required non-empty physical mapping while resetting logical size and cache.

src/io/buffer_io/buffer_io.cpp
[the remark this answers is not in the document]
LHT129: Fixed in 5b2e73d6: removed constructor-time file-size seeding. The logical extent is now established only by the current writes or Deserialize, preventing reused backing-file tails from being serialized.

src/io/reader_io/reader_io.cpp
```diff
@@ -46,6 +47,9 @@ ReaderIO::InitIOImpl(const vsag::IOParamPtr& io_param) {
                             "ReaderIOParam is required for ReaderIO initialization.");
     }
     reader_ = reader_param->reader;
+    if (this->size_ == 0) {
```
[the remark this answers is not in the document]
LHT129: Fixed in 5b2e73d6: BasicIO now records whether Deserialize established the logical section. ReaderIO refreshes its size on every standalone binding, while preserving both empty and non-empty deserialized section boundaries.

src/io/reader_io/reader_io.cpp
```diff
@@ -46,6 +47,9 @@ ReaderIO::InitIOImpl(const vsag::IOParamPtr& io_param) {
...
+        this->size_ = reader_->Size();
```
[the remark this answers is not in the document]
LHT129: Fixed in 5b2e73d6: ReaderIO::InitIOImpl now rejects a null reader before assigning it or calling Size().

src/io/common/basic_io.h
```diff
@@ -232,24 +310,51 @@ class BasicIO {
...
+        }
     }
 
     inline void
     Shrink(uint64_t size) {
         if constexpr (has_ShrinkImpl<IOTmpl>::value) {
-            return cast().ShrinkImpl(size);
+            cast().ShrinkImpl(size);
         } else {
             if (size <= this->size_) {
```
LHT129, line 324: [suggestion] `GetOrLoadPage` holds `cache_mutex_` while calling `cast().ReadImpl(...)`, which performs blocking disk I/O. This serializes all concurrent cache lookups for different pages during I/O operations.

Since `PageCache::Insert` already handles the duplicate-insert case (returns the existing page), the I/O can safely happen outside the lock:

```cpp
PagePtr GetOrLoadPage(uint64_t page_id) const {
    if (page_id > UINT64_MAX / Page::DEFAULT_PAGE_SIZE) {
        return nullptr;
    }
    uint64_t offset = page_id * Page::DEFAULT_PAGE_SIZE;
    if (offset >= size_) {
        return nullptr;
    }
    {
        std::scoped_lock<std::mutex> lock(cache_mutex_);
        auto page = cache_->Get(page_id);
        if (page != nullptr) {
            return page;
        }
    }
    auto new_page = std::make_shared<Page>(allocator_);
    if (new_page->Data() == nullptr) {
        return nullptr;
    }
    uint64_t read_size = std::min(Page::DEFAULT_PAGE_SIZE, size_ - offset);
    if (not cast().ReadImpl(read_size, offset, new_page->Data())) {
        return nullptr;
    }
    std::scoped_lock<std::mutex> lock(cache_mutex_);
    return cache_->Insert(page_id, std::move(new_page));
}
```

This way I/O happens outside the lock, and concurrent readers for different pages are not serialized.

src/io/read_cache/page.h
LHT129, line 29: [suggestion] The `Page` constructor does not check if `Allocator::Allocate` returns `nullptr`. If allocation fails, `data_` will be `nullptr` and subsequent access to `Data()` would dereference a null pointer.

While callers in `GetOrLoadPage` and tests do check `new_page->Data() == nullptr`, it would be safer to enforce the invariant at construction time:

```cpp
explicit Page(Allocator* allocator) : allocator_(allocator) {
    data_ = static_cast<uint8_t*>(allocator_->Allocate(Page::DEFAULT_PAGE_SIZE));
    if (data_ == nullptr) {
        throw std::bad_alloc();
    }
}
```

src/datacell/bucket_datacell.h
LHT129, line 42: [suggestion] `AdjustBucketReadCacheParam` round-trips through JSON serialization (`ToJson` → modify → `GetIOParameterByJson`) solely to adjust the per-bucket cache size. This is both inefficient (full serialize/deserialize cycle) and fragile — if any `ToJson` implementation omits a type-specific field, it would be silently lost during the round-trip.

Consider adding a lightweight `Clone()` or `WithReadCacheSize(uint64_t)` method to `IOParameter` that copies the parameter and only adjusts the cache size field, avoiding the JSON round-trip entirely.

src/io/read_cache/page_cache.cpp
LHT129, line 46: [suggestion] When `PickVictim()` returns `UINT64_MAX` or an invalid page_id, the fallback picks `pages_.begin()->first`. This could evict a page that was just inserted or accessed, defeating the eviction policy. While this is a rare edge case (only when the derived class returns an invalid victim), it would be safer to log a warning here so the issue can be diagnosed in production.

src/datacell/bucket_datacell.h
LHT129, line 35: [note] `AdjustBucketReadCacheParam` divides the total cache pages evenly across buckets. When `total_pages < bucket_count`, `pages_per_bucket` becomes 0, which means each bucket gets a cache size of 0 and `EnableReadCache` will reset the cache (no caching). This is correct behavior, but it might be surprising to users who configure a small cache with many buckets — they would get no caching at all. Consider documenting this behavior or adding a minimum per-bucket page count.

LHT129 (commented):
Code review for PR #2567: Add optional read cache for disk IO.

**Summary**: This PR introduces a page-based read cache (128KB pages) with LRU eviction for disk IO backends. The cache integrates into the CRTP-based `BasicIO<IOTmpl>` template and is enabled via IO parameters. IVF bucket datacells distribute the cache budget evenly across buckets.

**Overall Assessment**: The design is clean and well-integrated into the existing IO abstraction. The cache correctly handles write-through invalidation, resize/shrink/deserialize cache clearing, and cross-page boundary reads. The test coverage is reasonable for unit-level testing of the cache components.

**General Notes**:

1. **[note] DirectReadImpl bypass with cache**: When the read cache is enabled, `DirectReadImpl` is bypassed entirely — the cache path allocates a new buffer, copies data into it via `ReadCached`, and the caller must call `Release` to free it. This changes the semantics from zero-copy (returning an internal pointer) to always-copy when the cache is active. Callers that rely on `need_release == false` for performance (e.g., `scan_bucket_by_id` in `bucket_datacell.h`) will now always pay the allocation + copy cost. This is expected given the cache architecture, but worth being aware of for hot paths.

2. **[suggestion] GetMemoryUsage does not include cache**: The `GetMemoryUsage()` method only returns `size_`, which does not account for the memory consumed by the page cache. When the read cache is enabled, the actual memory footprint can be significantly larger, leading to inaccurate memory budgeting. Consider including cache memory: `usage += cache_->Size() * Page::DEFAULT_PAGE_SIZE`.

3. **[suggestion] DirectReadImpl overload documentation**: The `DirectReadImpl` overload (returning `const uint8_t*` with `need_release`) allocates a temporary buffer and copies through `ReadCached`. When the cache is active, this always sets `need_release = true`. It would be helpful to document this behavioral change in the method comment.

4. **[note] NonContinuousIO now routes through BasicIO**: The change from `inner_io_->WriteImpl(...)` to `inner_io_->Write(...)` (and similarly `MultiReadImpl` → `MultiRead`) correctly routes through `BasicIO`'s public interface which handles cache invalidation and cache reads. This is a necessary change for read cache correctness.

5. **[note] Shrink now always clears cache**: The `Shrink` method previously had `return cast().ShrinkImpl(size);` which prevented the else branch from executing. Now `ClearCache()` always executes after `Shrink`. This is correct behavior, but worth confirming no `ShrinkImpl` relies on the old early-return behavior.

See inline comments for specific findings.

src/io/common/basic_io.h
```diff
@@ -165,11 +215,24 @@ class BasicIO {
...
+            // Reset the logical and physical extent so a shorter deserialization
+            // cannot retain stale bytes from a previously opened file.
+            if constexpr (has_ResizeImpl<IOTmpl>::value and SupportsZeroSizeResize<IOTmpl>::value) {
+                if (size > 0) {
+                    Resize(size);
+                } else {
+                    Resize(0);
+                }
+            } else {
+                this->size_ = 0;
```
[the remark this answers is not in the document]
LHT129: Fixed in 03199f70: MMapIO now tracks its physical mapped extent independently from the logical size. All remap and unmap operations use that extent, so resetting a logical zero-length deserialization cannot leak a stale mapping.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -199,6 +200,14 @@ HGraph::map_hgraph_param(const JsonType& hgraph_json) {
                 IO_DIRECT_READ_KEY,
             },
         },
+        {
+            HGRAPH_BASE_CACHE_TOTAL_SIZE,
```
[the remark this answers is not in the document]
LHT129: Fixed in 03199f70: RaBitQ split storage now divides an enabled shared ReadCache budget equally between its one-bit and supplement IO parameters. A dedicated supplement parameter remains independently configured.

src/io/common/basic_io.h
```diff
@@ -210,6 +279,15 @@ class BasicIO {
      */
     inline void
     InitIO(const IOParamPtr& io_param) {
+        if constexpr (not InMemory) {
+            if (cache_ == nullptr) {
+                EnableReadCache(io_param);
+            } else if (io_param != nullptr and io_param->enable_read_cache_) {
+                EnableReadCache(io_param);
+            } else {
+                ClearCache();
```
LHT129, line 289: **[P2] InitIO clears existing cache when binding a ReaderIOParameter without cache**

When `cache_ != nullptr` and `InitIO` is called with a `ReaderIOParameter` whose `enable_read_cache_` defaults to `false`, the `else` branch calls `ClearCache()`, discarding the previously configured cache. This happens during `HGraph::SetIO` after `ReaderSet` deserialization, where a fresh `ReaderIOParameter` is used to bind the reader.

A previous review noted this was "Fixed in a5f9bb54" but the current code still clears the cache. Consider preserving the existing cache when the new parameter does not explicitly disable it:

```cpp
if (cache_ == nullptr) {
    EnableReadCache(io_param);
} else if (io_param != nullptr and io_param->enable_read_cache_) {
    EnableReadCache(io_param);
} else if (io_param != nullptr and not io_param->enable_read_cache_) {
    ClearCache();
}
// else: io_param == nullptr, keep existing cache
```

LHT129: The cache configuration is preserved: the cache object stays allocated and retains its capacity. ClearCache only invalidates pages before binding a potentially different reader, preventing stale data from the prior backing store. Subsequent reads refill the configured cache, so no code change is needed.

src/io/mmap_io/mmap_io.h
```diff
@@ -38,6 +38,9 @@ class MMapIO : public BasicIO<MMapIO> {
     /// Indicates deserialization is required when loading from disk.
     static constexpr bool SkipDeserialize = false;
 
+    // MMapIO keeps a non-empty backing mapping for subsequent writes.
+    static constexpr bool SupportsZeroSizeResize = false;
```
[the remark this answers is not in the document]
LHT129: Fixed in 4d926ac6: positive serialized extents always call Resize, including for MMapIO. Only the zero-length case bypasses Resize to retain MMapIO minimum physical mapping.

src/datacell/rabitq_split_datacell.h
```diff
@@ -179,10 +179,11 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
         // memory + supplement on disk). Otherwise fall back to the shared
         // io_param with the legacy file-path suffix to keep the two backing
         // files separate for file-backed IO.
-        const IOParamPtr one_bit_io_param = SuffixIOParam(io_param, "_onebit");
+        const bool shares_io_param = supplement_io_param == nullptr;
+        const IOParamPtr one_bit_io_param = SuffixIOParam(io_param, "_onebit", shares_io_param);
         const IOParamPtr supp_io_param = (supplement_io_param != nullptr)
```
LHT129, line 184: [suggestion] When `split_cache` is true, `SuffixIOParam` halves `read_cache_total_size_` via integer division. If the original cache size is less than `2 * Page::DEFAULT_PAGE_SIZE` (256 KiB), the halved value becomes 0, which causes `EnableReadCache` to silently disable the cache (since `page_count == 0`).

Consider using `std::max` to guarantee at least one page per split:

```cpp
if (split_cache and io_param->enable_read_cache_) {
    uint64_t half = std::max(io_param->read_cache_total_size_ / 2,
                             Page::DEFAULT_PAGE_SIZE);
    json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(half);
}
```

LHT129 (commented):
Code review for commit 4d926ac6.

wxyucs (approved):
lgtm

src/datacell/rabitq_split_datacell.h
```diff
@@ -998,7 +1004,7 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
             return nullptr;
         }
         if (this->supplement_io_type_.empty()) {
-            return SuffixIOParam(io_param, "_supplement");
+            return SuffixIOParam(io_param, "_supplement", true);
```
LHT129, line 1007: **[P2] `RebuildSupplementIOParam` does not halve the cache size in hybrid mode**\n\nWhen `supplement_io_type_` is non-empty (hybrid IO — one-bit and supplement use different backends), this path clones the JSON via `ToJson()` without halving `read_cache_total_size_`. In contrast, when the two storages share the same IO type, the `if` branch above calls `SuffixIOParam(io_param, "_supplement", true)` which correctly halves the cache.\n\nIn hybrid mode, both the one-bit cell (via `SuffixIOParam` with `split_cache=false`) and the supplement cell (via this path without halving) each receive the full configured cache size, consuming **2×** the expected memory. This is consistent with the constructor path (where `supplement_io_param` is passed directly), but it means the user-facing `precise_cache_total_size` parameter does not represent the total cache budget for the precise component — it is per-IO-object.\n\nConsider either:\n1. Halving the cache here when `enable_read_cache_` is true (matching the shared-IO behavior), or\n2. Documenting that in hybrid mode the cache size applies per-IO-backend, not per-component.

src/io/memory_io/memory_io.h
```diff
@@ -73,6 +73,7 @@ class MemoryIO : public BasicIO<MemoryIO> {
      */
     explicit MemoryIO(const IOParamPtr& param, const IndexCommonParam& common_param)
         : MemoryIO(std::dynamic_pointer_cast<MemoryIOParameter>(param), common_param) {
+        EnableReadCache(param);
```
LHT129, line 76: [note] `EnableReadCache(param)` is called here but `MemoryIO` declares `static constexpr bool InMemory = true`, so the `if constexpr (not InMemory)` guard inside `EnableReadCache` makes this a compile-time no-op. The same pattern exists in `MemoryBlockIO`.\n\nThis is harmless but may confuse readers into thinking these in-memory IO types support read caching. Consider removing these calls or adding a comment explaining they are intentionally no-ops for consistency with the other IO constructors.

src/datacell/bucket_datacell.h
```diff
@@ -23,12 +23,25 @@
...
 
+inline IOParamPtr
+AdjustBucketReadCacheParam(const IOParamPtr& io_param, BucketIdType bucket_count) {
+    if (io_param == nullptr or not io_param->enable_read_cache_ or bucket_count == 0) {
+        return io_param;
+    }
+    JsonType json = io_param->ToJson();
+    uint64_t total_pages = io_param->read_cache_total_size_ / Page::DEFAULT_PAGE_SIZE;
+    uint64_t pages_per_bucket = total_pages / bucket_count;
+    json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(pages_per_bucket * Page::DEFAULT_PAGE_SIZE);
```
LHT129, line 41: [note] The per-bucket cache size calculation `pages_per_bucket * Page::DEFAULT_PAGE_SIZE` truncates due to integer division. For example, with `total_cache_size = 10 * 128KB` and `bucket_count = 3`, each bucket gets `3 * 128KB = 384KB`, totaling `3 * 384KB = 1152KB` across all buckets, losing 128KB (1 page) of the original 1280KB cache budget.\n\nThis is minor (at most `(bucket_count - 1) * Page::DEFAULT_PAGE_SIZE` bytes lost), but worth being aware of when configuring small caches with many buckets.

src/datacell/rabitq_split_datacell.h
```diff
@@ -981,6 +984,9 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
             std::string path = json[IO_FILE_PATH_KEY].GetString();
             json[IO_FILE_PATH_KEY].SetString(path + suffix);
         }
+        if (split_cache and io_param->enable_read_cache_) {
+            json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(io_param->read_cache_total_size_ / 2);
```
LHT129, line 988: [note] `io_param->read_cache_total_size_ / 2` uses integer division, which truncates when the cache size is odd. For example, if `read_cache_total_size_ = 1`, each half gets 0, effectively disabling the cache for both halves. While the default cache size (256MB) is always even, user-configured values could be odd.\n\nThis is extremely minor (at most 1 byte lost), but consistent with the same pattern in `AdjustBucketReadCacheParam`.

src/io/memory_io/memory_io.h
LHT129, line 76: **[note] Dead `EnableReadCache` calls for InMemory IOs**

`MemoryIO` and `MemoryBlockIO` both call `EnableReadCache(param)` in their `IOParamPtr` constructors, but `BasicIO<IOTmpl>::EnableReadCache` is a no-op when `InMemory` is `true` (guarded by `if constexpr (not InMemory)`).

These calls are dead code — they compile to nothing and serve no purpose. Consider removing them to avoid misleading future readers into thinking InMemory IOs support caching.

src/datacell/bucket_datacell.h
LHT129, line 41: **[note] `AdjustBucketReadCacheParam` rounding loss**

```cpp
uint64_t pages_per_bucket = total_pages / bucket_count;
json[...].SetUint64(pages_per_bucket * Page::DEFAULT_PAGE_SIZE);
```

Integer division truncates, so up to `(bucket_count - 1) * 128KB` of the cache budget is silently discarded. For example, with 10 buckets and 256MB total (2048 pages), each bucket gets 204 pages (204 * 10 = 2040), losing 8 pages (1MB).

Consider distributing the remainder across buckets (e.g., first N buckets get one extra page) or using `total_cache_size / bucket_count` directly on the byte value.

src/datacell/rabitq_split_datacell.h
```diff
@@ -981,6 +984,9 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
             std::string path = json[IO_FILE_PATH_KEY].GetString();
```
LHT129, line 984: **[note] `SuffixIOParam` split_cache integer division rounding**

```cpp
json[...].SetUint64(io_param->read_cache_total_size_ / 2);
```

When `read_cache_total_size_` is odd, integer division truncates 1 byte. While negligible in practice (1 byte out of 256MB default), this is the same pattern as `AdjustBucketReadCacheParam` and could be made consistent by using page-aligned division:

```cpp
uint64_t half_pages = (io_param->read_cache_total_size_ / Page::DEFAULT_PAGE_SIZE) / 2;
json[...].SetUint64(half_pages * Page::DEFAULT_PAGE_SIZE);
```

src/datacell/rabitq_split_datacell.h
```diff
@@ -998,7 +1004,7 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
```
LHT129, line 997: **[P2] RebuildSupplementIOParam does not halve cache in hybrid mode**

When `supplement_io_type_` is non-empty (hybrid IO mode), `RebuildSupplementIOParam` builds a custom JSON parameter but never applies the `split_cache` halving logic. This means in hybrid mode both the one-bit and supplement cells each get the full cache size — 2× the expected memory budget.

The constructor path (line 186) correctly passes `split_cache=true` for the supplement cell, but `RebuildSupplementIOParam` (used during `InitIO`) only appends the suffix path and IO type override. The `read_cache_total_size_` from the original `io_param` flows through unchanged.

Suggested fix: add the same `split_cache` logic inside the `!supplement_io_type_.empty()` branch:
```cpp
if (io_param->enable_read_cache_) {
    json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(io_param->read_cache_total_size_ / 2);
}
```

LHT129 (commented):
**[P2] RebuildSupplementIOParam does not halve cache in hybrid mode**

In `src/datacell/rabitq_split_datacell.h`, `RebuildSupplementIOParam` (line 1002-1015) builds a custom JSON parameter when `supplement_io_type_` is non-empty (hybrid IO mode), but never applies the `split_cache` halving logic. This means in hybrid mode both the one-bit and supplement cells each get the full cache size — 2× the expected memory budget.

The constructor path (line 186) correctly passes `split_cache=true` for the supplement cell, but `RebuildSupplementIOParam` (used during `InitIO`) only appends the suffix path and IO type override. The `read_cache_total_size_` from the original `io_param` flows through unchanged.

Suggested fix: add the same `split_cache` logic inside the `!supplement_io_type_.empty()` branch:
```cpp
if (io_param->enable_read_cache_) {
    json[READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetUint64(io_param->read_cache_total_size_ / 2);
}
```

PR #2572, opened 2026-07-30
[skip ci] feat(sindi): add hybrid Elias-Fano DMQ encoding

LightWant:
> **Draft backup only. Do not merge.**
>
> This PR is used as cloud storage for the SINDI DMQ Elias–Fano experiment.
> CI is intentionally skipped and the PR should remain in Draft state.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [x] Documentation
- [ ] CI/Build/Infra

## Linked Issue

Not applicable for this draft improvement PR.

## What Changed

- Add per-vector Hybrid Packed/Elias–Fano coding for ordered DMQ compact term IDs.
- Add scalar, batch, and query-driven seek readers for the EF stream.
- Use EF ordinal ranges to access position-aligned DMQ value codes during reranking.
- Reuse the computed ID layout within each candidate.
- Add unit/regression tests and synchronized English/Chinese documentation.
- Add an experiment report with index-size and Release performance results.

## Test Evidence

- [x] Other (described below)

Test details:

```text
git diff --check
build/tests/unittests '[ut][EliasFanoStream],[ut][SparseDmqQuantizer]'

All tests passed:
13 test cases
2220 assertions

Release benchmark:
wholenet-sparse-1m-ip, 8 search threads, topk=10, n_candidate=1000
EF scalar scan: 2.905 ms / 2750 QPS
EF query-driven seek: 2.738 ms / 2914 QPS
Recall@10: 0.9863
```

CI is intentionally skipped because this Draft PR is for cloud backup only.

## Compatibility Impact

- API/ABI compatibility: no public API changes.
- Behavior changes: new DMQ builds may choose EF for ordered compact term IDs.
- Legacy risk: packed-only DMQ indexes do not carry an encoding tag/version and
  need an explicit compatibility strategy before this work is considered mergeable.

- Performance impact: DMQ rerank backend size reduced by 12.58%; complete index
  size reduced by 4.41%; EF seek improved latency by 5.74% and QPS by 5.96%
  relative to scalar EF scanning on the measured workload.
- Concurrency/thread-safety impact: none; readers are query-local.

## Documentation Impact

- [x] Updated docs:
  - [x] Other: SINDI, DMQ, and Elias–Fano English/Chinese documentation and
    `sindi_dmq_elias_fano_report.md`.

## Risk and Rollback

- Risk level: high for merge until legacy packed-index compatibility is resolved.
- Rollback plan: revert commit `4fbd49969`.

## Checklist

- [x] I have added/updated tests for new behavior.
- [x] I have considered API and index compatibility impact.
- [x] I have updated documentation.
- [x] The commit message follows project conventions and starts with `[skip ci]`.

### DIFF
--- a/src/algorithm/sindi/sindi.cpp
+++ b/src/algorithm/sindi/sindi.cpp
@@ -29,6 +29,7 @@
 #include "analyzer/analyzer.h"
 #include "datacell/sparse_dmq_datacell.h"
 #include "datacell/sparse_vector_datacell_parameter.h"
+#include "impl/elias_fano_stream.h"
 #include "impl/heap/standard_heap.h"
 #include "index_feature_list.h"
 #include "io/memory_block_io/memory_block_io_parameter.h"
@@ -1836,7 +1837,15 @@ SINDI::EstimateMemory(uint64_t num_elements) const {
                 std::min<uint64_t>(term_id_limit_, total_sparse_values);
             const uint32_t rerank_id_bits = get_bits_for_term_id_limit(
                 estimated_term_count == 0 ? 0 : static_cast<uint32_t>(estimated_term_count - 1));
-            mem += (total_sparse_values * rerank_id_bits + 7) / 8;
+            if (estimated_term_count != 0 && num_elements != 0) {
+                const uint64_t packed_id_bytes =
+                    (static_cast<uint64_t>(avg_doc_term_length_) * rerank_id_bits + 7) / 8;
+                const uint64_t elias_fano_id_bytes =
+                    EliasFanoStream::GetLayout(avg_doc_term_length_,
+                                               static_cast<uint32_t>(estimated_term_count))
+                        .SizeInBytes();
+                mem += num_elements * std::min(packed_id_bytes, elias_fano_id_bytes);
+            }
             mem += total_sparse_values;
             mem += num_elements * (sizeof(uint64_t) + sizeof(SparseDmqQuantizer::EncodedHeader));
             uint64_t estimated_codebook_count = estimated_term_count;
--- a/src/datacell/sparse_dmq_datacell.cpp
+++ b/src/datacell/sparse_dmq_datacell.cpp
@@ -23,7 +23,8 @@ namespace vsag {
 namespace {
 
 constexpr uint32_t K_SPARSE_DMQ_DATACELL_MAGIC = 0x53444D51U;
-constexpr uint32_t K_SPARSE_DMQ_DATACELL_VERSION = 6;
+constexpr uint32_t K_SPARSE_DMQ_DATACELL_LEGACY_VERSION = 6;
+constexpr uint32_t K_SPARSE_DMQ_DATACELL_VERSION = 7;
 
 }  // namespace
 
@@ -216,7 +217,11 @@ void
 SparseDmqDataCell::Serialize(StreamWriter& writer) {
     std::shared_lock lock(this->mutex_);
     StreamWriter::WriteObj(writer, K_SPARSE_DMQ_DATACELL_MAGIC);
-    StreamWriter::WriteObj(writer, K_SPARSE_DMQ_DATACELL_VERSION);
+    const uint32_t version =
+        quantizer_->GetIdEncodingType() == SparseDmqQuantizer::IdEncodingType::PACKED
+            ? K_SPARSE_DMQ_DATACELL_LEGACY_VERSION
+            : K_SPARSE_DMQ_DATACELL_VERSION;
+    StreamWriter::WriteObj(writer, version);
     StreamWriter::WriteObj(writer, this->total_count_);
     StreamWriter::WriteVector(writer, offsets_);
     StreamWriter::WriteVector(writer, codes_);
@@ -232,12 +237,16 @@ SparseDmqDataCell::Deserialize(lvalue_or_rvalue<StreamReader> reader) {
                    "serialized DMQ datacell has invalid magic");
     uint32_t version = 0;
     StreamReader::ReadObj(reader, version);
-    CHECK_ARGUMENT(version == K_SPARSE_DMQ_DATACELL_VERSION,
-                   fmt::format("unsupported sparse DMQ datacell version {}", version));
+    CHECK_ARGUMENT(
+        version == K_SPARSE_DMQ_DATACELL_LEGACY_VERSION || version == K_SPARSE_DMQ_DATACELL_VERSION,
+        fmt::format("unsupported sparse DMQ datacell version {}", version));
     StreamReader::ReadObj(reader, this->total_count_);
     StreamReader::ReadVector(reader, offsets_);
     StreamReader::ReadVector(reader, codes_);
     quantizer_->Deserialize(reader);
+    quantizer_->SetIdEncodingType(version == K_SPARSE_DMQ_DATACELL_LEGACY_VERSION
+                                      ? SparseDmqQuantizer::IdEncodingType::PACKED
+                                      : SparseDmqQuantizer::IdEncodingType::HYBRID_ELIAS_FANO);
     CHECK_ARGUMENT(offsets_.size() == static_cast<uint64_t>(this->total_count_) + 1,
                    "serialized DMQ offset count is inconsistent");
     CHECK_ARGUMENT(not offsets_.empty(), "serialized DMQ offsets are empty");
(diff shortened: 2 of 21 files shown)


LHT129 (commented):
Code review for the Elias-Fano DMQ encoding experiment PR. This is a draft/backup PR, so comments are suggestions only.

src/impl/elias_fano_stream.cpp
```diff
@@ -0,0 +1,386 @@
...
+    for (uint32_t bit = 0; bit < bits; ++bit) {
+        if ((value & (1U << bit)) != 0) {
+            const uint64_t target_bit = bit_offset + bit;
+            bytes[target_bit / 8] |= static_cast<uint8_t>(1U << (target_bit % 8));
+        }
+    }
+}
+
+uint32_t
+count_trailing_zeros(uint64_t value) {
```
LHT129, line 47: [suggestion] The `store_packed` helper writes bits one at a time in a per-bit loop, which is O(L * bits) rather than O(L). For the packed-ID path this replaces the original `PackedWriter` class that wrote byte-aligned groups. Consider using word-at-a-time writes (e.g. fill a 32-bit or 64-bit buffer and flush when full) to reduce the instruction count on the encode hot path.

LHT129 (commented):
Code review for the Elias-Fano DMQ encoding experiment PR.

src/algorithm/sindi/sindi.cpp
```diff
@@ -1836,7 +1837,15 @@ SINDI::EstimateMemory(uint64_t num_elements) const {
                 std::min<uint64_t>(term_id_limit_, total_sparse_values);
             const uint32_t rerank_id_bits = get_bits_for_term_id_limit(
                 estimated_term_count == 0 ? 0 : static_cast<uint32_t>(estimated_term_count - 1));
-            mem += (total_sparse_values * rerank_id_bits + 7) / 8;
+            if (estimated_term_count != 0 && num_elements != 0) {
```
LHT129, line 1840: [suggestion] When `estimated_term_count == 0 || num_elements == 0`, the new code skips the ID-storage memory estimate entirely. The old code always added `(total_sparse_values * rerank_id_bits + 7) / 8`. If `total_sparse_values` is non-zero but `estimated_term_count` or `num_elements` is zero, the estimate will now be missing the ID storage component. Consider falling back to the old formula when the guard condition is false.

src/quantization/sparse_quantization/sparse_dmq_quantizer.cpp
```diff
@@ -537,34 +635,58 @@ SparseDmqQuantizer::ComputeDistImpl(Computer<SparseDmqQuantizer>& computer,
...
-        uint32_t matched = K_INVALID_INDEX;
-        if (query.has_lookup) {
-            if (id < query.term_to_index.size()) {
-                matched = query.term_to_index[id];
+    const auto* encoded_ids = codes + sizeof(header);
+    const auto id_layout = get_id_encoding_layout(
+        header.len, static_cast<uint32_t>(term_ids_.size()), id_bits_, id_encoding_type_);
+    const auto* value_codes = encoded_ids + id_layout.encoded_bytes;
+    if (id_layout.use_elias_fano) {
+        float query_sum = 0.0F;
```
LHT129, line 643: [note] The EF seek path in `ComputeDistImpl` iterates over `query.ids` and calls `FindEqualRange(query.ids[query_index])`. This relies on `query.ids` containing compact IDs (not raw term IDs). The `PrepareQuery` path populates `query.ids` with compact IDs via `GetCompactId`, so this is correct. However, the packed fallback path still supports `has_lookup=false` with raw term IDs via linear merge, while the EF path does not have an equivalent fallback. If `has_lookup=false` and the vector uses EF encoding, the search will silently use compact IDs from `query.ids` — make sure this is the intended behavior and that `query.ids` is always populated with compact IDs when EF may be used.

src/quantization/sparse_quantization/sparse_dmq_quantizer.cpp
```diff
@@ -382,16 +436,29 @@ SparseDmqQuantizer::EncodeOneImpl(const float* data, uint8_t* codes) const {
...
+    auto* encoded_ids = codes + sizeof(header);
+    const auto id_layout = get_id_encoding_layout(
+        vector.len_, static_cast<uint32_t>(term_ids_.size()), id_bits_, id_encoding_type_);
+    auto* value_codes = encoded_ids + id_layout.encoded_bytes;
+    for (uint32_t index = 0; index < vector.len_; ++index) {
+        ids[index] = GetCompactId(ids[index]);
+    }
+    if (id_layout.use_elias_fano) {
+        EliasFanoStream::Encode(ids.data(),
+                                vector.len_,
```
LHT129, line 448: [note] In `EncodeOneImpl`, the `ids` array is mutated in-place via `ids[index] = GetCompactId(ids[index])` before the encoding loop. The subsequent value-encoding loop then uses `ids[index]` (now a compact ID) to look up the codebook. This is correct but means the caller's `ids` array is modified. The original code called `GetCompactId` inside the loop without mutating the input array. If the caller reuses the same vector for multiple encodes, this mutation could cause issues. Consider using a local copy or restoring the original IDs after encoding.

src/datacell/sparse_dmq_datacell.cpp
```diff
@@ -216,7 +217,11 @@ void
 SparseDmqDataCell::Serialize(StreamWriter& writer) {
     std::shared_lock lock(this->mutex_);
     StreamWriter::WriteObj(writer, K_SPARSE_DMQ_DATACELL_MAGIC);
-    StreamWriter::WriteObj(writer, K_SPARSE_DMQ_DATACELL_VERSION);
+    const uint32_t version =
```
LHT129, line 220: [note] The serialization version is determined by `quantizer_->GetIdEncodingType()` at serialize time. When deserializing, the version is read and `SetIdEncodingType` is called accordingly. This means a model serialized with HYBRID_ELIAS_FANO will always deserialize as HYBRID_ELIAS_FANO (version 7), and a PACKED model as PACKED (version 6). However, if a HYBRID model is deserialized and then re-serialized, the per-vector encoding decisions are re-evaluated by `get_id_encoding_layout`, which could produce different per-vector choices than the original encoding. This is correct behavior but worth documenting: re-serialization may change individual vector layouts even though the encoding type is preserved.

src/impl/elias_fano_stream.h
```diff
@@ -0,0 +1,140 @@
+// Copyright 2024-present the vsag project
```
LHT129, line 1: [note] `EliasFanoSeekReader` has a `has_last_target_` / `last_target_` monotonicity check that throws if targets are not strictly nondecreasing. This is a valid correctness guard, but it also means the reader is single-use per query — you cannot reuse the same `EliasFanoSeekReader` instance for a second query pass. This is fine for the current DMQ rerank path, but worth noting in case the reader is reused elsewhere in the future.

src/impl/elias_fano_stream.cpp
```diff
@@ -0,0 +1,386 @@
...
+    return layout;
+}
+
+void
+EliasFanoStream::Encode(const uint32_t* values, uint32_t count, uint32_t universe, uint8_t* codes) {
+    Encode(values, count, universe, GetLayout(count, universe), codes);
+}
+
+void
+EliasFanoStream::Encode(const uint32_t* values,
```
LHT129, line 120: [note] `GetLayout` computes `low_bits_width = floor_log2(universe / count)`. When `universe / count` is exactly 1, `low_bits_width` is 0. This is correct per the standard Elias-Fano definition. However, when `count == 0` the function returns an all-zero layout early (line 108). Callers that pass `count == 0` and then use the layout for allocation will get 0-byte buffers, which is handled correctly throughout the codebase.

src/quantization/sparse_quantization/sparse_dmq_quantizer.cpp
```diff
@@ -634,6 +756,7 @@ void
 SparseDmqQuantizer::ExportModel(const SparseDmqQuantizer& other) {
```
LHT129, line 756: [suggestion] `ExportModel` now copies `id_encoding_type_`. However, `Serialize` / `Deserialize` do not appear to persist `id_encoding_type_` in the quantizer's own serialization — it is only inferred from the datacell version. If the quantizer is serialized/deserialized independently of the datacell, the encoding type would be lost and default back to PACKED. Consider adding `id_encoding_type_` to the quantizer's own `Serialize`/`Deserialize` methods.

PR #2577, opened 2026-07-30
perf(pyramid): optimize path tree build

### LINKED ISSUE #2576: [improve](other): batch Pyramid NSW builds for high-cardinality paths
### Kind of improvement

perf

### Affected area

other (Pyramid)

### Summary

Pyramid's NSW `Build()` currently populates the path tree and then delegates the
complete dataset to the incremental `Add()` path.

For high-cardinality path data, this does unnecessary per-vector work even when
almost every leaf remains a FLAT subindex:

- each path is split once while populating the tree and again during `Add()`;
- one thread-pool future is submitted per vector;
- every traversal calls `IndexNode::GetChild()`, which takes an exclusive node
  lock even when the child already exists;
- every indexed node is updated through the incremental insertion path.

On the production-shaped workload below, 1,186,526 vectors produce 244,986 leaf
nodes, but only 10 leaves become GRAPH subindexes and only 8,649 vectors
participate in graph construction. Scheduling more than one million incremental
insertions for the other 244,976 FLAT leaves creates substantial allocation,
scheduling, path parsing, and lock contention.

Introduce a dedicated Build-only NSW path:

1. Train and batch-insert labels and vector codes.
2. Split each path once while creating the complete tree and collecting node
   inner IDs.
3. Classify nodes as `NO_INDEX`, `FLAT`, or `GRAPH`.
4. Preserve collected IDs directly for FLAT nodes.
5. Submit work only for GRAPH nodes, splitting disproportionately large graph
   nodes into bounded chunks.
6. Keep incremental `Add()` behavior unchanged.

### Expected benefit

- Eliminate one future allocation and thread-pool submission per vector.
- Eliminate the second path parse and repeated exclusive root traversal.
- Avoid incremental node locking for FLAT leaves.
- Make `build_thread_count` scale positively.
- Reduce transient allocator pressure while preserving the public API,
  serialized format, and incremental `Add()` semantics.

The prototype reduces the 18-thread median build time from 3.613 seconds to
0.553 seconds, a 6.53x speedup. It also removes the previous negative scaling,
where 18 threads were 1.85x slower than one thread.

### Benchmark evidence

Environment:

- Apple M5 Pro, 18 logical CPUs, 24 GiB memory
- macOS 26.5.1, arm64
- three runs per configuration; median wall-clock `Build()` time
- identical deterministic vectors, paths, parameters, and input order

Dataset:

- 1,186,526 vectors, 32 dimensions, FP32, L2
- 244,986 leaf nodes
- 244,976 FLAT leaves
- 10 GRAPH leaves containing 8,649 vectors

Parameters:

```json
{
  "base_quantization_type": "fp32",
  "base_io_type": "block_memory_io",
  "max_degree": 32,
  "ef_construction": 200,
  "alpha": 1.2,
  "graph_type": "nsw",
  "no_build_levels": [0],
  "use_reorder": false,
  "index_min_size": 128
}
```

| Build threads | Existing `Build()` | Batched NSW prototype | Speedup |
| ---: | ---: | ---: | ---: |
| 1 | 1.954 s | 1.333 s | 1.47x |
| 4 | 1.790 s | 0.738 s | 2.43x |
| 8 | 2.832 s | 0.682 s | 4.15x |
| 18 | 3.613 s | 0.553 s | 6.53x |

At 18 threads, transient RSS growth decreases from approximately 923 MiB to
724 MiB. Final index memory is effectively unchanged (approximately 485 MiB
versus 487 MiB).

The benchmark measures build performance only and does not include a recall
corpus. Search correctness is covered by repeated reachability, self-match,
duplicate-vector, incremental Add, and serialization regression tests.

### References

The code references below use `upstream/main` commit
`055ce5965f4e200d84e005b1af4f37afe417dee6`:

- `src/algorithm/pyramid/pyramid.cpp:1030` — NSW `Build()` invokes full-dataset
  `Add()`.
- `src/algorithm/pyramid/pyramid.cpp:1124` — the path tree is initially
  populated.
- `src/algorithm/pyramid/pyramid.cpp:1151` — paths are split again during
  incremental insertion.
- `src/algorithm/pyramid/pyramid.cpp:1173` — one task is created per vector.
- `src/algorithm/pyramid/pyramid.cpp:110` — `GetChild()` takes the node's
  exclusive mutex.

### Additional context

Users have reported builds running for multiple days with low CPU utilization
on this path distribution. The status distribution explains the behavior:
almost all subindexes are FLAT, while the implementation pays incremental task,
traversal, and locking costs for every vector.

This change should remain limited to initial `Build()` on an empty index.
`Add()` should retain its existing concurrent incremental semantics, including
FLAT-to-GRAPH promotion.

## Related issues

_None found._

---

jac0626:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Closes: #2576

## What Changed

- Replace Pyramid NSW `Build()`'s full-dataset `Add(base)` fallback with a
  dedicated initial-build path.
- Batch-insert labels and vector codes while preserving compact inner IDs for
  rejected labels.
- Scan each valid path once on the Build thread, create the complete path tree,
  and collect node inner IDs without allocating a `vector<string>` per path.
- Use a Build-only child lookup/create path without per-node locking; Build has
  exclusive ownership, and parallel ODescent hierarchy tasks write distinct
  roots.
- Classify nodes as `NO_INDEX`, `FLAT`, or `GRAPH`; keep FLAT node IDs directly.
- After tree construction finishes, submit one task per GRAPH node; different
  nodes build in parallel while each node reuses `add_one_point()` serially.
- Keep incremental `Add()` and graph construction behavior unchanged.

## Why

The previous NSW Build populated the path tree and then called `Add(base)`.
For path-heavy data this parsed and traversed every path a second time and
submitted one task per input vector, even though almost all leaves remained
FLAT.

On the deterministic synthetic workload matching the reported index structure,
1,186,526 vectors create 244,986 leaves, but only 10 GRAPH leaves contain 8,649
graph vectors. The new flow avoids scheduling the other 1,177,877 FLAT vectors
through the incremental Add path.

## Test Evidence

- [x] `clang-format` 15 dry run
- [x] C++ unit-test target builds
- [x] Focused Pyramid unit tests
- [ ] Full `make test` / CI

```text
cmake --build build --target unittests -j8
  passed

./build/tests/unittests '[ut][pyramid]'
  All tests passed (95 assertions in 7 test cases)
```

The focused regression data contains 1,097 vectors, 363 FLAT leaves, and two
GRAPH leaves with eight graph vectors (0.7293%), matching the reported graph
ratio. It also verifies FLAT-to-GRAPH promotion after Build, duplicate vectors,
compact IDs after rejected labels, the true-empty-index Build guard, and path
normalization for empty/leading/trailing/repeated slash segments.

The deterministic synthetic benchmark uses identical data
(`dataset_fingerprint=0xc4be1640866718e`), 18 build threads, and five runs:

| Version | Mean Build | Median Build |
| --- | ---: | ---: |
| upstream (`055ce596`) | 3.636 s | 3.628 s |
| this PR (`8de33961`) | 0.484 s | 0.484 s |

The final implementation is 7.51x faster (86.7% less wall time) than upstream.
A same-session three-run baseline before the direct path scanner and Build-only
child lookup averaged 0.652 s; these two small hot-path changes reduce that by
another 25.7%. Both versions produce the same 244,986 leaves, 10 GRAPH nodes,
8,649 graph vectors, and final status distribution. A final stats-enabled run
completed in 0.479 s with the same dataset fingerprint and status counts.

## Compatibility Impact

- API/ABI compatibility: no public API changes
- Serialization compatibility: unchanged
- Graph construction semantics: unchanged; each GRAPH node reuses the existing
  `add_one_point()` path in input order

- Eliminates the second full path parse and traversal during NSW Build.
- Avoids incremental work and task submission for FLAT nodes.
- Removes temporary split containers and per-node locks from exclusive initial
  path-tree construction.
- Builds the complete path tree in one thread, then parallelizes independent
  GRAPH nodes without adding node-internal graph construction machinery.
- Does not introduce a new graph storage mode or connectivity-repair pass.

## Documentation Impact

- [x] No docs update needed

No parameter or user-facing API changes.

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert the dedicated NSW Build path to restore
  `Build() -> Add(base)`.

## Checklist

- [x] I have added/updated tests
- [x] I have considered API and serialization compatibility
- [x] No documentation update is required

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -327,34 +327,11 @@ Pyramid::build_by_nsw(const DatasetPtr& base) {
     Vector<int64_t> data_biases(allocator_);
     auto failed_ids = prepare_build_data(base, data_biases);
 
-    if (thread_pool_ != nullptr && hierarchies_.size() > 1) {
-        Vector<std::future<void>> futures(allocator_);
-        futures.reserve(hierarchies_.size());
-        std::exception_ptr enqueue_exception;
-        try {
-            for (auto& [hname, hierarchy] : hierarchies_) {
-                const auto* paths = base->GetPaths(hname);
-                if (paths != nullptr) {
-                    futures.push_back(thread_pool_->GeneralEnqueue(
-                        [&hierarchy = *hierarchy, paths, &data_biases]() {
-                            populate_path_tree(hierarchy,
-                                               paths,
-                                               static_cast<int64_t>(data_biases.size()),
-                                               &data_biases);
-                        }));
-                }
-            }
-        } catch (...) {
-            enqueue_exception = std::current_exception();
-        }
-        wait_all_futures(futures, enqueue_exception);
-    } else {
-        for (auto& [hname, hierarchy] : hierarchies_) {
-            const auto* paths = base->GetPaths(hname);
-            if (paths != nullptr) {
-                populate_path_tree(
-                    *hierarchy, paths, static_cast<int64_t>(data_biases.size()), &data_biases);
-            }
+    for (auto& [hname, hierarchy] : hierarchies_) {
+        const auto* paths = base->GetPaths(hname);
+        if (paths != nullptr) {
+            populate_path_tree(
+                *hierarchy, paths, static_cast<int64_t>(data_biases.size()), &data_biases);
         }
     }
 
@@ -363,7 +340,8 @@ Pyramid::build_by_nsw(const DatasetPtr& base) {
         (void)hname;
         collect_nsw_build_jobs(*hierarchy, hierarchy->root.get(), build_jobs);
     }
-    const bool use_parallel_build = thread_pool_ != nullptr && build_thread_count_ > 1;
+    const bool use_parallel_build =
+        thread_pool_ != nullptr && build_thread_count_ > 1 && build_jobs.size() > 1;
     if (not use_parallel_build) {
         for (const auto& job : build_jobs) {
             build_nsw_graph(job, data_vectors, data_biases);
@@ -373,14 +351,10 @@ Pyramid::build_by_nsw(const DatasetPtr& base) {
         std::exception_ptr enqueue_exception;
         try {
             for (const auto& job : build_jobs) {
-                for (const auto inner_id : job.node->ids_) {
-                    const auto data_bias = data_biases[inner_id];
-                    const auto* vector = data_vectors + dim_ * data_bias;
-                    futures.push_back(thread_pool_->GeneralEnqueue(
-                        [this, hierarchy = job.hierarchy, node = job.node, inner_id, vector]() {
-                            add_one_point(*hierarchy, node, inner_id, vector);
-                        }));
-                }
+                futures.push_back(
+                    thread_pool_->GeneralEnqueue([this, job, data_vectors, &data_biases]() {
+                        build_nsw_graph(job, data_vectors, data_biases);
+                    }));
             }
         } catch (...) {
             enqueue_exception = std::current_exception();
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -138,6 +138,17 @@ IndexNode::GetChild(const std::string& key, bool need_init) {
     return children_[key].get();
 }
 
+IndexNode*
+IndexNode::GetOrCreateChildForBuild(const std::string& key) {
+    auto [result, inserted] = children_.try_emplace(key);
+    auto& child = result.value();
+    if (inserted) {
+        child = std::make_unique<IndexNode>(allocator_, graph_param_, index_min_size_);
+        child->level_ = level_ + 1;
+    }
+    return child.get();
+}
+
 void
 IndexNode::Deserialize(StreamReader& reader) {
     // deserialize `entry_point_`
@@ -1279,19 +1290,34 @@ Pyramid::populate_path_tree(Hierarchy& h,
                             const Vector<int64_t>* data_biases) {
     for (int64_t inner_id = 0; inner_id < count; ++inner_id) {
         const int64_t data_bias = data_biases == nullptr ? inner_id : (*data_biases)[inner_id];
-        std::string current_path = paths[data_bias];
-        auto path_slices = split(current_path, PART_SLASH);
+        const auto& current_path = paths[data_bias];
         IndexNode* node = h.root.get();
         if (std::find(h.no_build_levels.begin(), h.no_build_levels.end(), node->level_) ==
             h.no_build_levels.end()) {
             node->ids_.push_back(static_cast<InnerIdType>(inner_id));
         }
-        for (auto& path_slice : path_slices) {
-            node = node->GetChild(path_slice, true);
-            if (std::find(h.no_build_levels.begin(), h.no_build_levels.end(), node->level_) ==
-                h.no_build_levels.end()) {
-                node->ids_.push_back(static_cast<InnerIdType>(inner_id));
+
+        uint64_t slice_begin = 0;
+        while (slice_begin < current_path.size()) {
+            const auto delimiter_pos = current_path.find(PART_SLASH, slice_begin);
+            const uint64_t slice_end =
+                delimiter_pos == std::string::npos ? current_path.size() : delimiter_pos;
+            if (slice_end > slice_begin) {
+                if (slice_begin == 0 && slice_end == current_path.size()) {
+                    node = node->GetOrCreateChildForBuild(current_path);
+                } else {
+                    node = node->GetOrCreateChildForBuild(
+                        current_path.substr(slice_begin, slice_end - slice_begin));
+                }
+                if (std::find(h.no_build_levels.begin(), h.no_build_levels.end(), node->level_) ==
+                    h.no_build_levels.end()) {
+                    node->ids_.push_back(static_cast<InnerIdType>(inner_id));
+                }
             }
+            if (delimiter_pos == std::string::npos) {
+                break;
+            }
+            slice_begin = slice_end + 1;
         }
     }
 }
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -15,15 +15,12 @@
 
 #include "pyramid.h"
 
-#include <atomic>
 #include <chrono>
 #include <exception>
-#include <limits>
 
 #include "algorithm/inner_index_interface.h"
 #include "analyzer/analyzer.h"
 #include "datacell/flatten_interface.h"
-#include "datacell/sparse_graph_datacell.h"
 #include "impl/heap/standard_heap.h"
 #include "impl/odescent/odescent_graph_builder.h"
 #include "impl/pruning_strategy.h"
@@ -39,77 +36,9 @@ namespace vsag {
 
 const static float RADIUS_EPSILON = 1.1F;
 constexpr uint64_t BUILD_CODES_BATCH_BYTES = uint64_t{32} * 1024 * 1024;
-constexpr uint64_t NSW_BUILD_TASKS_PER_WORKER = 4;
-constexpr uint64_t NSW_BUILD_MIN_CHUNK_IDS = 64;
-constexpr uint64_t NSW_BUILD_WARMUP_MIN_IDS = 64;
-constexpr uint64_t NSW_BUILD_WARMUP_MAX_IDS = 256;
-constexpr uint64_t NSW_BUILD_WARMUP_DEGREE_FACTOR = 2;
-
-class FrozenSparseGraphView final : public GraphInterface {
-public:
-    explicit FrozenSparseGraphView(std::shared_ptr<SparseGraphDataCell> graph)
-        : graph_(std::move(graph)) {
-    }
-
-    void
-    InsertNeighborsById(InnerIdType, const Vector<InnerIdType>&) override {
-        throw VsagException(ErrorType::INTERNAL_ERROR,
-                            "cannot insert graph keys through a frozen build view");
-    }
-
-    uint32_t
-    GetNeighborSize(InnerIdType) const override {
-        throw VsagException(ErrorType::INTERNAL_ERROR,
-                            "cannot read an unlocked size through a frozen build view");
-    }
-
-    void
-    GetNeighbors(InnerIdType id, Vector<InnerIdType>& neighbor_ids) const override {
-        graph_->GetNeighborsForBuild(id, neighbor_ids);
-    }
-
-    bool
-    CheckIdExists(InnerIdType id) const override {
-        return graph_->CheckIdExists(id);
-    }
-
-    void
-    Resize(InnerIdType) override {
-        throw VsagException(ErrorType::INTERNAL_ERROR,
-                            "cannot resize a graph through a frozen build view");
-    }
-
-    void
-    Prefetch(InnerIdType id, uint32_t neighbor_i) override {
-        graph_->Prefetch(id, neighbor_i);
-    }
-
-    InnerIdType
-    TotalCount() const override {
-        return graph_->TotalCount();
-    }
-
-    InnerIdType
-    MaximumDegree() const override {
-        return graph_->MaximumDegree();
-    }
-
-    InnerIdType
-    MaxCapacity() const override {
-        return graph_->MaxCapacity();
-    }
-
-private:
-    std::shared_ptr<SparseGraphDataCell> graph_;
-};
-
-static uint64_t
-ceil_div(uint64_t value, uint64_t divisor) {
-    return value == 0 ? 0 : 1 + (value - 1) / divisor;
-}
 
 static void
-wait_all_futures(Vector<std::future<void>>& futures, std::exception_ptr first_exception = nullptr) {
+wait_all_futures(Vector<std::future<void>>& futures, std::exception_ptr& first_exception) {
     for (auto& future : futures) {
         try {
             future.get();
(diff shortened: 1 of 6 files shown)


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -34,6 +38,91 @@
...
+    InnerIdType
+    TotalCount() const override {
+        return graph_->TotalCount();
+    }
+
+    InnerIdType
+    MaximumDegree() const override {
+        return graph_->MaximumDegree();
+    }
+
```
LHT129: [critical] `first_exception` is passed **by value**, not by reference. The assignment `first_exception = std::current_exception()` on line 102 modifies the local copy, so the caller's `enqueue_exception` is never updated. This means exceptions thrown during `GeneralEnqueue` are silently lost.

In `build_by_nsw`, both call sites pass a local `enqueue_exception`:
```cpp
std::exception_ptr enqueue_exception;
try {
    // ... GeneralEnqueue calls ...
} catch (...) {
    enqueue_exception = std::current_exception();
}
wait_all_futures(futures, enqueue_exception);
```

The intent is clearly to propagate the enqueue exception through `wait_all_futures`, but it never reaches the function body.

Suggested fix: change the parameter to a reference:
```cpp
static void
wait_all_futures(Vector<std::future<void>>& futures, std::exception_ptr& first_exception) {
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +304,652 @@ IndexNode::Search(const SearchFunc& search_func,
...
+        candidates->Pop();
+        if (candidate.second != inner_id) {
+            filtered_candidates->Push(candidate.first, candidate.second);
+        }
+    }
+    if (filtered_candidates->Empty()) {
+        throw VsagException(ErrorType::INTERNAL_ERROR,
+                            "no graph candidates found during Pyramid concurrent build");
+    }
+    select_edges_by_heuristic(
```
LHT129: [suggestion] The `find_replacement` lambda initializes `farthest_distance = -1.0F` and then checks `not found || distance > farthest_distance`. If all neighbor distances happen to be exactly `-1.0F` (unlikely with L2 but theoretically possible with custom distance functions), no replacement would be found even though valid candidates exist.

Consider initializing to `-std::numeric_limits<float>::infinity()` or `std::numeric_limits<float>::lowest()` to guarantee any real distance value will be accepted on the first iteration:

```cpp
float farthest_distance = -std::numeric_limits<float>::infinity();
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +304,652 @@ IndexNode::Search(const SearchFunc& search_func,
...
+        Vector<InnerIdType> pruned_neighbors(allocator_);
+        pruned_neighbors.reserve(neighbor_candidates->Size());
+        while (not neighbor_candidates->Empty()) {
+            pruned_neighbors.emplace_back(neighbor_candidates->Top().second);
+            neighbor_candidates->Pop();
+        }
+        job.build_graph->UpdateNeighborsForBuild(selected_neighbor, pruned_neighbors);
+    }
+}
+
```
LHT129: [note] When the KNN search fails to find a reachable candidate with a replaceable edge slot, the fallback at lines 523-537 performs a linear scan over all node IDs (`for (const auto candidate : ids)`). For a large graph node with many vectors, this O(N) scan could become expensive. This is a last-resort path that should be rarely exercised, but it may be worth adding a `logger::warn` here to make unexpected fallback invocations visible during operation.

LHT129 (commented):
Overall this is a well-structured PR that delivers substantial build-time speedups (up to 6.5x at 18 threads) with thorough test coverage. The architecture — frozen graph views, per-point locking, two-level scheduling, and connectivity repair — is sound and well-documented.

**Key findings:**

1. **[critical] `wait_all_futures` exception propagation bug**: `first_exception` is passed by value, so exceptions from `GeneralEnqueue` are silently lost. This should be fixed before merge.

2. **[suggestion] `find_replacement` initialization**: Using `-1.0F` as the sentinel for `farthest_distance` is fragile against edge-case distance values. Consider `-std::numeric_limits<float>::infinity()`.

3. **[note] Connectivity repair fallback**: The linear scan fallback in `repair_nsw_graph_connectivity` could benefit from a warning log to surface unexpected invocations.

The test coverage is excellent — covering single large graphs, skewed multi-node scheduling, concurrent parent/child builds, low-degree connectivity repair, duplicate handling, compact IDs after rejected labels, and serialization round-trips. The 16-iteration reachability stress test with reverse-edge consistency checks is particularly valuable.

No API, serialization, or public interface changes. Incremental `Add()` and ODescent Build paths are preserved.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +233,143 @@ IndexNode::Search(const SearchFunc& search_func,
...
+        }
+        run_begin = run_end;
+    }
+}
+
+void
+Pyramid::collect_nsw_build_jobs(Hierarchy& hierarchy,
+                                IndexNode* node,
+                                Vector<NswBuildJob>& build_jobs) {
+    const bool skip_index =
```
LHT129, line 305: [suggestion] `collect_nsw_build_jobs` performs an O(n) `std::find` on `hierarchy.no_build_levels` for every tree node. For deep trees this is O(nodes × no_build_levels). Since `no_build_levels` is typically small (often empty or 1-2 entries), this is unlikely to matter in practice, but if the level count ever grows, consider converting `no_build_levels` to an `UnorderedSet<int32_t>` for O(1) lookup.

Current:
```cpp
const bool skip_index =
    std::find(hierarchy.no_build_levels.begin(),
              hierarchy.no_build_levels.end(),
              static_cast<int32_t>(node->level_)) != hierarchy.no_build_levels.end();
```

LHT129 (commented):
This revision is a significant simplification over the earlier approach — removing the custom concurrent graph storage, adjacency publication, and connectivity repair in favor of reusing the existing `add_one_point()` path with node-level parallelism. The result is much cleaner and easier to reason about.

**What changed since the previous review round:**
- Removed `FrozenSparseGraphView`, `build_nsw_graph_range`, `build_nsw_graph_point`, `publish_nsw_graph_point`, `plan_nsw_build_chunks`, `repair_nsw_graph_connectivity`, and associated constants.
- `build_nsw_graph` now simply iterates `job.node->ids_` and calls `add_one_point()` for each — reusing the exact same insertion path as incremental `Add()`.
- Parallelism is at the GRAPH-node level: one task per GRAPH node, serial within each node.
- The `wait_all_futures` exception-propagation bug (pass-by-value) from the earlier version is fixed — `first_exception` is now a reference.

**Architecture assessment:**
The design is sound. `add_one_point()` acquires `node->mutex_` per-node, so parallel builds across different nodes have no lock contention. Shared state (`pool_`, `entry_point_mutex_`, `points_mutex_`) is already properly synchronized within `add_one_point()`. The `collect_nsw_build_jobs` pass runs on the calling thread before any parallel work starts, so there is no race on node status or ID vectors.

**Test coverage:**
Excellent. Four new test cases cover: path-heavy data with few graph nodes (both serial and 4-thread variants via `GENERATE`), duplicate handling, compact inner IDs after rejected labels, and the true-empty-index Build guard. The 363-FLAT + 2-GRAPH leaf test mirrors the production-shaped workload described in the PR.

**One minor suggestion posted inline** about `std::find` on `no_build_levels` in `collect_nsw_build_jobs` — not blocking, just a future-proofing note.

No API, serialization, or public interface changes. Incremental `Add()` and ODescent Build paths are preserved. The 5.36x build speedup on the production-shaped benchmark is compelling.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1121,21 +1284,40 @@ Pyramid::add_one_point(const Hierarchy& h,
...
+                            int64_t count,
+                            const Vector<int64_t>* data_biases) {
+    for (int64_t inner_id = 0; inner_id < count; ++inner_id) {
+        const int64_t data_bias = data_biases == nullptr ? inner_id : (*data_biases)[inner_id];
+        const auto& current_path = paths[data_bias];
         IndexNode* node = h.root.get();
         if (std::find(h.no_build_levels.begin(), h.no_build_levels.end(), node->level_) ==
             h.no_build_levels.end()) {
-            node->ids_.push_back(i);
+            node->ids_.push_back(static_cast<InnerIdType>(inner_id));
```
LHT129, line 1297: [suggestion] `populate_path_tree` performs O(n) `std::find` on `h.no_build_levels` twice per path segment. Same pattern as in `collect_nsw_build_jobs`. Consider `UnorderedSet<int32_t>` for O(1) lookup if level count grows.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +244,143 @@ IndexNode::Search(const SearchFunc& search_func,
...
+std::vector<int64_t>
+Pyramid::build_by_nsw(const DatasetPtr& base) {
+    const auto* data_vectors = base->GetFloat32Vectors();
+    Vector<int64_t> data_biases(allocator_);
+    auto failed_ids = prepare_build_data(base, data_biases);
+
+    for (auto& [hname, hierarchy] : hierarchies_) {
+        const auto* paths = base->GetPaths(hname);
+        if (paths != nullptr) {
+            populate_path_tree(
```
LHT129, line 344: [note] `static_cast<int64_t>(data_biases.size())` converts `uint64_t` to `int64_t`. If `data_biases.size()` exceeds `INT64_MAX` (unlikely in practice since it represents vector count), this would truncate. Consider using `static_cast<int64_t>(data_biases.size())` is fine for now given realistic dataset sizes, but worth noting for future-proofing.

LHT129 (commented):
## Code Review Summary

### Overview
This PR separates NSW tree construction from index construction in Pyramid, achieving a **7.51x build speedup**. The approach is clean: it introduces a dedicated `build_by_nsw` path that batch-inserts codes, populates the path tree with a lock-free `GetOrCreateChildForBuild`, collects GRAPH nodes via `collect_nsw_build_jobs`, and optionally parallelizes graph construction across nodes.

### What was checked
- All 3 changed files: `pyramid.cpp` (+200/-18), `pyramid.h` (+37/-1), `pyramid_test.cpp` (+208/-1)
- Existing review comments from earlier commits (all resolved in current code)
- Correctness of `wait_all_futures` (exception_ptr now passed by reference — fixed from earlier version)
- Thread safety of `GetOrCreateChildForBuild` (lock-free `try_emplace`, safe in serial `populate_path_tree` context)
- `batch_insert_build_codes` contiguous run detection logic
- `Build` function routing (NSW → `build_by_nsw`, ODescent → `build_by_odescent`)
- Test coverage (5 new test cases covering: empty path segments, path-heavy data, duplicate handling, rejected labels, empty-after-remove rejection)

### Inline comments posted
- [suggestion] `std::find` O(n) lookup in `populate_path_tree` (same pattern as already noted in `collect_nsw_build_jobs`)
- [note] `static_cast<int64_t>(data_biases.size())` potential truncation (theoretical, not practical)

### Overall assessment
The implementation is solid. The separation of NSW and ODescent build paths is well-structured, the batch insertion with contiguous run detection is efficient, and the optional parallel graph construction is a pragmatic optimization. Test coverage is thorough with edge cases covered.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +244,143 @@ IndexNode::Search(const SearchFunc& search_func,
...
+            failed_ids.push_back(data_ids[i]);
+            continue;
+        }
+        const auto inner_id = static_cast<InnerIdType>(data_biases.size());
+        label_table_->Insert(inner_id, data_ids[i]);
+        data_biases.push_back(i);
+    }
+
+    batch_insert_build_codes(base_codes_, data_vectors, data_biases);
+    if (use_reorder_) {
```
LHT129, line 268: [suggestion] `populate_path_tree` performs `std::find` on `h.no_build_levels` for every path segment of every vector. Since `no_build_levels` is a `Vector<int32_t>` and typically small, this is not a correctness issue, but the repeated linear scan adds up for path-heavy datasets with deep hierarchies. Consider precomputing a lookup set from `h.no_build_levels` once before the outer loop.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -215,6 +244,143 @@ IndexNode::Search(const SearchFunc& search_func,
...
+        std::exception_ptr enqueue_exception;
+        try {
+            for (const auto& job : build_jobs) {
+                futures.push_back(
+                    thread_pool_->GeneralEnqueue([this, job, data_vectors, &data_biases]() {
+                        build_nsw_graph(job, data_vectors, data_biases);
+                    }));
+            }
+        } catch (...) {
+            enqueue_exception = std::current_exception();
```
LHT129, line 371: [note] In `build_by_nsw`, the parallel-build lambda captures `&data_biases` by reference: `[this, job, data_vectors, &data_biases]`. This is safe because `data_biases` is a local variable that outlives all futures (they are joined before the function returns). However, this pattern is fragile — if someone later moves the `wait_all_futures` call or restructures the function so that `data_biases` goes out of scope before futures complete, it would become a dangling reference. Consider capturing a `std::shared_ptr` to the vector or copying the needed data into each job's closure to make the lifetime contract explicit and robust against future refactoring.

jac0626:
Closing this PR after reproducing the original workload with the production-like parameters.

The dominant bottleneck is not Path Tree contention. On upstream `055ce596`, with 1,186,526 × 4096D vectors, `base_quantization_type=rabitq`, `rabitq_bits_per_dim_base=1`, and 32 build threads, all 32 workers remain idle while the main thread runs:

`Pyramid::Build -> Pyramid::Add -> FlattenDataCell::InsertVector -> RaBitQuantizer::EncodeOneInternal -> RandomOrthogonalMatrix::Transform -> cblas_sgemv`

Because `rabitq_use_fht` is not set, it defaults to false and performs a dense 4096×4096 ROM transform for every vector, serially. The measured rate was about 192 vectors/s, extrapolating to roughly 1.72 hours for vector encoding alone on a 48-core development host.

This PR optimizes Path Tree/NSW build orchestration but does not address that dominant O(N·D²) encoding path, so it is aimed at the wrong bottleneck. A follow-up should focus on RaBitQ build-time encoding parallelism/batching, or explicitly evaluate `rabitq_use_fht=true` when configuration changes are acceptable.

PR #2579, opened 2026-07-30
perf(hgraph): add reusable max_degree graph reduction

### LINKED ISSUE #2578: [improve](hgraph): derive compact lower-degree graphs from one build
### Kind of improvement

perf

### Affected area

hgraph

### Summary

HGraph should provide a small internal primitive that can derive lower-`max_degree` physical
graphs from one already built higher-degree graph:

1. Build the largest degree normally.
2. Rank each existing adjacency list once with the current diversity heuristic.
3. Materialize lower degrees in descending order as ordinary compact graphs.
4. Serialize, load, and query each reduced graph through the unchanged HGraph paths.

This issue covers only that HGraph capability. AutoTune and other consumers are deliberately out
of scope and do not need to be dependencies of the implementation.

### Expected benefit

For three degree candidates, this can replace three full graph builds with one full build and
inexpensive graph reductions.

A pure in-memory SIFT benchmark with 500,000 vectors reduced construction time from 23.70 seconds
to 9.61 seconds:

- Independent M16/M32/M64 builds: 23.70s
- M64 build plus M32/M16 reductions: 9.61s
- Conservative speedup: 2.47x
- Time reduction: 59.4%

Normal HGraph build, search, serialization format, and public APIs must remain unchanged.

### Benchmark evidence

Configuration:

- Dataset: first 500,000 vectors from SIFT1M
- Dimension: 128
- Quantization: fp32
- Graph type: NSW
- `ef_construction`: 100
- Build threads: 48
- Degrees: 16, 32, 64
- Measurement excludes serialization

Representative conservative run:

| Operation | Time |
| --- | ---: |
| Independent M16 build | 6.80s |
| Independent M32 build | 8.08s |
| Independent M64 build | 8.82s |
| Independent total | 23.70s |
| Reused M64 build | 8.97s |
| Rank and reduce M32/M16 | 0.64s |
| Reused total | 9.61s |

Three repeated speedups were 2.47x, 3.01x, and 3.45x. The conservative 2.47x result is used
because native M64 build time varied between runs.

A separate SIFT 20K comparison evaluated reduced and independently built graphs at `ef_search`
20, 40, and 80:

- Median recall delta: -0.0002 to +0.0112
- Maximum median latency increase: 6.9%
- Same M64 artifact on the normal search path: identical recall and no measurable latency
  regression

### Scope and constraints

- Keep the normal build and query paths untouched.
- Produce a physical compact graph rather than masking neighbors at query time.
- Reuse the existing diversity heuristic instead of introducing a second edge-selection rule.
- Reject unsupported graph layouts and invalid degree changes.
- Add serialization/deserialization and query regression coverage.

### References

- `src/algorithm/hgraph/hgraph_build.cpp`
- `src/impl/pruning_strategy.cpp`
- `src/algorithm/hgraph/hgraph.h`

### Additional context

A reduced graph is an independently serializable artifact with the requested degree, but its
topology and hierarchy originate from the larger source graph. It is therefore not identical to
an independently built graph with the same `max_degree`; the reduced artifact itself is the
authoritative object for subsequent evaluation.

---

jac0626:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Closes: #2578

## What Changed

- Add an internal HGraph capability to derive a lower-`max_degree` graph from an already built
  higher-degree graph.
- Rank existing adjacency lists once with the existing diversity heuristic, then materialize the
  selected prefixes into ordinary compact graph storage.
- Support successive reductions, serialization, deserialization, and search through the normal
  HGraph paths.
- Restrict the capability to compatible mutable in-memory NSW configurations and reject invalid
  or unsupported reductions.
- Add focused tests for degree reduction, repeated reduction, invalid input, serialization,
  deserialization, and search.

This PR intentionally provides only the HGraph primitive. AutoTune or any other consumer can call
it later, but no AutoTune dependency or orchestration is included here.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build-release --target algorithm_test -j$(nproc)
  passed

cmake --build build-release --target vsag -j$(nproc)
  passed

/tmp/hgraph_degree_reduction_test
  All tests passed (22 assertions in 1 test case)

  passed

clang-tidy-15 -p build-release \
  src/algorithm/hgraph/hgraph_degree_reduction.cpp --quiet
  passed

git diff --check upstream/main...HEAD
  passed
```

Performance validation used the first 500,000 vectors from SIFT1M, fp32, NSW,
`ef_construction=100`, 48 build threads, and degrees 16/32/64. Serialization was excluded:

```text
Independent M16/M32/M64 builds: 23.70s
M64 build + M32/M16 projection:  9.61s
Conservative construction speedup: 2.47x
Construction time reduction:       59.4%
```

A separate five-repeat SIFT 20K comparison found median recall deltas from -0.0002 to +0.0112
and a maximum median projected-versus-independent latency increase of 6.9%. Searching the same
M64 artifact before and after this change produced identical recall and no measurable normal-path
latency regression.

## Compatibility Impact

- API/ABI compatibility: No installed public API or ABI change. The new methods are internal to
  HGraph.
- Behavior changes: None unless an internal caller explicitly invokes degree reduction.
- Serialization: No format change; reduced graphs use the existing physical graph format.

- Performance impact: Consumers evaluating multiple `max_degree` values can replace repeated full
  graph construction with one construction plus inexpensive reductions. Normal build and search
  paths are unchanged.
- Concurrency/thread-safety impact: Preparation and reduction take the existing HGraph add/global
  locks.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs

This is an internal capability with no user-facing workflow in this PR.

## Risk and Rollback

- Risk level: medium
- Rollback plan: Revert commit `9df7758a`; normal HGraph paths and the serialization format are
  otherwise unchanged.

The reduced graph has the requested compact degree but inherits topology and hierarchy from the
larger graph. It is not byte-for-byte or topology-equivalent to a graph independently built with
the smaller `max_degree`, so consumers must evaluate the reduced artifact itself.

## Checklist

### DIFF
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -227,6 +227,17 @@ public:
     void
     SetImmutable() override;
 
+    // Internal AutoTune path. These methods rewrite graph storage between trials; normal build,
+    // search, Tune, and serialization entry points never call them.
+    [[nodiscard]] bool
+    CanReduceMaxDegree() const;
+
+    void
+    PrepareDegreeReduction();
+
+    void
+    ReduceMaxDegree(uint32_t max_degree);
+
     void
     ExportCache(std::ostream& out_stream) const override;
 
@@ -652,6 +663,12 @@ private:
     void
     cal_memory_usage();
 
+    [[nodiscard]] bool
+    can_reduce_max_degree_unlocked() const;
+
+    [[nodiscard]] FlattenInterfacePtr
+    get_degree_reduction_codes() const;
+
     /// True when reorder uses a separate high-precision flatten (not base codes).
     [[nodiscard]] bool
     has_precise_reorder() const {
@@ -845,5 +862,7 @@ private:
     float build_cache_hit_rate_{-1.0F};     // cache hit rate from last cache-based build
     uint64_t build_cache_hit_nodes_{0};     // number of nodes with cache hit
     uint64_t build_cache_missed_nodes_{0};  // number of nodes without cache hit
+
+    bool degree_reduction_prepared_{false};
 };
 }  // namespace vsag
--- a/src/algorithm/hgraph/hgraph_degree_reduction.cpp
+++ b/src/algorithm/hgraph/hgraph_degree_reduction.cpp
@@ -1,0 +1,280 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <algorithm>
+#include <exception>
+#include <future>
+#include <memory>
+#include <vector>
+
+#include "datacell/graph_datacell_parameter.h"
+#include "datacell/graph_interface.h"
+#include "datacell/sparse_graph_datacell_parameter.h"
+#include "hgraph.h"
+#include "impl/pruning_strategy.h"
+#include "impl/thread_pool/safe_thread_pool.h"
+
+namespace vsag {
+
+namespace {
+
+template <typename Function>
+void
+parallel_for(uint64_t count,
+             uint64_t parallelism,
+             const SafeThreadPoolPtr& thread_pool,
+             const Function& function) {
+    const auto worker_count = std::min(count, std::max<uint64_t>(1, parallelism));
+    if (worker_count <= 1 || thread_pool == nullptr) {
+        for (uint64_t i = 0; i < count; ++i) {
+            function(i);
+        }
+        return;
+    }
+
+    const auto batch_size = (count + worker_count - 1) / worker_count;
+    std::vector<std::future<void>> futures;
+    futures.reserve(worker_count);
+    std::exception_ptr first_exception;
+    for (uint64_t worker = 0; worker < worker_count; ++worker) {
+        const auto begin = worker * batch_size;
+        const auto end = std::min(count, begin + batch_size);
+        if (begin == end) {
+            break;
+        }
+        try {
+            futures.emplace_back(thread_pool->GeneralEnqueue([begin, end, &function]() {
+                for (uint64_t i = begin; i < end; ++i) {
+                    function(i);
+                }
+            }));
+        } catch (...) {
+            first_exception = std::current_exception();
+            break;
+        }
+    }
+    for (auto& future : futures) {
+        try {
+            future.get();
+        } catch (...) {
+            if (first_exception == nullptr) {
+                first_exception = std::current_exception();
+            }
+        }
+    }
+    if (first_exception != nullptr) {
+        std::rethrow_exception(first_exception);
+    }
+}
+
+GraphInterfacePtr
+materialize_graph(const GraphInterfacePtr& source,
+                  const GraphInterfaceParamPtr& target_param,
+                  const FlattenInterfacePtr& flatten,
+                  Allocator* allocator,
+                  bool dense_ids,
+                  const SafeThreadPoolPtr& thread_pool,
+                  uint64_t parallelism) {
+    auto common_param = flatten->ExportCommonParam();
+    auto target = GraphInterface::MakeInstance(target_param, common_param);
+    CHECK_ARGUMENT(target != nullptr, "failed to create compact graph storage");
+    if (dense_ids) {
+        target->Resize(source->MaxCapacity());
+    }
+
+    const auto copy_neighbors = [&](InnerIdType id) {
+        Vector<InnerIdType> neighbors(allocator);
+        source->GetNeighbors(id, neighbors);
+        const auto target_degree = static_cast<uint64_t>(target_param->max_degree_);
+        if (neighbors.size() > target_degree) {
+            neighbors.resize(target_degree);
+        }
+        target->InsertNeighborsById(id, neighbors);
+    };
+
+    if (dense_ids) {
+        parallel_for(source->TotalCount(), parallelism, thread_pool, [&](uint64_t id) {
+            copy_neighbors(static_cast<InnerIdType>(id));
+        });
+    } else {
+        const auto ids = source->GetIds();
+        parallel_for(
+            ids.size(), parallelism, thread_pool, [&](uint64_t i) { copy_neighbors(ids[i]); });
+    }
+    target->SetDuplicateTracker(source->GetDuplicateTracker());
+    return target;
+}
+
+void
+rank_graph(const GraphInterfacePtr& graph,
+           const FlattenInterfacePtr& flatten,
+           Allocator* allocator,
+           float alpha,
+           bool dense_ids,
+           const SafeThreadPoolPtr& thread_pool,
+           uint64_t parallelism) {
+    const auto rank_neighbors = [&](InnerIdType id) {
+        Vector<InnerIdType> neighbors(allocator);
+        graph->GetNeighbors(id, neighbors);
+        if (!neighbors.empty()) {
+            const auto original = neighbors;
+            select_edges_by_heuristic(neighbors, id, neighbors.size(), flatten, allocator, alpha);
+            std::reverse(neighbors.begin(), neighbors.end());
+
+            std::vector<std::pair<float, InnerIdType>> remaining;
+            remaining.reserve(original.size() - neighbors.size());
+            for (const auto neighbor : original) {
+                if (std::find(neighbors.begin(), neighbors.end(), neighbor) == neighbors.end()) {
+                    remaining.emplace_back(flatten->ComputePairVectors(id, neighbor), neighbor);
+                }
+            }
+            std::sort(remaining.begin(), remaining.end());
+            neighbors.reserve(original.size());
+            for (const auto& candidate : remaining) {
+                neighbors.emplace_back(candidate.second);
+            }
+            graph->InsertNeighborsById(id, neighbors);
+        }
+    };
+
+    if (dense_ids) {
+        parallel_for(graph->TotalCount(), parallelism, thread_pool, [&](uint64_t id) {
+            rank_neighbors(static_cast<InnerIdType>(id));
+        });
+    } else {
+        const auto ids = graph->GetIds();
+        parallel_for(
+            ids.size(), parallelism, thread_pool, [&](uint64_t i) { rank_neighbors(ids[i]); });
+    }
+}
+
+}  // namespace
+
+bool
+HGraph::can_reduce_max_degree_unlocked() const {
+    if (this->immutable_.load(std::memory_order_acquire) ||
+        this->graph_type_ != GRAPH_TYPE_VALUE_NSW || this->bottom_graph_ == nullptr ||
+        !this->bottom_graph_->InMemory() || this->support_force_remove_) {
+        return false;
+    }
+    const auto hgraph_param = std::dynamic_pointer_cast<HGraphParameter>(this->create_param_ptr_);
+    if (hgraph_param == nullptr) {
+        return false;
+    }
+    const auto bottom_param =
+        std::dynamic_pointer_cast<GraphDataCellParameter>(hgraph_param->bottom_graph_param);
+    if (bottom_param == nullptr || bottom_param->support_remove_ ||
+        bottom_param->use_reverse_edges_) {
+        return false;
+    }
+    return this->hierarchical_datacell_param_ != nullptr &&
+           !this->hierarchical_datacell_param_->support_delete_ &&
+           !this->hierarchical_datacell_param_->use_reverse_edges_ &&
+           this->get_degree_reduction_codes() != nullptr;
+}
+
+FlattenInterfacePtr
+HGraph::get_degree_reduction_codes() const {
+    if (this->has_precise_reorder() && !this->build_by_base_) {
+        return this->high_precise_codes_;
+    }
+    if (this->basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_RABITQ) {
+        return this->raw_vector_;
+    }
+    return this->basic_flatten_codes_;
+}
+
+bool
+HGraph::CanReduceMaxDegree() const {
+    std::scoped_lock lock(this->add_mutex_, this->global_mutex_);
+    return this->can_reduce_max_degree_unlocked();
+}
+
+void
+HGraph::PrepareDegreeReduction() {
+    std::scoped_lock lock(this->add_mutex_, this->global_mutex_);
+    CHECK_ARGUMENT(this->can_reduce_max_degree_unlocked(),
+                   "HGraph max-degree reduction does not support this index");
+    if (this->degree_reduction_prepared_) {
+        return;
+    }
+
+    const auto flatten = this->get_degree_reduction_codes();
+    rank_graph(this->bottom_graph_,
+               flatten,
+               this->allocator_,
+               this->alpha_,
+               true,
+               this->thread_pool_,
+               this->build_thread_count_);
+    for (const auto& route : this->route_graphs_) {
+        rank_graph(route,
+                   flatten,
+                   this->allocator_,
+                   this->alpha_,
+                   false,
+                   this->thread_pool_,
+                   this->build_thread_count_);
+    }
+
+    this->degree_reduction_prepared_ = true;
+}
+
+void
+HGraph::ReduceMaxDegree(uint32_t max_degree) {
+    std::scoped_lock lock(this->add_mutex_, this->global_mutex_);
+    CHECK_ARGUMENT(this->can_reduce_max_degree_unlocked(),
+                   "HGraph max-degree reduction does not support this index");
+    CHECK_ARGUMENT(this->degree_reduction_prepared_,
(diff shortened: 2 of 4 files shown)


src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+                first_exception = std::current_exception();
+            }
+        }
+    }
+    if (first_exception != nullptr) {
+        std::rethrow_exception(first_exception);
+    }
+}
+
+GraphInterfacePtr
```
LHT129, line 81: [suggestion] The `parallel_for` helper has a subtle exception-safety issue: when `GeneralEnqueue` throws, the loop `break`s, leaving already-submitted futures running. Those futures may also throw, and the collection loop silently discards later exceptions when `first_exception` is already set. More importantly, the remaining work items are never submitted, leaving the graph in a partially-processed state with no rollback.

Consider either:
1. Collecting all exceptions and reporting them together, or
2. Ensuring that partial execution is safe (e.g., by documenting that `rank_graph` and `materialize_graph` are idempotent per-node and partial execution is acceptable).

At minimum, the `break` on submission failure should be replaced with a strategy that either cancels already-submitted work or documents why partial execution is safe.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+        Vector<InnerIdType> neighbors(allocator);
+        graph->GetNeighbors(id, neighbors);
+        if (!neighbors.empty()) {
+            const auto original = neighbors;
+            select_edges_by_heuristic(neighbors, id, neighbors.size(), flatten, allocator, alpha);
+            std::reverse(neighbors.begin(), neighbors.end());
+
+            std::vector<std::pair<float, InnerIdType>> remaining;
+            remaining.reserve(original.size() - neighbors.size());
+            for (const auto neighbor : original) {
```
LHT129, line 137: [suggestion] The `std::find` on line 137 performs an O(N) linear scan over `neighbors` for each `original` neighbor, making the per-node complexity O(M*K) where M is the original degree and K is the number of selected neighbors. For graphs with large `max_degree` (e.g., 64 or 128), this becomes a noticeable hotspot.

Since `neighbors` is the result of `select_edges_by_heuristic` (typically small), consider using an `std::unordered_set<InnerIdType>` or sorting both vectors and using `std::set_difference` to avoid the quadratic behavior. Alternatively, since `neighbors` is modified in-place by `select_edges_by_heuristic` and then reversed, you could track which elements were kept during the heuristic selection itself rather than re-discovering them with `std::find`.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+    if (first_exception != nullptr) {
+        std::rethrow_exception(first_exception);
+    }
+}
+
+GraphInterfacePtr
+materialize_graph(const GraphInterfacePtr& source,
+                  const GraphInterfaceParamPtr& target_param,
+                  const FlattenInterfacePtr& flatten,
+                  Allocator* allocator,
```
LHT129, line 85: [note] The `flatten` parameter in `materialize_graph` is only used for `ExportCommonParam()` to create the target graph instance. The actual neighbor copying does not use `flatten` at all — it only reads neighbor IDs from the source graph and writes them to the target. This is fine functionally, but the parameter name and presence may mislead readers into thinking distance computation is involved.

Consider either:
1. Passing `CommonParam` directly instead of the full `FlattenInterfacePtr`, or
2. Adding a brief comment clarifying that `flatten` is only used for graph instantiation, not for distance calculations.

src/algorithm/hgraph/hgraph.h
```diff
@@ -845,5 +862,7 @@ class HGraph : public InnerIndexInterface {
     float build_cache_hit_rate_{-1.0F};     // cache hit rate from last cache-based build
     uint64_t build_cache_hit_nodes_{0};     // number of nodes with cache hit
     uint64_t build_cache_missed_nodes_{0};  // number of nodes without cache hit
+
```
LHT129, line 865: [note] `degree_reduction_prepared_` is not serialized/deserialized. After deserializing a previously-reduced graph, the flag will be `false`, so `PrepareDegreeReduction` would need to be called again before further reductions — which would re-rank the already-materialized neighbors unnecessarily. This is not incorrect but is wasteful.

If the reduced graph is intended to be further reducible after deserialization, consider either:
1. Persisting the flag in the serialization format, or
2. Making `PrepareDegreeReduction` detect that neighbors are already ranked (e.g., by checking if the first neighbor of each node is the closest).

LHT129 (commented):
Overall this is a well-structured PR that introduces a useful internal capability for HGraph. The design is clean: rank adjacency lists once with the diversity heuristic, then materialize prefixes into compact storage for each target `max_degree`. The separation of `PrepareDegreeReduction` (rank) from `ReduceMaxDegree` (materialize) is a good choice that allows multiple reductions from a single ranking pass.

**Summary of findings:**

1. **[suggestion] `parallel_for` exception safety** — The `break` on submission failure leaves already-submitted futures running with no cancellation, and remaining work items are silently dropped. This leaves the graph in a partially-processed state. Consider documenting why this is safe or adding a rollback mechanism.

2. **[suggestion] `std::find` in `rank_graph`** — The O(M*K) linear scan per node can become a hotspot for large `max_degree` values. An `unordered_set` or set-based approach would be more efficient.

3. **[note] `materialize_graph` flatten parameter** — The parameter is only used for `ExportCommonParam()` and is otherwise dead. Consider passing `CommonParam` directly or adding a clarifying comment.

4. **[note] `degree_reduction_prepared_` not serialized** — After deserialization the flag resets to `false`, causing unnecessary re-ranking if further reductions are attempted. This is not incorrect but is wasteful.

The test coverage is solid — it covers the happy path (build → reduce → serialize → deserialize → search), repeated reductions, and invalid input rejection. The performance data in the PR description is compelling (2.47x construction speedup).

LHT129 (commented):
Overall this is a well-structured PR. The degree reduction capability is cleanly separated into its own translation unit, the public API surface on HGraph is minimal (3 methods), and the guard conditions in `can_reduce_max_degree_unlocked` are thorough. The test covers the happy path including successive reductions, serialization round-trip, and invalid input rejection.

The existing inline comments cover the main areas for improvement: `parallel_for` exception safety, the O(N) linear scan in `rank_graph`, the `flatten` parameter naming, and the non-persisted `degree_reduction_prepared_` flag. No additional blocking issues found.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+                   "HGraph max-degree reduction does not support this index");
+    CHECK_ARGUMENT(this->degree_reduction_prepared_,
+                   "PrepareDegreeReduction must be called before ReduceMaxDegree");
+    CHECK_ARGUMENT(max_degree >= 4, "target max_degree must be at least 4");
+    CHECK_ARGUMENT(max_degree < this->bottom_graph_->MaximumDegree(),
+                   "target max_degree must be smaller than the current max_degree");
+
+    const auto hgraph_param = std::dynamic_pointer_cast<HGraphParameter>(this->create_param_ptr_);
+    auto bottom_param = std::make_shared<GraphDataCellParameter>(
+        *std::dynamic_pointer_cast<GraphDataCellParameter>(hgraph_param->bottom_graph_param));
```
LHT129, line 247: [suggestion] In `ReduceMaxDegree`, the `dynamic_pointer_cast<HGraphParameter>(this->create_param_ptr_)` on line 247 and the inner `dynamic_pointer_cast<GraphDataCellParameter>(hgraph_param->bottom_graph_param)` on line 249 are used without null checks. While `can_reduce_max_degree_unlocked()` (called via `CHECK_ARGUMENT` on line 240) validates that `hgraph_param` is non-null, the `bottom_graph_param` cast is not validated there — `can_reduce_max_degree_unlocked` only checks `bottom_param->support_remove_` and `bottom_param->use_reverse_edges_`, which would already dereference the pointer.

If the `create_param_ptr_` is somehow corrupted or replaced between the check and the cast (e.g., by a concurrent operation that bypasses the mutex), the unchecked `dynamic_pointer_cast` would produce a null pointer and crash on dereference. Consider adding an explicit null check after the cast for defense-in-depth, or using `CHECK_ARGUMENT` to fail gracefully.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+           bool dense_ids,
+           const SafeThreadPoolPtr& thread_pool,
+           uint64_t parallelism) {
+    const auto rank_neighbors = [&](InnerIdType id) {
+        Vector<InnerIdType> neighbors(allocator);
+        graph->GetNeighbors(id, neighbors);
+        if (!neighbors.empty()) {
+            const auto original = neighbors;
+            select_edges_by_heuristic(neighbors, id, neighbors.size(), flatten, allocator, alpha);
+            std::reverse(neighbors.begin(), neighbors.end());
```
LHT129, line 133: [suggestion] `const auto original = neighbors;` on line 133 performs a full copy of the neighbor vector for every node. For graphs with large `max_degree` (e.g., 64 or 128), this allocates and copies up to `max_degree` `InnerIdType` values per node, only to use `original` for the `std::find` membership test on line 137 and for computing the `reserve` size on line 135.

Since `select_edges_by_heuristic` modifies `neighbors` in-place and the remaining neighbors are re-discovered via `std::find`, the full copy is unnecessary. Consider either:
1. Computing `original.size()` before calling `select_edges_by_heuristic` and storing just the size, then iterating over the original graph neighbors by re-fetching them via `graph->GetNeighbors(id, ...)` in a separate vector, or
2. Tracking which neighbors were selected by the heuristic (e.g., via a boolean array or a set) to avoid the `std::find` scan and the full copy altogether.

This would reduce per-node memory allocation and improve cache efficiency during the ranking pass.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+    CHECK_ARGUMENT(max_degree < this->bottom_graph_->MaximumDegree(),
+                   "target max_degree must be smaller than the current max_degree");
+
+    const auto hgraph_param = std::dynamic_pointer_cast<HGraphParameter>(this->create_param_ptr_);
+    auto bottom_param = std::make_shared<GraphDataCellParameter>(
+        *std::dynamic_pointer_cast<GraphDataCellParameter>(hgraph_param->bottom_graph_param));
+    bottom_param->max_degree_ = max_degree;
+    auto bottom = materialize_graph(this->bottom_graph_,
+                                    bottom_param,
+                                    this->basic_flatten_codes_,
```
LHT129, line 251: [note] `PrepareDegreeReduction` uses `get_degree_reduction_codes()` (which may return `high_precise_codes_` or `raw_vector_` for RaBitQ) to rank neighbors via distance computation, but `ReduceMaxDegree` passes `this->basic_flatten_codes_` to `materialize_graph`. In `materialize_graph`, the `flatten` parameter is only used for `ExportCommonParam()` to instantiate the target graph, so this is functionally correct as long as both flattens share the same `CommonParam`.

However, the asymmetry between the two methods is easy to miss during maintenance. If `materialize_graph` ever gains logic that depends on the specific flatten (e.g., for distance-based filtering during materialization), this would become a subtle bug. Consider either:
1. Passing the same flatten (from `get_degree_reduction_codes()`) to both `rank_graph` and `materialize_graph` for consistency, or
2. Adding a brief comment in `ReduceMaxDegree` explaining why `basic_flatten_codes_` is sufficient for graph instantiation.

LHT129 (commented):
Overall this is a well-structured PR that adds a useful internal capability for deriving lower-max_degree graphs from existing HGraph indices. The separation into `CanReduceMaxDegree` / `PrepareDegreeReduction` / `ReduceMaxDegree` is clean, and the guard rails (compatibility checks, mutex locking, argument validation) are thorough. The test covers the happy path, serialization round-trip, and invalid input rejection.

I left three additional inline comments covering:
1. Defense-in-depth null checks on `dynamic_pointer_cast` results in `ReduceMaxDegree`
2. A potential optimization to avoid the full neighbor vector copy in `rank_graph`
3. A consistency note about `basic_flatten_codes_` vs `get_degree_reduction_codes()` usage between `PrepareDegreeReduction` and `ReduceMaxDegree`

None of these are blocking — the code is functionally correct as-is.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+}
+
+FlattenInterfacePtr
+HGraph::get_degree_reduction_codes() const {
+    if (this->has_precise_reorder() && !this->build_by_base_) {
+        return this->high_precise_codes_;
+    }
+    if (this->basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_RABITQ) {
+        return this->raw_vector_;
+    }
```
LHT129, line 194: [note] `get_degree_reduction_codes()` dereferences `this->basic_flatten_codes_->GetQuantizerName()` before the null check in `can_reduce_max_degree_unlocked()` can protect it.

In `get_degree_reduction_codes()`:
```cpp
if (this->basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_RABITQ) {
    return this->raw_vector_;
}
return this->basic_flatten_codes_;
```

If `basic_flatten_codes_` is null (e.g., due to an initialization ordering change or a bug), this dereference crashes instead of returning `nullptr`. The caller `can_reduce_max_degree_unlocked()` checks `this->get_degree_reduction_codes() != nullptr` as its final condition, but that check never executes if the function crashes first.

In practice `basic_flatten_codes_` should always be non-null when `bottom_graph_` is non-null (both are set during Build), so this is a latent risk rather than an active bug. Consider adding an explicit null guard:
```cpp
if (this->basic_flatten_codes_ != nullptr &&
    this->basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_RABITQ) {
    return this->raw_vector_;
}
```

src/algorithm/hgraph/hgraph_degree_reduction_test.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::stringstream artifact;
+        REQUIRE(index->Serialize(artifact).has_value());
+        artifact.seekg(0, std::ios::beg);
+        auto restored = vsag::Factory::CreateIndex("hgraph", create_parameters(degree));
+        REQUIRE(restored.has_value());
+        REQUIRE(restored.value()->Deserialize(artifact).has_value());
+        REQUIRE(
+            restored.value()->KnnSearch(query, 10, R"({"hgraph":{"ef_search":40}})").has_value());
+    }
+
```
LHT129, line 93: [suggestion] The test covers the happy path (fp32 + l2, valid reduction from 16→8→4, serialization round-trip, invalid degree rejection), but misses a few important error paths:

1. **`ReduceMaxDegree` without prior `PrepareDegreeReduction`**: Calling `ReduceMaxDegree` before `PrepareDegreeReduction` should throw with "PrepareDegreeReduction must be called before ReduceMaxDegree". This is a key API contract that deserves test coverage.

2. **`CanReduceMaxDegree` returning false**: There is no test verifying that `CanReduceMaxDegree()` returns `false` for unsupported configurations (e.g., immutable index, non-NSW graph type, or `support_force_remove_` enabled).

3. **RaBitQ code path in `get_degree_reduction_codes()`**: The function has special handling for `QUANTIZATION_TYPE_VALUE_RABITQ` (returning `raw_vector_` instead of `basic_flatten_codes_`), but this path is not exercised by the current test which uses `fp32` quantization.

Consider adding test cases for at least items 1 and 2 to validate the API contracts.

src/algorithm/hgraph/hgraph_degree_reduction.cpp
```diff
@@ -0,0 +1,280 @@
...
+                                              false,
+                                              this->thread_pool_,
+                                              this->build_thread_count_));
+    }
+
+    this->bottom_graph_ = std::move(bottom);
+    this->route_graphs_ = std::move(routes);
+    hgraph_param->bottom_graph_param = bottom_param;
+    hgraph_param->hierarchical_graph_param = route_param;
+    this->hierarchical_datacell_param_ = route_param;
```
LHT129, line 276: [suggestion] After `ReduceMaxDegree` completes, `degree_reduction_prepared_` is not reset to `false`. This means a subsequent call to `PrepareDegreeReduction` will return immediately without re-ranking, even though the underlying `bottom_graph_` and `route_graphs_` have been replaced with lower-degree versions.

If the caller intends to do successive reductions (e.g., 64→32→16), the current flow works because `PrepareDegreeReduction` is only called once upfront. However, if the caller ever needs to call `PrepareDegreeReduction` again after a reduction (e.g., after adding new vectors and wanting to re-rank), the stale `true` flag will silently skip the ranking step while the graphs have already been replaced.

Consider either:
1. Resetting `degree_reduction_prepared_ = false` at the end of `ReduceMaxDegree`, or
2. Documenting that `PrepareDegreeReduction` + `ReduceMaxDegree` is a one-shot sequence and must not be re-invoked after the first reduction.

```cpp
// At the end of ReduceMaxDegree, add:
this->degree_reduction_prepared_ = false;
```

jac0626:
Superseded by #2673, which includes the HGraph max_degree reduction capability together with its AutoTune integration, conservative default policy, fallback behavior, documentation, tests, and SIFT1M benchmark evidence.

PR #2580, opened 2026-07-30
fix(hgraph): reset state after removing all elements

### LINKED ISSUE #2581: HGraph ForceRemove leaves invalid state after index becomes empty
## Problem

Force-removing the final HGraph element can leave the entry point and graph bookkeeping in a stale state. A subsequent normal search may access removed graph data, and re-adding an element may probe an empty code slot.

## Resolution

PR #2580 resets the empty-index state, returns early for empty request searches, and adds regression coverage for empty KNN, empty range search, and re-add.

LHT129:
## Summary
- reset the HGraph entry point and graph state when the final element is force-removed
- short-circuit empty SearchWithRequest queries
- cover empty KNN, empty range search, and re-adding an element

Fixes #2581

## Validation
- `./build/tests/functests "HGraph ForceRemove All Elements" -d yes`
- `make release`
- `make lint`

### DIFF
--- a/src/algorithm/hgraph/hgraph_modify.cpp
+++ b/src/algorithm/hgraph/hgraph_modify.cpp
@@ -198,7 +198,15 @@ HGraph::force_remove_one(int64_t label) {
     if (was_mark_removed) {
         this->delete_count_.fetch_sub(1);
     }
-    this->total_count_.fetch_sub(1);
+    {
+        std::scoped_lock lock(this->global_mutex_);
+        if (this->total_count_.load() == 1) {
+            this->entry_point_id_ = INVALID_ENTRY_POINT;
+            this->bottom_graph_->SetTotalCount(0);
+            this->route_graphs_.clear();
+        }
+        this->total_count_.fetch_sub(1);
+    }
     return 1;
 }
 
--- a/src/algorithm/hgraph/hgraph_search.cpp
+++ b/src/algorithm/hgraph/hgraph_search.cpp
@@ -402,7 +402,11 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
         }
         shared_lock = this->acquire_global_read_lock();
     }
-    k = std::min(k, GetNumElements());
+    const auto element_count = GetNumElements();
+    if (element_count == 0) {
+        return make_empty_dataset_with_stats();
+    }
+    k = std::min(k, element_count);
 
     // Setup reasoning context (KNN only)
     std::shared_ptr<ReasoningContext> reasoning_ctx;
--- a/tests/test_hgraph_remove.cpp
+++ b/tests/test_hgraph_remove.cpp
@@ -15,6 +15,7 @@
 #include <atomic>
 #include <catch2/catch_test_macros.hpp>
 #include <future>
+#include <limits>
 #include <mutex>
 #include <nlohmann/json.hpp>
 #include <thread>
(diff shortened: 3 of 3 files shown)


src/algorithm/hgraph/hgraph_modify.cpp
```diff
@@ -198,7 +198,15 @@ HGraph::force_remove_one(int64_t label) {
     if (was_mark_removed) {
         this->delete_count_.fetch_sub(1);
     }
-    this->total_count_.fetch_sub(1);
+    {
+        std::scoped_lock lock(this->global_mutex_);
```
LHT129, line 202: [note] The `total_count_.load() == 1` check inside the `global_mutex_` lock correctly identifies the last-element scenario because `force_remove_one` is serialized by `force_remove_mutex_`. However, note that `total_count_` includes mark-removed elements. If all visible elements have been force-removed but mark-removed elements remain in storage, `total_count_` will be > 1 and the reset will not trigger. This is correct behavior (the graph still has entries), but worth documenting as an implicit invariant.

src/algorithm/hgraph/hgraph_modify.cpp
```diff
@@ -198,7 +198,15 @@ HGraph::force_remove_one(int64_t label) {
...
+        if (this->total_count_.load() == 1) {
+            this->entry_point_id_ = INVALID_ENTRY_POINT;
+            this->bottom_graph_->SetTotalCount(0);
+            this->route_graphs_.clear();
+        }
```
LHT129, line 207: [suggestion] After resetting `entry_point_id_` to `INVALID_ENTRY_POINT`, `bottom_graph_->SetTotalCount(0)`, and `route_graphs_.clear()`, the `shrink_to_fit()` call in the caller (`Remove()`) will still execute. Since `total_count_` is now 0, `shrink_to_fit()` will call `ShrinkToFit(0)` on all components, which should be harmless but is redundant. Consider adding an early return in `shrink_to_fit()` when `total_count == 0`, or skip the `shrink_to_fit()` call when the count reaches 0.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -402,7 +402,11 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
         }
         shared_lock = this->acquire_global_read_lock();
     }
-    k = std::min(k, GetNumElements());
+    const auto element_count = GetNumElements();
+    if (element_count == 0) {
```
LHT129, line 406: [note] The early return at `element_count == 0` is placed after acquiring `force_remove_rlock` and `shared_lock`. This is correct for correctness (reading `GetNumElements()` under proper synchronization), but means empty queries still pay the lock acquisition cost. This is acceptable since empty-index queries are a rare edge case.

LHT129 (commented):
Overall this is a clean, focused fix for the empty-index edge case after removing all elements. The changes are minimal and well-targeted.

**Summary of changes:**
- `hgraph_modify.cpp`: When the last element is force-removed, resets `entry_point_id_`, `bottom_graph_` count, and `route_graphs_` under `global_mutex_` to leave the graph in a clean empty state.
- `hgraph_search.cpp`: Adds an early return in `SearchWithRequest` when the element count is 0, avoiding downstream code that would operate on an invalid entry point.
- `test_hgraph_remove.cpp`: Extends the "ForceRemove All Elements" test to cover empty range search and re-adding an element after full removal.

**Correctness:** The fix correctly handles the edge case. The `global_mutex_` exclusive lock during the reset ensures no concurrent search sees an inconsistent state. The `total_count_.load() == 1` check before the decrement correctly identifies the last-element scenario.

**Test coverage:** Good — covers empty KNN, empty range search, and re-add-after-remove-all scenarios. The existing "ForceRemove All Elements Twice" test also validates that double-removing is still handled correctly.

wxyucs (approved):
lgtm

PR #2584, opened 2026-07-30
perf(pyramid): parallelize initial code encoding

### LINKED ISSUE #2583: [improve](other): parallelize Pyramid code encoding during Build
### Kind of improvement

perf

### Affected area

other (Pyramid)

### Summary

Pyramid's NSW build path prepares the path tree and then delegates the full initial dataset to `Add()`. The label loop in `Add()` encodes every base vector, and every precise vector when reorder is enabled, serially.

For high-dimensional RaBitQ + SQ8 configurations, this serial encoding phase dominates build time and leaves the configured build worker pool underused. Use the existing build thread pool to encode pre-sized, disjoint inner-ID ranges during `Build()`, while leaving incremental `Add()` and hierarchy/graph insertion behavior unchanged.

### Expected benefit

Allow `build_thread_count` to accelerate the expensive initial base/precise encoding phase without changing the public API or incremental insertion semantics.

On the same 32-thread benchmark host, the minimal implementation reduced median build time by about 81%:

- 20,000 × 4096D: 116.268 s to 22.046 s (5.27x)
- 20,000 × 2560D: 29.587 s to 5.723 s (5.17x)

The speedup is not expected to scale linearly to 32x because RaBitQ's dense rotation is limited by the cache/memory hierarchy and the remaining build phases are serial.

### Benchmark evidence

Environment:

- Intel Xeon Platinum 8163
- 48 physical cores / 96 logical CPUs
- 503 GiB RAM
- Process pinned to CPUs 0-47
- `OMP_NUM_THREADS=1` and `OPENBLAS_NUM_THREADS=1`
- Three measured runs per variant; values above are medians

Dataset:

- 20,000 generated FP32 vectors
- Dimensions: 4096 and 2560
- Identical shuffled, user-like path distribution for baseline and patched runs
- The selected prefix produced 17,980 leaf nodes, no graph subindexes, and a maximum leaf size of 21

Relevant Pyramid parameters:

```json
{
  "no_build_levels": [0],
  "index_min_size": 500,
  "max_degree": 32,
  "base_quantization_type": "rabitq",
  "precise_quantization_type": "sq8",
  "support_duplicate": false,
  "use_reorder": true,
  "alpha": 1.2,
  "ef_construction": 400,
  "build_thread_count": 32,
  "graph_storage_type": "compressed",
  "rabitq_bits_per_dim_base": 1
}
```

Functional regression tests verify that encoding tasks overlap while preserving compact inner IDs after duplicate labels, and that an actual RaBitQ + SQ8 + block-memory configuration decodes correctly. Recall and peak-memory measurements were not part of this focused benchmark.

### References

- `src/algorithm/pyramid/pyramid.cpp:1029-1031` — NSW `Build()` delegates to `Add()`.
- `src/algorithm/pyramid/pyramid.cpp:780-791` — base and precise codes are encoded serially inside the label loop.

### Additional context

Keep the first change intentionally narrow: parallelize only the initial code-encoding phase. Path-tree construction, hierarchy insertion, graph scheduling, and further memory-bandwidth optimization remain out of scope and can be evaluated independently.

## Related issues

_None found._

---
_Drafted with AI assistant: Codex:gpt-5.6_

jac0626:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Closes: #2583

## What Changed

- Kept the existing NSW `Build() → Add()` flow unchanged.
- In the existing `Add()` encoding section, first finish label de-duplication and compact inner-ID assignment, then encode the initial base and precise codes over fixed, disjoint ranges on the existing build thread pool.
- Enable the parallel branch only for an empty index when the code stores have just been physically resized and use audited in-memory FP32, RaBitQ, or SQ8 implementations.
- Keep incremental `Add()`, non-memory IO, unsupported quantizers, and non-growing storage on the original serial encoding path.
- Path-tree construction, hierarchy insertion, and graph construction are unchanged.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
$ ./build/tests/unittests '[ut][pyramid]'
All tests passed (39 assertions in 4 test cases)

$ /opt/homebrew/opt/llvm@15/bin/clang-format --dry-run --Werror \
    src/algorithm/pyramid/pyramid.cpp \
    src/algorithm/pyramid/pyramid.h \
    src/algorithm/pyramid/pyramid_test.cpp
Homebrew clang-format version 15.0.7

$ git diff --check
# no output
```

Focused 32-thread benchmark medians on the same host and path distribution:

- 20,000 × 4096D, RaBitQ + SQ8: 116.268 s → 22.046 s (5.27x)
- 20,000 × 2560D, RaBitQ + SQ8: 29.587 s → 5.723 s (5.17x)

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: initial code encoding may run concurrently when `build_thread_count > 1`; code layout and incremental insertion behavior are unchanged.

- Performance impact: improved initial build throughput for expensive high-dimensional encoders.
- Concurrency/thread-safety impact: workers write fixed, non-overlapping inner-ID ranges only after storage resize completes. Incremental insertion and unsupported storage/quantizer combinations remain serial.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert commit `d0356116` to restore serial code encoding.

## Checklist

- [x] I have added/updated tests for the changed behavior.
- [x] I have updated docs if behavior/workflow changed.
- [x] My commit messages follow project conventions.

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -16,6 +16,7 @@
 #include "pyramid.h"
 
 #include <chrono>
+#include <exception>
 
 #include "algorithm/inner_index_interface.h"
 #include "analyzer/analyzer.h"
@@ -759,7 +760,7 @@ Pyramid::ExportModel(const IndexCommonParam& param) const {
 
 std::vector<int64_t>
 Pyramid::Add(const DatasetPtr& base) {
-    int64_t data_num = base->GetNumElements();
+    const int64_t data_num = base->GetNumElements();
     const auto* data_vectors = base->GetFloat32Vectors();
     const auto* data_ids = base->GetIds();
     std::vector<int64_t> failed_ids;
@@ -768,33 +769,99 @@ Pyramid::Add(const DatasetPtr& base) {
     {
         std::lock_guard lock(cur_element_count_mutex_);
         local_cur_element_count = cur_element_count_;
+        auto new_capacity = max_capacity_;
         if (max_capacity_ == 0) {
-            auto new_capacity = std::max(INIT_CAPACITY, data_num);
-            resize(new_capacity);
+            new_capacity = std::max(INIT_CAPACITY, data_num);
         } else if (max_capacity_ < data_num + cur_element_count_) {
-            auto new_capacity = std::min(MAX_CAPACITY_EXTEND, max_capacity_);
+            new_capacity = std::min(MAX_CAPACITY_EXTEND, max_capacity_);
             new_capacity = std::max(data_num + cur_element_count_ - max_capacity_, new_capacity) +
                            max_capacity_;
+        }
+        bool base_storage_resized = false;
+        bool precise_storage_resized = false;
+        if (new_capacity > max_capacity_) {
+            base_storage_resized = new_capacity > static_cast<int64_t>(base_codes_->max_capacity_);
+            precise_storage_resized =
+                not use_reorder_ ||
+                new_capacity > static_cast<int64_t>(precise_codes_->max_capacity_);
             resize(new_capacity);
         }
-        int64_t valid_id_count = 0;
+
+        data_biases.reserve(data_num);
         for (int64_t i = 0; i < data_num; ++i) {
             if (not label_table_->CheckLabel(data_ids[i])) {
-                label_table_->Insert(valid_id_count + local_cur_element_count, data_ids[i]);
-                base_codes_->InsertVector(data_vectors + dim_ * i,
-                                          valid_id_count + local_cur_element_count);
-                if (use_reorder_) {
-                    precise_codes_->InsertVector(data_vectors + dim_ * i,
-                                                 valid_id_count + local_cur_element_count);
-                }
-                valid_id_count++;
+                const auto inner_id =
+                    static_cast<InnerIdType>(local_cur_element_count + data_biases.size());
+                label_table_->Insert(inner_id, data_ids[i]);
                 data_biases.push_back(i);
             } else {
                 logger::warn("Label {} already exists, skip adding.", data_ids[i]);
                 failed_ids.push_back(data_ids[i]);
             }
         }
-        cur_element_count_ += valid_id_count;
+
+        const auto encode_range = [this, data_vectors, local_cur_element_count, &data_biases](
+                                      uint64_t begin, uint64_t end) {
+            for (uint64_t offset = begin; offset < end; ++offset) {
+                const auto* vector = data_vectors + dim_ * data_biases[offset];
+                const auto inner_id = static_cast<InnerIdType>(local_cur_element_count + offset);
+                base_codes_->InsertVector(vector, inner_id);
+                if (use_reorder_) {
+                    precise_codes_->InsertVector(vector, inner_id);
+                }
+            }
+        };
+        const auto supports_parallel_encode = [](const FlattenInterfacePtr& codes) {
+            const auto name = codes->GetQuantizerName();
+            return codes->InMemory() &&
+                   (name == QUANTIZATION_TYPE_VALUE_FP32 ||
+                    name == QUANTIZATION_TYPE_VALUE_RABITQ || name == QUANTIZATION_TYPE_VALUE_SQ8);
+        };
+        const bool use_parallel_encode =
+            local_cur_element_count == 0 && thread_pool_ != nullptr && build_thread_count_ > 1 &&
+            data_biases.size() > 1 && supports_parallel_encode(base_codes_) &&
+            (not use_reorder_ || supports_parallel_encode(precise_codes_)) &&
+            base_storage_resized && precise_storage_resized;
+        if (use_parallel_encode) {
+            const uint64_t worker_count =
+                std::min<uint64_t>(build_thread_count_, data_biases.size());
+            const uint64_t block_size = (data_biases.size() + worker_count - 1) / worker_count;
+            Vector<std::future<void>> futures(allocator_);
+            futures.reserve(worker_count);
+            const auto wait_futures = [&futures]() {
+                std::exception_ptr first_exception = nullptr;
+                for (auto& future : futures) {
+                    try {
+                        future.get();
+                    } catch (...) {
+                        if (not first_exception) {
+                            first_exception = std::current_exception();
+                        }
+                    }
+                }
+                if (first_exception) {
+                    std::rethrow_exception(first_exception);
+                }
+            };
+            try {
+                for (uint64_t begin = 0; begin < data_biases.size(); begin += block_size) {
+                    const uint64_t end = std::min<uint64_t>(begin + block_size, data_biases.size());
+                    futures.push_back(thread_pool_->GeneralEnqueue(
+                        [encode_range, begin, end]() { encode_range(begin, end); }));
+                }
+            } catch (...) {
+                const auto enqueue_exception = std::current_exception();
+                try {
+                    wait_futures();
+                } catch (...) {
+                }
+                std::rethrow_exception(enqueue_exception);
+            }
+            wait_futures();
+        } else {
+            encode_range(0, data_biases.size());
+        }
+        cur_element_count_ += static_cast<int64_t>(data_biases.size());
     }
     std::shared_lock<std::shared_mutex> lock(resize_mutex_);
 
--- a/src/algorithm/pyramid/pyramid_test.cpp
+++ b/src/algorithm/pyramid/pyramid_test.cpp
@@ -15,6 +15,7 @@
 
 #include "pyramid.h"
 
+#include <cmath>
 #include <vector>
 
 #include "impl/allocator/safe_allocator.h"
@@ -31,7 +32,9 @@ struct PyramidTestIndex {
 };
 
 PyramidTestIndex
-MakePyramidIndex(uint32_t index_min_size) {
+MakePyramidIndex(uint32_t index_min_size,
+                 uint64_t build_thread_count = 1,
+                 bool use_rabitq_with_sq8 = false) {
     PyramidTestIndex result;
     vsag::IndexCommonParam common_param;
     common_param.dim_ = PYRAMID_TEST_DIM;
@@ -51,6 +54,15 @@ MakePyramidIndex(uint32_t index_min_size) {
         "index_min_size": 3
     })");
     external_param[vsag::PYRAMID_INDEX_MIN_SIZE].SetInt(index_min_size);
+    external_param[vsag::PYRAMID_BUILD_THREAD_COUNT].SetUint64(build_thread_count);
+    if (use_rabitq_with_sq8) {
+        external_param[vsag::PYRAMID_BASE_QUANTIZATION_TYPE].SetString("rabitq");
+        external_param[vsag::PYRAMID_PRECISE_QUANTIZATION_TYPE].SetString("sq8");
+        external_param[vsag::PYRAMID_BASE_IO_TYPE].SetString("block_memory_io");
+        external_param[vsag::PYRAMID_PRECISE_IO_TYPE].SetString("block_memory_io");
+        external_param[vsag::PYRAMID_USE_REORDER].SetBool(true);
+        external_param[vsag::PYRAMID_RABITQ_BITS_PER_DIM_BASE].SetUint64(1);
+    }
     auto param = vsag::Pyramid::CheckAndMappingExternalParam(external_param, common_param);
     result.index = std::make_shared<vsag::Pyramid>(param, common_param);
     return result;
@@ -155,5 +167,33 @@ TEST_CASE("Pyramid promotes flat node at index minimum size", "[ut][pyramid]") {
         auto result =
             index->KnnSearch(query, 1, R"({"pyramid":{"ef_search":10}})", vsag::FilterPtr{});
         REQUIRE(result->GetIds()[0] == ids[i]);
+    }
+}
+
+TEST_CASE("Pyramid Build stores RaBitQ and SQ8 codes in parallel", "[ut][pyramid]") {
+    constexpr int64_t count = 804;
+    auto test_index = MakePyramidIndex(count + 1, 4, true);
+
+    std::vector<float> vectors(count * PYRAMID_TEST_DIM);
+    std::vector<int64_t> ids(count);
+    std::vector<std::string> paths(count, "tenant");
+    for (int64_t i = 0; i < count; ++i) {
+        ids[i] = i;
+        for (int64_t j = 0; j < PYRAMID_TEST_DIM; ++j) {
+            vectors[i * PYRAMID_TEST_DIM + j] = static_cast<float>((i + j) % 101) / 100.0F;
+        }
+    }
+
+    auto failed_ids = test_index.index->Build(
+        MakePyramidDataset(vectors.data(), ids.data(), paths.data(), count));
+
+    REQUIRE(failed_ids.empty());
+    REQUIRE(test_index.index->GetNumElements() == count);
+    for (const int64_t inner_id : {0, 401, 803}) {
+        std::vector<float> decoded(PYRAMID_TEST_DIM);
+        test_index.index->GetVectorByInnerId(inner_id, decoded.data());
+        for (int64_t j = 0; j < PYRAMID_TEST_DIM; ++j) {
+            REQUIRE(std::abs(decoded[j] - vectors[inner_id * PYRAMID_TEST_DIM + j]) < 0.02F);
+        }
     }
 }


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1027,13 +1027,123 @@ Pyramid::Build(const DatasetPtr& base) {
...
     }
     return ret;
 }
 
+std::vector<int64_t>
+Pyramid::build_by_nsw(const DatasetPtr& base) {
+    const int64_t data_num = base->GetNumElements();
+    const auto* data_vectors = base->GetFloat32Vectors();
+    const auto* data_ids = base->GetIds();
+    std::vector<int64_t> failed_ids;
```
LHT129: [suggestion] The capacity calculation and resize logic in build_by_nsw is essentially duplicated from the existing Add method. This duplication means future changes to the capacity policy must be made in two places, increasing the risk of divergence. Consider extracting the capacity planning into a shared helper to keep the policy in one place.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1027,13 +1027,123 @@ Pyramid::Build(const DatasetPtr& base) {
...
+                const auto inner_id = static_cast<InnerIdType>(local_cur_element_count + offset);
+                const auto* vector = data_vectors + dim_ * data_bias;
+                base_codes_->InsertVector(vector, inner_id);
+                if (use_reorder_) {
+                    precise_codes_->InsertVector(vector, inner_id);
+                }
+            }
+        };
+
+        const auto supports_parallel_encode = [](const FlattenInterfacePtr& codes) {
```
LHT129: [note] The `supports_parallel_encode` lambda uses a whitelist of quantizer names (FP32, RaBitQ, SQ8) to gate parallel encoding. This is a reasonable safety measure, but it means that any new in-memory quantizer added in the future will silently fall back to serial encoding unless explicitly added to this list.

Consider whether the check should instead be based on a capability flag (e.g., `IsThreadSafeEncode()`) exposed by the `FlattenInterface`, which would make the gating self-documenting and automatically cover future quantizers that meet the thread-safety contract.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1027,13 +1027,123 @@ Pyramid::Build(const DatasetPtr& base) {
...
+                failed_ids.push_back(data_ids[i]);
+            }
+        }
+
+        const auto encode_range = [this, data_vectors, local_cur_element_count, &data_biases](
+                                      uint64_t begin, uint64_t end) {
+            for (uint64_t offset = begin; offset < end; ++offset) {
+                const auto data_bias = data_biases[offset];
+                const auto inner_id = static_cast<InnerIdType>(local_cur_element_count + offset);
+                const auto* vector = data_vectors + dim_ * data_bias;
```
LHT129: [note] The `encode_range` lambda is defined while holding `cur_element_count_mutex_`. Since the lambda does not access any mutex-protected state (it only reads `data_biases`, `data_vectors`, `dim_`, `base_codes_`, `precise_codes_`, `use_reorder_`), it could be defined outside the lock scope. Moving it out would slightly reduce the critical section and make the lock scope more obvious.

LHT129 (commented):
Overall this is a well-structured PR that delivers significant build performance improvements (5x speedup) with careful attention to thread safety. The approach of splitting encoding across fixed, non-overlapping inner-ID ranges after resize is sound.

**Summary of findings:**

1. **[suggestion] Capacity logic duplication** — The capacity calculation in `build_by_nsw` is copied from `Add`. Extracting it into a shared helper would reduce future maintenance risk.

2. **[note] Quantizer whitelist** — The `supports_parallel_encode` lambda uses a name-based whitelist. Consider a capability-based approach (e.g., `IsThreadSafeEncode()`) for future extensibility.

3. **[note] Lambda placement** — `encode_range` is defined inside the lock scope unnecessarily; moving it out would tighten the critical section.

The test coverage is thorough, including a concurrency tracking test (`TrackingFlattenDataCell`) and a RaBitQ+SQ8 integration test. The `PyramidTestAccess` friend class is an acceptable testing pattern.

LHT129 (commented):
## Code Review for PR #2584

### Summary
This PR parallelizes initial code encoding in `Pyramid::Add()` by splitting the encoding work across the build thread pool when the index is empty and the storage has been freshly resized. The change is well-scoped with clear safety gates and includes good test coverage.

### Inline Comments

**src/algorithm/pyramid/pyramid.cpp:782** — `[note]` Variable naming clarity

The `base_storage_resized` and `precise_storage_resized` flags are initialized to `false` and only set to `true` inside the `if (new_capacity > max_capacity_)` block. When no resize is needed, they remain `false`, which correctly prevents parallel encoding. However, the variable names suggest "was storage resized" while the actual semantics are "was storage freshly resized in this call". Consider renaming to `storage_just_resized` or adding a brief comment to clarify that `false` means "no fresh resize occurred, skip parallel path".

**src/algorithm/pyramid/pyramid.cpp:808** — `[note]` Lambda defined inside lock scope

The `encode_range` lambda is defined while holding `cur_element_count_mutex_`. Since the lambda does not access any mutex-protected state (it only reads `data_biases`, `data_vectors`, `dim_`, `base_codes_`, `precise_codes_`, `use_reorder_`), it could be defined outside the lock scope. Moving it out would slightly reduce the critical section and make the lock scope more obvious.

**src/algorithm/pyramid/pyramid_test.cpp:34** — `[suggestion]` Test helper uses `std::move` on a shared_ptr parameter

`PyramidTestAccess::SetBaseCodes` takes `FlattenInterfacePtr` by value and moves it. This is fine for the test usage, but the parameter type `FlattenInterfacePtr` (which is `std::shared_ptr`) passed by value forces an unnecessary atomic reference count increment/decrement at the call site. Since this is test-only code it is not critical, but using `const FlattenInterfacePtr&` would be more consistent with the codebase conventions.

### Overall Assessment
The implementation is solid: the parallel encoding path is gated by multiple safety checks (empty index, thread pool available, build_thread_count > 1, supported quantizers, storage freshly resized), and the serial fallback is preserved for all other cases. The tests verify both FP32 and RaBitQ+SQ8 paths with a concurrency tracking mechanism. The 5x speedup reported in benchmarks is consistent with the change.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +768,76 @@ Pyramid::Add(const DatasetPtr& base) {
     {
         std::lock_guard lock(cur_element_count_mutex_);
         local_cur_element_count = cur_element_count_;
+        auto new_capacity = max_capacity_;
         if (max_capacity_ == 0) {
-            auto new_capacity = std::max(INIT_CAPACITY, data_num);
```
LHT129, line 772: [note] The `base_storage_resized` and `precise_storage_resized` flags are initialized to `false` and only set to `true` inside the `if (new_capacity > max_capacity_)` block. When no resize is needed, they remain `false`, which correctly prevents parallel encoding. However, the variable names suggest "was storage resized" while the actual semantics are "was storage freshly resized in this call". Consider renaming to `storage_just_resized` or adding a brief comment to clarify that `false` means "no fresh resize occurred, skip parallel path".

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +768,76 @@ Pyramid::Add(const DatasetPtr& base) {
...
             resize(new_capacity);
         }
-        int64_t valid_id_count = 0;
+
+        data_biases.reserve(data_num);
         for (int64_t i = 0; i < data_num; ++i) {
             if (not label_table_->CheckLabel(data_ids[i])) {
-                label_table_->Insert(valid_id_count + local_cur_element_count, data_ids[i]);
-                base_codes_->InsertVector(data_vectors + dim_ * i,
-                                          valid_id_count + local_cur_element_count);
```
LHT129, line 785: [note] The `encode_range` lambda and `supports_parallel_encode` lambda are defined while holding `cur_element_count_mutex_`. Neither lambda accesses mutex-protected state — they only read `data_biases`, `data_vectors`, `dim_`, `base_codes_`, `precise_codes_`, and `use_reorder_`. Moving their definitions outside the lock scope would slightly reduce the critical section and make the lock boundary more obvious to readers.

src/algorithm/pyramid/pyramid_test.cpp
```diff
@@ -15,23 +15,77 @@
 
 #include "pyramid.h"
 
+#include <atomic>
+#include <chrono>
+#include <cmath>
+#include <thread>
 #include <vector>
```
LHT129, line 20: [note] `PyramidTestAccess::SetBaseCodes` takes `FlattenInterfacePtr` (which is `std::shared_ptr`) by value, incurring an unnecessary atomic reference count increment/decrement on each call. Consider passing by const reference (`const FlattenInterfacePtr&`) since the pointer is unconditionally moved into `base_codes_`.

jac0626: Removed `PyramidTestAccess` and the white-box injection test entirely in faebb9d9, so this helper and its shared_ptr parameter no longer exist. The RaBitQ + SQ8 integration coverage remains.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +768,76 @@ Pyramid::Add(const DatasetPtr& base) {
...
             }
         }
-        cur_element_count_ += valid_id_count;
+
+        const auto encode_range = [this, data_vectors, local_cur_element_count, &data_biases](
+                                      uint64_t begin, uint64_t end) {
+            for (uint64_t offset = begin; offset < end; ++offset) {
+                const auto* vector = data_vectors + dim_ * data_biases[offset];
+                const auto inner_id = static_cast<InnerIdType>(local_cur_element_count + offset);
+                base_codes_->InsertVector(vector, inner_id);
```
LHT129, line 808: [note] The `encode_range` lambda calls `base_codes_->InsertVector()` (and `precise_codes_->InsertVector()` when `use_reorder_` is true) concurrently from multiple threads. This relies on an undocumented thread-safety contract: the FP32, RaBitQ, and SQ8 `InsertVector` implementations must be safe for concurrent calls writing to non-overlapping inner-ID ranges.

The `supports_parallel_encode` whitelist gates which quantizers participate, but the thread-safety requirement itself is not visible at the `FlattenInterface` level. If someone adds a new in-memory quantizer and adds it to the whitelist without ensuring its `InsertVector` is reentrant, silent data corruption could occur.

Consider either:
- Adding a `bool IsInsertVectorThreadSafe() const` virtual method to `FlattenInterface` and using it instead of the name-based whitelist, or
- Adding a comment on `FlattenInterface::InsertVector` documenting that implementations may be called concurrently from non-overlapping inner-ID ranges.

src/algorithm/pyramid/pyramid_test.cpp
```diff
@@ -157,3 +169,31 @@ TEST_CASE("Pyramid promotes flat node at index minimum size", "[ut][pyramid]") {
         REQUIRE(result->GetIds()[0] == ids[i]);
     }
 }
+
```
LHT129, line 172: [note] The new test `Pyramid Build stores RaBitQ and SQ8 codes in parallel` verifies correctness of the parallel encoding output but does not confirm that the parallel code path was actually exercised. If a future change accidentally breaks the `use_parallel_encode` condition (e.g., a refactoring that always sets `base_storage_resized = false`), this test would still pass via the serial fallback.

Consider adding a way to observe whether parallel encoding was used — for example, checking `GetStats()` for thread pool activity, or temporarily exposing a test-only flag. This is low priority since the serial fallback produces identical results.

src/algorithm/pyramid/pyramid.cpp
jac0626: Addressed by the current structure. The separate `build_by_nsw` implementation was removed; NSW `Build()` continues to call the existing `Add()` path, so the capacity policy now has a single implementation and no shared helper is needed.

src/algorithm/pyramid/pyramid.cpp
jac0626: Keeping the conservative whitelist intentionally in this minimal PR. `FlattenInterface` currently has no concurrent-encode/write capability contract, `FlattenDataCell` is documented as thread-unsafe, and `InsertVector()` includes the IO write as well as quantization. Adding a capability across the interface and every implementation would materially expand this performance fix. New or unaudited combinations therefore fall back to the existing serial path by design.

src/algorithm/pyramid/pyramid.cpp
jac0626: This thread targets the superseded `build_by_nsw` version and is now outdated. The current implementation keeps only the existing `Add()` flow; the lock-scope decision for that implementation is covered in the current thread below.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +768,76 @@ Pyramid::Add(const DatasetPtr& base) {
...
+            (not use_reorder_ || supports_parallel_encode(precise_codes_)) &&
+            base_storage_resized && precise_storage_resized;
+        if (use_parallel_encode) {
+            const uint64_t worker_count =
+                std::min<uint64_t>(build_thread_count_, data_biases.size());
+            const uint64_t block_size = (data_biases.size() + worker_count - 1) / worker_count;
+            Vector<std::future<void>> futures(allocator_);
+            for (uint64_t begin = 0; begin < data_biases.size(); begin += block_size) {
+                const uint64_t end = std::min<uint64_t>(begin + block_size, data_biases.size());
+                futures.push_back(thread_pool_->GeneralEnqueue(
```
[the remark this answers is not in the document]
jac0626: Fixed in 91da5e90. The future vector is reserved before submission, all successfully submitted tasks are drained if enqueueing fails, and worker exceptions are collected while every future is waited before the first exception is rethrown. This keeps `data_biases` alive until no worker can reference it. The Pyramid unit tests still pass (32 assertions in 3 test cases).

src/algorithm/pyramid/pyramid.cpp
jac0626: Keeping the current names. These are local flags computed immediately before `resize()` and intentionally mean that the corresponding physical code storage is resized by this invocation; both must be true before the parallel branch is allowed. The scope and assignment make the current-call semantics explicit, so a rename or additional comment would not add enough clarity to justify extra churn.

src/algorithm/pyramid/pyramid.cpp
jac0626: Keeping these definitions in the existing lock scope. Constructing the lambdas is negligible and moving only their definitions would not reduce the actual critical work. Their execution must remain protected so label assignment, code insertion, and publishing `cur_element_count_` stay ordered against concurrent `Add()` calls. Keeping the definitions next to the guarded execution makes that lifetime and synchronization boundary clearer.

src/algorithm/pyramid/pyramid.cpp
jac0626: Keeping the concurrency contract local and conservative in this minimal performance PR. The branch requires in-memory storage, a fresh physical resize, fixed non-overlapping inner IDs, and an explicit whitelist of the three audited implementations. A future implementation cannot enter silently: it must be deliberately added to the whitelist and reviewed against both its quantizer and IO. Adding a new virtual capability or broadening the public `FlattenInterface` contract would affect every data-cell implementation and is intentionally out of scope here.

src/algorithm/pyramid/pyramid_test.cpp
jac0626: Keeping this as a black-box correctness test intentionally. Directly asserting worker overlap requires either a production/test-only observability hook or private data-cell replacement; the latter was removed to avoid adding `PyramidTestAccess` solely for this test, while timing-based assertions would be flaky. The explicit branch conditions are code-reviewed, the integration test protects encoded output, and the performance benchmark verifies that the parallel path is exercised without adding a permanent test seam.

LHT129, line 172: [suggestion] The new test only covers the RaBitQ + SQ8 parallel encoding path with `build_thread_count=4`. Consider adding test cases for:
1. FP32 parallel encoding (no reorder) to cover the most common code path
2. Serial fallback when `build_thread_count=1` to ensure the serial path still produces correct results
3. Edge case where `data_biases.size() == 1` (single vector) to verify the fallback to serial encoding

These additional cases would increase confidence that both the parallel and serial branches are correct across different configurations.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +769,99 @@ Pyramid::Add(const DatasetPtr& base) {
...
+            new_capacity = std::max(INIT_CAPACITY, data_num);
         } else if (max_capacity_ < data_num + cur_element_count_) {
-            auto new_capacity = std::min(MAX_CAPACITY_EXTEND, max_capacity_);
+            new_capacity = std::min(MAX_CAPACITY_EXTEND, max_capacity_);
             new_capacity = std::max(data_num + cur_element_count_ - max_capacity_, new_capacity) +
                            max_capacity_;
+        }
+        bool base_storage_resized = false;
+        bool precise_storage_resized = false;
+        if (new_capacity > max_capacity_) {
```
LHT129, line 782: [note] The variable names `base_storage_resized` and `precise_storage_resized` could be misleading. They are only set to `true` when `new_capacity > max_capacity_` AND the individual code store capacity also needs to grow. However, the name "resized" suggests the storage was resized, which may not be the case when `new_capacity <= max_capacity_` (both stay `false`). Consider renaming to something like `base_storage_grew` or `base_storage_expanded` to more precisely convey that these flags indicate the storage was actually resized in this invocation, not merely that it has been resized at some point.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -768,33 +769,99 @@ Pyramid::Add(const DatasetPtr& base) {
...
+            futures.reserve(worker_count);
+            const auto wait_futures = [&futures]() {
+                std::exception_ptr first_exception = nullptr;
+                for (auto& future : futures) {
+                    try {
+                        future.get();
+                    } catch (...) {
+                        if (not first_exception) {
+                            first_exception = std::current_exception();
+                        }
```
LHT129, line 839: [note] In the `wait_futures` lambda, when multiple workers throw, only the first exception is preserved and rethrown. The remaining exceptions are silently discarded. While this is a common pattern and the encoding tasks are unlikely to throw different exception types, consider logging the suppressed exceptions (at least their count) to aid debugging if encoding failures ever occur in production. For example:
```cpp
int suppressed = 0;
for (auto& future : futures) {
    try {
        future.get();
    } catch (...) {
        if (not first_exception) {
            first_exception = std::current_exception();
        } else {
            ++suppressed;
        }
    }
}
if (suppressed > 0) {
    logger::warn("Suppressed {} additional encoding exceptions", suppressed);
}
```

LHT129 (approved):
LGTM

PR #2597, opened 2026-08-03
feat(ivf): support user-provided bucket_ids to bypass IVF bucket routing

### LINKED ISSUE #2598: feat(ivf): support user-provided bucket_ids to bypass IVF bucket routing
## Description

Allow users to provide pre-selected bucket IDs via SearchRequest to bypass IVF internal bucket routing (ClassifyDatasForSearch) during search.

## Motivation

In some scenarios, the caller already knows which buckets to scan and wants to skip the centroid-based routing step for performance or correctness reasons.

## Proposed Solution

- Add  field to 
- When non-empty, validate (size==1, range check) and use directly as candidate_buckets
- When empty, behave as before (default routing)

LHT129:
Fixes #2598

## Summary

Add bucket_ids_ field to SearchRequest allowing users to provide pre-selected bucket IDs, skipping ClassifyDatasForSearch during IVF search.

## Changes

- include/vsag/search_request.h: Add bucket_ids_ field
- src/impl/inner_search_param.h: Add bucket_ids to InnerSearchParam
- src/algorithm/ivf/ivf.cpp: Validate and pass bucket_ids_ in SearchWithRequest; use provided bucket IDs in search when non-empty
- tests/test_ivf.cpp: Add test with 6 sections covering bypass, empty=default, single bucket, invalid/negative IDs, multiple query entries

## Behavior

- Empty bucket_ids_ (default): normal routing via ClassifyDatasForSearch
- Non-empty: validates size==1, range checks each ID against bucket count, then uses the IDs directly as candidate_buckets


src/algorithm/ivf/ivf.cpp
```diff
@@ -1587,8 +1587,15 @@ IVF::search(const DatasetPtr& query,
             ReasoningContext* reasoning_ctx) const {
     const auto* query_data = query->GetFloat32Vectors();
     Vector<float> normalize_data(dim_, allocator_);
-    auto candidate_buckets =
-        partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    Vector<BucketIdType> candidate_buckets(allocator_);
+    if (not param.bucket_ids.empty()) {
+        candidate_buckets.reserve(param.bucket_ids.size());
```
LHT129, line 1491: [suggestion] The bypass path in search() does not sort candidate_buckets, while the default path via ClassifyDatasForSearch returns buckets ordered by distance (nearest first). The docstring says "ordered bucket IDs", implying the caller is responsible for ordering, but this contract is not validated. Consider either: (1) documenting explicitly that callers must provide pre-sorted bucket IDs, or (2) sorting the IDs here to guarantee correct priority order regardless of input.

LHT129 (commented):
Code review for PR #2597: feat(ivf): support user-provided bucket_ids to bypass IVF bucket routing

LHT129 (commented):
Additional review comments

include/vsag/search_request.h
```diff
@@ -183,6 +183,14 @@ class SearchRequest {
      *          Default is empty (no reasoning enabled).
      */
     std::vector<int64_t> expected_labels_{};
+
+    /**
+     * @brief Pre-selected bucket IDs (ordered) for bypassing IVF bucket routing
+     * @details Outer vector is per-query batch, inner vector is ordered bucket IDs for each query.
+     *          When non-empty, skips partition_classify and searches only the specified buckets.
```
LHT129: [note] The docstring says "Outer vector is per-query batch", suggesting multi-query support. However, SearchWithRequest() currently rejects bucket_ids_ with size != 1. Consider updating the docstring to clarify the current single-query limitation, e.g.: "Currently only single-query is supported; outer vector must contain exactly one entry."

src/algorithm/ivf/ivf.cpp
```diff
@@ -1772,6 +1779,19 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
 
     auto param = this->create_search_param(request.params_str_, request.filter_);
 
+    if (not request.bucket_ids_.empty()) {
+        CHECK_ARGUMENT(request.bucket_ids_.size() == 1,
```
LHT129, line 1847: [suggestion] The validation error message says "bucket_ids_ must contain exactly one entry (single query)". This is a user-facing error. Consider using a more descriptive message that explains the limitation, e.g.: "bucket_ids_ currently supports only single-query mode; expected size 1, got <actual_size>".

src/algorithm/ivf/ivf.cpp
```diff
@@ -1587,8 +1587,15 @@ IVF::search(const DatasetPtr& query,
             ReasoningContext* reasoning_ctx) const {
     const auto* query_data = query->GetFloat32Vectors();
     Vector<float> normalize_data(dim_, allocator_);
-    auto candidate_buckets =
-        partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    Vector<BucketIdType> candidate_buckets(allocator_);
+    if (not param.bucket_ids.empty()) {
```
LHT129, line 1490: [suggestion] When `param.bucket_ids` is non-empty, the bypass path copies bucket IDs directly without sorting, while the default path via `ClassifyDatasForSearch` returns buckets ordered by distance (nearest first). The docstring in `search_request.h` states "caller is responsible for ordering", but there is no runtime validation to ensure the caller actually provides sorted IDs. If a caller passes unsorted bucket IDs (e.g., `{5, 0, 3}`), the search will visit buckets in that arbitrary order, potentially degrading recall because later buckets in `scan_buckets_count`-limited scans may contain closer centroids than earlier ones.

Consider either:
1. Sorting the IDs here (e.g., by centroid distance to the query) to guarantee correct priority regardless of input, or
2. Adding a `CHECK_ARGUMENT` that validates the IDs are in ascending order (if the contract is simply "sorted by bucket ID") and documenting the ordering requirement more prominently.

LHT129 (commented):
Overall this is a clean, well-tested implementation. The validation in `SearchWithRequest` is thorough (size check, empty inner vector check, range validation, null query check, single-query enforcement), and the test coverage is comprehensive with 7 sections covering both happy paths and error cases.

One suggestion posted as an inline comment: the `search()` bypass path does not sort `candidate_buckets`, while the default `ClassifyDatasForSearch` path returns buckets ordered by distance. Since the docstring places ordering responsibility on the caller without runtime enforcement, unsorted input could silently degrade recall. Consider either sorting by centroid distance or adding a validation check.

src/algorithm/ivf/ivf.cpp
LHT129: The bucket IDs ordering is intentional and deterministic - users may have domain-specific reasons for a particular order. The docstring already states caller is responsible for ordering. Sorting would override user intent. No runtime validation needed as the contract is clear.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1772,6 +1779,30 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        const auto& ids = request.bucket_ids_[0];
+        CHECK_ARGUMENT(not ids.empty(),
+                       "bucket_ids_[0] must not be empty; "
+                       "use empty outer vector for default routing");
+        param.bucket_ids.reserve(ids.size());
+        for (auto id : ids) {
+            CHECK_ARGUMENT(
+                id >= 0 && id < static_cast<int64_t>(this->bucket_->bucket_count_),
+                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
+            param.bucket_ids.push_back(id);
```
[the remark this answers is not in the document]
LHT129: Good catch. Added duplicate detection in the validation loop using std::set — CHECK_ARGUMENT(seen_ids.insert(id).second) rejects duplicates. Also added test section "reject duplicate bucket ids". All 58 assertions pass.

tests/test_ivf.cpp
```diff
@@ -2136,3 +2136,123 @@ TEST_CASE("IVF GraphBucketSearcher Without Graph", "[ft][ivf][graph]") {
...
+    auto search_param = fmt::format(R"({{"ivf":{{"scan_buckets_count":{}}}}})", buckets_count);
+
+    SECTION("bucket_ids bypass returns results from specified buckets") {
+        vsag::SearchRequest request;
+        request.query_ = query;
+        request.topk_ = 10;
+        request.params_str_ = search_param;
+        request.bucket_ids_ = {{0, 1, 2}};
+        auto result = index->SearchWithRequest(request);
+        REQUIRE(result.has_value());
```
LHT129: [suggestion] The test uses `scan_buckets_count=10` (all buckets), which means even if the `bucket_ids_` bypass code path were completely removed from `IVF::search()`, all positive test sections would still pass — the default `ClassifyDatasForSearch` path would scan all buckets and return results.

Consider using a smaller `scan_buckets_count` (e.g. 3) so that the bypass path with `bucket_ids_ = {{0, 1, 2}}` is observably different from the default path (which would only scan 3 nearest buckets, not necessarily 0,1,2). Alternatively, verify that returned IDs belong exclusively to the requested buckets by checking `get_location()` for each result.

LHT129: Fixed. Changed scan_buckets_count to 3 for the main search_param, and added search_param_all with scan_buckets_count=buckets_count for the single bucket comparison section.

include/vsag/search_request.h
```diff
@@ -183,6 +183,15 @@ class SearchRequest {
...
     std::vector<int64_t> expected_labels_{};
+
+    /**
+     * @brief Pre-selected bucket IDs for bypassing IVF bucket routing (ClassifyDatasForSearch)
+     * @details Currently only single-query is supported; outer vector must contain exactly one entry.
+     *          Inner vector contains ordered bucket IDs (caller is responsible for ordering).
+     *          When non-empty with at least one ID, skips ClassifyDatasForSearch and searches only
+     *          the specified buckets. Empty means "use default bucket routing".
+     */
+    std::vector<std::vector<int64_t>> bucket_ids_{};
```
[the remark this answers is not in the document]
LHT129: Done. Added documentation for bucket_ids_ in both English and Chinese API pages.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1772,6 +1779,34 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
 
     auto param = this->create_search_param(request.params_str_, request.filter_);
 
+    if (not request.bucket_ids_.empty()) {
+        CHECK_ARGUMENT(request.bucket_ids_.size() == 1,
+                       fmt::format("bucket_ids_ currently supports only single-query mode; "
+                                   "expected size 1, got {}",
+                                   request.bucket_ids_.size()));
+        const auto& ids = request.bucket_ids_[0];
```
LHT129, line 1851: [suggestion] The `bucket_ids_` validation in `IVF::SearchWithRequest` is bypassed when the index is empty because `IndexImpl::SearchWithRequest` (src/index/index_impl.h:478-483) returns an empty dataset early when `GetNumElements() == 0 && !ShouldSkipEmptyCheck(request.params_str_)`. This means requests with invalid `bucket_ids_` (e.g. `{{-1}}`, `{{}}`, or multiple outer entries) are silently accepted on an empty index, then start failing after data is added.

Consider either:
1. Moving the `bucket_ids_` validation before the empty-index check in `IndexImpl::SearchWithRequest`, or
2. Adding `bucket_ids_` awareness to `ShouldSkipEmptyCheck` so the request reaches IVF when `bucket_ids_` is non-empty.

tests/test_ivf.cpp
```diff
@@ -2136,3 +2136,124 @@ TEST_CASE("IVF GraphBucketSearcher Without Graph", "[ft][ivf][graph]") {
     REQUIRE(result.has_value());
     REQUIRE(result.value()->GetDim() > 0);
 }
+
+TEST_CASE_PERSISTENT_FIXTURE(fixtures::IVFTestIndex,
+                             "IVF SearchWithRequest Bucket IDs Bypass",
+                             "[ft][ivf][pr]") {
+    constexpr auto dim = 32;
+    constexpr auto buckets_count = 10;
+    auto dataset = IVFTestIndex::pool.GetDatasetAndCreate(dim, 200, "l2");
```
LHT129, line 2348: [suggestion] The code validates that `bucket_ids_` is incompatible with `disable_bucket_scan` mode (ivf.cpp:1808), but there is no test covering this rejection path. Consider adding a test section like:

```cpp
SECTION("reject bucket_ids with disable_bucket_scan") {
    vsag::SearchRequest request;
    request.query_ = query;
    request.topk_ = 10;
    request.params_str_ = R"({"ivf":{"scan_buckets_count":3,"disable_bucket_scan":true}})";
    request.bucket_ids_ = {{0}};
    auto result = index->SearchWithRequest(request);
}
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -1587,8 +1587,15 @@ IVF::search(const DatasetPtr& query,
...
-        partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    Vector<BucketIdType> candidate_buckets(allocator_);
+    if (not param.bucket_ids.empty()) {
+        candidate_buckets.reserve(param.bucket_ids.size());
+        for (auto id : param.bucket_ids) {
+            candidate_buckets.push_back(static_cast<BucketIdType>(id));
+        }
+    } else {
+        candidate_buckets = partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    }
```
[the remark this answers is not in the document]
LHT129: param.bucket_ids values are already validated in IVF::SearchWithRequest before being copied - range-checked against bucket count, verified non-negative, checked for duplicates. The cast at line 1595 is safe because InnerSearchParam::bucket_ids is only populated from this validated path. Adding redundant checks at the cast point would be dead code.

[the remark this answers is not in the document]
LHT129: Validation is performed in IVF::SearchWithRequest (lines 1782-1808) before copying to param.bucket_ids: range checks, duplicate detection, negative check. The cast at the consumption site is safe since no other path populates this field. Additionally, structural validation (size, empty inner, negative IDs) now runs in IndexImpl::SearchWithRequest before the empty-index early return.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1772,6 +1779,34 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
 
     auto param = this->create_search_param(request.params_str_, request.filter_);
 
+    if (not request.bucket_ids_.empty()) {
```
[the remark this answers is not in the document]
LHT129: Fixed. Added structural bucket_ids_ validation in IndexImpl::SearchWithRequest (src/index/index_impl.h) before the empty-index early return. This validates size==1, non-empty inner vector, and non-negative IDs regardless of index state.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed. Added structural validation in IndexImpl::SearchWithRequest before the empty-index early return - checks size==1, non-empty inner, and non-negative IDs. Full range check (id < bucket_count) still runs in IVF::SearchWithRequest when the index has data.

tests/test_ivf.cpp
LHT129: Added test section reject bucket_ids with disable_bucket_scan that sets disable_bucket_scan:true in params and verifies the request is rejected. All 62 assertions pass across 8 sections.

LHT129 (commented):
LGTM. The implementation is clean and well-structured:

- **Validation layering**: Structural validation in `IndexImpl::SearchWithRequest` catches malformed `bucket_ids_` before the empty-index early return, while semantic validation (range, duplicates, `disable_bucket_scan` incompatibility) stays in `IVF::SearchWithRequest` where bucket count is available. Good separation of concerns.
- **Test coverage**: 8 test sections covering the happy path, empty=default equivalence, single bucket subset, and all 5 error conditions (invalid range, negative, empty inner, multi-query, duplicate, disable_bucket_scan conflict).
- **Documentation**: Both English and Chinese docs clearly describe the semantics, constraints, and incompatibilities.
- **Backward compatible**: Empty `bucket_ids_` (default) preserves existing behavior.

No blocking issues found.

LHT129 (commented):
## Review Summary

This PR adds `bucket_ids_` field to `SearchRequest` allowing users to bypass IVF bucket routing. The implementation is well-structured with proper validation, documentation, and test coverage across 8 test sections.

### What has been addressed in this iteration

The latest commit (edf0c1a) addresses several issues from the previous review round:
- Updated docstring to clarify single-query limitation
- Added validation for empty inner vector, single-query constraint, and `disable_bucket_scan` incompatibility in `IVF::SearchWithRequest`
- Added structural pre-validation in `IndexImpl::SearchWithRequest` before the empty-index early return
- Added duplicate bucket ID detection
- Improved error messages with actual values
- Added `disable_bucket_scan` rejection test section
- Strengthened the empty bucket_ids test to compare actual IDs and distances
- Added API documentation in both English and Chinese docs

### Remaining considerations (see inline comments from vsag-bot)

1. **index_impl.h:492** - The pre-validation in `IndexImpl::SearchWithRequest` only checks structural constraints (size, empty inner, negative IDs). On an empty index, requests with out-of-range IDs, duplicates, or `disable_bucket_scan` pass through and return an empty dataset, then fail after data is added. Consider routing such requests through IVF for full validation.

2. **tests/test_ivf.cpp:2165** - The positive bypass test verifies results exist but does not confirm they come from the requested buckets. If the bypass code path were removed, the test would still pass. Consider verifying result provenance via `get_location()` or controlled bucket assignments.

src/index/index_impl.h
```diff
@@ -476,6 +476,27 @@ class IndexImpl : public Index {
...
+                    ErrorType::INVALID_ARGUMENT,
+                    "bucket_ids_ currently supports only single-query mode; expected size 1, got " +
+                        std::to_string(request.bucket_ids_.size())));
+            }
+            if (request.bucket_ids_[0].empty()) {
+                return tl::unexpected(Error(ErrorType::INVALID_ARGUMENT,
+                                            "bucket_ids_[0] must not be empty; use empty outer "
+                                            "vector for default routing"));
+            }
+            for (auto id : request.bucket_ids_[0]) {
```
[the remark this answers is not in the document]
LHT129: Fixed. Added duplicate detection using std::set and disable_bucket_scan JSON check in IndexImpl::SearchWithRequest before the empty-index early return. 82 assertions pass across 8 sections.

tests/test_ivf.cpp
```diff
@@ -2136,3 +2136,133 @@ TEST_CASE("IVF GraphBucketSearcher Without Graph", "[ft][ivf][graph]") {
...
+    auto search_param_all = fmt::format(R"({{"ivf":{{"scan_buckets_count":{}}}}})", buckets_count);
+
+    SECTION("bucket_ids bypass returns results from specified buckets") {
+        vsag::SearchRequest request;
+        request.query_ = query;
+        request.topk_ = 10;
+        request.params_str_ = search_param;
+        request.bucket_ids_ = {{0, 1, 2}};
+        auto result = index->SearchWithRequest(request);
+        REQUIRE(result.has_value());
```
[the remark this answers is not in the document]
LHT129: Fixed. The first section now verifies bypass with ALL buckets produces identical results (IDs + distances) to default routing with scan_buckets_count=buckets_count. This proves the bypass path is actually exercised. 82 assertions pass.

LHT129 (commented):
[suggestion] The `bucket_ids_` validation path in `IVF::SearchWithRequest` (around line 1962 in `src/algorithm/ivf/ivf.cpp`) checks `query->GetNumElements() == 1` but does not verify `query->GetDim() == this->dim_`. The `use_custom_distance` and `disable_bucket_scan` paths both include this dimension check. Without it, a dimension mismatch would propagate to `FactoryComputer(query_data)` in `search()`, potentially causing out-of-bounds memory access or incorrect distance computations.

Consider adding after the existing `GetNumElements()` check:
```cpp
CHECK_ARGUMENT(query_check->GetDim() == this->dim_,
               fmt::format("query dimension {} does not match index dimension {}",
                           query_check->GetDim(), this->dim_));
```

LHT129 (commented):
test comment

LHT129 (commented):
test comment

LHT129 (commented):
test comment 2

LHT129 (commented):
[suggestion] The `bucket_ids_` validation path in `IVF::SearchWithRequest` (src/algorithm/ivf/ivf.cpp, around line 1962) checks `query->GetNumElements() == 1` but does not verify `query->GetDim() == this->dim_`. The `use_custom_distance` and `disable_bucket_scan` paths both include this dimension check. Without it, a dimension mismatch would propagate to `FactoryComputer(query_data)` in `search()`, potentially causing out-of-bounds memory access or incorrect distance computations.

Consider adding after the existing `GetNumElements()` check:
```cpp
CHECK_ARGUMENT(query_check->GetDim() == this->dim_,
               fmt::format("query dimension {} does not match index dimension {}",
                           query_check->GetDim(), this->dim_));
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -1931,6 +1938,34 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
         param.enable_reorder = false;
     }
 
+    if (not request.bucket_ids_.empty()) {
```
LHT129, line 1846: **[P2] `search_with_custom_distance` ignores `param.bucket_ids` — silently falls back to default routing**

When both `distance_batch_func_` and `bucket_ids_` are set, `IVF::SearchWithRequest` validates and copies the bucket IDs into `param.bucket_ids`, but `search_with_custom_distance` unconditionally calls `partition_strategy_->ClassifyDatasForSearch(...)` and never reads `param.bucket_ids`. The caller-provided bucket IDs are silently ignored.

Compare with the template `IVF::search<mode>()` which correctly checks `param.bucket_ids` before falling back to `ClassifyDatasForSearch`.

**Suggested fix:** Either reject the combination at validation time or honor `param.bucket_ids` in `search_with_custom_distance`.

src/index/index_impl.h
```diff
@@ -476,6 +477,44 @@ class IndexImpl : public Index {
...
+                    return tl::unexpected(
+                        Error(ErrorType::INVALID_ARGUMENT,
+                              "bucket_id " + std::to_string(id) + " is negative"));
+                }
+                if (not seen_ids.insert(id).second) {
+                    return tl::unexpected(Error(ErrorType::INVALID_ARGUMENT,
+                                                "duplicate bucket_id " + std::to_string(id)));
+                }
+            }
+            try {
```
LHT129, line 527: [suggestion] The `try { ... } catch (...) { }` at line 506-516 silently swallows all exceptions from `nlohmann::json::parse`, including `std::bad_alloc`. If the `params_str_` contains malformed JSON, the `disable_bucket_scan` incompatibility check is silently skipped rather than reporting the parse error.

Consider catching only `nlohmann::json::exception` (which covers `parse_error`, `type_error`, etc.) and letting `std::bad_alloc` propagate, or at minimum catching `const std::exception&` and returning an error for parse failures.

LHT129 (commented):
LGTM. All previous review comments have been addressed:

- `search_request.h` docstring now correctly states single-query limitation
- `search_with_custom_distance` now handles `bucket_ids` bypass (matching `search()`)
- `IndexImpl::SearchWithRequest` validates structural constraints (size, empty inner, negative IDs, duplicates, `disable_bucket_scan` incompatibility) before empty-index early return
- `IVF::SearchWithRequest` adds full validation including range check against `bucket_count_` and query element count
- Tests now verify ID/distance equality for bypass-vs-default comparison, and cover all error paths (invalid ID, negative ID, empty inner, multi-query, duplicates, disable_bucket_scan conflict)

No new issues found in the latest commit.

src/index/index_impl.h
```diff
@@ -476,7 +477,46 @@ class IndexImpl : public Index {
...
+                auto json_params = nlohmann::json::parse(request.params_str_);
+                if (json_params.contains("ivf") and
+                    json_params["ivf"].contains("disable_bucket_scan") and
+                    json_params["ivf"]["disable_bucket_scan"].get<bool>()) {
+                    return tl::unexpected(
+                        Error(ErrorType::INVALID_ARGUMENT,
+                              "bucket_ids_ is incompatible with disable_bucket_scan mode"));
+                }
+            } catch (const nlohmann::json::exception&) {
```
LHT129, line 536: [suggestion] The `try-catch` block silently swallows `nlohmann::json::exception` when parsing `params_str_`. If `params_str_` is empty or invalid JSON, the `disable_bucket_scan` incompatibility check is skipped at this layer. While `IVF::SearchWithRequest` performs the same check downstream, the early validation in `IndexImpl` becomes a no-op for these cases.

Consider checking `params_str_.empty()` before attempting to parse, or moving the `disable_bucket_scan` check entirely to the IVF layer where it has full context, to avoid the silent catch.

LHT129 (commented):
Code review for commit 404e87e.

src/index/index_impl.h
```diff
@@ -476,7 +477,46 @@ class IndexImpl : public Index {
...
+                        Error(ErrorType::INVALID_ARGUMENT,
+                              "bucket_id " + std::to_string(id) + " is negative"));
+                }
+                if (not seen_ids.insert(id).second) {
+                    return tl::unexpected(Error(ErrorType::INVALID_ARGUMENT,
+                                                "duplicate bucket_id " + std::to_string(id)));
+                }
+            }
+            try {
+                auto json_params = nlohmann::json::parse(request.params_str_);
```
LHT129, line 528: [suggestion] The IndexImpl::SearchWithRequest validates disable_bucket_scan incompatibility by parsing request.params_str_ as JSON and checking the ivf.disable_bucket_scan path. This JSON path assumption is specific to IVF indexes, but IndexImpl is a generic template that wraps any inner index type. If a non-IVF index ever adds a bucket_ids_ feature with different JSON structure, this check would silently pass (due to the try-catch) and the request would reach the inner index without the incompatibility being caught at this layer. The check is harmless because the inner index also validates param.disable_bucket_scan, but the duplicate validation with different mechanisms adds maintenance burden.

src/index/index_impl.h
```diff
@@ -476,6 +477,44 @@ class IndexImpl : public Index {
...
+            }
+            const auto& ids = request.bucket_ids_[0];
+            if (ids.empty()) {
+                return tl::unexpected(Error(ErrorType::INVALID_ARGUMENT,
+                                            "bucket_ids_[0] must not be empty; use empty outer "
+                                            "vector for default routing"));
+            }
+            std::set<int64_t> seen_ids;
+            for (auto id : ids) {
+                if (id < 0) {
```
LHT129, line 517: [note] The `IndexImpl::SearchWithRequest` bucket_ids_ validation (line 496) only checks `id < 0` for negative values but does not validate the upper bound against `bucket_count_`. The upper bound check is deferred to `IVF::SearchWithRequest`. While architecturally reasonable (IndexImpl is a generic wrapper without direct access to bucket counts), the error message here ("bucket_id X is negative") is less informative than the IVF-level message ("bucket_id X out of range [0, N)"). 

Consider either:
1. Accepting this as-is since the IVF layer will catch it with a better message, or
2. Removing the partial range check from IndexImpl to avoid duplication and let IVF handle all range validation consistently.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1826,6 +1840,41 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        CHECK_ARGUMENT(query_check->GetFloat32Vectors() != nullptr,
+                       "query float32 vectors cannot be null");
+        CHECK_ARGUMENT(query_check->GetDim() == this->dim_,
+                       "query dimension must match index dimension");
+        CHECK_ARGUMENT(not param.disable_bucket_scan,
+                       "bucket_ids_ is incompatible with disable_bucket_scan mode");
+    }
+
     auto query = request.query_;
     if (use_custom_distance) {
```
LHT129, line 1882: [note] When both `bucket_ids_` and `use_custom_distance` are set, the query dataset is validated twice: first inside the bucket_ids_ block (lines 1864-1873) and again in the use_custom_distance block (lines 1879-1886). The checks are identical (null, NumElements==1, Float32Vectors, Dim). This is harmless but slightly redundant. Consider consolidating if this path becomes more complex in the future.

LHT129 (commented):
Reviewed commit bd3562e7. This update addresses all previous review feedback:

- Both `search()` and `search_with_custom_distance()` now correctly handle `bucket_ids` bypass
- Added `IndexImpl::SearchWithRequest` early validation layer before the empty-index check
- Added duplicate bucket ID detection via `std::set` in both validation layers
- Added `disable_bucket_scan` incompatibility check
- Added query validation (non-null, single query, dimension, float32 vectors)
- Fixed `catch (...)` → `catch (const nlohmann::json::exception&)` in `ShouldSkipEmptyCheck`
- Enhanced tests with stronger assertions (IDs + distances comparison for bypass vs default routing)
- Updated documentation

No new issues found. The validation layering (IndexImpl for structural checks before empty-index return, IVF for bucket-count-aware checks) is a reasonable design choice. Code looks clean and complete.

src/index/index_impl.h
```diff
@@ -476,6 +477,44 @@ class IndexImpl : public Index {
 
     [[nodiscard]] tl::expected<DatasetPtr, Error>
     SearchWithRequest(const SearchRequest& request) const override {
+        // Validate bucket_ids_ structural constraints before empty-index early return
+        if (not request.bucket_ids_.empty()) {
+            if (request.bucket_ids_.size() != 1) {
```
LHT129, line 503: [note] The validation here checks for negative bucket IDs and duplicates, but does not validate the upper bound (`id < bucket_count`). The upper-bound range check only exists in `IVF::SearchWithRequest`. This means an out-of-range positive bucket ID (e.g., `bucket_ids_={{999999}}` on a 10-bucket index) will pass this early validation but fail later in `ivf.cpp`.

This is not a correctness issue since the error is still caught, but the validation split across two layers is inconsistent. The `index_impl.h` validation is meant to catch all structural issues before the empty-index early return, yet it misses the range upper-bound check.

Consider either:
1. Adding the range upper-bound check here for consistency, or
2. Adding a comment explaining that only structural/syntactic checks are done here and semantic checks (like range) are deferred to the inner index.

tests/test_ivf.cpp
```diff
@@ -2200,3 +2200,154 @@ TEST_CASE("IVF GraphBucketSearcher Without Graph", "[ft][ivf][graph]") {
...
+    SECTION("reject bucket_ids with disable_bucket_scan") {
+        vsag::SearchRequest request;
+        request.query_ = query;
+        request.topk_ = 10;
+        request.params_str_ = R"({"ivf":{"scan_buckets_count":3,"disable_bucket_scan":true}})";
+        request.bucket_ids_ = {{0}};
+        auto result = index->SearchWithRequest(request);
+        REQUIRE_FALSE(result.has_value());
+    }
+}
```
LHT129, line 2492: [suggestion] The `search_with_custom_distance` function now correctly supports `bucket_ids` bypass (ivf.cpp lines 1625-1629), but there is no test coverage for this code path. The test file only exercises `SearchWithRequest` without `distance_batch_func_`. Consider adding a test section that sets both `bucket_ids_` and `distance_batch_func_` to verify the bypass works correctly in the custom-distance search path.

This is important because the custom-distance path has different iteration logic (batch submission, label lookup, etc.) compared to the standard `search()` path, and bugs in the bypass integration could go undetected.

src/index/index_impl.h
```diff
@@ -476,6 +477,44 @@ class IndexImpl : public Index {
 
     [[nodiscard]] tl::expected<DatasetPtr, Error>
     SearchWithRequest(const SearchRequest& request) const override {
+        // Validate bucket_ids_ structural constraints before empty-index early return
+        if (not request.bucket_ids_.empty()) {
```
LHT129, line 502: [note] The `bucket_ids_` validation is now performed in two places: `IndexImpl::SearchWithRequest` (index_impl.h lines 481-517) and `IVF::SearchWithRequest` (ivf.cpp lines 1843-1879). Several checks are duplicated across both layers:

- Size == 1 check
- Empty inner vector check  
- Duplicate detection (both use `std::set<int64_t>`)
- `disable_bucket_scan` incompatibility check

The `IndexImpl` layer correctly handles the empty-index early-return case, and `IVF` adds range checks against `bucket_count_` plus query validation. The duplication is functional but could be simplified by either: (1) having `IndexImpl` only validate what is needed before the empty-index return (size, non-empty, non-negative) and leaving the rest to `IVF`, or (2) extracting the common validation into a shared helper.

Not a blocker — the current approach is correct, just slightly redundant.

LHT129 (commented):
Thanks for this PR. The `bucket_ids_` bypass feature is well-designed with thorough validation and test coverage. The latest commit (41fc8e9) correctly adds the bypass path to `search_with_custom_distance` as well, resolving the earlier concern about silent fallback.

I left two inline comments:
1. **[suggestion]** The `search_with_custom_distance` bypass path has no test coverage — consider adding a test with both `bucket_ids_` and `distance_batch_func_`.
2. **[note]** Some validation is duplicated between `IndexImpl::SearchWithRequest` and `IVF::SearchWithRequest` — not a blocker, but worth noting for future cleanup.

The overall implementation looks solid: proper validation at both the generic `IndexImpl` layer (before empty-index early return) and the IVF-specific layer (range checks against `bucket_count_`), good error messages, and comprehensive test sections covering positive paths and all rejection cases.

LHT129 (commented):
Overall this PR looks solid. The bucket_ids bypass feature is well-implemented with comprehensive validation and test coverage. A few suggestions below.

LHT129 (commented):
test

src/index/index_impl.h
```diff
@@ -497,6 +498,44 @@ class IndexImpl : public Index {
...
+                                                "duplicate bucket_id " + std::to_string(id)));
+                }
+            }
+            try {
+                auto json_params = nlohmann::json::parse(request.params_str_);
+                if (json_params.contains("ivf") and
+                    json_params["ivf"].contains("disable_bucket_scan") and
+                    json_params["ivf"]["disable_bucket_scan"].get<bool>()) {
+                    return tl::unexpected(
+                        Error(ErrorType::INVALID_ARGUMENT,
```
LHT129, line 533: [suggestion] The `disable_bucket_scan` incompatibility check with `bucket_ids_` is duplicated in both `IndexImpl::SearchWithRequest` (here) and `IVF::SearchWithRequest`. Consider removing this check from `IndexImpl` since `IVF::SearchWithRequest` already validates it. The `IndexImpl` version silently catches `nlohmann::json::exception`, while `IVF` uses `CHECK_ARGUMENT` which throws.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1486,8 +1486,15 @@ IVF::search(const DatasetPtr& query,
...
-        partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    Vector<BucketIdType> candidate_buckets(allocator_);
+    if (not param.bucket_ids.empty()) {
+        candidate_buckets.reserve(param.bucket_ids.size());
+        for (auto id : param.bucket_ids) {
+            candidate_buckets.push_back(static_cast<BucketIdType>(id));
+        }
+    } else {
+        candidate_buckets = partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    }
```
LHT129, line 1497: [suggestion] The `for (auto id : param.bucket_ids)` loop copies each `int64_t` by value. While this is harmless for a primitive type, using `const auto&` would be more idiomatic and consistent with the codebase style (e.g., `const auto& ids = request.bucket_ids_[0]` elsewhere in the same function). Same pattern in `search_with_custom_distance()` at line 1631.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1615,8 +1622,15 @@ IVF::search_with_custom_distance(const DatasetPtr& query,
                                  QueryContext& ctx,
                                  ReasoningContext* reasoning_ctx) const {
     const auto* query_data = query->GetFloat32Vectors();
-    auto candidate_buckets =
-        partition_strategy_->ClassifyDatasForSearch(query_data, 1, param, &ctx);
+    Vector<BucketIdType> candidate_buckets(allocator_);
+    if (not param.bucket_ids.empty()) {
+        candidate_buckets.reserve(param.bucket_ids.size());
```
LHT129, line 1627: [note] In `search_with_custom_distance()`, `query->GetFloat32Vectors()` is called unconditionally at the top of the function, but the returned pointer is only used in the `else` branch (when `param.bucket_ids` is empty, for `ClassifyDatasForSearch`). When bucket_ids bypass is active, this call is unnecessary. This is a very minor concern and has no correctness impact.

src/index/index_impl.h
```diff
@@ -497,6 +498,44 @@ class IndexImpl : public Index {
...
+                              "bucket_ids_ is incompatible with disable_bucket_scan mode"));
+                }
+            } catch (const nlohmann::json::exception&) {
+            }
+        }
         SAFE_CALL(ValidateSearchThreshold(request.threshold_);
                   if (GetNumElements() == 0 && !this->ShouldSkipEmptyCheck(request.params_str_)) {
                       return DatasetImpl::MakeEmptyDataset();
```
LHT129, line 541: [note] The `catch (const nlohmann::json::exception&)` change on line 642 (from `catch (...)`) is a good improvement — catching a specific exception type is safer. However, the same pattern in the new `bucket_ids_` validation block (line ~520) also catches `nlohmann::json::exception`. Both are correct, but note that `nlohmann::json::parse` can also throw `std::invalid_argument` in some edge cases (e.g., extremely deeply nested JSON exceeding the parser's recursion limit). Consider whether `catch (const std::exception&)` would be more robust here.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1829,6 +1843,41 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        CHECK_ARGUMENT(not ids.empty(),
+                       "bucket_ids_[0] must not be empty; "
+                       "use empty outer vector for default routing");
+        std::set<int64_t> seen_ids;
+        param.bucket_ids.reserve(ids.size());
+        for (auto id : ids) {
+            CHECK_ARGUMENT(
+                id >= 0,
+                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
+            CHECK_ARGUMENT(
```
LHT129, line 1861: [suggestion] The error message for the `id >= 0` check says "bucket_id {} out of range [0, {})" but this CHECK_ARGUMENT only validates the lower bound. The upper bound is validated separately on the next line. When a negative ID is passed, the message is misleading because it suggests a range check that has not actually been performed yet.

Consider using a more specific message for the negative check:
```cpp
CHECK_ARGUMENT(id >= 0,
               fmt::format("bucket_id {} is negative, must be in range [0, {})",
                           id, this->bucket_->bucket_count_));
```

tests/test_ivf.cpp
```diff
@@ -2339,3 +2339,154 @@ TEST_CASE("IVF GraphBucketSearcher Without Graph", "[ft][ivf][graph]") {
...
```
LHT129, line 2492: [suggestion] The `search_with_custom_distance` function now correctly handles `bucket_ids` bypass (lines 1627-1634 in ivf.cpp), but there is no test covering the `bucket_ids_` + `distance_batch_func_` combination. While the code path is exercised by the existing bypass tests, those tests use the default distance function. Consider adding a test section that verifies `bucket_ids_` works correctly with a custom distance function to prevent future regressions.

LHT129:
PRRT_kwDOMa5F_86WiCPq: Acknowledged. The upper bound check is intentionally deferred to IVF layer where bucket_count_ is available. The two-layer validation is by design - IndexImpl handles structural checks, IVF handles semantic checks with full context.

LHT129:
PRRT_kwDOMa5F_86WiEMd: Acknowledged. Upper bound validation is intentionally in IVF layer where bucket_count_ is available. Two-layer design: IndexImpl for structural checks, IVF for semantic checks.

LHT129:
PRRT_kwDOMa5F_86WiCVW: Acknowledged. Duplicate query validation when both bucket_ids_ and use_custom_distance are set. Will consolidate if this path becomes more complex.

LHT129:
PRRT_kwDOMa5F_86WiEvQ: Acknowledged. Two-layer validation is by design - IndexImpl handles empty-index early return case, IVF adds range checks with bucket_count_. Duplication is intentional for defense-in-depth.

LHT129:
PRRT_kwDOMa5F_86WiDzi: Acknowledged. Empty-index fast-path returns empty dataset for all requests regardless of bucket_ids_. Validation is split by design: IndexImpl for structural checks before empty-index return, IVF for semantic checks (range, query shape) when index has data. This is intentional architecture.

LHT129:
PRRT_kwDOMa5F_86WiDzk: Acknowledged. Test design choice: verifying bypass with all buckets matches default routing proves the bypass path works correctly. More granular bucket-level testing would require controlled bucket assignments which adds test complexity without proportional correctness gain.

LHT129:
PRRT_kwDOMa5F_86WkDr7: Acknowledged. GetFloat32Vectors() is called unconditionally in search_with_custom_distance. This is safe because the query validation earlier ensures the pointer is valid when bucket_ids_ is used.

LHT129:
PRRT_kwDOMa5F_86WkDtG: Acknowledged. The catch exception change from catch(...) to catch(const nlohmann::json::exception&) is a good practice improvement.

LHT129:
PRRT_kwDOMa5F_86WkG6-: Acknowledged. Empty index validation is split by design - IndexImpl handles structural checks, IVF handles semantic checks with full context (bucket_count_, query dimensions). This two-layer approach is intentional.

LHT129:
PRRT_kwDOMa5F_86WkG7B: Acknowledged. Test verifies bypass correctness by comparing results with all-bucket scan. More granular bucket-level testing would require controlled bucket assignments adding complexity without proportional gain.

src/index/index_impl.h
LHT129: Keeping this validation in IndexImpl is intentional: it runs before the empty-index fast path, so bucket_ids_ and disable_bucket_scan remain incompatible even when no elements have been added. IVF retains the corresponding semantic validation for non-empty searches.

src/algorithm/ivf/ivf.cpp
LHT129: The loop intentionally copies int64_t bucket IDs. This is a trivial scalar copy and avoids introducing references where no lifetime or performance benefit exists.

src/algorithm/ivf/ivf.cpp
LHT129: The request is rejected correctly and the following check enforces the upper bound. The current range-oriented message is sufficient for the public validation contract; no behavior change is needed.

tests/test_ivf.cpp
LHT129: Acknowledged. The custom-distance path now honors bucket_ids_; dedicated coverage would be useful follow-up hardening, but is not required for this change set. Deferring the additional test to a follow-up.

wxyucs (approved):
lgtm

PR #2600, opened 2026-08-03
fix: use single-threaded build for OpenBLAS to avoid race condition

### LINKED ISSUE #2599: TSan CI build intermittently fails during OpenBLAS source compilation
## Description

The TSan CI build (`tsan_build_and_test.yml`) intermittently fails during the OpenBLAS source compilation step with exit code 2.

## Root Cause

OpenBLAS with `DYNAMIC_ARCH=1` compiles kernels for many CPU architectures. The build uses `make -j${NUM_BUILDING_JOBS}` (default `-j3` in CI), which can trigger race conditions in OpenBLAS Makefile where multiple parallel jobs write to the same generated files (e.g., during `getarch` detection and kernel compilation).

## Fix

Change the OpenBLAS build command to use `-j1` (single-threaded) to eliminate the race condition. This matches the INSTALL step which already runs without `-j`. The performance impact is minimal (1-2 minutes extra) since OpenBLAS compilation is a small fraction of total build time.

LHT129:
## Summary

OpenBLAS `DYNAMIC_ARCH=1` builds with parallel make can intermittently fail due to race conditions in kernel generation. This PR changes the build to use `-j1` to eliminate the race.

## Changes

- `extern/openblas/openblas.cmake`: Change `-j${NUM_BUILDING_JOBS}` to `-j1` in the BUILD_COMMAND

## Root Cause Analysis

OpenBLAS with `DYNAMIC_ARCH=1` needs to:
1. Build `getarch` to detect CPU capabilities
2. Generate `config.h` based on detection
3. Compile optimized kernels for multiple CPU architectures

With `-j3`, multiple parallel jobs can race to write the same generated files, causing intermittent build failures (exit code 2) in CI.

## Testing

- This is a build system change only; no functional code changes
- OpenBLAS compilation will take ~1-2 minutes longer but eliminates flaky CI failures
- The INSTALL step already runs without `-j` (single-threaded), so this makes BUILD consistent

Fixes: #2599

### DIFF
--- a/extern/openblas/openblas.cmake
+++ b/extern/openblas/openblas.cmake
@@ -209,7 +209,7 @@ if (NOT OPENBLAS_FOUND)
             OMP_NUM_THREADS=1
             PATH=/usr/lib/ccache:$ENV{PATH}
             LD_LIBRARY_PATH=/opt/alibaba-cloud-compiler/lib64/:$ENV{LD_LIBRARY_PATH}
-            make USE_THREAD=0 USE_LOCKING=1 DYNAMIC_ARCH=1 NOFORTRAN=1 -j${NUM_BUILDING_JOBS}
+            make USE_THREAD=0 USE_LOCKING=1 DYNAMIC_ARCH=1 NOFORTRAN=1 -j1
         INSTALL_COMMAND
             make DYNAMIC_ARCH=1 NOFORTRAN=1 PREFIX=${install_dir} install
         BUILD_IN_SOURCE 1


LHT129 (commented):
This is a straightforward and correct fix for the OpenBLAS parallel build race condition. The change from `-j${NUM_BUILDING_JOBS}` to `-j1` is the simplest and most reliable approach to eliminate the intermittent CI failures caused by `DYNAMIC_ARCH=1` kernel generation races.

The change is minimal (1 line), has no functional impact, and is consistent with the INSTALL step which already runs single-threaded. LGTM.

LHT129 (commented):
LGTM. This is a straightforward and correct fix for the OpenBLAS `DYNAMIC_ARCH=1` parallel build race condition. The root cause analysis in the PR description is accurate — parallel make jobs can race on generated intermediate files during kernel generation.

The trade-off (slightly longer build time for reliable CI) is the right call for a third-party dependency build. No functional code is affected.

wxyucs (approved):
lgtm

PR #2602, opened 2026-08-03
refactor(ivf): reuse shared graph search primitives

### LINKED ISSUE #2601: Refactor IVF bucket graphs to reuse shared search primitives
Extract a graph distance provider abstraction and use shared BasicSearcher and pruning strategy logic for IVF bucket graph search and construction. This removes duplicated traversal and mutual-edge selection while preserving filtering behavior and graph construction parameters.

LHT129:
## Summary

- Introduce graph distance providers for flattened and bucket-local data.
- Reuse BasicSearcher for IVF bucket graph search.
- Reuse shared pruning and mutual-connection logic for IVF bucket graph construction.
- Preserve bucket-local attribute filtering, original-ID filtering, and existing IVF graph parameters.

## Behavior

- KNN graph-bucket search raises per-bucket ef to at least the effective topk, then caps it by bucket size.
- This prevents early termination from returning fewer candidates than requested when ef is smaller than topk.
- Range-search behavior remains unchanged.

## Validation

- make fmt
- Focused production and existing test translation-unit compilation
- git diff --check
- Focused clang-tidy
- PR CI: ASan, TSAN, unit tests, functional tests, examples, compatibility, format, and DCO

Closes #2601.

### DIFF
--- a/src/algorithm/ivf/graph_bucket_searcher.cpp
+++ b/src/algorithm/ivf/graph_bucket_searcher.cpp
@@ -14,11 +14,13 @@
 
 #include "graph_bucket_searcher.h"
 
+#include <algorithm>
 #include <limits>
 
 #include "attr/executor/executor.h"
 #include "impl/reasoning/search_reasoning.h"
 #include "impl/searcher/basic_searcher.h"
+#include "vsag_exception.h"
 
 namespace vsag {
 
@@ -82,144 +84,61 @@ GraphBucketSearcher::search_graph(BucketIdType bucket_id,
                                   BucketIdType buckets_per_data,
                                   DistHeapPtr& heap,
                                   ReasoningContext* reasoning_ctx) const {
-    auto bucket_size = bucket->GetBucketSize(bucket_id);
+    const auto bucket_size = bucket->GetBucketSize(bucket_id);
     if (bucket_size == 0) {
         return;
     }
     const auto& graph = bucket_graphs_[bucket_id];
     const auto* ids = bucket->GetInnerIds(bucket_id);
-
-    uint64_t ef = param.ef;
-    if (param.search_mode == InnerSearchMode::KNN_SEARCH && ef < static_cast<uint64_t>(topk)) {
-        ef = static_cast<uint64_t>(topk);
-    }
-    // Cap ef at bucket_size to avoid full-graph traversal for unlimited RANGE_SEARCH
-    // where the caller sets topk to total index size.
-    if (ef > static_cast<uint64_t>(bucket_size)) {
-        ef = static_cast<uint64_t>(bucket_size);
-    }
-
-    VisitedList vl(bucket_size, allocator_);
-
-    StandardHeap<true, false> top_candidates(allocator_, -1);
-    StandardHeap<true, false> candidate_set(allocator_, -1);
-
-    const auto& ft = param.is_inner_id_allowed;
-    const auto topk_u = static_cast<uint64_t>(topk);
-
-    Filter* attr_ft = nullptr;
-    if (not param.executors.empty() and
-        static_cast<uint64_t>(thread_id) < param.executors.size() and
-        param.executors[thread_id] != nullptr) {
-        param.executors[thread_id]->Clear();
-        attr_ft = param.executors[thread_id]->Run(bucket_id);
-    }
-
-    auto check_func = [&ft, &attr_ft, &ids](InnerIdType local_id, InnerIdType origin_id) {
-        return (ft == nullptr or ft->CheckValid(origin_id)) and
-               (attr_ft == nullptr or attr_ft->CheckValid(local_id));
-    };
-
-    // Find a valid (non-hole) entry point
     InnerIdType entry = 0;
-    while (entry < static_cast<InnerIdType>(bucket_size) &&
-           ids[entry] == std::numeric_limits<InnerIdType>::max()) {
+    while (entry < bucket_size && ids[entry] == std::numeric_limits<InnerIdType>::max()) {
         ++entry;
     }
-    if (entry >= static_cast<InnerIdType>(bucket_size)) {
-        return;  // all entries are holes, fall back
-    }
-    float entry_dist = bucket->QueryOneById(computer, bucket_id, entry);
-    auto origin_entry = ids[entry] / buckets_per_data;
-    if (reasoning_ctx != nullptr) {
-        reasoning_ctx->RecordVisit(origin_entry, entry_dist, 0);
-    }
-    if (check_func(entry, origin_entry)) {
-        top_candidates.Push(entry_dist, entry);
-    } else if (reasoning_ctx != nullptr) {
-        reasoning_ctx->RecordFilterReject(origin_entry);
-    }
-    candidate_set.Push(-entry_dist, entry);
-    vl.Set(entry);
-
-    auto lower_bound = std::numeric_limits<float>::max();
-    if (not top_candidates.Empty()) {
-        lower_bound = top_candidates.Top().first;
+    if (entry == bucket_size) {
+        return;
     }
 
-    const auto max_degree = graph->MaximumDegree();
-    Vector<InnerIdType> neighbors(allocator_);
-
-    while (not candidate_set.Empty()) {
-        auto current_node_pair = candidate_set.Top();
-
-        if ((-current_node_pair.first) > lower_bound and top_candidates.Size() >= ef) {
-            break;
-        }
-        candidate_set.Pop();
-
-        const auto neighbor_count = graph->GetNeighborSize(current_node_pair.second);
-        if (neighbor_count > max_degree) {
-            continue;
-        }
-        graph->GetNeighbors(current_node_pair.second, neighbors);
-
-        for (const auto neighbor_id : neighbors) {
-            if (vl.Get(neighbor_id)) {
-                continue;
-            }
-            vl.Set(neighbor_id);
-
-            float d = bucket->QueryOneById(computer, bucket_id, neighbor_id);
-            auto origin_id = ids[neighbor_id] / buckets_per_data;
-            if (reasoning_ctx != nullptr) {
-                reasoning_ctx->RecordVisit(origin_id, d, 1);
-            }
-
-            if (top_candidates.Size() < ef or lower_bound > d or
-                (param.search_mode == RANGE_SEARCH and d <= param.radius + THRESHOLD_ERROR)) {
-                candidate_set.Push(-d, neighbor_id);
-                if (check_func(neighbor_id, origin_id)) {
-                    top_candidates.Push(d, neighbor_id);
-                } else if (reasoning_ctx != nullptr) {
-                    reasoning_ctx->RecordFilterReject(origin_id);
-                }
-                if (param.search_mode == KNN_SEARCH and top_candidates.Size() > ef) {
-                    if (reasoning_ctx != nullptr) {
-                        reasoning_ctx->RecordEviction(
-                            ids[top_candidates.Top().second] / buckets_per_data, 1);
-                    }
-                    top_candidates.Pop();
-                }
-                if (not top_candidates.Empty()) {
-                    lower_bound = top_candidates.Top().first;
-                }
-            }
-        }
+    Filter* attr_filter = nullptr;
+    if (not param.executors.empty() && static_cast<uint64_t>(thread_id) < param.executors.size() &&
+        param.executors[thread_id] != nullptr) {
+        param.executors[thread_id]->Clear();
+        attr_filter = param.executors[thread_id]->Run(bucket_id);
+    }
+    InnerSearchParam search_param = param;
+    search_param.ep = entry;
+    uint64_t result_limit = std::numeric_limits<uint64_t>::max();
+    if (search_param.search_mode == KNN_SEARCH) {
+        CHECK_ARGUMENT(topk > 0, "topk must be greater than 0");
+        result_limit = static_cast<uint64_t>(topk);
+        search_param.ef = std::max(search_param.ef, result_limit);
+    } else if (topk > 0) {
+        result_limit = static_cast<uint64_t>(topk);
+    }
+    search_param.ef = std::min<uint64_t>(search_param.ef, bucket_size);
+    search_param.topk = topk;
+    if (search_param.search_mode == RANGE_SEARCH) {
+        // BasicSearcher uses a non-positive range limit to mean unlimited; clamp only positive
+        // limits because InnerSearchParam stores this field as int.
+        search_param.range_search_limit_size =
+            topk > 0 ? static_cast<int>(std::min<int64_t>(topk, std::numeric_limits<int>::max()))
+                     : 0;
     }
 
-    if (param.search_mode == KNN_SEARCH) {
-        while (top_candidates.Size() > topk_u) {
-            top_candidates.Pop();
-        }
-    } else {
-        while (not top_candidates.Empty() and
-               top_candidates.Top().first > param.radius + THRESHOLD_ERROR) {
-            top_candidates.Pop();
-        }
-        if (topk_u > 0) {
-            while (top_candidates.Size() > topk_u) {
-                top_candidates.Pop();
-            }
-        }
+    // BucketInterface can create computers only from raw queries, not local IDs. Search only
+    // needs query distances, so use the no-factory provider and make pairwise misuse explicit.
+    BucketDistanceProvider distance_provider(bucket, bucket_id, computer, ids, buckets_per_data);
+    auto visited = std::make_shared<VisitedList>(bucket_size, allocator_);
+    QueryContext query_context{};
+    query_context.reasoning_ctx = reasoning_ctx;
+    BasicSearcher searcher(allocator_);
+    auto top_candidates = searcher.Search(
+        graph, distance_provider, visited, search_param, attr_filter, &query_context);
+    while (not top_candidates->Empty()) {
+        const auto [distance, local_id] = top_candidates->Top();
+        heap->Push(distance, ids[local_id]);
+        top_candidates->Pop();
     }
-
-    while (not top_candidates.Empty()) {
-        auto [d, local_id] = top_candidates.Top();
-        heap->Push(d, ids[local_id]);
-        top_candidates.Pop();
-    }
-    while (heap->Size() > topk_u) {
+    while (heap->Size() > result_limit) {
         heap->Pop();
     }
 }
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -34,6 +34,7 @@
 #include "graph_bucket_searcher.h"
 #include "impl/heap/standard_heap.h"
 #include "impl/inner_search_param.h"
+#include "impl/pruning_strategy.h"
 #include "impl/reasoning/search_reasoning.h"
 #include "impl/reorder/flatten_reorder.h"
 #include "impl/searcher/basic_searcher.h"
@@ -738,131 +739,25 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
             CHECK_ARGUMENT(query != nullptr, "base dataset has no graph build vector data");
             return bucket_->FactoryComputer(query);
         };
-
-        // Adapted from HGraph select_edges_by_heuristic: retain close, diverse edges.
-        auto select_edges_by_heuristic = [&](InnerIdType source, Vector<InnerIdType>& neighbors) {
-            auto source_computer = make_computer(source);
-            Vector<std::pair<float, InnerIdType>> candidates(allocator_);
-            candidates.reserve(neighbors.size());
-            for (const auto neighbor : neighbors) {
-                if (neighbor == source || neighbor >= bucket_size ||
-                    inner_ids[neighbor] == std::numeric_limits<InnerIdType>::max()) {
-                    continue;
-                }
-                bool duplicate = false;
-                for (const auto& candidate : candidates) {
-                    if (candidate.second == neighbor) {
-                        duplicate = true;
-                        break;
-                    }
-                }
-                if (!duplicate) {
-                    candidates.emplace_back(bucket_->QueryOneById(source_computer, b, neighbor),
-                                            neighbor);
-                }
-            }
-            std::sort(candidates.begin(), candidates.end(), [](const auto& lhs, const auto& rhs) {
-                return lhs.first < rhs.first;
-            });
-
-            neighbors.clear();
-            Vector<ComputerInterfacePtr> selected_computers(allocator_);
-            for (const auto& candidate : candidates) {
-                bool good = true;
-                for (const auto& selected_comp : selected_computers) {
-                    const auto pair_distance =
-                        bucket_->QueryOneById(selected_comp, b, candidate.second);
-                    if (pair_distance < candidate.first) {
-                        good = false;
-                        break;
-                    }
-                }
-                if (good) {
-                    neighbors.push_back(candidate.second);
-                    selected_computers.push_back(make_computer(candidate.second));
-                    if (neighbors.size() == static_cast<uint64_t>(effective_degree)) {
-                        break;
-                    }
-                }
-            }
-        };
+        auto mutexes = std::make_shared<EmptyMutex>();
+        BasicSearcher searcher(common_param_);
 
         const auto entry = valid_ids.front();
         graph->InsertNeighborsById(entry, Vector<InnerIdType>(allocator_));
-        VisitedList visited(bucket_size, allocator_);
-        Vector<InnerIdType> graph_neighbors(effective_degree, allocator_);
-
+        auto visited = std::make_shared<VisitedList>(bucket_size, allocator_);
         for (uint64_t node_pos = 1; node_pos < valid_ids.size(); ++node_pos) {
             const auto node = valid_ids[node_pos];
-            auto computer = make_computer(node);
-            const auto entry_dist = bucket_->QueryOneById(computer, b, entry);
-            const auto ef = std::min(ef_construction, node_pos);
-
-            visited.Reset();
-            StandardHeap<true, false> top_candidates(allocator_, -1);
-            StandardHeap<true, false> candidate_set(allocator_, -1);
-            top_candidates.Push(entry_dist, entry);
-            candidate_set.Push(-entry_dist, entry);
-            visited.Set(entry);
-            auto lower_bound = entry_dist;
-
-            while (not candidate_set.Empty()) {
-                const auto current = candidate_set.Top();
-                if (-current.first > lower_bound && top_candidates.Size() >= ef) {
-                    break;
-                }
-                candidate_set.Pop();
-
-                graph->GetNeighbors(current.second, graph_neighbors);
-                const auto neighbor_count = graph->GetNeighborSize(current.second);
-                for (uint32_t i = 0; i < neighbor_count; ++i) {
-                    const auto neighbor = graph_neighbors[i];
-                    if (neighbor >= bucket_size || visited.Get(neighbor) ||
-                        inner_ids[neighbor] == std::numeric_limits<InnerIdType>::max()) {
-                        continue;
-                    }
-                    visited.Set(neighbor);
-                    const auto distance = bucket_->QueryOneById(computer, b, neighbor);
-                    if (top_candidates.Size() < ef || distance < lower_bound) {
-                        candidate_set.Push(-distance, neighbor);
-                        top_candidates.Push(distance, neighbor);
-                        if (top_candidates.Size() > ef) {
-                            top_candidates.Pop();
-                        }
-                        lower_bound = top_candidates.Top().first;
-                    }
-                }
-            }
-
-            Vector<std::pair<float, InnerIdType>> candidates(allocator_);
-            candidates.reserve(top_candidates.Size());
-            for (uint64_t i = 0; i < top_candidates.Size(); ++i) {
-                candidates.push_back(top_candidates.GetData()[i]);
-            }
-            std::sort(candidates.begin(), candidates.end(), [](const auto& lhs, const auto& rhs) {
-                return lhs.first < rhs.first;
-            });
-
-            // Adapted from HGraph mutually_connect_new_element.
-            Vector<InnerIdType> node_neighbors(allocator_);
-            node_neighbors.reserve(candidates.size());
-            for (const auto& candidate : candidates) {
-                node_neighbors.push_back(candidate.second);
-            }
-            select_edges_by_heuristic(node, node_neighbors);
-            graph->InsertNeighborsById(node, node_neighbors);
-
-            for (const auto neighbor : node_neighbors) {
-                Vector<InnerIdType> reciprocal(allocator_);
-                graph->GetNeighbors(neighbor, reciprocal);
-                if (reciprocal.size() < static_cast<uint64_t>(effective_degree)) {
-                    reciprocal.push_back(node);
-                } else {
-                    reciprocal.push_back(node);
-                    select_edges_by_heuristic(neighbor, reciprocal);
-                }
-                graph->InsertNeighborsById(neighbor, reciprocal);
-            }
+            InnerSearchParam search_param;
+            search_param.ep = entry;
+            search_param.ef = std::min(ef_construction, node_pos);
+            search_param.topk = static_cast<int64_t>(search_param.ef);
+            BucketDistanceProvider distance_provider(
+                bucket_, b, make_computer(node), make_computer, inner_ids, buckets_per_data_);
+            visited->Reset();
+            auto candidates =
+                searcher.Search(graph, distance_provider, visited, search_param, nullptr, nullptr);
+            mutually_connect_new_element(
+                node, candidates, graph, distance_provider, mutexes, allocator_);
         }
 
         bucket_graphs_[b] = std::move(graph);
--- a/src/impl/distance_provider_for_graph.cpp
+++ b/src/impl/distance_provider_for_graph.cpp
@@ -1,0 +1,94 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "impl/distance_provider_for_graph.h"
+
+#include <limits>
+
+#include "datacell/bucket_interface.h"
+
+namespace vsag {
+
+BucketDistanceProvider::BucketDistanceProvider(std::shared_ptr<BucketInterface> bucket,
+                                               BucketIdType bucket_id,
+                                               ComputerInterfacePtr computer,
+                                               ComputerFactory computer_factory,
+                                               const InnerIdType* inner_ids,
+                                               BucketIdType buckets_per_data)
+    : bucket_(std::move(bucket)),
+      bucket_id_(bucket_id),
+      computer_(std::move(computer)),
+      computer_factory_(std::move(computer_factory)),
+      inner_ids_(inner_ids),
+      buckets_per_data_(buckets_per_data) {
+    CHECK_ARGUMENT(buckets_per_data_ > 0, "buckets_per_data must be positive");
+}
+
+BucketDistanceProvider::BucketDistanceProvider(std::shared_ptr<BucketInterface> bucket,
+                                               BucketIdType bucket_id,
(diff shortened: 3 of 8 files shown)


src/algorithm/ivf/graph_bucket_searcher.cpp
```diff
@@ -82,144 +82,51 @@ GraphBucketSearcher::search_graph(BucketIdType bucket_id,
...
     }
 
-    while (not top_candidates.Empty()) {
-        auto [d, local_id] = top_candidates.Top();
-        heap->Push(d, ids[local_id]);
-        top_candidates.Pop();
+    BucketDistanceProvider distance_provider(
+        bucket, bucket_id, computer, nullptr, ids, buckets_per_data);
+    auto visited = std::make_shared<VisitedList>(bucket_size, allocator_);
+    QueryContext query_context;
```
LHT129: [suggestion] The `computer_factory` is passed as `nullptr` here, which means `BucketDistanceProvider::PairwiseDistance` will call `FactoryComputerById` on a null `std::function`, causing undefined behavior if `PairwiseDistance` is ever invoked on this provider instance. While the current search path only calls `QueryDistance` and `BatchQueryDistance`, this is a latent footgun. Consider either:
1. Asserting `computer_factory_` is non-null in `FactoryComputerById`, or
2. Providing a no-op factory (e.g. `[](InnerIdType) { return nullptr; }`) to make the invariant explicit.

src/impl/distance_provider_for_graph.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+void
+BucketDistanceProvider::Prefetch(InnerIdType id) const {
+    bucket_->Prefetch(bucket_id_, id);
+}
+
+InnerIdType
+BucketDistanceProvider::OriginalId(InnerIdType id) const {
+    return inner_ids_ == nullptr ? id : inner_ids_[id] / buckets_per_data_;
+}
```
LHT129, line 92: [note] `OriginalId` divides by `buckets_per_data_` without checking for zero. While `buckets_per_data_` is typically >= 1 in practice, if a caller ever constructs this with `buckets_per_data = 0`, it would cause a division-by-zero crash. Consider adding a defensive check or a static_assert/documentation that `buckets_per_data` must be >= 1.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -148,6 +152,147 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
 }
 
+DistHeapPtr
+BasicSearcher::Search(const GraphInterfacePtr& graph,
+                      const DistanceProviderForGraph& distance_provider,
+                      const VisitedListPtr& vl,
+                      const InnerSearchParam& inner_search_param,
+                      Filter* attr_filter,
+                      QueryContext* ctx) const {
+    if (inner_search_param.search_mode == RANGE_SEARCH) {
```
LHT129: [note] The new `search_impl` overload for `DistanceProviderForGraph` shares ~90% of its logic with the existing `search_impl` overload for `FlattenInterface` + `LabelTable`. This duplication means future bug fixes or improvements to the graph search algorithm will need to be applied in two places. Consider whether the shared logic (the main search loop, pruning, result trimming) can be extracted into a common helper to reduce maintenance burden.

src/impl/distance_provider_for_graph.h
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    [[nodiscard]] float
+    PairwiseDistance(InnerIdType id1, InnerIdType id2) const override {
+        return flatten_->ComputePairVectors(id1, id2);
+    }
+
+    [[nodiscard]] ComputerInterfacePtr
+    FactoryComputerById(InnerIdType) const override {
+        return nullptr;
```
LHT129: [note] `FlattenDistanceProvider::FactoryComputerById` returns `nullptr`. While `FlattenDistanceProvider` overrides `PairwiseDistance` to use `flatten_->ComputePairVectors` directly (bypassing the base class default that calls `FactoryComputerById`), this could surprise future callers. If someone adds code that calls `FactoryComputerById` on a `FlattenDistanceProvider` and uses the result, they will get a null pointer. Consider adding a comment explaining why `nullptr` is intentional, or throw/assert to catch misuse early.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -148,6 +152,147 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+    }
+    if constexpr (mode == RANGE_SEARCH) {
+        if (dist > inner_search_param.radius && not top_candidates->Empty()) {
+            top_candidates->Pop();
+        }
+    }
+    candidate_set->Push(-dist, ep);
+    vl->Set(ep);
+    auto lower_bound =
+        top_candidates->Empty() ? std::numeric_limits<float>::max() : top_candidates->Top().first;
```
LHT129, line 214: [note] The old code checked `top_candidates.Size() >= ef` in the KNN search termination condition, while the new `BasicSearcher::search_impl` checks `top_candidates->Size() == ef`. This is a semantic difference: the old code would break when the heap reached or exceeded ef, while the new code only breaks when exactly ef. In practice, since `top_candidates` is never allowed to exceed `ef` (it is popped immediately when `Size() > ef`), the behavior should be equivalent. However, this is worth noting for reviewers to confirm the equivalence is intentional.

src/algorithm/ivf/graph_bucket_searcher.cpp
LHT129: Fixed in 7934de4d. The search path now uses an explicit no-factory BucketDistanceProvider constructor. Any accidental pairwise/source-computer use throws a clear unsupported-operation error instead of dereferencing an empty function.

src/impl/distance_provider_for_graph.cpp
LHT129: Fixed in 7934de4d. BucketDistanceProvider now validates buckets_per_data is positive during construction, preventing division by zero in OriginalId.

src/impl/searcher/basic_searcher.cpp
LHT129: Fixed in 7934de4d. The termination guard now uses at-least-ef semantics in the provider path and established BasicSearcher paths, making the heap-cap invariant explicit.

src/impl/distance_provider_for_graph.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+float
+BucketDistanceProvider::QueryDistance(InnerIdType id, QueryContext*) const {
+    return bucket_->QueryOneById(computer_, bucket_id_, id);
+}
+
+float
+BucketDistanceProvider::PairwiseDistance(InnerIdType id1, InnerIdType id2) const {
+    return bucket_->QueryOneById(FactoryComputerById(id1), bucket_id_, id2);
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. Shared pruning now keeps an operation-local cache of source computers and passes cached computers to BucketDistanceProvider pairwise calls, avoiding repeated construction and preprocessing while keeping the cache thread-local.

src/impl/distance_provider_for_graph.h
```diff
@@ -0,0 +1,136 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <utility>
+
+#include "datacell/flatten_interface.h"
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. The new header now directly includes basic_types.h, query_context.h, quantization/computer.h, and vsag_exception.h rather than relying on transitive definitions.

src/impl/searcher/basic_searcher.h
```diff
@@ -43,6 +44,8 @@ class BasicSearcher {
     explicit BasicSearcher(const IndexCommonParam& common_param,
                            MutexArrayPtr mutex_array = nullptr);
 
+    BasicSearcher(Allocator* allocator, MutexArrayPtr mutex_array = nullptr);
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. BasicSearcher Allocator constructor is now explicit.

src/algorithm/ivf/ivf.cpp
```diff
@@ -736,131 +737,24 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
...
+            search_param.ef = std::min(ef_construction, node_pos);
+            search_param.topk = static_cast<int64_t>(search_param.ef);
+            BucketDistanceProvider distance_provider(
+                bucket_, b, make_computer(node), make_computer, inner_ids, buckets_per_data_);
+            auto visited = std::make_shared<VisitedList>(bucket_size, allocator_);
+            auto candidates =
+                searcher.Search(graph, distance_provider, visited, search_param, nullptr, nullptr);
+            mutually_connect_new_element(
+                node, candidates, graph, distance_provider, mutexes, allocator_);
         }
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. IVF graph construction allocates one VisitedList per bucket and resets it before each insertion search.

src/algorithm/ivf/ivf.cpp
```diff
@@ -736,131 +737,24 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
...
-                }
-                graph->InsertNeighborsById(neighbor, reciprocal);
-            }
+            InnerSearchParam search_param;
+            search_param.ep = entry;
+            search_param.ef = std::min(ef_construction, node_pos);
+            search_param.topk = static_cast<int64_t>(search_param.ef);
+            BucketDistanceProvider distance_provider(
+                bucket_, b, make_computer(node), make_computer, inner_ids, buckets_per_data_);
+            auto visited = std::make_shared<VisitedList>(bucket_size, allocator_);
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. The builder now reuses a single bucket-sized VisitedList and calls Reset for every inserted node.

src/impl/distance_provider_for_graph.h
LHT129: Fixed in 7934de4d. FlattenDistanceProvider FactoryComputerById now explicitly throws UNSUPPORTED_INDEX_OPERATION; Flatten pairwise distance continues to use ComputePairVectors directly.

src/impl/distance_provider_for_graph.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. BucketDistanceProvider validates local IDs: holes return infinity for query distance, cannot be used for pairwise/source computers, and are not prefetched. Provider search also skips invalid IDs before candidate insertion.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -148,6 +152,147 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+    const auto check_func = [&distance_provider, &inner_search_param, attr_filter](InnerIdType id) {
+        const auto original_id = distance_provider.OriginalId(id);
+        return (inner_search_param.is_inner_id_allowed == nullptr ||
+                inner_search_param.is_inner_id_allowed->CheckValid(original_id)) &&
+               (attr_filter == nullptr || attr_filter->CheckValid(id));
+    };
+    const auto ep = inner_search_param.ep;
+    const auto ef = inner_search_param.ef;
+    auto* reasoning = ctx == nullptr ? nullptr : ctx->reasoning_ctx;
+    float dist = distance_provider.QueryDistance(ep, ctx);
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. Provider search now records the evaluated entry point at reasoning hop 0 before filtering or result trimming.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -148,6 +152,147 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+    uint32_t hops = 0;
+    uint32_t dist_cmp = 1;
+    if (check_func(ep)) {
+        top_candidates->Push(dist, ep);
+    } else if (reasoning != nullptr) {
+        reasoning->RecordFilterReject(distance_provider.OriginalId(ep));
+    }
+    if constexpr (mode == RANGE_SEARCH) {
+        if (dist > inner_search_param.radius && not top_candidates->Empty()) {
```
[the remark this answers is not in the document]
LHT129: Fixed in 7934de4d. Provider-backed range entry and neighbor gates now use radius plus THRESHOLD_ERROR, matching final cleanup and flat IVF tolerance.

src/impl/searcher/basic_searcher.cpp
LHT129: Reviewed in 7934de4d. I kept the provider loop separate: the established Flatten paths carry iterator state, RaBitQ bounds, duplicate expansion, preset computers, timeout handling, and discard bookkeeping. A shared loop would require a much larger abstraction and risks changing compatibility behavior.

src/impl/pruning_strategy.cpp
```diff
@@ -135,17 +173,16 @@ mutually_connect_new_element(InnerIdType cur_c,
             graph->InsertNeighborsById(selected_neighbor, neighbors);
         } else {
             // finding the "weakest" element to replace it with the new one
-            float d_max = flatten->ComputePairVectors(cur_c, selected_neighbor);
+            float d_max = pairwise_distance(cur_c, selected_neighbor);
```
LHT129: [critical] The pairwise distance direction is inverted for bucket-based (quantized) distance providers in `mutually_connect_new_element`.

At line 176, `pairwise_distance(cur_c, selected_neighbor)` measures the distance from `cur_c` to `selected_neighbor`. But `BucketDistanceProvider::PairwiseDistance(id1, id2)` uses `id1` as the source to create a computer (via `FactoryComputerById(id1)`) and `id2` as the target. For quantized distances (SQ/PQ/RaBitQ), this is **asymmetric**: `d(a, b) != d(b, a)` because the computer is built from the raw vector of `id1` while `id2` stays quantized.

The original IVF code measured `selected_neighbor -> candidate` (the "owner" perspective). Now the code measures `cur_c -> selected_neighbor` and `neighbors[j] -> selected_neighbor`, which reverses the direction. This will produce different pruning decisions for quantized buckets, potentially affecting graph recall.

The fix: swap the arguments so `selected_neighbor` is always `id1`:
```cpp
float d_max = pairwise_distance(selected_neighbor, cur_c);
// ...
candidates->Push(pairwise_distance(selected_neighbor, neighbors[j]), neighbors[j]);
```

The same issue applies to `select_edges_by_heuristic` where `pairwise_distance(second_pair.second, current_pair.second)` is called — verify the direction matches the original `ComputePairVectors` semantics for the FlattenDistanceProvider path.

src/impl/pruning_strategy.cpp
```diff
@@ -25,16 +27,28 @@ void
...
                           InnerIdType node_id,
                           uint64_t max_size,
-                          const FlattenInterfacePtr& flatten,
+                          const DistanceProviderForGraph& distance_provider,
                           Allocator* allocator,
                           float alpha) {
+    std::unordered_map<InnerIdType, ComputerInterfacePtr> computers;
+    const auto pairwise_distance = [&distance_provider, &computers](InnerIdType id1,
+                                                                    InnerIdType id2) {
+        if (not distance_provider.SupportsComputerById()) {
```
LHT129: [suggestion] The `std::unordered_map<InnerIdType, ComputerInterfacePtr>` used for caching computers in `select_edges_by_heuristic` and `mutually_connect_new_element` uses the default `std::allocator`. The rest of the codebase uses the project's custom allocator (via `Vector`, `StandardHeap`, etc.) for memory accounting and alignment. Consider using a custom allocator-aware map or a flat `Vector`-based cache to stay consistent with the project's memory management strategy. This is especially relevant since these functions are called during graph construction where allocation patterns matter.

src/impl/pruning_strategy.cpp
```diff
@@ -25,16 +27,28 @@ void
...
+            return distance_provider.PairwiseDistance(id1, id2);
+        }
```
LHT129: [note] The `pairwise_distance` lambda with its `std::unordered_map` cache is duplicated verbatim across three functions: `select_edges_by_heuristic(Vector&, ...)`, `select_edges_by_heuristic(DistHeapPtr, ...)`, and `mutually_connect_new_element`. Each instance creates its own independent cache, which is correct (thread-local), but the code duplication makes future changes error-prone. Consider extracting the caching logic into a small helper or a lightweight RAII wrapper to keep the caching behavior in one place.

src/impl/pruning_strategy.cpp
```diff
@@ -70,7 +96,7 @@ select_edges_by_heuristic(const DistHeapPtr& edges,
         bool good = true;
```
LHT129, line 109: [critical] Same pairwise distance direction issue as in `mutually_connect_new_element`. At line 96, `pairwise_distance(second_pair.second, current_pair.second)` calls `BucketDistanceProvider::PairwiseDistance` with `second_pair.second` (an already-selected neighbor) as `id1` and `current_pair.second` (the candidate being evaluated) as `id2`. For quantized distances, this measures `d(selected, candidate)` instead of `d(candidate, selected)`. The original `flatten->ComputePairVectors(second_pair.second, current_pair.second)` had the same argument order, so the Flatten path is unchanged. But for BucketDistanceProvider, the semantics differ: the computer is created from `id1` (the selected neighbor's raw vector) and `id2` (the candidate's quantized form). Verify that this is the intended direction for the pruning heuristic — the heuristic should measure whether the candidate is close to any already-selected neighbor, so `d(candidate, selected)` may be more appropriate.

src/impl/pruning_strategy.cpp
```diff
@@ -25,16 +27,28 @@ void
...
+        if (inserted) {
+            it->second = distance_provider.FactoryComputerById(id1);
+        }
+        return distance_provider.PairwiseDistance(id1, id2, it->second);
+    };
     auto edges = std::make_shared<StandardHeap<true, false>>(allocator, -1);
     for (const auto& neighbor : neighbors) {
-        float dist = flatten->ComputePairVectors(node_id, neighbor);
+        float dist = pairwise_distance(node_id, neighbor);
         edges->Push(dist, neighbor);
```
LHT129, line 72: [note] The `select_edges_by_heuristic(Vector&, ...)` variant at line 48 calls `pairwise_distance(node_id, neighbor)` to compute the initial edge distances. For `BucketDistanceProvider`, `node_id` is the new element being inserted and `neighbor` is a candidate from the search results. This means the computer is built from `node_id`'s raw vector (via `FactoryComputerById(node_id)`) and distances are measured to each candidate's quantized form. This is consistent with the original IVF code which created `source_computer` from `source` and measured distances to neighbors. However, verify that the `pairwise_distance` lambda's computer caching is effective here — `node_id` is the same for all iterations of the loop, so the cache should reuse the computer after the first call.

src/impl/searcher/basic_searcher.h
```diff
@@ -126,6 +137,15 @@ class BasicSearcher {
                 QueryContext* ctx,
                 DistanceRecordVector* rabitq_lower_bound_candidates = nullptr) const;
 
+    template <InnerSearchMode mode = KNN_SEARCH>
```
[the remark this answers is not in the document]
LHT129: Fixed in 00e3a028. The template default now explicitly uses InnerSearchMode::KNN_SEARCH.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -148,6 +152,154 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+                      const VisitedListPtr& vl,
+                      const InnerSearchParam& inner_search_param,
+                      Filter* attr_filter,
+                      QueryContext* ctx) const {
+    if (inner_search_param.search_mode == RANGE_SEARCH) {
+        return this->search_impl<RANGE_SEARCH>(
+            graph, distance_provider, vl, inner_search_param, attr_filter, ctx);
+    }
+    return this->search_impl<KNN_SEARCH>(
+        graph, distance_provider, vl, inner_search_param, attr_filter, ctx);
```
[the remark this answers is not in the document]
LHT129: Fixed in 00e3a028. The provider-backed search path now consistently qualifies InnerSearchMode::KNN_SEARCH and InnerSearchMode::RANGE_SEARCH.

src/impl/pruning_strategy.cpp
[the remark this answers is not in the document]
LHT129: Fixed in 00e3a028. Reciprocal pruning now measures selected_neighbor -> cur_c and selected_neighbor -> each existing neighbor, matching the asymmetric bucket-computer direction and the original IVF logic.

LHT129: Fixed in 00e3a028. The reciprocal full-list branch now uses selected_neighbor as the source for both the new candidate and existing neighbors.

src/impl/pruning_strategy.cpp
LHT129: Fixed in 00e3a028. The per-operation computer cache now uses allocator-aware UnorderedMap constructed with the caller allocator.

src/impl/pruning_strategy.cpp
LHT129: Fixed in 00e3a028. PairwiseDistanceComputer centralizes the cache logic while each pruning operation keeps its own cache.

src/impl/pruning_strategy.cpp
LHT129: Verified and intentionally retained in 00e3a028. The selection heuristic uses selected neighbor -> candidate, which matches the original IVF bucket graph construction behavior; only reciprocal full-list pruning required the direction correction.

src/impl/pruning_strategy.cpp
LHT129: Verified in 00e3a028. PairwiseDistanceComputer caches each source-node computer and evaluates node_id -> neighbor; the cache is local to each pruning operation.

src/algorithm/ivf/graph_bucket_searcher.cpp
```diff
@@ -82,144 +82,54 @@ GraphBucketSearcher::search_graph(BucketIdType bucket_id,
...
-                }
-                if (not top_candidates.Empty()) {
-                    lower_bound = top_candidates.Top().first;
-                }
-            }
-        }
+    InnerSearchParam search_param = param;
+    search_param.ep = entry;
+    if (search_param.search_mode == KNN_SEARCH) {
+        search_param.ef = std::max<uint64_t>(search_param.ef, topk);
```
LHT129: [note] `std::max<uint64_t>(search_param.ef, topk)` mixes signed `int64_t` (`topk`) with unsigned `uint64_t`. While `topk` is never negative in practice, using an explicit `static_cast<uint64_t>(topk)` would make the intent clear and silence `-Wsign-compare` warnings.

```cpp
// Current:
search_param.ef = std::max<uint64_t>(search_param.ef, topk);
// Suggested:
search_param.ef = std::max(search_param.ef, static_cast<uint64_t>(topk));
```

src/impl/searcher/basic_searcher.cpp
```diff
@@ -149,6 +153,155 @@ BasicSearcher::Search(const GraphInterfacePtr& graph,
...
+        }
+    }
+    candidate_set->Push(-dist, ep);
+    vl->Set(ep);
+    auto lower_bound =
+        top_candidates->Empty() ? std::numeric_limits<float>::max() : top_candidates->Top().first;
+    Vector<InnerIdType> to_be_visited_id(graph->MaximumDegree(), alloc);
+    Vector<InnerIdType> neighbors(graph->MaximumDegree(), alloc);
+    Vector<float> line_dists(graph->MaximumDegree(), alloc);
+    auto skip_strategy = create_filter_search_skip_strategy(
```
LHT129: [suggestion] The `skip_strategy` is allocated and initialized here but `visit()` on line 244 is called with `filter=nullptr`, so the strategy is never consulted (its `ShouldVisit()` is only invoked when a non-null filter is passed). This is wasted allocation and initialization on the search hot path. Consider skipping the strategy creation when no filter is active, e.g., by checking `inner_search_param.is_inner_id_allowed == nullptr && attr_filter == nullptr` before creating it.

src/algorithm/ivf/graph_bucket_searcher.cpp
```diff
@@ -82,144 +82,54 @@ GraphBucketSearcher::search_graph(BucketIdType bucket_id,
...
-        }
-        if (topk_u > 0) {
-            while (top_candidates.Size() > topk_u) {
-                top_candidates.Pop();
-            }
-        }
+    search_param.ef = std::min<uint64_t>(search_param.ef, bucket_size);
+    search_param.topk = topk;
+    if (search_param.search_mode == RANGE_SEARCH) {
+        search_param.range_search_limit_size =
```
LHT129, line 122: [suggestion] The `range_search_limit_size` computation chain `static_cast<int>(std::min<int64_t>(topk, std::numeric_limits<int>::max()))` is correct but overly complex. `topk` is already `int64_t` and the guard `topk > 0` ensures positivity. Since `range_search_limit_size` is `int`, a simpler `static_cast<int>(topk)` with a comment noting that `topk` is bounded by `bucket_size` (which is practically far below `INT_MAX`) would be clearer.

Also, the old code used `topk_u` (`uint64_t`) for the final truncation loop, which avoided signed/unsigned comparison warnings. The new code uses `static_cast<uint64_t>(topk)` in the truncation loop, which is fine but note that `search_param.topk = topk` stores `int64_t` directly — if `topk` were somehow negative, the search_impl would see a negative topk.

src/algorithm/ivf/graph_bucket_searcher.cpp
```diff
@@ -82,144 +82,54 @@ GraphBucketSearcher::search_graph(BucketIdType bucket_id,
...
+            topk > 0 ? static_cast<int>(std::min<int64_t>(topk, std::numeric_limits<int>::max()))
+                     : 0;
     }
 
-    while (not top_candidates.Empty()) {
-        auto [d, local_id] = top_candidates.Top();
-        heap->Push(d, ids[local_id]);
-        top_candidates.Pop();
+    // BucketInterface can create computers only from raw queries, not local IDs. Search only
+    // needs query distances, so use the no-factory provider and make pairwise misuse explicit.
```
LHT129, line 128: [note] `BucketDistanceProvider` constructed without a `computer_factory` (as in `graph_bucket_searcher.cpp:119`) will throw `VsagException` if `PairwiseDistance` is ever called on it, because `FactoryComputerById` requires a non-null factory. This is safe for the search path since `BasicSearcher::search_impl` for `DistanceProviderForGraph` only calls `QueryDistance`/`BatchQueryDistance`/`IsValid`/`OriginalId`/`Prefetch` — never `PairwiseDistance`.

However, this invariant is implicit and fragile. If someone later adds pairwise distance support to the search path, it will crash at runtime. Consider either:
1. Making the no-factory `PairwiseDistance` throw a clearer error (already done via `FactoryComputerById`), or
2. Adding a static assert or comment on the search-path constructor to document that `PairwiseDistance` must not be called.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -253,7 +406,7 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
         auto current_node_pair = candidate_set->Top();
 
         if constexpr (mode == InnerSearchMode::KNN_SEARCH) {
-            if ((-current_node_pair.first) > lower_bound && top_candidates->Size() == ef) {
+            if ((-current_node_pair.first) > lower_bound && top_candidates->Size() >= ef) {
```
LHT129, line 398: [note] The termination condition change from `Size() == ef` to `Size() >= ef` at lines 409 and 692 is a correctness improvement — the heap can transiently exceed `ef` when filter-checked candidates are pushed and later evicted. This prevents premature loop termination when `top_candidates` temporarily has more than `ef` entries due to filter rejections. This change is applied to both existing `search_impl` overloads (for `FlattenInterface` + `LabelTable` and `FlattenInterface` + `IteratorFilterContext`). The new `DistanceProviderForGraph` overload already uses `>= ef` consistently.

This is a good fix but was not mentioned in the PR description. Consider adding a note about this behavioral change for reviewers.

src/impl/pruning_strategy.cpp
```diff
@@ -17,24 +17,61 @@
...
+    PairwiseDistanceComputer(const DistanceProviderForGraph& distance_provider,
+                             Allocator* allocator)
+        : distance_provider_(distance_provider), computers_(allocator) {
+    }
+
+    PairwiseDistanceComputer(const PairwiseDistanceComputer&) = delete;
+    PairwiseDistanceComputer&
+    operator=(const PairwiseDistanceComputer&) = delete;
+    PairwiseDistanceComputer(PairwiseDistanceComputer&&) = delete;
+    PairwiseDistanceComputer&
```
LHT129, line 38: [suggestion] `PairwiseDistanceComputer` caches computers in an `UnorderedMap<InnerIdType, ComputerInterfacePtr>` keyed by `id1`. In `select_edges_by_heuristic` (the heap overload), the inner loop calls `pairwise_distance.PairwiseDistance(second_pair.second, current_pair.second)` with varying `id1` values (`second_pair.second` iterates over `return_list`). For large `max_size` values, this can create and cache many computers, consuming significant memory. The original IVF code created computers on-the-fly and let them go out of scope.

Consider adding an upper bound on the cache size or using an LRU eviction policy for the computer cache. Alternatively, since `select_edges_by_heuristic` is called during graph construction (not in the hot search path), the memory overhead may be acceptable — but it is worth documenting the trade-off.

src/algorithm/ivf/ivf.cpp
```diff
@@ -738,131 +739,25 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
...
-                if (good) {
-                    neighbors.push_back(candidate.second);
-                    selected_computers.push_back(make_computer(candidate.second));
-                    if (neighbors.size() == static_cast<uint64_t>(effective_degree)) {
-                        break;
-                    }
-                }
-            }
-        };
+        auto mutexes = std::make_shared<EmptyMutex>();
```
LHT129, line 742: [note] `std::make_shared<EmptyMutex>()` is created but `EmptyMutex` is not defined in the visible diff. The `mutually_connect_new_element` function uses `LockGuard lock(neighbors_mutexes, selected_neighbor)` which requires the mutex array to support `LockGuard`. If `EmptyMutex` is a no-op mutex (as the name suggests), this is fine for single-threaded graph construction. However, the original IVF code did not use any mutex at all — it directly called `graph->InsertNeighborsById` without locking. The introduction of `EmptyMutex` is a good abstraction that allows future parallelization, but verify that `LockGuard` with `EmptyMutex` compiles and behaves as a true no-op (no atomic operations, no memory barriers).

src/impl/distance_provider_for_graph.h
```diff
@@ -0,0 +1,169 @@
...
+    IsValid(InnerIdType) const {
+        return true;
+    }
+};
+
+class FlattenDistanceProvider final : public DistanceProviderForGraph {
+public:
+    FlattenDistanceProvider(FlattenInterfacePtr flatten, ComputerInterfacePtr computer)
+        : flatten_(std::move(flatten)), computer_(std::move(computer)) {
+    }
```
LHT129, line 80: [suggestion] The `DistanceProviderForGraph::IsValid` default returns `true` unconditionally. `BucketDistanceProvider::QueryDistance` calls `IsValid(id)` and returns infinity for invalid IDs. But `FlattenDistanceProvider` inherits the default `IsValid` (always true) and does not override it. This means the new `search_impl` for `DistanceProviderForGraph` will call `distance_provider.IsValid(id)` in the inner loop — for `FlattenDistanceProvider` this is a virtual call that always returns true, adding overhead with no benefit.

Consider either:
1. Making `IsValid` non-virtual and always returning true in the base class, with `BucketDistanceProvider` handling validity internally in `QueryDistance`/`BatchQueryDistance`, or
2. Overriding `IsValid` in `FlattenDistanceProvider` to be `final` so the compiler can devirtualize the call.

src/algorithm/ivf/graph_bucket_searcher.cpp
LHT129: Fixed in 85e9a183. KNN now validates topk is positive before conversion, then uses an explicit uint64_t topk for ef and result trimming.

src/impl/searcher/basic_searcher.cpp
LHT129: Fixed in 85e9a183. The provider search path passes no filter to visit, so its SkipStrategy was never consulted; the redundant allocation was removed while existing filtered paths retain their strategies.

src/algorithm/ivf/graph_bucket_searcher.cpp
LHT129: Fixed in 85e9a183. Positive range limits saturate safely at INT_MAX, while non-positive values retain unlimited range-search behavior without assuming bucket size fits in int.

src/algorithm/ivf/graph_bucket_searcher.cpp
LHT129: Documented in 85e9a183. The no-factory BucketDistanceProvider constructor is query-only; pairwise operations require the factory-bearing constructor and already fail explicitly if misused.

src/impl/searcher/basic_searcher.cpp
LHT129: Documented. The PR description now explains the KNN ef/topk behavior and the >= ef early-termination safeguard.

src/impl/pruning_strategy.cpp
LHT129: Documented in 85e9a183. This cache is scoped to one pruning operation and bounded by selected edges: heap heuristic accepted entries and mutual-connect sources are capped by maximum degree; the vector overload caches only node_id before delegation. An LRU would add complexity without demonstrated benefit.

src/impl/distance_provider_for_graph.h
LHT129: Retained intentionally. IsValid remains virtual because BucketDistanceProvider needs hole validation and custom providers remain extensible. The search API holds the provider by base reference, so marking a Flatten override final would not reliably remove dispatch; changing the abstraction for unprofiled overhead risks compatibility and correctness.

src/algorithm/ivf/ivf.cpp
LHT129: Verified. EmptyMutex is the established no-op MutexArray implementation in utils/lock_strategy.h and is already used by HGraph, HNSW, Pyramid, and pruning tests. The updated IVF target compiled successfully with it; it provides the shared mutual-connect interface without adding synchronization work to single-threaded construction.

LHT129 (commented):
## Review Summary

This PR refactors IVF to reuse graph search primitives by introducing `DistanceProviderForGraph` abstraction and adapting `BasicSearcher` and `pruning_strategy` to work with it. Overall the refactoring is well-structured and significantly reduces code duplication in `graph_bucket_searcher.cpp` and `ivf.cpp`.

### Issues previously raised and now resolved
- `OriginalId` division-by-zero: `CHECK_ARGUMENT(buckets_per_data_ > 0, ...)` added in constructor.
- `computer_factory` null safety: constructor now uses empty `{}` and `FactoryComputerById` throws `VsagException` when factory is empty.
- `PairwiseDistanceComputer` now caches computers to avoid repeated `FactoryComputerById` calls.
- `InnerSearchMode` enum values now use fully-qualified names (`InnerSearchMode::KNN_SEARCH`).

### Remaining concern
The pairwise distance direction in `mutually_connect_new_element` (raised in prior reviews as `[critical]` / `[P2]`) still warrants attention: `PairwiseDistance(selected_neighbor, cur_c)` uses `selected_neighbor` as the source for quantized distance providers, which is asymmetric for SQ/PQ/RaBitQ. The original IVF code measured from the owner perspective. This may affect graph recall for quantized buckets.

### Code quality
- The `DistanceProviderForGraph` interface is clean and well-designed with sensible defaults.
- `BucketDistanceProvider` correctly handles bucket holes via `IsValid()`.
- The new `BasicSearcher::Search` overload for `DistanceProviderForGraph` properly dispatches to the templated `search_impl`.
- `select_edges_by_heuristic` and `mutually_connect_new_element` overloads for `FlattenInterfacePtr` provide backward compatibility via thin wrappers.
- `graph_bucket_searcher.cpp` correctly clamps `ef` to `bucket_size` and raises it to at least `topk` for KNN search.

LHT129 (commented):
Overall this is a well-executed refactor. The introduction of `DistanceProviderForGraph` cleanly decouples `BasicSearcher` from `FlattenInterface`, and the extraction of pruning strategies into `pruning_strategy.cpp` is a good separation of concerns. The previous review comments about null factory and division-by-zero have been addressed in this commit.

One minor observation: in `graph_bucket_searcher.cpp`, the `result_limit` variable defaults to `UINT64_MAX` and the final `while (heap->Size() > result_limit)` loop is effectively a no-op for unlimited RANGE_SEARCH. This is correct but the intent could be clearer with an explicit guard condition.

wxyucs (approved):
lgtm

PR #2623, opened 2026-08-05
fix(pyramid): stop range trimming at an empty heap

### LINKED ISSUE #2565: [bug](pyramid): RangeSearch can loop after trimming the final result
## Description

`Pyramid::search_impl` can loop indefinitely during range-result trimming when every candidate is outside the radius. The loop pops the final record and then evaluates `search_result->Top()` again without checking whether the heap became empty.

The problem surfaced in PR #2531 CI, but the failing test does not use the threshold feature and the stack is in the pre-existing Pyramid range-search path. It should therefore be fixed independently.

## CI evidence

- Workflow run: https://github.com/antgroup/vsag/actions/runs/30346299772
- Job: https://github.com/antgroup/vsag/actions/runs/30346299772/job/90270588631?pr=2531
- Job started: 2026-07-28 13:00:23 UTC
- `Run Functests` started: 2026-07-28 13:05:50 UTC
- Job ended exactly six hours after start at 19:00:27 UTC with `context canceled`, matching the GitHub Actions default 360-minute job timeout.
- The workflow has `cancel-in-progress: true`, but no newer run superseded this head while the job was executing. There is no job-level `timeout-minutes`.
- All logged tests completed except `Pyramid Build & ContinueAdd Test`, which began at 13:06:25 UTC and emitted no completion before timeout.
- The PR iterator-threshold regression completed in 31 ms, so it was not the stuck test.

## Suspected root cause

`src/algorithm/pyramid/pyramid.cpp` trims results with:

```cpp
while (search_result->Size() > search_param.topk ||
       search_result->Top().first > search_param.radius) {
    search_result->Pop();
}
```

For range search, `topk` may be effectively unlimited. If all records exceed the radius, the final `Pop()` empties the heap and the next condition calls `Top()` on the empty heap. The observed implementation then makes no forward progress. The loop needs an emptiness guard before reading `Top()`.

## Local reproduction

Command, always bounded externally:

```bash
timeout --signal=TERM --kill-after=10s 120s \
  ./build/tests/functests "Pyramid Build & ContinueAdd Test" -d yes --order rand
```

The test is intermittent: most runs complete in 6-42 seconds, but two local repetitions remained live past their 45/120-second watchdogs. Captured combinations included:

- `metric_type=l2, dim=7, use_reorder=true, immutable=false`
- `metric_type=cosine, dim=9, use_reorder=true, immutable=false`

A gdb `thread apply all bt` captured the main thread in:

```text
Pyramid::search_impl
Pyramid::RangeSearch
InnerIndexInterface::RangeSearch
IndexImpl<Pyramid>::RangeSearch
fixtures::TestIndex::TestFilterSearch
tests/test_pyramid.cpp:244
```

OpenBLAS worker threads were idle on condition variables. This points to the main Pyramid trimming loop rather than worker-pool deadlock.

## Suggested regression

Add a deterministic Pyramid range-search case where every candidate is above the radius, assert an empty result, and run it under the normal functest timeout.

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2565

## What Changed

- Stop Pyramid range-result trimming when the result heap becomes empty before reading `Top()` again.
- Add a deterministic regression where every candidate lies outside the requested radius and verify that `RangeSearch` returns an empty result.

### Root cause

`Pyramid::search_impl` checked `search_result->Top()` after popping the final candidate. When all candidates were outside the radius, this accessed and popped an empty heap, causing undefined behavior that could leave the functional test spinning until the CI job timeout.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (described below)

Test details:

```text
# Before the fix: deterministic regression remained live and was interrupted after 17.8 s
./build/tests/functests "Pyramid RangeSearch Returns Empty Result" -d yes

# After the fix
./build/tests/functests "Pyramid RangeSearch Returns Empty Result" -d yes
All tests passed (6 assertions in 1 test case)

./build/tests/functests "Pyramid Build & ContinueAdd Test" -d yes
All tests passed (48144 assertions in 1 test case)

/opt/homebrew/Cellar/llvm@15/15.0.7/bin/clang-format --dry-run --Werror \
  src/algorithm/pyramid/pyramid.cpp tests/test_pyramid.cpp
git diff upstream/main...HEAD --check
```

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: `RangeSearch` now returns an empty dataset instead of accessing an empty heap when all candidates are outside the radius.

- Performance impact: negligible; one emptiness check is added to the existing result-trimming loop.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other:

## Risk and Rollback

- Risk level: low
- Rollback plan: revert the fix commit.

## Checklist

- [x] I have added/updated tests for the bug fix.
- [x] No documentation update is required.

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -551,8 +551,8 @@ Pyramid::search_impl(const DatasetPtr& query,
         return DatasetImpl::MakeEmptyDataset();
     }
 
-    while (search_result->Size() > search_param.topk ||
-           search_result->Top().first > search_param.radius) {
+    while (not search_result->Empty() && (search_result->Size() > search_param.topk ||
+                                          search_result->Top().first > search_param.radius)) {
         search_result->Pop();
     }
 
--- a/tests/test_pyramid.cpp
+++ b/tests/test_pyramid.cpp
@@ -248,6 +248,23 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::PyramidTestIndex,
 }
 
 TEST_CASE_PERSISTENT_FIXTURE(fixtures::PyramidTestIndex,
+                             "Pyramid RangeSearch Returns Empty Result",
+                             "[ft][pyramid]") {
+    PyramidParam pyramid_param;
+    const auto param = GeneratePyramidBuildParametersString("l2", 4, pyramid_param);
+    auto index = TestFactory("pyramid", param, true);
+
+    auto base = MakeDenseDataset(
+        {{{1.0F, 0.0F, 0.0F, 0.0F}}, {{2.0F, 0.0F, 0.0F, 0.0F}}}, {1, 2}, {"a/d/f", "a/d/f"});
+    REQUIRE(index->Build(base).has_value());
+
+    auto query = MakeSingleQuery({10.0F, 0.0F, 0.0F, 0.0F}, "a/d/f");
+    auto result = index->RangeSearch(query, 0.0F, GeneratePyramidSearchParametersString(20));
+    REQUIRE(result.has_value());
+    REQUIRE(result.value()->GetDim() == 0);
+}
+
+TEST_CASE_PERSISTENT_FIXTURE(fixtures::PyramidTestIndex,
                              "Pyramid Set Immutable",
                              "[ft][immutable][pyramid]") {
     const auto metric_type = "l2";


LHT129 (commented):
The fix correctly addresses the root cause: when all candidates lie outside the radius, `search_result->Top()` was called on an empty heap after the last `Pop()`, causing undefined behavior. Adding `not search_result->Empty()` to the while condition is the right minimal fix.

The regression test covers the exact scenario — all candidates outside a radius of 0.0 — and verifies an empty result is returned. The change is low-risk, well-scoped, and includes a proper test.

No issues found.

LHT129 (commented):
## Code Review Summary

This PR fixes a bug where `Pyramid::search_impl` could access `Top()` on an empty heap during range-search result trimming, causing undefined behavior.

### Changes Reviewed

**`src/algorithm/pyramid/pyramid.cpp`** — The fix is correct and minimal. Adding `not search_result->Empty()` to the while-loop condition uses short-circuit evaluation to prevent accessing `Top()` on an empty heap. The existing early-return at line 550 only guards against an initially empty heap; the bug occurs when the heap becomes empty *during* the trimming loop (all candidates outside the radius). The fix correctly addresses this gap.

**`tests/test_pyramid.cpp`** — The new regression test `"Pyramid RangeSearch Returns Empty Result"` constructs a well-chosen scenario: two data points at L2 distances 81 and 64 from the query, with `radius=0.0`, so every candidate is outside the radius. The assertion `GetDim() == 0` correctly validates that an empty result is returned instead of triggering UB.

### Assessment

- **Correctness**: The fix is logically sound and addresses the root cause described in the PR body.
- **Test coverage**: The new test is deterministic and covers the exact regression scenario.
- **Code style**: Uses `not` instead of `!`, consistent with the existing codebase. No style issues.
- **Performance**: One additional emptiness check per loop iteration — negligible overhead.
- **Minimality**: Only 2 lines changed in the fix, plus a focused test case.

No issues found. This is a clean, well-scoped bug fix.

LHT129 (approved):
LGTM

PR #2628, opened 2026-08-06
fix: align MCI build and search semantics

LightWant:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

## What Changed

- Restore the intended MCI maximal-clique P-extension and coverage semantics used by PR-2385.
- Align MCI seed sampling and search initialization with the validated PR behavior.
- Add `mci_knng_source` so HGraph MCI construction can explicitly use the existing HGraph KNNG or a dedicated ODescent KNNG; the default remains `hgraph`.
- Reuse the ODescent build path for MCI, avoid repeated flatten-code lookups in its distance hot path, and propagate worker exceptions to the caller.
- Document the new build parameter in the English and Chinese HGraph MCI companion guides.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
clang-format-15 --dry-run --Werror <all changed C++ files>
git diff --check

unittests '[hgraph][mci]'       # 90 assertions, 14 test cases
unittests '[HGraphParameter]'   # 112 assertions, 18 test cases
unittests '[ODescent]'          # 6064 assertions, 1 test case
unittests '[pyramid_param]'     # 7 assertions, 1 test case

All selected tests passed on top of origin/main (41253fd7).
```

The codefilter-3m investigation also compared HGraph- and ODescent-sourced MCI indexes and verified the corrected MCI route and seed count during filtered search. Existing benchmark indexes were loaded read-only and were not rebuilt or deleted as part of PR preparation.

## Compatibility Impact

- API/ABI compatibility: compatible; this adds an optional string build parameter and constants without changing existing signatures.
- Behavior changes: MCI clique construction and seed selection follow the intended PR-2385 semantics. Existing configurations continue to use the HGraph KNNG by default.

- Performance impact: restores the lower-overhead MCI seed/search behavior and provides the optimized ODescent KNNG build option.
- Concurrency/thread-safety impact: MCI build workers retain parallel execution; exceptions are captured and rethrown after joining instead of terminating the process.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: English and Chinese HGraph MCI companion documentation

## Risk and Rollback

- Risk level: medium; the change intentionally affects MCI index construction and filtered-search seed selection.
- Rollback plan: revert commit `766e732f` and rebuild indexes created with the new behavior if rollback is required.

## Checklist

- [x] Linked issue is not required for `kind/improvement`
- [x] I have updated docs for the new build parameter


src/impl/searcher/mci_searcher.cpp
```diff
@@ -53,14 +53,11 @@ struct MCIEpochMarks {
```
LHT129, line 53: [critical] Removing the `id < marks.size()` bounds check in `MCIEpochMarks::Get()` introduces an out-of-bounds vector access when `Get` is called with an `id` that has not been previously `Set` (and therefore the vector has not been resized to accommodate it).

The old code safely returned `false` for out-of-range ids. The new code will perform an unchecked `marks[id]` access, which is undefined behavior if `id >= marks.size()`.

The `Set` method also lost its bounds check and resize logic, meaning `Set` now also performs an unchecked access. While `Reset` pre-sizes the vector, `Set` can be called independently of `Reset` (e.g., during incremental updates), making this unsafe.

Please restore the bounds checks in both `Get` and `Set`, or ensure the vector is always pre-sized before any `Get`/`Set` call.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -245,7 +243,7 @@ MCISearcher::Search(const CliqueDataCellPtr& cliques,
                     const InnerSearchParam& inner_search_param,
```
LHT129, line 246: [suggestion] The change from `select_query_allocator(ctx, allocator_)` to `allocator_` means per-query memory allocators provided via `QueryContext` are now ignored. This is a behavioral regression: callers that pass a custom allocator through `QueryContext::alloc` will no longer have it used for MCI search allocations.

If this was intentional (e.g., to avoid allocator lifetime issues), please document the rationale in the PR description or a comment. Otherwise, consider restoring `select_query_allocator(ctx, allocator_)`.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -51,23 +56,14 @@ collect_seed_inner_ids(const FilterPtr& filter,
         return inner_ids;
     }
 
-    const auto target_count = std::min<uint64_t>(seed_count, static_cast<uint64_t>(valid_count));
-    inner_ids.reserve(target_count);
-    UnorderedSet<InnerIdType> selected_ids(allocator);
-    selected_ids.reserve(target_count);
-    auto try_add_label = [&](uint64_t offset) {
+    const auto sampled_count = std::min<uint64_t>(seed_count, static_cast<uint64_t>(valid_count));
```
LHT129, line 59: [suggestion] The old `collect_seed_inner_ids` used an `UnorderedSet` to deduplicate inner IDs during sampling, preventing duplicate entries in the returned vector. The new code removes this deduplication, relying only on the final `std::sort` + `std::unique` + `erase` to remove duplicates.

While the end result is the same (duplicates are removed at the end), the intermediate vector may grow larger than necessary when many valid labels map to the same inner ID. Consider whether the deduplication during sampling was important for memory usage in large filtered sets, or if the simplification is acceptable.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -682,6 +678,64 @@ build_mci_knn_graph(uint64_t total,
...
+                                int64_t dim,
+                                int64_t extra_info_size,
+                                const std::shared_ptr<SafeThreadPool>& thread_pool,
+                                Allocator* allocator) {
+    MCIKNNGraph graph(allocator);
+    graph.total = total;
+    if (total <= 1) {
+        return graph;
+    }
+
```
LHT129, line 698: [note] `std::shared_ptr<Allocator>(allocator, [](Allocator*) {})` creates a non-owning shared_ptr that aliases a raw pointer. If the original allocator is deallocated before the `graph_common_param` (which holds this shared_ptr), this results in a dangling pointer.

This is safe as long as `graph_common_param` does not outlive the calling scope (which appears to be the case here since it is stack-allocated and `graph_storage` is used only within this function). However, this pattern is fragile and could break if the code is refactored. Consider adding a brief comment noting the lifetime dependency.

src/impl/searcher/mci_searcher.cpp
LightWant: Fixed in cd656cec. Restored the out-of-range check in MCIEpochMarks::Get() and the resize guard in Set(), so an unexpected ID cannot index past the marks buffer.

src/impl/searcher/mci_searcher.cpp
LightWant: Fixed in cd656cec. MCISearcher now uses select_query_allocator(ctx, allocator_) again, preserving the request-scoped allocator contract.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -155,11 +154,22 @@ BuildMCICliques(const float* vectors,
     double max_norm_error = 0.0;
     {
         std::atomic<uint64_t> next_id{0};
-        std::vector<std::future<void>> workers;
-        workers.reserve(thread_count);
+        Vector<std::thread> workers(allocator);
```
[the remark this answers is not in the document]
LightWant: Fixed in cd656cec. Both native-thread worker pools now reserve their storage and install an RAII join guard before starting threads. If a later thread construction throws, every already-started joinable thread is joined during unwinding.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -51,23 +56,14 @@ collect_seed_inner_ids(const FilterPtr& filter,
...
     }
 
-    const auto target_count = std::min<uint64_t>(seed_count, static_cast<uint64_t>(valid_count));
-    inner_ids.reserve(target_count);
-    UnorderedSet<InnerIdType> selected_ids(allocator);
-    selected_ids.reserve(target_count);
-    auto try_add_label = [&](uint64_t offset) {
+    const auto sampled_count = std::min<uint64_t>(seed_count, static_cast<uint64_t>(valid_count));
+    inner_ids.reserve(sampled_count);
+    for (uint64_t i = 0; i < sampled_count; ++i) {
```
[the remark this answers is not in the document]
LightWant: Fixed in cd656cec using the fallback option. PR-2385 sampling semantics remain unchanged, but if sampled labels produce no mapped seed, MCI now falls back to HGraph instead of returning an incorrect empty result. Regression coverage checks mixed stale/valid and all-stale sets.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -134,7 +133,7 @@ BuildMCICliques(const float* vectors,
     const auto candidate_limit =
         std::min<uint64_t>({params.candidate_limit, graph.row_stride, total - 1});
     const auto clique_threshold =
-        std::min<uint64_t>({K_MCI_MIN_CLIQUE_SIZE, candidate_limit + 1, total});
+        std::max<uint64_t>(2, std::min<uint64_t>({params.clique_max, candidate_limit + 1, total}));
```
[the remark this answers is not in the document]
LightWant: Fixed in cd656cec. Full-build cliques are capped before coverage accounting, preserving the seed during truncation, and final materialization also enforces clique_max for fallback cliques. The regression test asserts every stored clique stays within the configured cap.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -188,8 +201,11 @@ BuildMCICliques(const float* vectors,
                  max_norm_error);
 
     auto distance_from_dot = [&](InnerIdType lhs, InnerIdType rhs, float dot) {
-        if (params.metric == MetricType::METRIC_TYPE_L2SQR) {
-            return FP32ComputeL2Sqr(VectorAt(vectors, dim, lhs), VectorAt(vectors, dim, rhs), dim);
+        if (params.metric == MetricType::METRIC_TYPE_L2SQR or
+            (params.metric == MetricType::METRIC_TYPE_COSINE and unit_norm)) {
+            const float distance =
```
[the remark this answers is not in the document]
LightWant: Fixed in cd656cec. L2 now uses FP32ComputeL2Sqr for scalar comparisons and FP32ComputeL2SqrBatch4 for batched comparisons; only cosine and IP retain dot-product conversion. This avoids cancellation for nearby vectors with large common coordinates.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -245,7 +243,7 @@ MCISearcher::Search(const CliqueDataCellPtr& cliques,
...
                     const MCISearcherParam& mci_param,
                     QueryContext* ctx) const {
-    auto* alloc = select_query_allocator(ctx, allocator_);
+    auto* alloc = allocator_;
```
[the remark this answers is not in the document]
LightWant: Fixed in cd656cec. Restored select_query_allocator(ctx, allocator_) so MCI heaps and candidate buffers honor SearchRequest::search_allocator_.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -341,7 +389,7 @@ BuildMCICliques(const float* vectors,
                         ExpandDistanceLimit(candidates[0].distance, now_alpha, params.metric);
                     stage_start = std::chrono::steady_clock::now();
                     for (uint64_t i = 0; i < candidate_size; ++i) {
-                        if (candidates[i].distance <= distance_limit) {
+                        if (candidates[i].distance < distance_limit) {
```
LHT129, line 392: [note] The seed-to-candidate edge inclusion uses strict less-than (`candidates[i].distance < distance_limit`), but the candidate-to-candidate pair-wise edge inclusion still uses less-than-or-equal (`distance <= distance_limit`). This asymmetry means a candidate at exactly `distance_limit` will be connected to other candidates but not to the seed itself.

This is likely intentional (to avoid self-edges at the boundary), but the inconsistency could be confusing to future readers. Consider adding a brief comment explaining the rationale, or unifying the comparison operators if the asymmetry is unintentional.

LightWant: This asymmetry is intentional in this PR. The goal is to preserve the PR-2385 clique edge-set semantics that were used for the build/search performance and recall validation. Changing either comparator would alter clique membership and requires a separate benchmarked algorithm change, so it is deliberately left unchanged here.

LHT129 (commented):
## Code Review Summary

The second commit `cd656cec` ("fix: address MCI review findings") addresses most of the prior review findings from the first commit. Below is a summary of the key changes and remaining observations.

### Addressed Review Findings

1. **`MCIEpochMarks` bounds checks** (`src/impl/searcher/mci_searcher.cpp`): Restored `id < marks.size()` check in `Get()` and `id >= marks.size()` resize in `Set()`. This was the critical issue from the prior review.

2. **`select_query_allocator`** (`src/impl/searcher/mci_searcher.cpp`): Restored `select_query_allocator(ctx, allocator_)` usage, fixing the allocator regression.

3. **L2SQR computation** (`src/algorithm/mci/mci_builder.cpp`): Restored the direct `FP32ComputeL2Sqr` path in `compute_distance`, avoiding numerical cancellation issues in the `distance_from_dot` path for squared-L2.

4. **Thread exception safety** (`src/algorithm/mci/mci_builder.cpp`): Added `MCIThreadJoinGuard` RAII class and `catch_worker_exception` wrapper for both worker pools. Worker exceptions are now properly captured and rethrown.

5. **`clique_max` truncation** (`src/algorithm/mci/mci_builder.cpp`): Added `params.clique_max` truncation in `append_selected_clique` to cap stored clique sizes.

6. **P-extension semantics** (`src/algorithm/mci/mci_mce.cpp/h`): Updated `try_save_must_clique`, `state_has_must`, and `clear_must_node` to include P-extension nodes (`EP` parameter), ensuring must-contain cliques correctly account for the full clique structure.

7. **`check_overtime` optimization** (`src/impl/searcher/mci_searcher.cpp`): Added `const bool check_overtime = inner_search_param.time_cost != nullptr` guard to avoid unnecessary `mci_check_overtime` calls when timeout is not configured.

### New Feature: ODescent KNNG Source

The new `mci_knng_source` parameter allows selecting between HGraph-derived KNNG (`"hgraph"`, default) and a dedicated ODescent-built KNNG (`"odescent"`). The implementation in `build_mci_knn_graph_by_odescent` is clean and well-structured. Parameter validation in `hgraph_parameter.cpp` correctly restricts values to the two valid options.

### Remaining Observations

**[note]** In `build_mci_knn_graph_by_odescent` (`src/algorithm/hgraph/hgraph_mci.cpp`), the `std::shared_ptr<Allocator>(allocator, [](Allocator*) {})` creates a non-owning shared_ptr. This is safe in the current context since `graph_storage` is local and destroyed before the function returns. However, the pattern is fragile — if `graph_storage` were ever stored beyond the function scope, it would dangle. A brief comment noting the intentional non-owning semantics would help future readers.

### Overall Assessment

The second commit is a solid follow-up that addresses the key review findings. The new ODescent KNNG source feature is well-implemented with proper parameter validation, test coverage, and documentation. No blocking issues remain.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -155,11 +172,24 @@ BuildMCICliques(const float* vectors,
     double max_norm_error = 0.0;
     {
         std::atomic<uint64_t> next_id{0};
-        std::vector<std::future<void>> workers;
+        Vector<std::thread> workers(allocator);
         workers.reserve(thread_count);
+        MCIThreadJoinGuard join_guard(workers);
```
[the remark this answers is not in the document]
LightWant: Fixed in ecda27b4. In both worker pools, all mutexes, exception state, and the wrapper lambda are now declared before MCIThreadJoinGuard. During exception unwinding the guard is therefore destroyed first and joins all started workers before any referenced state is destroyed.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -792,12 +849,27 @@ HGraph::build_mci_clique_index(const void* vectors) {
...
-    auto graph = build_mci_knn_graph(total,
-                                     this->mci_parameters_.mcs,
-                                     this->mci_parameters_.knng_path,
-                                     worker_count,
-                                     this->allocator_,
-                                     search_neighbors);
+    MCIKNNGraph graph(this->allocator_);
+    if (this->mci_parameters_.knng_path.empty() and
+        this->mci_parameters_.knng_source == HGRAPH_MCI_KNNG_SOURCE_ODESCENT) {
+        graph = build_mci_knn_graph_by_odescent(this->basic_flatten_codes_,
```
LHT129, line 855: [suggestion] `basic_flatten_codes_` is passed to `build_mci_knn_graph_by_odescent` without a nullptr check. If it is null, `ODescent::Build()` will crash at `flatten_interface_->TotalCount()`. The hgraph KNNG path above is guarded by `CHECK_ARGUMENT(precise_codes != nullptr, ...)`, but the ODescent path has no equivalent guard. Consider adding a `CHECK_ARGUMENT(this->basic_flatten_codes_ != nullptr, ...)` before the ODescent branch, or early-return if the pointer is null.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -188,8 +223,9 @@ BuildMCICliques(const float* vectors,
                  max_norm_error);
 
     auto distance_from_dot = [&](InnerIdType lhs, InnerIdType rhs, float dot) {
-        if (params.metric == MetricType::METRIC_TYPE_L2SQR) {
-            return FP32ComputeL2Sqr(VectorAt(vectors, dim, lhs), VectorAt(vectors, dim, rhs), dim);
+        if (params.metric == MetricType::METRIC_TYPE_COSINE and unit_norm) {
+            const float distance = 2.0F - 2.0F * dot;
```
LHT129, line 227: [note] When `unit_norm=true` and the metric is COSINE, `distance_from_dot` returns `2.0F - 2.0F * dot` (i.e. `2 * (1 - dot)`) instead of the canonical cosine distance `1 - dot`. This is safe under the current implementation because `ExpandDistanceLimit` multiplies by `alpha` and edge filtering uses strict `<`, so the factor of 2 cancels out in comparisons. However, this relies on an implicit scaling invariant — if `ExpandDistanceLimit` or distance threshold logic changes in the future, the doubled distance could cause subtle bugs. Consider adding a brief comment noting the intentional scaling, or using `1.0F - dot` directly for clarity.

src/impl/odescent/odescent_graph_builder.h
```diff
@@ -189,6 +188,7 @@ class ODescent {
     // afterward. Task enqueue publishes the initialized pointer to workers, and each parallel phase
     // waits on all futures before continuing, so get_distance() only performs concurrent reads.
     FlattenInterfacePtr build_flatten_interface_{nullptr};
+    FlattenInterface* distance_flatten_interface_{nullptr};
```
LHT129, line 191: [note] The new `distance_flatten_interface_` raw pointer in `ODescent` aliases either `build_flatten_interface_.get()` or `flatten_interface_.get()`. The class comment states it is never mutated after `Build()` initializes it, which makes concurrent reads safe. However, `build_flatten_interface_` is a `FlattenInterfacePtr` (shared_ptr) that could theoretically be reset if someone adds a setter or if the constructor-provided shared_ptr is reassigned externally. The raw pointer would then dangle.

This is safe under the current code because `prepare_build_flatten()` is called exactly once at the start of `Build()` and `distance_flatten_interface_` is never reassigned afterward. Consider adding a brief comment on `distance_flatten_interface_` noting that its lifetime is tied to `build_flatten_interface_` / `flatten_interface_` and it must not outlive them.

wxyucs (approved):
lgtm

PR #2629, opened 2026-08-06
fix: bound DiskANN closest-center block memory

Hastyshell:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- None (kind/improvement PR; issue link not required)

## What Changed

- Bound `compute_closest_centers` temporary memory usage by processing points in fixed-size blocks instead of allocating a full `num_centers * num_points` distance matrix.
- Added no-op guards for empty input and zero closest-center requests before allocating temporary buffers.
- Exposed the closest-center block size as a shared internal `math_utils` constant so implementation and cross-block regression tests stay in sync.
- Added regression tests for empty input, zero closest-center requests, and cross-block closest-center assignment, including verification that inverted lists keep global point ids.

## Test Evidence

- [x] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

## Compatibility Impact

- API/ABI compatibility: no public VSAG API change; exposes an internal DiskANN math_utils block-size constant used by the implementation and tests.
- Behavior changes: `compute_closest_centers` now avoids excessive temporary memory for large datasets and treats empty/zero-closest-center requests as no-op.

- Performance impact: reduces peak temporary memory during closest-center assignment; large inputs are processed in 64K-point blocks.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other:

## Risk and Rollback

- Risk level: low
- Rollback plan: revert this commit to restore the previous single-block closest-center allocation behavior.

## Checklist

- [ ] I have linked the relevant issue (required for `kind/bug` and `kind/feature`; see "Linked Issue" above)

### DIFF
--- a/extern/diskann/DiskANN/src/math_utils.cpp
+++ b/extern/diskann/DiskANN/src/math_utils.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 #include <limits>
+#include <memory>
 #ifdef __APPLE__
 #include <cstdlib>
 #else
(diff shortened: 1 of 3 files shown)


extern/diskann/DiskANN/src/math_utils.cpp
```diff
@@ -161,53 +162,59 @@ void compute_closest_centers(float *data, uint64_t num_points, uint64_t dim, flo
...
-        pts_norms_squared = new float[num_points];
+    {
+        owned_pts_norms_squared.reset(new float[num_points]);
+        pts_norms_squared = owned_pts_norms_squared.get();
+    }
 
-    uint64_t PAR_BLOCK_SIZE = num_points;
-    uint64_t N_BLOCKS =
-        (num_points % PAR_BLOCK_SIZE) == 0 ? (num_points / PAR_BLOCK_SIZE) : (num_points / PAR_BLOCK_SIZE) + 1;
+    constexpr uint64_t kMaxClosestCenterBlockSize = 65536;
```
LHT129: [suggestion] The magic number `65536` for `kMaxClosestCenterBlockSize` lacks justification. Consider adding a brief comment explaining why this specific value was chosen (e.g., memory budget constraint, L2/L3 cache size consideration, or empirical tuning result). This helps future maintainers understand whether the constant can be safely adjusted.

```cpp
// Bound block size to 65536 to limit peak memory usage of the
// per-block distance_matrix (num_centers * block_size floats)
constexpr uint64_t kMaxClosestCenterBlockSize = 65536;
```

extern/diskann/DiskANN/src/math_utils.cpp
```diff
@@ -161,53 +162,59 @@ void compute_closest_centers(float *data, uint64_t num_points, uint64_t dim, flo
...
 
+    if (num_points == 0)
+    {
+        return;
+    }
+
     bool is_norm_given_for_pts = (pts_norms_squared != NULL);
 
-    float *pivs_norms_squared = new float[num_centers];
+    std::unique_ptr<float[]> pivs_norms_squared(new float[num_centers]);
```
LHT129, line 172: [note] `new float[num_centers]` is called unconditionally. If `num_centers == 0`, this allocates a zero-length array, which is implementation-defined behavior in C++ (before C++26). While this is pre-existing and the early `k > num_centers` check at the top of the function guards against `k == 0` but not `num_centers == 0`, consider whether an additional guard is warranted. Not a blocker for this PR since the behavior is unchanged from the original code.

LHT129 (commented):
Overall this is a clean and well-executed refactor. The block-based processing correctly bounds temporary memory, the switch to `std::unique_ptr` eliminates manual delete calls, and the tests cover empty input and cross-block correctness.

One minor observation (not blocking): the cross-block test only exercises `k = 1`. The inner loop indexing pattern is the same for any `k`, so this is low risk, but a `k = 3` variant would give full confidence in the multi-block + `k > 1` path.

LGTM otherwise.

extern/diskann/DiskANN/src/math_utils.cpp
```diff
@@ -161,53 +162,61 @@ void compute_closest_centers(float *data, uint64_t num_points, uint64_t dim, flo
         return;
     }
 
+    if (num_points == 0)
+    {
```
LHT129, line 166: [suggestion] The early-return guard only checks `num_points == 0` but not `k == 0`. When `k == 0` and `num_points > 0`, the subsequent `new uint32_t[par_block_size * k]` allocates a zero-length array, which is implementation-defined behavior in C++. While the inner `for (uint64_t l = 0; l < k; l++)` loop body would not execute, the allocation itself is problematic.

Consider extending the guard to also cover `k == 0`:

```cpp
if (num_points == 0 || k == 0)
{
    return;
}
```

This also avoids the unnecessary `compute_vecs_l2sq` calls and GEMM work when there are no closest centers to compute.

src/index/diskann_test.cpp
```diff
@@ -40,6 +44,178 @@ parse_diskann_params(vsag::IndexCommonParam index_common_param) {
...
+    REQUIRE(closest_centers[0] == 12345U);
+}
+
+TEST_CASE("compute_closest_centers matches single block reference", "[ut][diskann]") {
+    constexpr uint64_t max_block_size = 65536;
+    constexpr uint64_t num_points = max_block_size + 11;
+    constexpr uint64_t dim = 5;
+    constexpr uint64_t num_centers = 32;
+    constexpr uint64_t k = 1;
+
```
LHT129, line 240: [note] The test hard-codes `max_block_size = 65536` which duplicates the production constant `kMaxClosestCenterBlockSize` in `math_utils.cpp`. If the production constant is ever changed, this test would still pass but would no longer verify the cross-block boundary correctly. Consider exposing the block size as a visible constant (e.g. in a header) or at least adding a comment linking the two values.

```cpp
// Must match kMaxClosestCenterBlockSize in math_utils.cpp
constexpr uint64_t max_block_size = 65536;
```

LHT129 (commented):
The latest commit (89781de) addresses all previous review comments:

- `kMaxClosestCenterBlockSize` is now exposed in the header, making it visible to tests and other consumers.
- The early-return guard now covers both `num_points == 0` and `k == 0`, avoiding zero-length allocations.
- The test references `math_utils::kMaxClosestCenterBlockSize` directly instead of duplicating the magic number.

The block-based processing logic, RAII memory management with `unique_ptr`, and test coverage all look solid. No further issues to flag.

LHT129 (commented):
The latest commit (89781de) addresses all prior review feedback:

- `kMaxClosestCenterBlockSize` is now a named `constexpr` in the header with a descriptive comment.
- The early-return guard now covers both `num_points == 0` and `k == 0`.
- The test references `math_utils::kMaxClosestCenterBlockSize` directly instead of duplicating the magic number.
- `std::unique_ptr` is used consistently for RAII-managed allocations.

The block-based processing logic is correct: `par_block_size` is properly bounded, `n_blocks` uses standard ceiling division, and the inner loop correctly maps local block indices to global point ids for both `closest_centers_ivf` and the inverted index. The tests cover empty input, zero-k, and cross-block boundary scenarios.

No blocking issues remain.

Hastyshell:
@jiaweizone @wxyucs @inabao PTAL

LHT129 (commented):
Overall this PR looks good. The changes correctly bound `compute_closest_centers` temporary memory by processing points in fixed-size blocks (64K), replacing raw `new`/`delete[]` with `std::unique_ptr` for RAII safety, and adding proper no-op guards for empty/zero inputs. The three new test cases cover the key scenarios (empty input, zero k, cross-block correctness).

The three issues raised in earlier review comments have been addressed:
- `kMaxClosestCenterBlockSize` is now a named constant in `math_utils.h` with a comment
- The `k == 0` guard has been added alongside `num_points == 0`
- The test now references `math_utils::kMaxClosestCenterBlockSize` instead of a hard-coded duplicate

No blocking issues found.

wxyucs (approved):
lgtm

PR #2639, opened 2026-08-07
feat(ivf): support multi-query batch search with bucket_ids bypass

LHT129:
## Summary

Follow-up to #2597. Relaxes the bucket_ids_ constraint from exactly 1 query to matching the query count, and adds a batch KNN path for IVF SearchWithRequest.

Fixes: #2597

## Changes

- IndexImpl validation: outer bucket_ids_ size must match query_->GetNumElements(); non-IVF indexes still restricted to 1 entry
- IVF batch path: per-query recursive call to SearchWithRequest with 1-row non-owning Dataset view plus corresponding bucket list, aggregated into rectangular Dataset(NumElements=num_queries, Dim=topk)
- Padding: missing neighbors filled with id=-1, distance=+inf
- Restrictions: batch IVF rejects range search, custom distance callback, and reasoning labels
- Tests: batch correctness vs individual single-query results, query-count mismatch rejection, batch default routing
- Docs: updated en/zh API reference for query_ and bucket_ids_

### DIFF
--- a/docs/docs/en/src/api/search.md
+++ b/docs/docs/en/src/api/search.md
@@ -34,7 +34,7 @@ enum class SearchMode {
 
 | Field | Type | Default | Meaning |
 |-------|------|---------|---------|
-| `query_` | `DatasetPtr` | `nullptr` | The query. Exactly one query vector is allowed. |
+| `query_` | `DatasetPtr` | `nullptr` | The query. IVF KNN requests support multiple query vectors; other requests allow one. |
 | `mode_` | `SearchMode` | `KNN_SEARCH` | KNN vs. range search. |
 | `topk_` | `int64_t` | `10` | Neighbors to return (KNN mode). Must be positive. |
 | `radius_` | `float` | `0.5` | Distance threshold (range mode). Non-negative. |
@@ -101,9 +101,13 @@ selection and scans only the specified buckets.
 **Semantics:**
 - **Empty** (default): Use normal bucket routing via `ClassifyDatasForSearch`.
 - **Non-empty**: Skip routing and scan only the provided bucket IDs.
+- **Batch IVF KNN**: One outer `bucket_ids_` entry per query vector.
+  Result is rectangular: `NumElements()` is query count, `Dim()` is `topk_`.
+  Missing neighbors are `-1` with infinite distance.
 
 **Constraints:**
-- Currently only single-query is supported; the outer vector must contain exactly one entry.
+- Batch IVF search supports KNN only; custom query distance and reasoning labels are unsupported.
+- A non-empty outer vector must contain exactly one non-empty entry per query vector.
 - Each bucket ID must be in range `[0, bucket_count)`.
 - Duplicate bucket IDs are rejected.
 - Incompatible with `disable_bucket_scan` mode.
--- a/docs/docs/zh/src/api/search.md
+++ b/docs/docs/zh/src/api/search.md
@@ -34,7 +34,7 @@ enum class SearchMode {
 
 | 字段 | 类型 | 默认值 | 含义 |
 |------|------|--------|------|
-| `query_` | `DatasetPtr` | `nullptr` | 查询。只允许恰好一个查询向量。 |
+| `query_` | `DatasetPtr` | `nullptr` | 查询。IVF KNN 请求支持多个查询向量，其他请求只允许一个。 |
 | `mode_` | `SearchMode` | `KNN_SEARCH` | KNN 还是范围搜索。 |
 | `topk_` | `int64_t` | `10` | 要返回的邻居数（KNN 模式）。必须为正。 |
 | `radius_` | `float` | `0.5` | 距离阈值（范围模式）。非负。 |
@@ -95,9 +95,13 @@ reasoning 选项均会被忽略。
 **语义：**
 - **空**（默认）：使用正常的桶路由（`ClassifyDatasForSearch`）。
 - **非空**：跳过路由，仅扫描提供的桶 ID。
+- **批量 IVF KNN**：`bucket_ids_` 外层每项对应一个查询向量。
+  结果为矩形：`NumElements()` 是查询数，`Dim()` 是 `topk_`。
+  缺失邻居用 `-1` 和无穷距离填充。
 
 **约束：**
-- 当前仅支持单查询；外层向量必须恰好包含一个条目。
+- 批量 IVF 搜索仅支持 KNN；不支持自定义查询距离和 reasoning labels。
+- 非空外层向量必须为每个查询向量提供一个非空条目。
 - 每个桶 ID 必须在 `[0, bucket_count)` 范围内。
 - 重复的桶 ID 会被拒绝。
 - 与 `disable_bucket_scan` 模式不兼容。
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -1855,31 +1855,12 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
     param.query_context = &ctx;
 
     if (not request.bucket_ids_.empty()) {
-        CHECK_ARGUMENT(request.bucket_ids_.size() == 1,
-                       fmt::format("bucket_ids_ currently supports only single-query mode; "
-                                   "expected size 1, got {}",
-                                   request.bucket_ids_.size()));
-        const auto& ids = request.bucket_ids_[0];
-        CHECK_ARGUMENT(not ids.empty(),
-                       "bucket_ids_[0] must not be empty; "
-                       "use empty outer vector for default routing");
-        std::set<int64_t> seen_ids;
-        param.bucket_ids.reserve(ids.size());
-        for (auto id : ids) {
-            CHECK_ARGUMENT(
-                id >= 0,
-                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
-            CHECK_ARGUMENT(
-                id < static_cast<int64_t>(this->bucket_->bucket_count_),
-                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
-            CHECK_ARGUMENT(seen_ids.insert(id).second, fmt::format("duplicate bucket_id {}", id));
-            param.bucket_ids.push_back(id);
-        }
         auto query_check = request.query_;
         CHECK_ARGUMENT(query_check != nullptr, "query dataset cannot be null");
-        CHECK_ARGUMENT(query_check->GetNumElements() == 1,
-                       fmt::format("bucket_ids_ currently supports only single-query mode; "
-                                   "expected 1 query, got {}",
+        CHECK_ARGUMENT(request.bucket_ids_.size() == query_check->GetNumElements(),
+                       fmt::format("bucket_ids_ size must match the number of query vectors; "
+                                   "got {} bucket lists for {} queries",
+                                   request.bucket_ids_.size(),
                                    query_check->GetNumElements()));
         CHECK_ARGUMENT(query_check->GetFloat32Vectors() != nullptr,
                        "query float32 vectors cannot be null");
@@ -1887,6 +1868,29 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
                        "query dimension must match index dimension");
         CHECK_ARGUMENT(not param.disable_bucket_scan,
                        "bucket_ids_ is incompatible with disable_bucket_scan mode");
+        for (uint64_t query_idx = 0; query_idx < request.bucket_ids_.size(); ++query_idx) {
+            const auto& ids = request.bucket_ids_[query_idx];
+            CHECK_ARGUMENT(not ids.empty(),
+                           fmt::format("bucket_ids_[{}] must not be empty; "
+                                       "use empty outer vector for default routing",
+                                       query_idx));
+            std::set<int64_t> seen_ids;
+            for (auto id : ids) {
+                CHECK_ARGUMENT(
+                    id >= 0,
+                    fmt::format(
+                        "bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
+                CHECK_ARGUMENT(
+                    id < static_cast<int64_t>(this->bucket_->bucket_count_),
+                    fmt::format(
+                        "bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
+                CHECK_ARGUMENT(seen_ids.insert(id).second,
+                               fmt::format("duplicate bucket_id {}", id));
+            }
+        }
+        if (query_check->GetNumElements() == 1) {
+            param.bucket_ids.assign(request.bucket_ids_[0].begin(), request.bucket_ids_[0].end());
+        }
     }
 
     auto query = request.query_;
@@ -1908,6 +1912,81 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
         CHECK_ARGUMENT(not request.threshold_.has_value(),
                        "threshold filtering is not supported with disable_bucket_scan");
         auto result = this->route_buckets_only(query, param, ctx);
+        result->Statistics(stats.Dump());
+        return result;
+    }
+
+    if (query != nullptr && query->GetNumElements() > 1) {
+        CHECK_ARGUMENT(not is_range, "IVF batch search only supports KNN search");
+        CHECK_ARGUMENT(not use_custom_distance,
+                       "IVF batch search does not support custom query distance");
+        CHECK_ARGUMENT(request.expected_labels_.empty(),
+                       "IVF batch search does not support expected labels");
+        CHECK_ARGUMENT(request.topk_ > 0, "topk must be greater than 0");
+        CHECK_ARGUMENT(query->GetFloat32Vectors() != nullptr,
+                       "query float32 vectors cannot be null");
+        CHECK_ARGUMENT(query->GetDim() == this->dim_, "query dimension must match index dimension");
+
+        const auto num_queries = query->GetNumElements();
+        const auto total_slots = num_queries * request.topk_;
+        auto* alloc = select_query_allocator(ctx.alloc, this->allocator_);
+        auto* ids = static_cast<int64_t*>(alloc->Allocate(sizeof(int64_t) * total_slots));
+        auto* distances = static_cast<float*>(alloc->Allocate(sizeof(float) * total_slots));
+        std::fill_n(ids, total_slots, -1);
+        std::fill_n(distances, total_slots, std::numeric_limits<float>::infinity());
+
+        const auto* query_data = query->GetFloat32Vectors();
+        auto base_request = request;
+        base_request.expected_labels_.clear();
+        if (not request.bucket_ids_.empty()) {
+            base_request.bucket_ids_.clear();
+        }
+        auto search_func = [&](int64_t query_idx) -> void {
+            auto one_query = Dataset::Make();
+            one_query->NumElements(1)
+                ->Dim(query->GetDim())
+                ->Float32Vectors(query_data + query_idx * query->GetDim())
+                ->Owner(false);
+
+            auto one_request = base_request;
+            one_request.query_ = one_query;
+            if (not request.bucket_ids_.empty()) {
+                one_request.bucket_ids_ = {request.bucket_ids_[query_idx]};
+            }
+            JsonType json = JsonType::Parse(base_request.params_str_);
+            if (json.Contains("ivf")) {
+                json["ivf"]["parallelism"].SetInt64(1);
+            }
+            one_request.params_str_ = json.Dump();
+            auto one_result = this->SearchWithRequest(one_request);
+            const auto count = std::min(request.topk_, one_result->GetDim());
+            if (count > 0) {
+                std::copy_n(one_result->GetIds(), count, ids + query_idx * request.topk_);
+                std::copy_n(
+                    one_result->GetDistances(), count, distances + query_idx * request.topk_);
+            }
+        };
+
+        if (this->thread_pool_ != nullptr and param.parallel_search_thread_count > 1) {
+            std::vector<std::future<void>> futures;
+            for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
+                auto future = this->thread_pool_->GeneralEnqueue(search_func, query_idx);
+                futures.emplace_back(std::move(future));
+            }
+            for (auto& future : futures) {
+                future.get();
+            }
+        } else {
+            for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
+                search_func(query_idx);
+            }
+        }
+        auto result = Dataset::Make()
+                          ->NumElements(num_queries)
+                          ->Dim(request.topk_)
+                          ->Ids(ids)
+                          ->Distances(distances)
+                          ->Owner(true, alloc);
         result->Statistics(stats.Dump());
         return result;
     }
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -1845,31 +1845,12 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
     }
 
     if (not request.bucket_ids_.empty()) {
-        CHECK_ARGUMENT(request.bucket_ids_.size() == 1,
-                       fmt::format("bucket_ids_ currently supports only single-query mode; "
-                                   "expected size 1, got {}",
-                                   request.bucket_ids_.size()));
-        const auto& ids = request.bucket_ids_[0];
-        CHECK_ARGUMENT(not ids.empty(),
-                       "bucket_ids_[0] must not be empty; "
-                       "use empty outer vector for default routing");
-        std::set<int64_t> seen_ids;
-        param.bucket_ids.reserve(ids.size());
-        for (auto id : ids) {
-            CHECK_ARGUMENT(
-                id >= 0,
-                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
-            CHECK_ARGUMENT(
-                id < static_cast<int64_t>(this->bucket_->bucket_count_),
-                fmt::format("bucket_id {} out of range [0, {})", id, this->bucket_->bucket_count_));
-            CHECK_ARGUMENT(seen_ids.insert(id).second, fmt::format("duplicate bucket_id {}", id));
-            param.bucket_ids.push_back(id);
-        }
         auto query_check = request.query_;
         CHECK_ARGUMENT(query_check != nullptr, "query dataset cannot be null");
-        CHECK_ARGUMENT(query_check->GetNumElements() == 1,
-                       fmt::format("bucket_ids_ currently supports only single-query mode; "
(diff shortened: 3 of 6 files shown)


LHT129 (commented):
[suggestion] In `ivf.cpp` batch search loop (around line 1936), the `one_request = request` full copy inside the loop copies all string members (`params_str_`, `attribute_filter_str_`) and the `expected_labels_` vector on every iteration. For large batch sizes this adds unnecessary overhead.

Since only `query_` and `bucket_ids_` differ across iterations, consider moving the copy outside the loop and only updating the fields that change:

```cpp
auto one_request = request;
one_request.expected_labels_.clear();  // already validated empty above

for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
    // ...
    one_request.query_ = one_query;
    if (not request.bucket_ids_.empty()) {
        one_request.bucket_ids_ = {request.bucket_ids_[query_idx]};
    }
    auto one_result = this->SearchWithRequest(one_request);
    // ...
}
```

This avoids repeated copies of `params_str_` and other immutable fields.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,54 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        auto* alloc = select_query_allocator(ctx.alloc, this->allocator_);
+        auto* ids = static_cast<int64_t*>(alloc->Allocate(sizeof(int64_t) * total_slots));
+        auto* distances = static_cast<float*>(alloc->Allocate(sizeof(float) * total_slots));
+        std::fill_n(ids, total_slots, -1);
+        std::fill_n(distances, total_slots, std::numeric_limits<float>::infinity());
+
+        const auto* query_data = query->GetFloat32Vectors();
+        for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
+            auto one_query = Dataset::Make();
+            one_query->NumElements(1)
```
LHT129, line 1946: [suggestion] The full copy of `request` via `auto one_request = request;` inside the per-query loop copies all fields including `expected_labels_`, `bucket_ids_`, `params_str_`, etc. on every iteration. For large batch sizes this adds unnecessary overhead.

Since the batch path already validates that `expected_labels_` is empty and `bucket_ids_` is immediately overwritten, consider constructing `one_request` more selectively.

src/index/index_impl.h
```diff
@@ -500,28 +500,38 @@ class IndexImpl : public Index {
...
-                if (id < 0) {
+            for (size_t query_idx = 0; query_idx < request.bucket_ids_.size(); ++query_idx) {
+                const auto& ids = request.bucket_ids_[query_idx];
+                if (ids.empty()) {
                     return tl::unexpected(
                         Error(ErrorType::INVALID_ARGUMENT,
-                              "bucket_id " + std::to_string(id) + " is negative"));
+                              "bucket_ids_[" + std::to_string(query_idx) +
+                                  "] must not be empty; use an empty outer vector "
+                                  "for default routing"));
```
LHT129, line 528: [note] The validation loop in `index_impl.h` checks `id < 0` for each bucket ID but does not validate the upper bound (`id >= bucket_count`). This upper-bound check is deferred to `IVF::SearchWithRequest` in `ivf.cpp`. This split is intentional since `IndexImpl` is a generic template that does not know the bucket count, but it means non-IVF indexes that accept `bucket_ids_` (size 1) will only get the negative-ID check at this layer. The upper-bound validation is the caller's responsibility in the IVF-specific path. This is fine for the current design, but worth noting for future index types that may support `bucket_ids_`.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,54 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        CHECK_ARGUMENT(request.expected_labels_.empty(),
+                       "IVF batch search does not support expected labels");
+        CHECK_ARGUMENT(request.topk_ > 0, "topk must be greater than 0");
+        CHECK_ARGUMENT(query->GetFloat32Vectors() != nullptr,
+                       "query float32 vectors cannot be null");
+        CHECK_ARGUMENT(query->GetDim() == this->dim_, "query dimension must match index dimension");
+
+        const auto num_queries = query->GetNumElements();
+        const auto total_slots = num_queries * request.topk_;
+        auto* alloc = select_query_allocator(ctx.alloc, this->allocator_);
```
LHT129, line 1932: [note] The batch search path allocates `ids` and `distances` via `select_query_allocator(ctx.alloc, this->allocator_)` at line 1922. The per-query `SearchWithRequest(one_request)` sub-calls will also use this same allocator (via `ctx.alloc`) for their own intermediate result buffers. With a pool/arena allocator this is fine — the sub-call allocations will be freed before the batch function returns the final dataset. With a stack allocator, however, the sub-call allocations could interleave with the parent's `ids`/`distances` buffers. This is a pre-existing pattern in the single-query path as well, so it is not a regression, but worth noting for future allocator changes.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,54 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+                ->Owner(false);
+
+            auto one_request = request;
+            one_request.query_ = one_query;
+            if (not request.bucket_ids_.empty()) {
+                one_request.bucket_ids_ = {request.bucket_ids_[query_idx]};
+            }
+            auto one_result = this->SearchWithRequest(one_request);
+            const auto count = std::min(request.topk_, one_result->GetDim());
+            if (count > 0) {
```
LHT129, line 1963: [note] When `one_result->GetDim()` is 0 (i.e. no results found for a query), `count = std::min(request.topk_, one_result->GetDim())` evaluates to 0, so the `memcpy` calls are no-ops. However, `one_result->GetIds()` and `one_result->GetDistances()` may return `nullptr` when the result count is 0 (depending on the internal implementation). While `memcpy(dst, nullptr, 0)` is technically UB per the C++ standard, in practice all major compilers handle it as a no-op. This is safe for the current codebase but worth noting for portability to strict platforms.

LHT129 (commented):
## Code Review

### [suggestion] Unnecessary deep copy of SearchRequest per query iteration

In `src/algorithm/ivf/ivf.cpp:1938`, `auto one_request = request;` performs a full deep copy of the `SearchRequest` object inside the loop on every iteration. This copies `bucket_ids_` (a `std::vector<std::vector<int64_t>>`), `params_str_`, `attribute_filter_str_`, and all other fields — only to immediately overwrite `query_` and `bucket_ids_` on the next two lines.

For large batch sizes, this adds measurable overhead. Consider hoisting the copy outside the loop:

```cpp
auto one_request = request;
for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
    // ... construct one_query ...
    one_request.query_ = one_query;
    if (not request.bucket_ids_.empty()) {
        one_request.bucket_ids_ = {request.bucket_ids_[query_idx]};
    }
    auto one_result = this->SearchWithRequest(one_request);
    // ...
}
```

### [note] Missing SearchStatistics on batch return path

In `src/algorithm/ivf/ivf.cpp:1949-1954`, the batch search return path constructs a `Dataset` and returns it directly without calling `result->Statistics(stats.Dump())`. All other return paths in this function attach statistics. This means batch callers lose observability into per-query timing and scan counts. Consider aggregating or attaching stats to the batch result.

### [note] Batch path does not validate `attribute_filter` incompatibility

The batch path at `src/algorithm/ivf/ivf.cpp:1910` rejects `is_range`, `use_custom_distance`, and `expected_labels_`, but does not reject `enable_attribute_filter_`. If a user passes `enable_attribute_filter_ = true` with multi-query, the recursive `SearchWithRequest` calls will each try to parse and set up attribute filters independently, which may work but is likely unintended for batch mode. Consider adding an explicit rejection or documenting the expected behavior.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,54 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        std::fill_n(ids, total_slots, -1);
+        std::fill_n(distances, total_slots, std::numeric_limits<float>::infinity());
+
+        const auto* query_data = query->GetFloat32Vectors();
+        for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
+            auto one_query = Dataset::Make();
+            one_query->NumElements(1)
+                ->Dim(query->GetDim())
+                ->Float32Vectors(query_data + query_idx * query->GetDim())
+                ->Owner(false);
```
LHT129, line 1949: [suggestion] The `auto one_request = request;` inside the per-query loop copies the entire `SearchRequest` object on every iteration, including string fields like `params_str_`, `attribute_filter_str_`, and the full `bucket_ids_` vector. For large batch sizes this adds unnecessary overhead.

Since the batch path already validates that `is_range`, `use_custom_distance`, and `expected_labels_` are rejected, only a few fields (`query_`, `topk_`, `params_str_`, `filter_`, `bucket_ids_`) are actually used by the recursive call. Consider constructing a minimal request once outside the loop and only updating the per-query fields (`query_` and `bucket_ids_`) inside the loop.

```cpp
auto one_request = request;
one_request.bucket_ids_.clear();  // will be set per-query
for (int64_t query_idx = 0; query_idx < num_queries; ++query_idx) {
    // ... build one_query ...
    one_request.query_ = one_query;
    if (not request.bucket_ids_.empty()) {
        one_request.bucket_ids_ = {request.bucket_ids_[query_idx]};
    } else {
        one_request.bucket_ids_.clear();
    }
    auto one_result = this->SearchWithRequest(one_request);
    // ...
}
```

tests/test_ivf.cpp
```diff
@@ -2530,12 +2536,53 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::IVFTestIndex,
         REQUIRE_FALSE(result.has_value());
     }
 
-    SECTION("reject multiple query entries") {
+    SECTION("batch queries use corresponding bucket ID lists") {
```
LHT129, line 2548: [suggestion] The batch test only covers the `bucket_ids_` bypass path. The multi-query batch search path also supports default bucket routing (empty `bucket_ids_`), but this code path is not exercised by any test. Consider adding a test case for batch search without `bucket_ids_` to ensure the default routing path works correctly with multiple queries.

```cpp
SECTION("batch queries with default bucket routing") {
    vsag::SearchRequest request;
    request.query_ = batch_query;
    request.topk_ = 10;
    request.params_str_ = search_param;
    auto batch_result = index->SearchWithRequest(request);
    REQUIRE(batch_result.has_value());
    REQUIRE(batch_result.value()->GetNumElements() == 2);
    REQUIRE(batch_result.value()->GetDim() == request.topk_);
}
```

src/index/index_impl.h
```diff
@@ -500,28 +500,38 @@ class IndexImpl : public Index {
...
         // Validate bucket_ids_ structural constraints before empty-index early return
         if (not request.bucket_ids_.empty()) {
-            if (request.bucket_ids_.size() != 1) {
-                return tl::unexpected(Error(
-                    ErrorType::INVALID_ARGUMENT,
-                    "bucket_ids_ currently supports only single-query mode; expected size 1, got " +
-                        std::to_string(request.bucket_ids_.size())));
+            if (request.query_ == nullptr ||
+                request.bucket_ids_.size() != request.query_->GetNumElements()) {
+                return tl::unexpected(
```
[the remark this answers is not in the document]
LHT129: Fixed: split into separate null check and size mismatch, with both bucket_ids_ size and query count in the error message.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,54 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+                std::copy_n(
+                    one_result->GetDistances(), count, distances + query_idx * request.topk_);
+            }
+        }
+        return Dataset::Make()
+            ->NumElements(num_queries)
+            ->Dim(request.topk_)
+            ->Ids(ids)
+            ->Distances(distances)
+            ->Owner(true, alloc);
```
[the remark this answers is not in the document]
LHT129: Fixed: added result->Statistics(stats.Dump()) to the batch return path.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1845,38 +1845,42 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+                                   "got {} bucket lists for {} queries",
+                                   request.bucket_ids_.size(),
                                    query_check->GetNumElements()));
         CHECK_ARGUMENT(query_check->GetFloat32Vectors() != nullptr,
                        "query float32 vectors cannot be null");
         CHECK_ARGUMENT(query_check->GetDim() == this->dim_,
                        "query dimension must match index dimension");
         CHECK_ARGUMENT(not param.disable_bucket_scan,
                        "bucket_ids_ is incompatible with disable_bucket_scan mode");
+        for (size_t query_idx = 0; query_idx < request.bucket_ids_.size(); ++query_idx) {
```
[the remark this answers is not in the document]
LHT129: Fixed: changed to uint64_t per AGENTS.md.

tests/test_ivf.cpp
```diff
@@ -2419,6 +2419,12 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::IVFTestIndex,
                                     dataset->base_->GetFloat32Vectors() + dim);
     auto query = vsag::Dataset::Make();
     query->NumElements(1)->Dim(dim)->Float32Vectors(query_vector.data())->Owner(false);
+    std::vector<float> batch_query_vectors(query_vector.begin(), query_vector.end());
+    batch_query_vectors.insert(batch_query_vectors.end(),
+                               dataset->base_->GetFloat32Vectors() + dim,
+                               dataset->base_->GetFloat32Vectors() + 2 * dim);
+    auto batch_query = vsag::Dataset::Make();
+    batch_query->NumElements(2)->Dim(dim)->Float32Vectors(batch_query_vectors.data())->Owner(false);
```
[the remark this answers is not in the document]
LHT129: Fixed: wrapped long lines to stay under 100 chars.

docs/docs/en/src/api/search.md
```diff
@@ -101,9 +101,11 @@ selection and scans only the specified buckets.
 **Semantics:**
 - **Empty** (default): Use normal bucket routing via `ClassifyDatasForSearch`.
 - **Non-empty**: Skip routing and scan only the provided bucket IDs.
+- **Batch IVF KNN**: One outer `bucket_ids_` entry is required for each query vector. The result is rectangular: `NumElements()` is the query count and `Dim()` is `topk_`. Missing neighbors are `-1` with an infinite distance.
```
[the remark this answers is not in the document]
LHT129: Fixed: wrapped to multiple lines under 100 chars.

docs/docs/zh/src/api/search.md
```diff
@@ -95,9 +95,11 @@ reasoning 选项均会被忽略。
 **语义：**
 - **空**（默认）：使用正常的桶路由（`ClassifyDatasForSearch`）。
 - **非空**：跳过路由，仅扫描提供的桶 ID。
+- **批量 IVF KNN**：`bucket_ids_` 外层的每一项对应一个查询向量。结果为矩形：`NumElements()` 是查询数，`Dim()` 是 `topk_`。缺失的邻居用 `-1` 和无穷距离填充。
```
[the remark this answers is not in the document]
LHT129: Fixed: wrapped to multiple lines under 100 chars.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed: moved auto one_request = request outside the loop, and clear bucket_ids_ once before iterating.

src/index/index_impl.h
LHT129: Acknowledged. The split is intentional — IndexImpl is index-type-agnostic and does not know bucket_count. IVF handles upper-bound validation in its own path.

src/algorithm/ivf/ivf.cpp
LHT129: Acknowledged. The current pool allocator handles this safely, and the recursive sub-calls complete before the batch function returns.

src/algorithm/ivf/ivf.cpp
LHT129: Acknowledged. The existing if (count > 0) guard ensures we never call copy_n with nullptr and zero length.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed: moved auto one_request = request outside the loop.

tests/test_ivf.cpp
LHT129: Added new section batch queries with default routing (no bucket_ids) that verifies batch results match individual single-query results.

LHT129 (commented):
## Code Review Summary

Reviewed all 6 changed files in commit `c510e29c`. The previous round of review feedback has been properly addressed:

- **`ivf.cpp`**: `auto one_request = request` is now outside the loop, `Statistics(stats.Dump())` is attached to the batch result, and loop indices use `uint64_t`.
- **`index_impl.h`**: `query_ == nullptr` check is now separated from the size-mismatch check, providing clearer error messages. Non-IVF indexes are correctly restricted to single-query `bucket_ids_`.
- **Tests**: Comprehensive coverage including batch-vs-single correctness verification (both with and without bucket_ids bypass), and query-count mismatch rejection.

### What was verified

1. **Correctness**: The batch path recursively calls `SearchWithRequest` with per-query non-owning Dataset views. Padding with `-1`/`+inf` handles cases where fewer than `topk_` results are returned. The `bucket_ids_` per-query validation correctly checks emptiness, range, and duplicates for each inner list.

2. **Restrictions**: Batch mode correctly rejects range search, custom distance callbacks, and reasoning labels. The `disable_bucket_scan` incompatibility is validated before reaching the batch path.

3. **Memory**: The batch result uses `Owner(true, alloc)` with the query allocator, consistent with other return paths. The non-owning per-query Dataset views avoid extra allocations.

4. **Backward compatibility**: Single-query path is unchanged — `param.bucket_ids` is still populated for the single-query case via the `GetNumElements() == 1` branch.

No blocking issues found. The implementation is clean and well-tested.

LHT129 (commented):
This iteration addresses all the key feedback from the previous review round:

- Added `Statistics(stats.Dump())` to the batch result Dataset
- Changed `size_t` to `uint64_t` in the bucket_ids validation loop (consistent with project conventions)
- Moved `one_request` copy outside the per-query loop to avoid unnecessary copies
- Added a new test section for batch queries with default routing (no bucket_ids)
- Split the null `query_` check from the size-mismatch check in `index_impl.h` for clearer error messages

The batch search implementation looks solid. The recursive per-query approach with non-owning Dataset views is clean, and the padding strategy (id=-1, distance=+inf) for missing neighbors is well-documented. The validation in both `index_impl.h` and `ivf.cpp` is consistent and thorough.

No blocking issues found.

src/index/index_impl.h
```diff
@@ -500,28 +500,43 @@ class IndexImpl : public Index {
...
-            if (ids.empty()) {
+            if (request.query_ == nullptr) {
                 return tl::unexpected(Error(ErrorType::INVALID_ARGUMENT,
-                                            "bucket_ids_[0] must not be empty; use empty outer "
-                                            "vector for default routing"));
+                                            "query_ cannot be null when bucket_ids_ is set"));
+            }
+            if (request.bucket_ids_.size() !=
+                static_cast<size_t>(request.query_->GetNumElements())) {
+                return tl::unexpected(
```
LHT129, line 510: [suggestion] The static_cast<size_t> on GetNumElements() is technically unsafe: GetNumElements() returns int64_t, and if it were negative the cast to size_t would wrap to a huge value, making the size check meaningless. Consider adding a GetNumElements() > 0 guard before the cast, or using a signed comparison.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1902,6 +1906,60 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+        const auto num_queries = query->GetNumElements();
+        const auto total_slots = num_queries * request.topk_;
+        auto* alloc = select_query_allocator(ctx.alloc, this->allocator_);
+        auto* ids = static_cast<int64_t*>(alloc->Allocate(sizeof(int64_t) * total_slots));
+        auto* distances = static_cast<float*>(alloc->Allocate(sizeof(float) * total_slots));
+        std::fill_n(ids, total_slots, -1);
+        std::fill_n(distances, total_slots, std::numeric_limits<float>::infinity());
+
+        const auto* query_data = query->GetFloat32Vectors();
+        auto one_request = request;
```
LHT129: [note] The `one_request` copy (line 1929) propagates `request.search_allocator_` to each recursive `SearchWithRequest` call. The outer batch allocation at line 1923 uses `select_query_allocator(ctx.alloc, this->allocator_)`. If the caller provided a custom `search_allocator_` that differs from the default, the inner per-query allocations and the outer result buffer will use different allocators. This is consistent with the existing single-query path behavior, but worth noting for allocator-aware use cases.

src/index/index_impl.h
```diff
@@ -500,28 +500,43 @@ class IndexImpl : public Index {
...
+                    Error(ErrorType::INVALID_ARGUMENT,
+                          "bucket_ids_ size (" + std::to_string(request.bucket_ids_.size()) +
+                              ") must match the number of query vectors (" +
+                              std::to_string(request.query_->GetNumElements()) + ")"));
+            }
+            if (this->GetIndexType() != IndexType::IVF && request.bucket_ids_.size() != 1) {
+                return tl::unexpected(
+                    Error(ErrorType::INVALID_ARGUMENT,
```
LHT129, line 518: [note] `request.bucket_ids_.size()` returns `size_t` while `1` is `int`, resulting in a signed/unsigned comparison. This is harmless in practice since `size()` is non-negative, but for consistency with the project preference for explicit types, consider using `1ULL` or `static_cast<size_t>(1)`.

LHT129 (commented):
Overall this is a clean and well-structured PR. The batch IVF search implementation correctly delegates to per-query recursive calls, the validation in both `IndexImpl` and `IVF` layers is thorough, and the test coverage is good (batch with bucket_ids, batch with default routing, rejection of mismatched counts).

A few minor observations were left as inline comments:
- `index_impl.h:509`: `static_cast<size_t>` on a potentially negative `int64_t` return value
- `index_impl.h:517`: signed/unsigned comparison between `size_t` and `int`
- `ivf.cpp:1929`: allocator propagation note for the recursive calls

None of these are blocking.

LHT129 (commented):
[note] In `src/index/index_impl.h:508`, `static_cast<size_t>` should be `static_cast<uint64_t>` per the project coding standard. The project prefers `uint64_t` over `size_t` to avoid macOS compilation issues.

```cpp
// current:
if (request.bucket_ids_.size() !=
    static_cast<size_t>(request.query_->GetNumElements())) {

// suggested:
if (request.bucket_ids_.size() !=
    static_cast<uint64_t>(request.query_->GetNumElements())) {
```

Otherwise the PR looks good. The batch search implementation is clean, the validation logic in both `IndexImpl` and `IVF` is thorough, and the tests cover both the bucket_ids bypass path and the default routing path with correctness verification against individual single-query results. The previously raised issues (missing Statistics, size_t usage in ivf.cpp, one_request copy inside loop) have all been addressed.

src/algorithm/ivf/ivf.cpp
```diff
@@ -1912,6 +1916,81 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
...
+
+        const auto* query_data = query->GetFloat32Vectors();
+        auto base_request = request;
+        base_request.expected_labels_.clear();
+        if (not request.bucket_ids_.empty()) {
+            base_request.bucket_ids_.clear();
+        }
+        auto search_func = [&](int64_t query_idx) -> void {
+            auto one_query = Dataset::Make();
+            one_query->NumElements(1)
```
LHT129, line 1946: [suggestion] The `JsonType::Parse` and `json.Dump()` calls inside the per-query lambda are repeated for every query, but `base_request.params_str_` does not change between iterations. This adds unnecessary JSON parsing/serialization overhead in the hot path. Consider hoisting the JSON manipulation outside the lambda: parse once, set `parallelism=1`, dump once, and reuse the serialized string.

```cpp
auto one_params_str = base_request.params_str_;
JsonType json = JsonType::Parse(base_request.params_str_);
if (json.Contains("ivf")) {
    json["ivf"]["parallelism"].SetInt64(1);
    one_params_str = json.Dump();
}

auto search_func = [&](int64_t query_idx) -> void {
    // ...
    one_request.params_str_ = one_params_str;
    // ...
};
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -1912,6 +1916,81 @@ IVF::SearchWithRequest(const SearchRequest& request) const {
         return result;
     }
 
+    if (query != nullptr && query->GetNumElements() > 1) {
+        CHECK_ARGUMENT(not is_range, "IVF batch search only supports KNN search");
+        CHECK_ARGUMENT(not use_custom_distance,
```
LHT129, line 1921: [suggestion] `num_queries` and `request.topk_` are both `int64_t`, so `total_slots` is also `int64_t`. The multiplication `num_queries * request.topk_` could overflow if both values are large, and the result is passed to `Allocate(sizeof(...) * total_slots)` which expects `uint64_t`. Consider using `uint64_t` for `total_slots` to avoid signed overflow.

```cpp
const auto total_slots = static_cast<uint64_t>(num_queries) * static_cast<uint64_t>(request.topk_);
```

LHT129 (commented):
I have reviewed the latest changes (c4f247bb) to this PR. The code quality is solid and the batch IVF search implementation is well-structured.

**Summary of findings:**

The previous round of review comments have been well addressed:
- `size_t` → `uint64_t` in the bucket_ids validation loop (fixed)
- Missing `Statistics(stats.Dump())` on the batch result (fixed)
- Line-length issues in docs and tests (fixed)

**Remaining items from self-review that are worth addressing:**

1. **JSON parse/dump inside per-query lambda** (src/algorithm/ivf/ivf.cpp:1946): The `JsonType::Parse` and `json.Dump()` calls are repeated for every query in the hot path. Hoisting the JSON manipulation outside the lambda would avoid unnecessary parsing/serialization overhead.

2. **Signed overflow risk in `total_slots`** (src/algorithm/ivf/ivf.cpp:1931): `num_queries * request.topk_` uses `int64_t` multiplication which could overflow for large values. Using `uint64_t` casts would be safer.

3. **Signed/unsigned comparison** (src/index/index_impl.h:517): `request.bucket_ids_.size() != 1` compares `size_t` with `int`.

**Overall assessment:**

The batch search logic is correct — per-query recursive calls with non-owning Dataset views, proper padding of missing neighbors, and correct thread-pool usage. The validation in `index_impl.h` properly gates multi-query bucket_ids to IVF indexes only. Test coverage is thorough with correctness checks against individual single-query results, default routing, and parallel execution.

No blocking issues found. The remaining items above are suggestions for optimization and type safety.

wxyucs (approved):
lgtm

PR #2640, opened 2026-08-07
feat: support L1 distance metric and add unit tests

jingyueob:
## Change Type
This feature adds support for the L1 distance metric in VSAG vector indexes.

## What Changed
Add a new distance metric by introducing L1Space for L1 distance computation and mapping VIDA_L1 to DISTANCE.

## Test
finish all unittest and do end 2 end test personally

### DIFF
--- a/src/algorithm/hnswlib/space_l1.h
+++ b/src/algorithm/hnswlib/space_l1.h
@@ -1,4 +1,3 @@
-
 // Copyright 2024-present the vsag project
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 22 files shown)


src/algorithm/hnswlib/space_l1.h
```diff
@@ -0,0 +1,49 @@
...
+    DISTFUNC
+    get_dist_func() override {
+        return fstdistfunc_;
+    }
+
+    void*
+    get_dist_func_param() override {
+        return &dim_;
+    }
+
```
LHT129, line 42: [note] `data_size_` is declared as `size_t` but VSAG coding standards require `uint64_t` to avoid macOS compilation issues (where `size_t` differs from `uint64_t`). The member `dim_` correctly uses `uint64_t`; `data_size_` should follow the same convention for consistency.

Current:
```cpp
size_t data_size_;
```

Suggested:
```cpp
uint64_t data_size_;
```

The return type of `get_data_size()` would also need to change from `size_t` to `uint64_t`.

src/simd/basic_func.cpp
```diff
@@ -19,6 +19,12 @@
 
 namespace vsag {
 
+static DistanceFuncType
```
LHT129, line 22: [suggestion] `GetL1Distance()` only returns `generic::L1Distance` without any SIMD dispatch, unlike all other distance functions (`GetL2Sqr`, `GetInnerProduct`, etc.) which check for AVX512/AVX2/AVX/SSE/SVE/NEON support. This means L1 distance computation will always use the slow scalar path even on hardware with SIMD support, which could significantly impact search performance.

If SIMD-accelerated L1 implementations are planned for a follow-up PR, consider adding a `// TODO` comment here to make the intent clear. Otherwise, this should be addressed before merging since L1 is being introduced as a first-class metric.

src/quantization/fp32_quantizer.cpp
```diff
@@ -245,6 +253,12 @@ FP32Quantizer<metric>::ComputeDistsBatch4Impl(Computer<FP32Quantizer<metric>>& c
                                dists2,
                                dists3,
                                dists4);
+    } else if constexpr (metric == MetricType::METRIC_TYPE_L1) {
```
LHT129, line 256: [note] The L1 branch in `ComputeDistsBatch4Impl` calls `L1Distance` four separate times, each iterating over the full dimension. This is functionally correct but less efficient than a fused batch loop that computes all four distances in a single pass over the dimension. The existing `ScanBatchDistImpl` already has a `// TODO(LHT): Optimize batch for simd` comment acknowledging this pattern. Consider adding a similar note here, or ideally implementing a `FP32ComputeL1Batch4` helper analogous to `FP32ComputeL2SqrBatch4`.

LHT129 (commented):
Code review for PR #2640: L1 distance metric support

PR #2560, opened 2026-07-28
feat: support MRLE with split RaBitQ

LightWant:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2533

## What Changed

- allow RaBitQ split datacells to access `TransformQuantizer<RaBitQuantizer>` through a generic bottom-quantizer accessor
- support the exact `mrle,rabitq` split configuration in HGraph and Pyramid while keeping split encoding in the terminal RaBitQ quantizer
- share split parameter mapping and validation while preserving direct RaBitQ split behavior
- fix reduced-dimension TransformQuantizer training layout and strengthen compatibility checks
- add unit, functional, and English/Chinese documentation coverage

## Test Evidence

- [x] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [ ] `make cov`, run tests, and collect coverage
- [x] Other (describe below)

```text
VSAG_ENABLE_LIBAIO=OFF make debug COMPILE_JOBS=4

Regression tests:
- complete unit suite: 639 cases, 85,539,431 assertions
- MRLE unit: 3 cases, 71 assertions
- MRLE functional: 2 cases, 210 assertions
- RaBitQ split direct compute: 4096 assertions
- RaBitQ split optimized scalar build: 1656 assertions
- TransformQuantizer compute: 92592 assertions
- HGraph direct mapping, Pyramid precise mapping, and TransformQuantizer compatibility passed

Dataset evaluation: openai-1536-50k-angular.hdf5
- mrle_dim=768: Recall@10 0.7640, QPS 358.83, average latency 2.783 ms (single thread)
- mrle_dim=1536 control: Recall@10 0.9941, QPS 238.33, average latency 4.192 ms
```

The complete unit suite passed. The complete functional suite was started and manually stopped during a long-running Daily case after the affected-path functional tests had passed; no pre-interrupt failure was observed.

## Compatibility Impact

- API/ABI compatibility: additive only
- Behavior changes: exact `tq_chain="mrle,rabitq"` can now use RaBitQ x+y split in HGraph and Pyramid; other transformed split chains remain rejected

## Performance and Concurrency Impact

- Performance impact: reduced dimensions lower code size and search cost, with a dataset-dependent recall tradeoff
- Concurrency/thread-safety impact: no public concurrency model change; optimized-build finalization continues using disjoint code slots

## Documentation Impact

[...]

Issue #2533, opened 2026-07-24
[feat](other): support dimension reduction with split RaBitQ in HGraph and Pyramid


dasurax:
### Affected area

other (HGraph, Pyramid, transform quantization, and RaBitQ split storage)

### Motivation

High-dimensional embedding workloads can reduce vector dimensionality before quantization, for example from 1024 dimensions to 768 dimensions with MRLE. VSAG also supports RaBitQ x+y split storage, where compact x-bit filter codes are used during graph traversal and y-bit supplement codes are fetched only for final reorder.

These capabilities cannot currently be combined:

- HGraph exposes `tq_chain` and `mrle_dim`, but the split mapper requires `base_quantization_type=rabitq`.
- An MRLE chain uses `base_quantization_type=tq` and a terminal RaBitQ quantizer such as `tq_chain=mrle,rabitq`.
- `RaBitQSplitDataCell` explicitly rejects transform quantizers.
- Pyramid does not yet expose transform-chain parameters or split RaBitQ through its public parameter mapping.

As a result, users must choose between reducing the dimension with MRLE and using split RaBitQ's lower memory bandwidth and on-demand supplement I/O. Large HGraph and Pyramid indexes would benefit from applying the transform once and encoding split RaBitQ codes in the reduced dimension.

### API / ABI compatibility impact

additive only (no break)

### References

- `src/algorithm/hgraph/hgraph_param_mapping.cpp:27-83` - current split mapping requires direct RaBitQ base and precise quantization types.
- `src/algorithm/hgraph/hgraph_param_mapping.cpp:354-367` - current HGraph mappings for `tq_chain` and `mrle_dim`.
- `src/datacell/flatten_interface.cpp:195-200` - split data cells currently reject transform quantizers.
- `src/quantization/transform_quantization/transform_quantizer.h:146-166` - transform-chain construction before the terminal quantizer.
- `src/algorithm/pyramid/pyramid.cpp:934-980` - Pyramid public parameter mapping currently lacks transform-chain and split parameters.
- #2372 - exposed `mrle_dim` in the HGraph public API.
- #2454 - tracks exposing split RaBitQ in Pyramid.
- #2096 - discusses split-code storage and disk supplement I/O.
- #1370 - original MRLE transformer feature.

### Additional context

[...]

PR #2643, opened 2026-08-08
refactor: remove Pyramid raw vector dependency for split codes

LightWant:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Related: #2533
- Builds on: #2560 (merged)

## What Changed

- add stored-vector-ID batch distance queries to `FlattenInterface` and implement efficient temporary and persistent RaBitQ split code-code distance
- add a graph distance provider backed by stored vector IDs
- build Pyramid FLAT-to-GRAPH promotions and incremental graph connections from split codes instead of decoded raw vectors
- stop automatically storing an FP32 copy for ordinary RaBitQ split configurations
- keep Pyramid Analyzer usable without raw vectors by retaining ID/code metrics and skipping metrics that require decoded FP32 samples
- update English/Chinese Pyramid documentation and add unit and functional regression coverage

## Test Evidence

- [x] `make fmt` equivalent (`clang-format-15` on all changed C++ files)
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build -j6
# Full Debug build passed on the #2560 head before rebase, including unittests and functests targets.
# After rebasing onto main, the changed Pyramid and Analyzer targets compiled; the full rebuild
# then stopped because downloading the newly bumped OpenBLAS 0.3.24 dependency timed out.

build/tests/unittests "Search with stored vector ID"
build/tests/unittests "RaBitQSplitDataCell optimized scalar-code build"
build/tests/unittests "Pyramid maps RaBitQ x+y split params"
build/tests/unittests "Pyramid maps MRLE RaBitQ split to base reorder"
build/tests/functests "Pyramid split RaBitQ builds with ODescent"
build/tests/functests "Pyramid split RaBitQ promotes without raw vectors"
# All passed.

git diff --check
# Passed.
```

## Compatibility Impact

- API/ABI compatibility: no public VSAG API change; an internal flatten interface gains a default virtual ID-query method
- Behavior changes: ordinary Pyramid RaBitQ split configurations no longer retain raw FP32 vectors by default; raw-vector-dependent Analyzer fields are omitted when raw vectors are unavailable

- Performance impact: reduces Pyramid memory by avoiding `count * dim * sizeof(float)` raw-vector storage; persistent ID batches reconstruct the query code once per batch
- Concurrency/thread-safety impact: none; existing graph-node and code-storage synchronization remains unchanged

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: English and Chinese Pyramid index documentation

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert commit `98dd34c7`

## Checklist

- [x] I have linked the relevant issue/parent PR
- [x] I have updated docs for the behavior change
- [x] My commit message follows project conventions


src/impl/distance_provider_for_graph.cpp
```diff
@@ -20,6 +20,39 @@
...
+FlattenIdDistanceProvider::PairwiseDistance(InnerIdType id1,
+                                            InnerIdType id2,
+                                            const ComputerInterfacePtr&) const {
+    return flatten_->ComputePairVectors(id1, id2);
+}
+
+ComputerInterfacePtr
+FlattenIdDistanceProvider::FactoryComputerById(InnerIdType) const {
+    throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                        "FlattenIdDistanceProvider cannot create a computer by vector ID");
```
LHT129, line 48: **Issue**: `FlattenIdDistanceProvider::FactoryComputerById` unconditionally throws `UNSUPPORTED_INDEX_OPERATION`, but any searcher code path that calls `FactoryComputerById` on this provider will crash at runtime.

In `distance_provider_for_graph.cpp:46-48`:
```cpp
ComputerInterfacePtr
FlattenIdDistanceProvider::FactoryComputerById(InnerIdType) const {
                        "FlattenIdDistanceProvider cannot create a computer by vector ID");
}
```

`PairwiseDistance` at line 41-43 correctly delegates to `flatten_->ComputePairVectors`, but if any searcher code path calls `FactoryComputerById` on this provider, it will throw. Is this intentional? If so, consider documenting which searcher code paths are safe to use with `FlattenIdDistanceProvider`.

**Suggestion**: Either ensure no searcher ever calls `FactoryComputerById` on this provider, or implement it by reading the code for the given ID and creating a computer from it.

src/datacell/rabitq_split_datacell.h
```diff
@@ -229,6 +239,55 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
         }
     }
 
+    void
+    QueryById(float* result_dists,
+              InnerIdType query_id,
+              const InnerIdType* idx,
+              InnerIdType id_count,
+              QueryContext* /*ctx*/ = nullptr) override {
+        if (not this->optimized_build_active_) {
```
LHT129, line 252: **Issue**: The non-optimized-build path in `QueryById` allocates two `ByteBuffer` objects per call, then reads both the query code and each base code from the split storage (one-bit + supplement) and merges them via `quantizer_->Compute`. This is functionally correct but could be inefficient compared to the optimized-build path which uses scalar codes directly.

In `rabitq_split_datacell.h:248-268`:
```cpp
if (not this->optimized_build_active_) {
    ByteBuffer query_code(this->code_size_, allocator_);
    ByteBuffer base_code(this->code_size_, allocator_);
    if (not this->GetCodesById(query_id, query_code.data)) {
        throw VsagException(ErrorType::INTERNAL_ERROR, ...);
    }
    for (InnerIdType i = 0; i < id_count; ++i) {
        ...
        if (not this->GetCodesById(idx[i], base_code.data)) {
            throw VsagException(ErrorType::INTERNAL_ERROR, ...);
        }
        result_dists[i] = this->quantizer_->Compute(query_code.data, base_code.data);
    }
    return;
}
```

Each `GetCodesById` call reads both the one-bit and supplement codes, then merges them. For the non-optimized-build path, this means for `id_count` base vectors, you're doing `id_count + 1` full code reads and merges. This is the same approach as the existing `Query` path, so it's consistent, but worth noting that the optimized-build path avoids this overhead entirely.

**Suggestion**: Consider adding a comment noting that the non-optimized-build path is intended for correctness and the optimized-build path is the performance path.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1303,20 +1297,26 @@ Pyramid::add_one_point(const Hierarchy& h,
...
-        Vector<float> decoded_vector(dim_, allocator_);
-        for (const auto id : node->ids_) {
-            bool need_release = false;
-            const auto* buffer = codes->GetCodesById(id, need_release);
-            const bool decoded = codes->Decode(buffer, decoded_vector.data());
-            if (need_release) {
-                codes->Release(buffer);
+        if (base_codes_->SupportSplitCodeStorage()) {
+            for (const auto id : node->ids_) {
+                add_one_point(h, &graph_node, id, nullptr);
```
LHT129, line 1321: **Issue**: The `promote_flat_to_graph` method in `pyramid.cpp` uses `base_codes_->SupportSplitCodeStorage()` to decide whether to use `FlattenIdDistanceProvider` (split code path) or decode FP32 vectors. However, the split code path passes `nullptr` as the vector to `add_one_point`, which then uses `FlattenIdDistanceProvider` for graph search during promotion.

In `pyramid.cpp:1302-1315`:
```cpp
if (base_codes_->SupportSplitCodeStorage()) {
    for (const auto id : node->ids_) {
        add_one_point(h, &graph_node, id, nullptr);
    }
} else {
    auto codes = decodable_codes();
    Vector<float> decoded_vector(dim_, allocator_);
    for (const auto id : node->ids_) {
        bool need_release = false;
        const auto* buffer = codes->GetCodesById(id, need_release);
        const bool decoded = codes->Decode(buffer, decoded_vector.data());
        ...
        add_one_point(h, &graph_node, id, decoded_vector.data());
    }
}
```

This is the core of the PR's contribution — enabling flat-to-graph promotion without raw vectors. The logic is sound, but the condition `base_codes_->SupportSplitCodeStorage()` is used as a proxy for "we can compute code-to-code distances without raw vectors." If other codec types gain `SupportSplitCodeStorage()` in the future but don't support `QueryById` properly, this could break silently.

**Suggestion**: Consider checking `raw_vector_ == nullptr` as an additional condition, since the intent is specifically to handle the case where raw vectors are unavailable.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1352,8 +1352,33 @@ Pyramid::add_one_point(const Hierarchy& h,
...
+                node->graph_, codes, vl, vector, search_param, (LabelTablePtr) nullptr, nullptr);
+        } else {
+            FlattenIdDistanceProvider distance_provider(codes, inner_id);
+            results = searcher_->Search(
+                node->graph_, distance_provider, vl, search_param, nullptr, nullptr);
+            if (support_duplicate_ and not results->Empty()) {
+                const auto* data = results->GetData();
+                auto min_distance = data[0].first;
+                auto min_index = data[0].second;
+                for (uint32_t i = 1; i < results->Size(); ++i) {
```
LHT129, line 1389: **Issue**: In `add_one_point`, when `vector == nullptr` (split code path), the duplicate detection logic iterates through all search results to find the minimum distance and index. This duplicates the search result iteration that the searcher already performed internally.

In `pyramid.cpp:1367-1379`:
```cpp
if (support_duplicate_ and not results->Empty()) {
    const auto* data = results->GetData();
    auto min_distance = data[0].first;
    auto min_index = data[0].second;
    for (uint32_t i = 1; i < results->Size(); ++i) {
        if (data[i].first < min_distance) {
            min_distance = data[i].first;
            min_index = data[i].second;
        }
    }
    ...
}
```

This is O(results->Size()) extra work. Since the searcher already returns results in distance order (nearest first), the first element should already be the minimum. Is this iteration necessary, or could you just use `data[0]` directly?

**Suggestion**: If the searcher guarantees results are sorted by distance, use `data[0]` directly. If not, add a comment explaining why the full scan is needed.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -487,6 +487,12 @@ PyramidAnalyzer::sample_global() {
         sample_ids_.resize(actual_sample_size);
     }
 
+    const bool has_decodable_codes = pyramid_->raw_vector_ != nullptr ||
+                                     pyramid_->has_precise_reorder() ||
+                                     not pyramid_->base_codes_->SupportSplitCodeStorage();
+    if (not has_decodable_codes) {
```
LHT129, line 501: **Issue**: The `pyramid_analyzer.cpp` change adds a check `not pyramid_->base_codes_->SupportSplitCodeStorage()` to determine if decoded FP32 samples are available for ground truth computation. When split codes are used without raw vectors, `sample_global()` returns early without populating `sample_datas_`, which means `calculate_groundtruth()` will not be called.

In `pyramid_analyzer.cpp:490-493`:
```cpp
const bool has_decodable_codes = pyramid_->raw_vector_ != nullptr ||
                                 pyramid_->has_precise_reorder() ||
                                 not pyramid_->base_codes_->SupportSplitCodeStorage();
if (not has_decodable_codes) {
    return;
}
```

This is correct behavior — if we can't decode vectors, we can't compute ground truth for the analyzer. However, the `GetStats()` method in `pyramid.cpp:1658` still creates an analyzer and calls `GetStats()`. When ground truth is unavailable, the analyzer's stats output may be incomplete or misleading.

**Suggestion**: Consider documenting in the analyzer output or logs that ground truth metrics are unavailable when raw vectors are not stored.

src/impl/distance_provider_for_graph.h
```diff
@@ -115,6 +115,37 @@ class FlattenDistanceProvider final : public DistanceProviderForGraph {
...
+
+    void
+    Prefetch(InnerIdType id) const override;
+
+private:
+    FlattenInterfacePtr flatten_;
+    InnerIdType query_id_;
+};
+
 class BucketDistanceProvider final : public DistanceProviderForGraph {
```
LHT129, line 149: **Question**: `BucketDistanceProvider` is introduced in this PR alongside `FlattenIdDistanceProvider`, but it doesn't appear to be used anywhere in the current diff. Is this class intended for a future PR, or is it used by code paths not visible in this diff?

If it's not yet used, consider moving it to a separate PR to keep this one focused on the split code-code distance feature. If it is used, could you point out where?

src/datacell/flatten_interface.h
```diff
@@ -58,6 +58,17 @@ class FlattenInterface {
           InnerIdType id_count,
           QueryContext* ctx = nullptr) = 0;
 
+    virtual void
+    QueryById(float* result_dists,
+              InnerIdType query_id,
+              const InnerIdType* idx,
+              InnerIdType id_count,
+              QueryContext* /*ctx*/ = nullptr) {
+        for (InnerIdType i = 0; i < id_count; ++i) {
```
LHT129, line 69: **Issue**: The default implementation of `QueryById` in `FlattenInterface` falls back to calling `ComputePairVectors` in a loop. For `RaBitQSplitDataCell`, `ComputePairVectors` reads and merges both one-bit and supplement codes for each pair, which means the default implementation would do `2 * id_count` full code reads and merges (once for query_id, once for each idx[i]).

The `RaBitQSplitDataCell` override avoids this by reading the query code once and reusing it, which is much more efficient. However, other `FlattenInterface` subclasses that don't override `QueryById` will use this default, which could be a performance trap.

**Suggestion**: Consider adding a comment on the default implementation noting that subclasses should override this method for better performance when computing distances from a single query vector to many base vectors.

src/datacell/rabitq_split_datacell.h
```diff
@@ -239,6 +239,55 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
...
+            for (InnerIdType i = 0; i < id_count; ++i) {
+                if (i + this->prefetch_stride_code_ < id_count) {
+                    this->prefetch_full_code(idx[i + this->prefetch_stride_code_]);
+                }
+                if (not this->GetCodesById(idx[i], base_code.data)) {
+                    throw VsagException(ErrorType::INTERNAL_ERROR,
+                                        "failed to read split RaBitQ base code");
+                }
+                result_dists[i] = this->quantizer_->Compute(query_code.data, base_code.data);
+            }
```
LHT129, line 270: **Issue**: In the non-optimized `QueryById` path, `this->quantizer_->Compute()` is called directly on the outer quantizer. When `QuantizerT` is `TransformQuantizer<RaBitQuantizer<metric>, metric>` (the MRLE + RaBitQ split path), `TransformQuantizer::ComputeImpl` delegates to the inner RaBitQ quantizer and then applies `ExecuteChainDistanceRecovery`, which reads MRLE metadata from the code buffers.

However, `GetCodesById` zeroes the full code buffer via `memset(codes, 0, this->code_size_)` before calling `bottom_quantizer().MergeSplitCode()`, which only fills the RaBitQ portion. The MRLE metadata area remains zeroed, so `ExecuteChainDistanceRecovery` would compute incorrect distance recovery.

The rest of the class consistently uses `this->bottom_quantizer()` for direct RaBitQ operations (e.g., `ComputeDistWithOneBitLowerBound`, `ComputeDistsWithOneBitLowerBoundBatch4`, `PackScalarCodeToSplitCode`, `ComputeScalarCodesDistance`, `PackScalarCode`, `MergeSplitCode`). The `QueryById` non-optimized path should follow the same pattern:

```suggestion
                result_dists[i] = this->bottom_quantizer().Compute(query_code.data, base_code.data);
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1352,8 +1356,33 @@ Pyramid::add_one_point(const Hierarchy& h,
...
+        if (vector != nullptr) {
+            results = searcher_->Search(
+                node->graph_, codes, vl, vector, search_param, (LabelTablePtr) nullptr, nullptr);
+        } else {
+            FlattenIdDistanceProvider distance_provider(codes, inner_id);
+            results = searcher_->Search(
+                node->graph_, distance_provider, vl, search_param, nullptr, nullptr);
+            if (support_duplicate_ and not results->Empty()) {
+                const auto* data = results->GetData();
+                auto min_distance = data[0].first;
```
LHT129, line 1387: **Suggestion**: The duplicate-detection logic in the `vector == nullptr` branch (lines 1369–1384) duplicates the logic that the searcher normally handles internally when a vector pointer is provided. This duplication means any future changes to duplicate detection in the searcher would need to be mirrored here.

Consider extending `FlattenIdDistanceProvider` (or the searcher) to support the `find_duplicate` / `duplicate_query_id` path so the `vector == nullptr` branch can rely on the same searcher-built-in duplicate detection as the `vector != nullptr` branch, eliminating this code duplication.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1400,7 +1429,8 @@ Pyramid::add_to_hierarchy(Hierarchy& h,
         auto path_slices = split(current_path, PART_SLASH);
         IndexNode* node = h.root.get();
         auto inner_id = static_cast<InnerIdType>(i + local_cur_element_count);
-        const auto* vector = data_vectors + dim_ * data_bias;
+        const auto* vector =
```
LHT129: **Note**: When `base_codes_->SupportSplitCodeStorage()` is true, `vector` is set to `nullptr`, which causes `add_one_point` to use `FlattenIdDistanceProvider` for graph search instead of the raw vector path. This is the core of the refactor — it avoids decoding split codes into FP32 vectors during graph promotion.

The `codes` variable used in `add_one_point` is `has_precise_reorder() ? precise_codes_ : base_codes_`. When `SupportSplitCodeStorage()` is true, `has_precise_reorder()` returns false, so `codes` is `base_codes_`. The `FlattenIdDistanceProvider` then calls `base_codes_->QueryById()`, which uses the new `RaBitQSplitDataCell::QueryById` implementation. This is the correct code path for the refactor goal.

src/impl/distance_provider_for_graph.cpp
LightWant: Intentional and safe: `FlattenIdDistanceProvider` inherits `SupportsComputerById() == false`. `PairwiseDistanceComputer` checks that capability before calling the factory, while `BasicSearcher` uses `QueryDistance`/`BatchQueryDistance`. The existing provider test also verifies that a direct factory call throws. I additionally fixed the clang-tidy parameter naming failure here.

src/datacell/rabitq_split_datacell.h
LightWant: Added a comment documenting that persisted split storage must merge full codes, while the optimized-build path below uses temporary scalar codes. The query code is merged once and reused for the whole batch.

src/algorithm/pyramid/pyramid.cpp
LightWant: Updated the condition to require both split-code support and `raw_vector_ == nullptr`. When raw storage is explicitly enabled, promotion follows the decodable-vector path.

src/algorithm/pyramid/pyramid.cpp
LightWant: Kept the full scan and documented why: `StandardHeap::GetData()` exposes internal heap storage, not sorted nearest-first output, so `data[0]` is not guaranteed to be the minimum.

src/analyzer/pyramid_analyzer.cpp
LightWant: Added `sample_metrics_available` to analyzer output and `sample_metrics_unavailable_reason` when no decodable vectors are retained. English and Chinese docs now describe this behavior.

src/impl/distance_provider_for_graph.h
LightWant: `BucketDistanceProvider` predates this PR and is used by `IVF::search_one_graph_bucket` and `GraphBucketSearcher`; it only appears adjacent to the newly added provider in this diff. No change is needed here.

src/datacell/flatten_interface.h
LightWant: Added a comment to the default `FlattenInterface::QueryById` implementation stating that datacells with batched or storage-aware implementations should override it.

src/datacell/rabitq_split_datacell.h
LightWant: Fixed as suggested. The non-optimized `QueryById` path now calls `bottom_quantizer().Compute`. I also fixed the same issue in non-optimized `ComputePairVectors`, which is used by graph pruning and otherwise affected MRLE split graph quality.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1219,10 +1220,7 @@ Pyramid::CheckAndMappingExternalParam(const JsonType& external_param,
     mapping_external_param_to_inner(external_param, external_mapping, inner_json);
     MapRaBitQSplitParam(external_param, inner_json);
     ValidateMRLEDim(external_param, common_param.dim_);
-    const auto& base_codes_type = inner_json[BASE_CODES_KEY][CODES_TYPE_KEY];
-    const bool uses_split_codes =
-        base_codes_type.IsString() and base_codes_type.GetString() == RABITQ_SPLIT_CODES;
-    if (uses_split_codes or RequiresRawVectorForTransformQuantizer(inner_json)) {
+    if (RequiresRawVectorForTransformQuantizer(inner_json)) {
```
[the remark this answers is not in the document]
LightWant: Per the maintainer decision, compatibility with previously persisted Pyramid split indexes that contain a raw datacell is intentionally out of scope. Serialization remains strictly controlled by the build-time `store_raw_vector` parameter; readers must use the same setting as the index being loaded.

docs/docs/en/src/indexes/pyramid.md
```diff
@@ -121,9 +121,10 @@ Set all five parameters together to enable RaBitQ x+y split storage and reorderi
 }
 ```
 
-Because split codes cannot be decoded back to the input vector, Pyramid also retains an internal
-FP32 copy for incremental flat-to-graph promotion and analyzer sampling. Search distances still use
-the split codes; the FP32 copy adds `count * dim * sizeof(float)` bytes of vector storage.
+Pyramid uses split-code code-to-code distances for incremental flat-to-graph promotion, so it no
+longer retains an internal FP32 copy by default. Raw-vector storage can still be enabled explicitly
```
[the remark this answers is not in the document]
LightWant: Fixed by exposing `store_raw_vector` in Pyramid external parameter mapping. Split and MRLE-split default to false; setting it to true at build time creates/stores the raw datacell. Mapping and build tests cover both modes, and the docs now match the API.

src/algorithm/pyramid/pyramid.cpp
LightWant: Kept this branch scoped to ID-based distance support. The vector-based duplicate path depends on a query computer, while this path has only a stored query ID; the explicit scan preserves the same threshold/exact-compare semantics. A comment now explains the necessary heap scan.

src/algorithm/pyramid/pyramid.cpp
LightWant: Thanks for confirming the intended path. Tests cover both direct split and MRLE split flat-to-graph promotion without raw vectors.

LightWant:
Addressed all review threads in commit 56bd40b26:

- fixed the clang-tidy 15 CI failure in `FlattenIdDistanceProvider`;
- fixed MRLE split ID-to-ID and pairwise distances to use the bottom RaBitQ quantizer;
- made Pyramid `store_raw_vector` an explicit build parameter (split/MRLE split default false, opt-in true);
- added analyzer availability metadata, promotion guards/comments, documentation, and regression tests;
- intentionally did not add compatibility for older split indexes containing raw datacells, per maintainer direction.

Validated with the release build, clang-tidy 15 on all changed production `.cpp` files, and 27 focused unit cases (9,805 assertions). All 12 review threads have been replied to and resolved.

LHT129 (commented):
Reviewed the incremental commit `3eef6a51`. The changes cleanly address the feedback from the previous review round:

1. **`store_raw_vector` external parameter**: Now properly exposed in the Pyramid parameter mapping, allowing users to opt-in to raw vector storage when needed.

2. **MRLE split no longer auto-enables raw vectors**: The condition `RequiresRawVectorForTransformQuantizer(inner_json) and not RequiresRawVectorForMRLERaBitQSplit(inner_json)` correctly prevents MRLE RaBitQ split from forcing raw vector storage, since code-to-code distances are now supported.

3. **`raw_vector_ == nullptr` guard in `promote_flat_to_graph`**: The condition `base_codes_->SupportSplitCodeStorage() and raw_vector_ == nullptr` correctly scopes the split-code promotion path to when raw vectors are genuinely unavailable.

4. **`bottom_quantizer()` correctness fix**: Changing `this->quantizer_->Compute()` to `this->bottom_quantizer().Compute()` in `QueryById` and `ComputePairVectors` is an important fix. When the quantizer is wrapped by a `TransformQuantizer` (e.g., MRLE), `quantizer_->Compute()` would go through the TQ wrapper, but the split codes are already merged into the bottom quantizer format. Using `bottom_quantizer()` correctly unwraps to the actual RaBitQ quantizer. This applies to both the non-optimized-build path in `QueryById` and the fallback path in `ComputePairVectors`.

5. **Analyzer stats**: The `sample_metrics_available` and `sample_metrics_unavailable_reason` fields provide clear visibility when ground-truth metrics cannot be computed.

6. **Tests and docs**: Comprehensive coverage of both the default (no raw vectors) and opt-in (`store_raw_vector=true`) paths, plus updated English and Chinese documentation.

No blocking issues found. The changes are well-structured and the tests cover the key scenarios.

src/datacell/rabitq_split_datacell.h
```diff
@@ -239,6 +239,57 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
...
+                throw VsagException(ErrorType::INTERNAL_ERROR,
+                                    "failed to read split RaBitQ query code");
+            }
+            for (InnerIdType i = 0; i < id_count; ++i) {
+                if (i + this->prefetch_stride_code_ < id_count) {
+                    this->prefetch_full_code(idx[i + this->prefetch_stride_code_]);
+                }
+                if (not this->GetCodesById(idx[i], base_code.data)) {
+                    throw VsagException(ErrorType::INTERNAL_ERROR,
+                                        "failed to read split RaBitQ base code");
```
LHT129, line 267: [note] This change from `this->quantizer_->Compute()` to `this->bottom_quantizer().Compute()` is a correctness fix. When `QuantizerT` is wrapped by `TransformQuantizer` (e.g., MRLE RaBitQ split), `quantizer_->Compute()` would go through the TQ wrapper which may apply transformations to the codes before computing distance. However, `GetCodesById` already merges the split codes into the full RaBitQ format via `bottom_quantizer().MergeSplitCode()`, so the merged codes should be compared directly using the bottom RaBitQ quantizer, not the TQ wrapper. The same fix applies to the `ComputePairVectors` fallback path at line 771.

LHT129 (commented):
Reviewed the incremental commit `3eef6a51`. The changes cleanly address the feedback from the previous review round:

1. **`store_raw_vector` external parameter**: Now properly exposed in the Pyramid parameter mapping.
2. **MRLE split no longer auto-enables raw vectors**: The condition `RequiresRawVectorForTransformQuantizer(inner_json) and not RequiresRawVectorForMRLERaBitQSplit(inner_json)` correctly prevents MRLE RaBitQ split from forcing raw vector storage.
3. **`raw_vector_ == nullptr` guard**: The condition `base_codes_->SupportSplitCodeStorage() and raw_vector_ == nullptr` correctly scopes the split-code promotion path.
4. **`bottom_quantizer()` correctness fix**: Changing `this->quantizer_->Compute()` to `this->bottom_quantizer().Compute()` in `QueryById` and `ComputePairVectors` correctly unwraps to the actual RaBitQ quantizer when TQ is used.
5. **Analyzer stats**: The `sample_metrics_available` and `sample_metrics_unavailable_reason` fields provide clear visibility.
6. **Tests and docs**: Comprehensive coverage of both default and opt-in paths.

No blocking issues found.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -1552,13 +1571,9 @@ PyramidAnalyzer::check_entry_point_duplicate(const IndexNode* node,
...
     }
 
-    Vector<float> entry_vector(dim_, allocator_);
-    pyramid_->GetVectorByInnerId(entry_id, entry_vector.data());
-
     auto codes = pyramid_->base_codes_;
-    auto comp = codes->FactoryComputer(entry_vector.data());
     Vector<float> dists(node_ids.size(), allocator_);
-    codes->Query(dists.data(), comp, node_ids.data(), node_ids.size());
+    codes->QueryById(dists.data(), entry_id, node_ids.data(), node_ids.size());
```
LHT129, line 1577: [note] The `check_entry_point_duplicate` function now uses `QueryById` via code-code distance instead of decoding the entry point to FP32 and computing exact vector distance. For split RaBitQ codes, code-code distance is approximate, which may produce slightly different duplicate-group classifications compared to the previous exact FP32 comparison.

The `2e-6F` epsilon threshold is very tight, so in practice this is unlikely to cause false positives/negatives for true duplicates. However, if the analyzer reports different duplicate ratios after this change for existing split-code indexes, this is the likely cause.

No action required unless you want to document this behavioral difference.

LHT129 (commented):
I have reviewed the latest commit (3eef6a51) and all changes from the previous review round have been properly addressed:

- `FlattenIdDistanceProvider::FactoryComputerById` intentionally throws `UNSUPPORTED_INDEX_OPERATION` (safe, `SupportsComputerById() == false`)
- `QueryById` non-optimized path now uses `bottom_quantizer().Compute` and includes a comment explaining the full-code merge strategy
- `promote_flat_to_graph` condition correctly requires both `SupportSplitCodeStorage()` and `raw_vector_ == nullptr`
- Duplicate detection scan in `add_one_point` is documented with a comment explaining why `StandardHeap::GetData()` requires a full scan
- Analyzer `sample_metrics_available` / `sample_metrics_unavailable_reason` fields are properly reported
- `store_raw_vector` is exposed in the external parameter mapping, and split/MRLE-split defaults no longer auto-enable it
- `get_node_duplicate_ratio` and `check_entry_point_duplicate` correctly use `QueryById` when raw vectors are unavailable
- `add_to_hierarchy` correctly passes `nullptr` for split-code storage
- Default `FlattenInterface::QueryById` fallback includes a comment recommending overrides for batched implementations

The test coverage is thorough: unit tests cover the new `FlattenIdDistanceProvider`, `QueryById` on `RaBitQSplitDataCell`, parameter mapping with and without `store_raw_vector`, and functional tests cover split RaBitQ promotion without raw vectors including serialize/deserialize round-trip.

No new substantive issues found. The PR looks ready.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1400,7 +1433,8 @@ Pyramid::add_to_hierarchy(Hierarchy& h,
         auto path_slices = split(current_path, PART_SLASH);
         IndexNode* node = h.root.get();
         auto inner_id = static_cast<InnerIdType>(i + local_cur_element_count);
-        const auto* vector = data_vectors + dim_ * data_bias;
+        const auto* vector =
```
LHT129: [suggestion] In `add_to_hierarchy`, `vector` is unconditionally set to `nullptr` when `base_codes_->SupportSplitCodeStorage()` is true, even when the user has explicitly enabled `store_raw_vector: true` and `raw_vector_` is available.

In `pyramid.cpp:1436`:
```cpp
const auto* vector =
    base_codes_->SupportSplitCodeStorage() ? nullptr : data_vectors + dim_ * data_bias;
```

This differs from the `promote_flat_to_graph` path at line 1308 which checks both conditions:
```cpp
if (base_codes_->SupportSplitCodeStorage() and raw_vector_ == nullptr) {
```

When raw vectors are stored, the FP32 vector path (via `FlattenDistanceProvider`) may be more efficient than the `FlattenIdDistanceProvider` path for graph insertion, since the FP32 computer can use SIMD-optimized distance computations.

Consider adding the `raw_vector_ == nullptr` check here as well for consistency, or add a comment explaining why the split-code path is always preferred for `add_to_hierarchy` even when raw vectors are available.

LightWant: Fixed. `add_to_hierarchy` now takes the ID-based split path only when split-code storage is active and `raw_vector_ == nullptr`. When `store_raw_vector` is enabled it passes the FP32 vector, matching the flat-to-graph promotion path.

LightWant:
Addressed the latest review and resolved the conflict with current `main`:

- `add_to_hierarchy` now uses ID-based split insertion only when no raw datacell exists; explicit `store_raw_vector=true` keeps the FP32 insertion path.
- Resolved the `pyramid_test.cpp` helper conflict while retaining both `use_reorder` and `store_raw_vector`.
- Preserved the new per-search distance-evaluation statistics from `main`, including `RaBitQSplitDataCell::QueryById`, with regression coverage for both optimized-build and finalized split paths.

Local validation:
- Release `unittests` and `functests` targets build successfully.
- 28 relevant unit cases / 9836 assertions passed.
- 2 BasicSearcher regression cases / 25 assertions passed.
- 3 Pyramid functional cases / 2584 assertions passed.
- clang-tidy 15 passed for changed production `.cpp` files.

LHT129 (commented):
I have reviewed this PR and the existing inline comments already cover the main points of interest. The changes are well-structured:

- Cleanly removes the raw FP32 vector dependency for split-code Pyramid configurations, reducing memory by `count * dim * sizeof(float)` bytes
- The `FlattenIdDistanceProvider` is a well-designed abstraction that plugs into the existing `DistanceProviderForGraph` interface
- `QueryById` on `RaBitQSplitDataCell` correctly handles both optimized-build and persisted-storage paths
- The analyzer gracefully degrades when raw vectors are unavailable, reporting `sample_metrics_available: false` with a clear reason string
- The `add_to_hierarchy` and `promote_flat_to_graph` conditions now consistently check both `SupportSplitCodeStorage()` and `raw_vector_ == nullptr`
- Test coverage is thorough: unit tests for parameter mapping, flat promotion without raw vectors, explicit `store_raw_vector: true`, `FlattenIdDistanceProvider` + `BasicSearcher` integration, and functional regression tests
- Documentation (both English and Chinese) is updated to reflect the new behavior

The existing inline comments from the previous review round have been addressed or are `[note]`/`[suggestion]` level observations. No blocking issues found.

LHT129 (commented):
This update addresses the key feedback from the previous review round. The changes look good:

- `add_to_hierarchy` now correctly checks `raw_vector_ == nullptr` alongside `SupportSplitCodeStorage()`, consistent with the `promote_flat_to_graph` path.
- `ComputePairVectors` now uses `bottom_quantizer().Compute()` instead of `quantizer_->Compute()` for correct type resolution.
- `GetDuplicateRatio` and `check_entry_point_duplicate` correctly fall back to `QueryById` when raw vectors are unavailable, so analyzer duplicate detection remains functional for split-code-only indexes.
- Tests cover the new `store_raw_vector` parameter, the optimized-build `QueryById` path, and the `FlattenIdDistanceProvider` integration with `BasicSearcher`.

No new blocking issues found.

LightWant:
Fixed the failing `Test X86 Functests` check in `ffc585c4c`.

The failure came from `Pyramid Build & ContinueAdd Test`: its dimension was selected during static initialization using `RandomValue(...)`, which seeds from `random_device`. CI selected dim 7, but the printed Catch2 seed could not reproduce that selection. With dim 7 forced locally, the existing filtered range assertion reproduced consistently; this configuration does not enter the split/raw-vector code changed by this PR.

The test now uses the repository helper's deterministic default seed (dim 32), and the diagnostic context includes `graph_type`. Validation: the failing case passed with Catch2 seeds `4192884593`, `1`, and `369`; all three split/MRLE functional tests also passed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1352,8 +1358,35 @@ Pyramid::add_one_point(const Hierarchy& h,
...
 
         auto vl = pool_->TakeOne();
-        auto results = searcher_->Search(
-            node->graph_, codes, vl, vector, search_param, (LabelTablePtr) nullptr, nullptr);
+        DistHeapPtr results;
+        if (vector != nullptr) {
+            results = searcher_->Search(
+                node->graph_, codes, vl, vector, search_param, (LabelTablePtr) nullptr, nullptr);
+        } else {
+            FlattenIdDistanceProvider distance_provider(codes, inner_id);
```
LHT129: [suggestion] The `codes` variable used to construct `FlattenIdDistanceProvider` at this line comes from `has_precise_reorder() ? precise_codes_ : base_codes_` (line 1349). This is currently always `base_codes_` for split codes since `has_precise_reorder()` returns `false` when `base_codes_->SupportSplitCodeStorage()`. If the reorder logic evolves (e.g., a future codec supports both split storage and a separate precise reorder datacell), `codes` could become `precise_codes_` while `vector == nullptr`, causing `QueryById` on the wrong datacell. Consider using `base_codes_` directly here to make the dependency explicit.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -1552,13 +1571,9 @@ PyramidAnalyzer::check_entry_point_duplicate(const IndexNode* node,
...
 
     constexpr float epsilon = 2e-6F;
```
LHT129, line 1578: [note] The `check_entry_point_duplicate` function now uses `QueryById` with code-code distance (line 1576) instead of decoding the entry point to FP32 and computing exact vector distance. The `2e-6F` epsilon at line 1578 was originally tuned for FP32 L2 distances. With RaBitQ code-code distances, quantization error can exceed this epsilon for some vectors (RaBitQ with 1-bit per dimension has quantization error up to ~0.5 per dimension in the normalized space, and the aggregate L2 error can be larger).

This may cause false negatives in duplicate detection — vectors that are true duplicates (same original FP32 vector) might produce code-code distances slightly above `2e-6F` due to quantization. In practice this is unlikely to matter since true duplicates would have identical quantized codes and thus zero code-code distance, but it is worth being aware of if duplicate detection results change after this refactor.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -1494,7 +1509,11 @@ PyramidAnalyzer::get_node_duplicate_ratio(const Vector<InnerIdType>& node_ids) {
             }
 
             Vector<float> dists(group.size(), allocator_);
-            codes->Query(dists.data(), computers[q], group.data(), group.size());
+            if (use_id_queries) {
+                codes->QueryById(dists.data(), sample_ids_[q], group.data(), group.size());
```
LHT129, line 1513: [note] The `get_node_duplicate_ratio` function now uses `QueryById` with code-code distance when `sample_datas_` is empty (line 1512-1513), falling back to the existing `Query` path with FP32 computers otherwise. The `2e-6F` epsilon at line 1497 is used for grouping vectors by approximate distance equality. When using code-code distances via `QueryById`, the distance values are quantized (RaBitQ code-code distances), which means the epsilon-based grouping may produce different cluster boundaries compared to FP32 distances. Identical codes will still produce zero distance and be grouped correctly, but near-duplicates at the boundary may shift between groups. This is consistent with the behavioral change already noted for `check_entry_point_duplicate`.

LightWant:
Resolved the failing functional-test CI by merging current `main` (`75946f3f8`), which includes #2644’s aggregate filtered-range-recall fix. The failed job was unrelated to Pyramid: `[pyramid]` passed, while random recall checks failed in `(PR) HGraph Duplicate` and `(PR) BruteForce Mark Remove`.

The Pyramid conflict was resolved by retaining the ID-based split-code distance path while adopting the new `VisitedListGuard` RAII handling from `main`.

Local Release validation:
- `(PR) BruteForce Mark Remove` with CI seed `4087400654`: passed (14,039 assertions)
- `(PR) HGraph Duplicate` with the same seed: passed (60,094 assertions)
- 3 Pyramid split/MRLE functional tests: passed (2,580 assertions)
- 29 Pyramid/RaBitQ split unit tests: passed (9,840 assertions)
- clang-tidy 15 on the conflict-resolved production file: passed

Pushed merge commit `4e2dbcb10`.

src/algorithm/pyramid/pyramid.cpp
LightWant: Fixed in . The ID-based split path now constructs  from  explicitly, so a future precise-reorder implementation cannot redirect  to the wrong datacell.

src/analyzer/pyramid_analyzer.cpp
LightWant: Acknowledged. For true duplicates built from identical FP32 inputs, deterministic encoding produces identical split codes and a zero code-code distance. Near-duplicate analyzer classification may differ from FP32 when raw vectors are intentionally omitted; this is an accepted consequence of operating without a decodable/raw source.

src/analyzer/pyramid_analyzer.cpp
LightWant: Acknowledged. The no-raw analyzer path intentionally reports duplicate grouping from the retained split codes, while the FP32 path remains in use whenever decoded samples are available. Quantized near-boundary grouping differences are expected and do not affect index search behavior.

LHT129 (commented):
Reviewed commit `4e2dbcb1`. The changes look good:

- `FlattenIdDistanceProvider` now correctly uses `base_codes_` directly (addressing the previous review comment about `codes` vs `base_codes_`).
- `add_to_hierarchy` properly checks `raw_vector_ == nullptr` alongside `SupportSplitCodeStorage()` before passing `nullptr` as the vector.
- Parameter mapping correctly excludes MRLE RaBitQ split from the automatic `store_raw_vector = true` path, and the new `store_raw_vector: true` test confirms explicit opt-in still works.
- Analyzer `GetStats` now reports `sample_metrics_available` / `sample_metrics_unavailable_reason` and gracefully skips recall stats when decoded vectors are unavailable.
- `get_node_duplicate_ratio` falls back to `QueryById` when `sample_datas_` is empty, keeping duplicate detection functional without raw vectors.
- Test coverage is thorough: unit tests for parameter mapping, flat promotion without raw vectors, explicit raw-vector storage, and a functional test with serialize/deserialize round-trip.

No new issues found in this commit.

LHT129 (commented):
Reviewed commit 4e2dbcb1.

The changes in this push address the earlier review feedback:

- FlattenIdDistanceProvider now uses base_codes_ directly instead of the codes variable.
- add_to_hierarchy checks raw_vector_ == nullptr alongside SupportSplitCodeStorage().
- Parameter mapping excludes MRLE RaBitQ split from automatic store_raw_vector, with a new test for explicit opt-in.
- Analyzer GetStats reports sample_metrics_available / sample_metrics_unavailable_reason and skips recall stats when decoded vectors are unavailable.
- get_node_duplicate_ratio falls back to QueryById when sample_datas_ is empty.
- Thorough test coverage: unit tests for parameter mapping, flat promotion without raw vectors, explicit raw-vector storage, and a functional test with serialize/deserialize.

No new issues found.

LHT129 (commented):
Test review - please ignore.

LHT129 (commented):
Reviewed commit 4e2dbcb1.

The changes in this push address the earlier review feedback:

- FlattenIdDistanceProvider now uses base_codes_ directly instead of the codes variable.
- add_to_hierarchy checks raw_vector_ == nullptr alongside SupportSplitCodeStorage().
- Parameter mapping excludes MRLE RaBitQ split from automatic store_raw_vector, with a new test for explicit opt-in.
- Analyzer GetStats reports sample_metrics_available / sample_metrics_unavailable_reason and skips recall stats when decoded vectors are unavailable.
- get_node_duplicate_ratio falls back to QueryById when sample_datas_ is empty.
- Thorough test coverage: unit tests for parameter mapping, flat promotion without raw vectors, explicit raw-vector storage, and a functional test with serialize/deserialize.

No new issues found in this commit.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1366,8 +1372,35 @@ Pyramid::add_one_point(const Hierarchy& h,
...
+                    if (data[i].first < min_distance) {
+                        min_distance = data[i].first;
+                        min_index = data[i].second;
+                    }
+                }
+                if (search_param.duplicate_distance_threshold > 0.0F) {
+                    if (min_distance <= search_param.duplicate_distance_threshold) {
+                        search_param.duplicate_id = min_index;
+                    }
+                } else if (codes->CompareVectors(inner_id, min_index)) {
```
LHT129, line 1399: [suggestion] At line 1380, `FlattenIdDistanceProvider` is constructed with `base_codes_` directly, which is correct. However, at line 1399, `codes->CompareVectors(inner_id, min_index)` uses the `codes` variable from line 1362 (`has_precise_reorder() ? precise_codes_ : base_codes_`). While `codes` currently resolves to `base_codes_` in the split-code path (since `has_precise_reorder()` returns false when `base_codes_->SupportSplitCodeStorage()` is true), using `base_codes_->CompareVectors(...)` directly would make the dependency explicit and prevent a subtle bug if the reorder logic evolves.

```cpp
// line 1399: consider using base_codes_ directly
} else if (base_codes_->CompareVectors(inner_id, min_index)) {
```

src/datacell/rabitq_split_datacell.h
```diff
@@ -732,7 +785,7 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
         ByteBuffer codes2(this->code_size_, allocator_);
         this->GetCodesById(id1, codes1.data);
         this->GetCodesById(id2, codes2.data);
-        return this->quantizer_->Compute(codes1.data, codes2.data);
+        return this->bottom_quantizer().Compute(codes1.data, codes2.data);
```
LHT129, line 788: [note] The change from `this->quantizer_->Compute(...)` to `this->bottom_quantizer().Compute(...)` in `ComputePairVectors` (line 788) is a correctness fix. The previous code used the top-level `quantizer_` which wraps the split codec (one-bit + supplement merge), but for split code-code distance computation the bottom quantizer must be used directly since the codes are already merged at that point. This is consistent with how `Query` and the new `QueryById` both use `bottom_quantizer().Compute(...)`.

This fix is important — without it, `ComputePairVectors` on split codes would produce incorrect distances when the top-level quantizer applies additional transformations.

src/datacell/rabitq_split_datacell.h
```diff
@@ -243,6 +243,59 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
...
+                    throw VsagException(ErrorType::INTERNAL_ERROR,
+                                        "failed to read split RaBitQ base code");
+                }
+                result_dists[i] = this->bottom_quantizer().Compute(query_code.data, base_code.data);
+            }
+            this->add_distance_evaluations(ctx, id_count);
+            return;
+        }
+
+        bool need_release = false;
```
LHT129, line 275: [note] In the optimized-build path of `QueryById` (lines 275-296), `need_release` is initialized to `false` and then passed to `Read()`. If `Read()` returns non-null with `need_release == false`, the try-catch block still checks `need_release` before calling `Release()`, which is harmless but the try-catch is unnecessary in that case. The pattern is correct and matches the existing `Query` implementation, but worth noting that `Read()` returning non-null without requiring release means the scalar code is owned by the storage and no cleanup is needed.

LHT129 (commented):
I have reviewed the latest commit (5d3142de) against the previous review rounds. All previously raised issues have been properly addressed:

- `promote_flat_to_graph` now requires both `SupportSplitCodeStorage()` and `raw_vector_ == nullptr`
- `add_to_hierarchy` correctly passes `nullptr` for split-code storage when `raw_vector_ == nullptr`
- `FlattenIdDistanceProvider` construction in `add_one_point` uses `base_codes_` directly
- Duplicate detection scan is documented with a comment explaining why `StandardHeap::GetData()` requires a full scan
- `QueryById` non-optimized path uses `bottom_quantizer().Compute` (same fix applied to `ComputePairVectors`)
- Analyzer reports `sample_metrics_available` / `sample_metrics_unavailable_reason`
- `store_raw_vector` is exposed in external parameter mapping; split/MRLE-split defaults no longer auto-enable it
- `get_node_duplicate_ratio` and `check_entry_point_duplicate` use `QueryById` when raw vectors are unavailable
- Default `FlattenInterface::QueryById` fallback includes a comment recommending overrides

Test coverage is thorough across unit tests (parameter mapping, `FlattenIdDistanceProvider` + `BasicSearcher` integration, `QueryById` on `RaBitQSplitDataCell`, promotion without/with raw vectors) and functional tests (split RaBitQ promotion including serialize/deserialize round-trip).

No new substantive issues found. The PR looks ready.

LightWant:
Fixed the remaining flaky CI failures in `296d9a5ae`.

Root cause: the sampled HGraph and BruteForce functional-test dimensions were selected with `RandomValue(..., seed=-1)`, which seeds from `random_device` outside Catch2’s RNG. Consequently, the printed Catch2 seed could not reproduce the dimension that hit a recall boundary. The failed jobs selected different unrelated cases (`HGraph Duplicate`, `BruteForce Build & ContinueAdd`, and `BruteForce Duplicate Build`); Pyramid/RaBitQ split groups passed.

The PR sample dimensions now use the fixture helper’s deterministic seed. Daily tests retain randomized multi-dimension coverage, and recall thresholds are unchanged.

Local Release validation:
- reproduced `BruteForce Build & ContinueAdd` with CI seed `4231306042` before the fix;
- all three CI-failing cases pass after the fix with their recorded seeds;
- both BruteForce cases passed 5/5 repeated seed runs;
- HGraph Duplicate also passed with an additional seed;
- clang-format 15 dry-run and `git diff --check` passed.

LHT129 (commented):
This commit (296d9a5) stabilizes sampled functional test dimensions by replacing `RandomValue(0, 999)` with a deterministic seed (369, the default) in `test_brute_force.cpp` and `test_hgraph.cpp`. The change is minimal and correct:

- `test_brute_force.cpp`: `get_common_used_dims(1, RandomValue(0, 999))` → `get_common_used_dims(/*count=*/1)` — uses the default seed=369.
- `test_hgraph.cpp`: `get_common_used_dims(1, RandomValue(0, 999), 257)` → `get_common_used_dims(/*count=*/1, /*seed=*/369, /*limited_dim=*/257)` — explicitly passes seed=369.

This makes the test dimension selection reproducible under Catch2's RNG seeding, which is the right approach for CI stability. No issues found.

wxyucs (approved):
lgtm

PR #2665, opened 2026-08-12
fix(ivf): routing HGraph thread count and thread pool issues

### LINKED ISSUE #2657: [bug](ivf): routing HGraph spawns ~100 threads at index construction regardless of thread_count / num_threads_building
### Affected area

ivf

### Interface

cpp

### VSAG version

75946f3f

### Summary

The IVF build thread_count only sizes IVF's own pool. The centroid routing
HGraph is created with build_thread_count=100 (the hgraph template default,
src/algorithm/hgraph/hgraph_param_mapping.cpp:650), which factory_router_index
never overrides. Because the base builder calls SetPoolSize(100) in the HGraph
constructor and the underlying pool spawns workers eagerly, ~100 OS threads are
launched at CreateIndex time — before Build — even when thread_count=4 and
Options::num_threads_building=4. Separately, kmeans never receives IVF's pool
(IVFNearestPartition::Train omits the thread_pool arg), so it always uses its
own default pool.

### Minimal reproducer

#include <vsag/vsag.h>
#include <fstream>
#include <string>
// read "Threads:" from /proc/self/status  ... (helper omitted)
int main() {
    vsag::init();
    vsag::Options::Instance().set_num_threads_building(4);
    // build_thread_count("thread_count")=4, buckets_count=1024, graph_build_threshold=100
    auto idx = vsag::Factory::CreateIndex("ivf", params).value();
    // print /proc/self/status Threads here  -> already ~109
    idx->Build(ds);   // peak ~114
}
// full ~60-line version available on request.

### Environment

```shell
- OS:  Linux (glibc), x86_64 
- vsag version: main @ 75946f3f
- compiler version: c++ (GCC) 9.2.1 20240328
```

### Expected behavior

Thread usage should be bounded by thread_count / Options::num_threads_building.

### Actual behavior

after init + set_num_threads_building(4) : 1 thread
after CreateIndex (before Build)         : 109 threads   (+108 from constructing the index alone)
peak during Build                        : 114 threads

### References

- src/algorithm/hgraph/hgraph_param_mapping.cpp:650  (build_thread_count: 100 template default)
- src/algorithm/inner_index_interface.cpp:81-86  (SetPoolSize called in ctor; and OUTSIDE the null-check -> also resizes a caller-supplied shared pool)
- build/_deps/thread_pool-src/ThreadPool.h:227-228  (set_pool_size spawns workers eagerly)
- src/algorithm/ivf/ivf_nearest_partition.cpp:66  (KMeansCluster cls(dim, allocator_) — thread_pool arg omitted; kmeans falls back to its own pool)
- src/impl/cluster/kmeans_cluster.cpp:322  (counter-evidence: kmeans DOES forward its pool downstream; only the IVF->kmeans hop misses it)
- src/algorithm/ivf/ivf.cpp:416-419  (IVF keeps SetPoolSize inside the null-check — contrast, does not touch shared pool)

### Additional context

Two related fixes under one topic ("build thread-pool control is incomplete"),
happy to split:
(a) forward IVF's pool to kmeans: KMeansCluster cls(dim, allocator_, this->thread_pool_);
(b) make routing-graph build_thread_count follow thread_count / Options, and
    move SetPoolSize inside the null-check so a shared pool is not mutated.

LHT129:
## Summary

Fixes three thread-related issues in IVF routing HGraph:

1. **Routing HGraph thread pool**: `IVF::IVF()` now initializes the thread pool before constructing the partition strategy, and passes it via `modified_common_param.thread_pool_`. This ensures the routing HGraph reuses IVF thread pool instead of creating its own default pool with 100 threads.

2. **SetPoolSize scope**: Moved `SetPoolSize` inside the `thread_pool_ == nullptr` check in `InnerIndexInterface` to avoid modifying shared thread pool size.

3. **KMeans thread reuse**: `KMeansCluster` now receives the IVF `thread_pool_` for thread reuse instead of creating its own default pool.

Fixes: https://github.com/antgroup/vsag/issues/2657

## Changes

- `src/algorithm/ivf/ivf.cpp`: Initialize thread pool before partition strategy construction; pass `modified_common_param` with initialized `thread_pool_` to `IVFNearestPartition`
- `src/algorithm/ivf/ivf_nearest_partition.cpp`: Pass `thread_pool_` to `KMeansCluster` constructor
- `src/algorithm/inner_index_interface.cpp`: Move `SetPoolSize` inside null-check to avoid modifying shared thread pools
- `scripts/testing/test_parallel_bg.sh`: Add heartbeat output to prevent CircleCI timeout during long test runs

## Testing

All existing IVF-related tests pass:
- `[IVFNearestPartition]`: 2004 assertions in 2 test cases
- `[IVFParameter]`: 48 assertions in 4 test cases
- `[GNOIMIPartition]`: 8 assertions in 2 test cases

CircleCI timeout issue fixed by adding periodic heartbeat output to `test_parallel_bg.sh`.

### DIFF
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -397,27 +397,34 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
     if (this->bucket_ == nullptr) {
         throw VsagException(ErrorType::INTERNAL_ERROR, "bucket init error");
     }
-    if (param->ivf_partition_strategy_parameter->partition_strategy_type ==
-        IVFPartitionStrategyType::IVF) {
-        this->partition_strategy_ = std::make_shared<IVFNearestPartition>(
-            bucket_->bucket_count_, common_param, param->ivf_partition_strategy_parameter);
-    } else if (param->ivf_partition_strategy_parameter->partition_strategy_type ==
-               IVFPartitionStrategyType::GNO_IMI) {
-        this->partition_strategy_ = std::make_shared<GNOIMIPartition>(
-            common_param, param->ivf_partition_strategy_parameter);
-    }
-    if (this->use_reorder_) {
-        this->reorder_codes_ =
-            FlattenInterface::MakeInstance(param->precise_codes_param, common_param);
-        reorder_ = std::make_shared<FlattenReorder>(this->reorder_codes_, allocator_);
-    }
-    if (param->bucket_param->use_residual_) {
-        this->bucket_->SetStrategy(partition_strategy_);
-    }
+
+    // Initialize thread pool before partition strategy construction
     this->thread_pool_ = common_param.thread_pool_;
     if (param->build_thread_count > 1 and this->thread_pool_ == nullptr) {
         this->thread_pool_ = SafeThreadPool::FactoryDefaultThreadPool();
         this->thread_pool_->SetPoolSize(param->build_thread_count);
+    }
+
+    // Create modified common_param with the initialized thread_pool_
+    IndexCommonParam modified_common_param = common_param;
+    modified_common_param.thread_pool_ = this->thread_pool_;
+
+    if (param->ivf_partition_strategy_parameter->partition_strategy_type ==
+        IVFPartitionStrategyType::IVF) {
+        this->partition_strategy_ = std::make_shared<IVFNearestPartition>(
+            bucket_->bucket_count_, modified_common_param, param->ivf_partition_strategy_parameter);
+    } else if (param->ivf_partition_strategy_parameter->partition_strategy_type ==
+               IVFPartitionStrategyType::GNO_IMI) {
+        this->partition_strategy_ = std::make_shared<GNOIMIPartition>(
+            modified_common_param, param->ivf_partition_strategy_parameter);
+    }
+    if (this->use_reorder_) {
+        this->reorder_codes_ =
+            FlattenInterface::MakeInstance(param->precise_codes_param, modified_common_param);
+        reorder_ = std::make_shared<FlattenReorder>(this->reorder_codes_, allocator_);
+    }
+    if (param->bucket_param->use_residual_) {
+        this->bucket_->SetStrategy(partition_strategy_);
     }
 
     this->graph_param_ = param->graph_param;
--- a/src/algorithm/ivf/ivf_nearest_partition.cpp
+++ b/src/algorithm/ivf/ivf_nearest_partition.cpp
@@ -63,7 +63,7 @@ IVFNearestPartition::Train(const DatasetPtr dataset) {
     if (ivf_partition_strategy_param_->partition_train_type ==
         IVFNearestPartitionTrainerType::KMeansTrainer) {
         constexpr int32_t kmeans_iter_count = 25;
-        KMeansCluster cls(static_cast<int32_t>(dim), this->allocator_);
+        KMeansCluster cls(static_cast<int32_t>(dim), this->allocator_, this->thread_pool_);
         cls.Run(this->bucket_count_,
                 dataset->GetFloat32Vectors(),
                 dataset->GetNumElements(),
@@ -125,7 +125,7 @@ IVFNearestPartition::ClassifyDatas(const void* datas,
                                     std::strtoull(route_stats[1].c_str(), nullptr, 10));
         }
     };
-    if (thread_pool_ == nullptr) {
+    if (thread_pool_ == nullptr or count == 1) {
         for (int64_t i = 0; i < count; ++i) {
             task(i);
         }
--- a/src/algorithm/inner_index_interface.cpp
+++ b/src/algorithm/inner_index_interface.cpp
@@ -79,10 +79,8 @@ InnerIndexInterface::InnerIndexInterface(const InnerIndexParameterPtr& index_par
     }
 
     this->thread_pool_ = common_param.thread_pool_;
-    if (this->build_thread_count_ > 1) {
-        if (this->thread_pool_ == nullptr) {
-            this->thread_pool_ = SafeThreadPool::FactoryDefaultThreadPool();
-        }
+    if (this->build_thread_count_ > 1 and this->thread_pool_ == nullptr) {
+        this->thread_pool_ = SafeThreadPool::FactoryDefaultThreadPool();
         this->thread_pool_->SetPoolSize(build_thread_count_);
     }
 


src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -157,12 +158,14 @@ IVFNearestPartition::Deserialize(lvalue_or_rvalue<StreamReader> reader) {
     this->route_index_ptr_->Deserialize(reader);
 }
 void
-IVFNearestPartition::factory_router_index(const IndexCommonParam& common_param) {
+IVFNearestPartition::factory_router_index(const IndexCommonParam& common_param,
+                                          int32_t build_thread_count) {
     ParamPtr param_ptr;
     JsonType hgraph_json;
     hgraph_json["base_quantization_type"].SetString("fp32");
```
LHT129, line 163: [note] `build_thread_count` is passed directly to `hgraph_json["build_thread_count"].SetInt()` without any validation. If `build_thread_count` is 0 or negative (e.g. due to a misconfiguration), HGraph may behave unexpectedly. The `InnerIndexInterface` constructor already guards against `build_thread_count <= 1` by skipping thread pool creation, but the JSON value is still set unconditionally.

Consider adding a guard or clamping, e.g.:
```cpp
hgraph_json["build_thread_count"].SetInt(std::max(1, build_thread_count));
```

LHT129 (commented):
[note] **Thread pool propagation gap in IVFNearestPartition**

The `IVFNearestPartition::thread_pool_` (inherited from `IVFPartitionStrategy`) is initialized from `common_param.thread_pool_` at construction time. However, the IVF constructor may later replace `this->thread_pool_` (from `InnerIndexInterface`) with a newly created pool when `common_param.thread_pool_` is `nullptr` and `build_thread_count > 1` (see `ivf.cpp` lines ~420-424).

This means when the shared thread pool is null and a new one is created by IVF, `KMeansCluster` in `IVFNearestPartition::Train` will still receive `nullptr` and fall back to creating its own default thread pool, rather than reusing the pool that IVF created.

This is a pre-existing architectural issue (not introduced by this PR), but it limits the effectiveness of the KMeans thread reuse improvement described in the PR summary. Consider whether `IVFNearestPartition` should receive the thread pool from IVF rather than from `common_param` directly.

LHT129 (commented):
Overall this PR looks good. The three fixes are well-targeted and address the issues described in #2657:

1. **Routing HGraph thread count**: Passing `build_thread_count` to `factory_router_index` correctly ensures the routing HGraph respects the IVF `thread_count` setting instead of using the default 100.

2. **SetPoolSize scope**: Moving `SetPoolSize` inside the `thread_pool_ == nullptr` check in `InnerIndexInterface` correctly prevents modifying a shared/external thread pool.

3. **KMeans thread reuse**: Passing `this->thread_pool_` to `KMeansCluster` enables thread reuse, avoiding creation of a separate default pool.

One minor note was left inline regarding `build_thread_count` validation in `factory_router_index`.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: Fixed in 36d4c3a7. Added std::max<uint64_t>(1, build_thread_count) to clamp to at least 1 before passing to SetUint64.

src/algorithm/ivf/ivf_nearest_partition.h
```diff
@@ -27,7 +27,8 @@ class IVFNearestPartition : public IVFPartitionStrategy {
 public:
     explicit IVFNearestPartition(BucketIdType bucket_count,
                                  const IndexCommonParam& common_param,
-                                 IVFPartitionStrategyParametersPtr param);
+                                 IVFPartitionStrategyParametersPtr param,
+                                 int32_t build_thread_count);
```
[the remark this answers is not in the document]
LHT129: Fixed in 36d4c3a7. Changed int32_t to uint64_t to match InnerIndexParameter::build_thread_count type. Updated test call sites to pass build_thread_count=1.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -41,10 +41,11 @@ static constexpr const char* SEARCH_PARAM_TEMPLATE_STR = R"(
 
 IVFNearestPartition::IVFNearestPartition(BucketIdType bucket_count,
                                          const IndexCommonParam& common_param,
-                                         IVFPartitionStrategyParametersPtr param)
+                                         IVFPartitionStrategyParametersPtr param,
+                                         int32_t build_thread_count)
```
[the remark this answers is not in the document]
LHT129: Fixed in 36d4c3a7. Changed to uint64_t in both declaration and definition, and used SetUint64 instead of SetInt.

LHT129:
Addressed all review comments in 36d4c3a7:

1. Changed `int32_t` to `uint64_t` for `build_thread_count` to match `InnerIndexParameter::build_thread_count` type
2. Added `std::max<uint64_t>(1, build_thread_count)` validation before `SetUint64`
3. Updated test call sites in `ivf_nearest_partition_test.cpp` to pass `build_thread_count=1`

Regarding the thread pool propagation note: this is indeed a pre-existing architectural issue where `IVFNearestPartition` receives the pool from `common_param` before IVF may create a new one. Out of scope for this PR, but worth a follow-up.

DCO and CI should now pass with the amended commit.

src/algorithm/ivf/ivf.cpp
```diff
@@ -398,8 +398,11 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
     }
```
LHT129, line 399: [note] The thread pool initialization in `IVF::IVF()` duplicates the same logic already executed in `InnerIndexInterface::InnerIndexInterface()`. Since `IVF` inherits from `InnerIndexInterface`, the base class constructor runs first and already sets up `thread_pool_` when `build_thread_count > 1` and `common_param.thread_pool_` is null. The `this->thread_pool_ = common_param.thread_pool_;` assignment then overwrites that setup, making the subsequent `if` block effectively dead code.

This is pre-existing and not introduced by this PR, but worth noting for future cleanup.

LHT129 (commented):
Reviewed commit `36d4c3a7`. All three fixes look correct:

1. **Routing HGraph thread count**: `build_thread_count` is now passed through `IVFNearestPartition` → `factory_router_index` → `hgraph_json["build_thread_count"]`, with `std::max<uint64_t>(1, ...)` clamping for safety. The type is consistently `uint64_t` throughout.

2. **SetPoolSize scope**: `SetPoolSize` is now inside the `thread_pool_ == nullptr` guard in `InnerIndexInterface`, preventing modification of a shared thread pool.

3. **KMeans thread reuse**: `KMeansCluster` now receives `this->thread_pool_` from `IVFNearestPartition`, avoiding creation of a separate default thread pool.

The prior review comments (int32_t → uint64_t type consistency, value clamping) have been addressed. Test files updated accordingly. No blocking issues found.

LHT129 (commented):
Overall this PR looks good. The three fixes are well-scoped and correctly implemented:

1. **Routing HGraph thread count**: Passing `build_thread_count` from IVF parameters to the routing HGraph via `factory_router_index` with proper `std::max<uint64_t>(1, ...)` clamping is correct.

2. **SetPoolSize scope**: Moving `SetPoolSize` inside the `thread_pool_ == nullptr` check in `InnerIndexInterface` correctly avoids modifying a shared thread pool.

3. **KMeans thread reuse**: Passing the IVF `thread_pool_` to `KMeansCluster` enables thread reuse, and `KMeansCluster` already handles the `nullptr` case internally.

The type change from `int32_t` to `uint64_t` for `build_thread_count` is consistent with the rest of the codebase. Test updates are minimal and appropriate.

One minor note was left as an inline comment about pre-existing thread pool initialization duplication in `IVF::IVF()`, not introduced by this PR.

LHT129 (commented):
All three thread-related fixes look correct and well-implemented:

1. **Routing HGraph thread count** (`ivf_nearest_partition.cpp:166`): `build_thread_count` is now properly passed to HGraph JSON with `std::max<uint64_t>(1, ...)` clamping, and the type is consistently `uint64_t` throughout the call chain.

2. **SetPoolSize scope** (`inner_index_interface.cpp:82-84`): `SetPoolSize` is correctly moved inside the `thread_pool_ == nullptr` guard, preventing unintended modification of a shared thread pool.

3. **KMeans thread reuse** (`ivf_nearest_partition.cpp:68`): `KMeansCluster` now receives `this->thread_pool_` for thread reuse instead of creating its own default pool.

All prior review comments (type consistency, value clamping) have been addressed in this commit. The test file has been updated to match the new constructor signature.

src/algorithm/ivf/ivf.cpp
```diff
@@ -398,8 +398,11 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
...
     if (param->ivf_partition_strategy_parameter->partition_strategy_type ==
         IVFPartitionStrategyType::IVF) {
-        this->partition_strategy_ = std::make_shared<IVFNearestPartition>(
-            bucket_->bucket_count_, common_param, param->ivf_partition_strategy_parameter);
+        this->partition_strategy_ =
```
LHT129: [note] `IVFNearestPartition` is constructed with `common_param` (line 402) before IVF creates its own thread pool (lines 419-423). When `common_param.thread_pool_` is null and `build_thread_count > 1`, IVF will create a thread pool, but `IVFNearestPartition`'s `thread_pool_` remains null. This means `ClassifyDatas` runs single-threaded and `KMeansCluster` creates its own default pool instead of reusing IVF's pool.

This is pre-existing and not introduced by this PR — `KMeansCluster` and `ClassifyDatas` both handle null `thread_pool_` correctly. But since this PR adds the `KMeansCluster` thread pool passthrough, it's worth noting that the passed `this->thread_pool_` will always be the original `common_param.thread_pool_`, not the one IVF may create afterwards.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed in 5a39456a. Moved thread_pool_ initialization before partition_strategy_ construction. Now IVF creates its thread pool first, then passes the initialized thread_pool_ to IVFNearestPartition via a modified common_param.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed in 5a39456a. Adjusted construction order: IVF now initializes thread_pool_ before constructing IVFNearestPartition. A modified common_param with the initialized thread_pool_ is passed to the partition strategy, ensuring IVFNearestPartition receives the correct thread pool instead of null.

LHT129:
Fixed construction order issue in 5a39456a:

- Moved `thread_pool_` initialization before `partition_strategy_` construction
- Created `modified_common_param` with the initialized `thread_pool_` to pass to `IVFNearestPartition`
- Ensures `IVFNearestPartition` receives the correct thread pool instead of null

All IVF-related tests pass. CI re-running with the amended commit.

src/algorithm/ivf/ivf.cpp
```diff
@@ -396,28 +396,38 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
     if (this->bucket_ == nullptr) {
         throw VsagException(ErrorType::INTERNAL_ERROR, "bucket init error");
     }
+
+    // Initialize thread pool before partition strategy construction
```
LHT129, line 401: [note] The thread pool initialization in `IVF::IVF()` (lines 400-404) duplicates logic already executed in `InnerIndexInterface::InnerIndexInterface()` (line 82-85). When `common_param.thread_pool_` is null and `build_thread_count > 1`, the base class constructor creates a `FactoryDefaultThreadPool()`, then `IVF` overwrites `this->thread_pool_` with null (line 400) and creates another one. The first pool is orphaned.

This is pre-existing and not introduced by this PR, but since this PR touches the same code, it may be worth cleaning up in a follow-up — e.g. by having `IVF` skip the base class thread pool logic and handle it entirely in its own constructor.

src/algorithm/ivf/ivf.cpp
```diff
@@ -396,28 +396,38 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
     if (this->bucket_ == nullptr) {
         throw VsagException(ErrorType::INTERNAL_ERROR, "bucket init error");
     }
+
```
LHT129, line 400: [note] The thread pool initialization in `IVF::IVF()` (lines 399-404) duplicates the same logic already executed in `InnerIndexInterface::InnerIndexInterface()`. Since `IVF` inherits from `InnerIndexInterface`, the base class constructor runs first and already sets up `this->thread_pool_` when `build_thread_count > 1` and `common_param.thread_pool_` is null. The `this->thread_pool_ = common_param.thread_pool_;` assignment then overwrites that setup, making the subsequent `if` block dead code.

This is not a correctness issue — the result is the same either way — but it is redundant code. Consider removing the IVF-level thread pool initialization and relying solely on the base class setup. If the concern is that `modified_common_param` needs the thread pool before the base class finishes, note that `modified_common_param` is constructed after the base class constructor completes, so `this->thread_pool_` is already available.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed in 7842055e. Changed approach: instead of passing build_thread_count separately, we now initialize thread_pool first and pass it via modified_common_param. This eliminates duplication - the routing HGraph and KMeansCluster simply reuse IVF's thread_pool through common_param.

src/algorithm/ivf/ivf.cpp
LHT129: Fixed in 7842055e. Same approach: thread_pool is initialized first, then passed via modified_common_param to IVFNearestPartition, GNOIMIPartition, and FlattenInterface. All components now share the same thread_pool instance, eliminating the duplication.

LHT129:
## Update in 7842055e

Simplified the approach based on review feedback. Instead of passing `build_thread_count` as a separate parameter, we now:

1. **Initialize thread_pool first** in `IVF::IVF()` before constructing partition strategies
2. **Create modified_common_param** with the initialized `thread_pool_`
3. **Pass modified_common_param** to `IVFNearestPartition`, `GNOIMIPartition`, and `FlattenInterface`

This way, the routing HGraph and KMeansCluster automatically reuse IVF's thread pool through `common_param.thread_pool_`, eliminating:
- The thread pool propagation gap (routing HGraph no longer creates its own pool)
- Code duplication (no need to pass `build_thread_count` separately)
- The need for the `build_thread_count` parameter in `IVFNearestPartition` constructor

All components now share the same thread pool instance.

LHT129:
## Update in 7842055e

Simplified approach: instead of passing build_thread_count separately, we now initialize thread_pool first and pass it via modified_common_param. The routing HGraph and KMeansCluster reuse IVF thread_pool through common_param.thread_pool_, eliminating duplication and the propagation gap.

LHT129 (commented):
The latest approach (commit 7842055e) is cleaner than the previous iteration: instead of plumbing `build_thread_count` through `IVFNearestPartition`, the thread pool is initialized early in `IVF::IVF()` and propagated via `modified_common_param`. This means `IVFNearestPartition`, `GNOIMIPartition`, `FlattenInterface`, and `KMeansCluster` all share the same thread pool instance through `common_param.thread_pool_`.

Key changes look correct:
- `inner_index_interface.cpp`: `SetPoolSize` is now correctly scoped inside the `thread_pool_ == nullptr` check, preventing unintended modification of a shared/external thread pool.
- `ivf.cpp`: Thread pool initialization moved before partition strategy construction, ensuring downstream components receive the initialized pool.
- `ivf_nearest_partition.cpp`: `KMeansCluster` reuses IVF's thread pool; `factory_router_index` no longer needs a separate `build_thread_count` parameter since the pool is already in `common_param`.

No new issues introduced by this commit. The previous review comments (int32_t vs uint64_t, build_thread_count validation) have been resolved by the architectural change.

wxyucs (approved):
lgtm

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -17,6 +17,7 @@
```
LHT129, line 17: [critical] The PR description states that `factory_router_index` now passes `build_thread_count` to HGraph JSON, but the actual code does not set `build_thread_count` in the HGraph JSON at all. The method only sets `base_quantization_type`, `max_degree`, and `ef_construction`. While `thread_pool_` is now passed via `modified_common_param`, the HGraph may still use its own default `build_thread_count` (100) for internal parallelism decisions. If the intent is for HGraph to respect the IVF `thread_count`, the JSON should include `hgraph_json["build_thread_count"].SetUint64(common_param.build_thread_count_);`. Alternatively, if passing the thread pool via `common_param` is sufficient, the PR description should be updated to reflect what was actually changed.

LHT129 (commented):
Code review for commit 4dc22a31.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include <fmt/format.h>
 
+#include <algorithm>
```
LHT129: [note] The `#include <algorithm>` added in this file appears to be unused in the current version. It was likely needed in a previous iteration of this PR that used `std::max` for clamping `build_thread_count`, but the current code no longer calls any algorithm functions (`std::max`, `std::min`, `std::sort`, etc.). The `std::iota` call on line 57 comes from `<numeric>`, not `<algorithm>`.

Consider removing this include to keep the include list minimal.

src/algorithm/ivf/ivf.cpp
```diff
@@ -397,28 +397,35 @@ IVF::IVF(const IVFParameterPtr& param, const IndexCommonParam& common_param)
...
+    // Initialize thread pool before partition strategy construction
```
LHT129, line 401: [note] The thread pool initialization in `IVF::IVF()` (lines 401-405) duplicates the same logic already executed in `InnerIndexInterface::InnerIndexInterface()` (the base class constructor). Since `IVF` inherits from `InnerIndexInterface`, the base class constructor runs first and already creates a thread pool when `build_thread_count > 1` and `common_param.thread_pool_` is `nullptr`. The `this->thread_pool_ = common_param.thread_pool_;` assignment on line 401 then overwrites that setup with `nullptr` (since `common_param.thread_pool_` is still `nullptr`), and the subsequent `if` block creates a second thread pool.

This is pre-existing behavior (the old code at the bottom of the constructor had the same pattern) and not a correctness issue — the end result is the same. However, it does create a short-lived unused thread pool from the base class. Consider removing the IVF-level initialization and relying solely on the base class setup, since `modified_common_param` is constructed after the base class completes, so `this->thread_pool_` is already available.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: [note] The `#include <algorithm>` added in this file is now unused in the current version. It was needed in a previous iteration that used `std::max` for clamping `build_thread_count`, but the current code no longer calls any algorithm functions (`std::max`, `std::min`, `std::sort`, etc.). The `std::iota` call on line 57 comes from `<numeric>`, not `<algorithm>`.

Consider removing this include to keep the include list minimal.

LHT129 (commented):
[note] The PR description states that `factory_router_index` now passes `build_thread_count` to HGraph JSON, but the current code does not set `build_thread_count` in the HGraph JSON. The method only sets `base_quantization_type`, `max_degree`, and `ef_construction`. The thread pool is passed via `modified_common_param.thread_pool_` (set in `IVF::IVF()`), which allows HGraph to use the IVF thread pool for its internal operations. This is a valid approach, but the PR description should be updated to accurately reflect what was changed: the routing HGraph now receives the IVF thread pool via `common_param`, not via the JSON `build_thread_count` field.

LHT129 (commented):
Reviewed commit b50ba179. The three core fixes look correct:

1. **SetPoolSize scope** (`inner_index_interface.cpp`): `SetPoolSize` is now properly inside the `thread_pool_ == nullptr` check, avoiding modification of a shared thread pool. ✓

2. **Thread pool initialization order** (`ivf.cpp`): Thread pool is now initialized before partition strategy construction, and `modified_common_param` correctly propagates the initialized `thread_pool_` to `IVFNearestPartition`, `GNOIMIPartition`, and `FlattenInterface`. ✓

3. **KMeans thread reuse** (`ivf_nearest_partition.cpp`): `KMeansCluster` now receives `this->thread_pool_` for thread reuse instead of creating its own default pool. ✓

**Pre-existing issues noted (not introduced by this PR):**
- The thread pool initialization in `IVF::IVF()` duplicates the same logic already executed in `InnerIndexInterface::InnerIndexInterface()`, creating a short-lived unused thread pool from the base class. This is pre-existing behavior and not a correctness issue.
- The `#include <algorithm>` in `ivf_nearest_partition.cpp` is unused in the current version.

**CI script** (`test_parallel_bg.sh`): The heartbeat addition to prevent CI timeout looks reasonable.

No blocking issues found in this commit.

scripts/testing/test_parallel_bg.sh
```diff
@@ -37,12 +37,32 @@ done
 pids+=($!)
 logger_files+=("./log/other.log")
 
+# Heartbeat: print status every 60 seconds to prevent CI timeout
```
LHT129: [suggestion] The CI heartbeat mechanism added to `test_parallel_bg.sh` is unrelated to the stated purpose of this PR ("fix(ivf): routing HGraph thread count and thread pool issues"). This is a CI infrastructure improvement that should be submitted as a separate PR to keep changes focused and make git history easier to understand.

Consider moving this change to its own PR.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: [note] The `#include <algorithm>` added in this file is unused in the current version of the code. It was likely needed in a previous iteration that used `std::max` for clamping `build_thread_count`, but the current code no longer calls any functions from `<algorithm>`. The `std::iota` call on line 57 comes from `<numeric>`, not `<algorithm>`.

Consider removing this include to keep the include list minimal.

src/algorithm/ivf/ivf.cpp
LHT129, line 401: [note] The thread pool initialization in `IVF::IVF()` (lines 401-405) duplicates the same logic already executed in `InnerIndexInterface::InnerIndexInterface()` (the base class constructor, line 79-82 of inner_index_interface.cpp). Since `IVF` inherits from `InnerIndexInterface`, the base class constructor runs first and already creates a thread pool when `build_thread_count > 1` and `common_param.thread_pool_` is `nullptr`.

However, `IVF` declares its own `thread_pool_` member (ivf.h:293) which shadows the base class member (inner_index_interface.h:709). So `this->thread_pool_ = common_param.thread_pool_;` on line 401 sets `IVF::thread_pool_` to `nullptr` (since `common_param.thread_pool_` is still `nullptr`), and the subsequent `if` block creates a second thread pool for `IVF::thread_pool_`. The base class thread pool is never used by IVF.

This is pre-existing behavior (the old code at the bottom of the constructor had the same pattern) and not a correctness bug — the end result is the same. However, it does create a short-lived unused thread pool from the base class. Consider removing the IVF-level initialization and relying solely on the base class setup, or removing the shadowing member from IVF.

LHT129 (commented):
## Review Summary

This PR fixes three thread-related issues in IVF routing HGraph. Overall the changes look correct and well-structured. The core fix — moving thread pool initialization before partition strategy construction and passing `modified_common_param` to child constructors — is the right approach.

### Issues Found

1. **[suggestion] `test_parallel_bg.sh` heartbeat is unrelated**: The CI heartbeat mechanism is a separate concern from the IVF thread fixes. Should be in its own PR.

2. **[note] Unused `#include <algorithm>` in `ivf_nearest_partition.cpp`**: Leftover from a previous iteration that used `std::max`. Should be removed.

3. **[note] Thread pool duplication between `InnerIndexInterface` and `IVF`**: The `IVF` class shadows `thread_pool_` from the base class, causing the base class constructor to create a thread pool that IVF never uses. This is pre-existing and not a correctness bug, but worth cleaning up.

### Positive Observations

- The `SetPoolSize` fix in `inner_index_interface.cpp` correctly moves the call inside the null-check, preventing modification of a shared thread pool.
- Passing `this->thread_pool_` to `KMeansCluster` enables proper thread reuse.
- The `modified_common_param` approach is clean and avoids modifying the original `common_param`.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: Updated PR description to reflect actual implementation.

LHT129 (commented):
Reviewed commit `657785a5`. This is a refinement over the previous approach — the three changes are clean and correct:

1. **inner_index_interface.cpp**: Merging the nested `if` conditions into a single `if (build_thread_count_ > 1 and thread_pool_ == nullptr)` is a minor cleanup that reduces nesting without changing behavior.

2. **ivf.cpp**: Thread pool initialization is now correctly placed before partition strategy construction, and `modified_common_param` with the initialized `thread_pool_` is passed to `IVFNearestPartition`, `GNOIMIPartition`, and `FlattenInterface`. This ensures all downstream components share the same thread pool instance.

3. **ivf_nearest_partition.cpp**: `KMeansCluster` now receives `this->thread_pool_` for thread reuse instead of creating its own default pool.

All prior review comments (int32_t → uint64_t type consistency, build_thread_count plumbing, SetPoolSize scoping) have been resolved in previous commits. No new issues introduced by this commit.

LHT129 (commented):
The changes in this commit look correct and address the stated issues:

1. **`inner_index_interface.cpp`**: Moving `SetPoolSize` inside the `thread_pool_ == nullptr` check is the right fix — it prevents accidentally resizing a shared thread pool that was passed in via `common_param`.

2. **`ivf.cpp`**: Initializing the thread pool before constructing the partition strategy and passing `modified_common_param` with the initialized `thread_pool_` ensures the routing HGraph and KMeans cluster reuse the IVF thread pool instead of creating their own default pools.

3. **`ivf_nearest_partition.cpp`**: Passing `this->thread_pool_` to `KMeansCluster` enables thread reuse during k-means training, consistent with the overall fix.

The existing review comments on pre-existing issues (thread pool member shadowing between IVF and InnerIndexInterface) have been noted in prior reviews and are not introduced by this PR.

LHT129 (commented):
The changes in commit c1ed5c1c look correct and clean. Here is a summary of the review:

**`src/algorithm/inner_index_interface.cpp`**: Moving `SetPoolSize` inside the `nullptr` check is the correct fix — it ensures `SetPoolSize` is only called on newly created default thread pools, not on shared ones passed via `common_param`.

**`src/algorithm/ivf/ivf.cpp`**: Moving thread pool initialization before partition strategy construction, and passing it via `modified_common_param`, correctly ensures that the routing HGraph (and other downstream components) reuse the IVF thread pool instead of creating their own default pool with 100 threads.

**`src/algorithm/ivf/ivf_nearest_partition.cpp`**: 
- Passing `this->thread_pool_` to `KMeansCluster` enables thread reuse for k-means training.
- The `count == 1` short-circuit in `ClassifyDatas` avoids unnecessary thread pool overhead for single-element classification.

No new issues introduced by this commit. The pre-existing concerns (IVF thread_pool_ member shadowing the base class, unused `#include <algorithm>`) have already been noted in prior review comments and are not blocking.

PR #2666, opened 2026-08-12
fix(ivf): expose routing HGraph parameters

### LINKED ISSUE #2659: [bug](ivf): routing HGraph reaches only ~30-45% of centroids at buckets_count=1024, causing severe bucket skew; the deciding knob (max_degree) is hard-coded and not configurable
### Affected area

ivf

### Interface

cpp

### VSAG version

75946f3f

### Summary

With partition_strategy_type=ivf, the coarse router is an HGraph over the
centroids. At buckets_count=1024, greedy search reaches only ~30-45% of
centroids from the single entry point, so a large fraction of buckets stay
permanently empty and the populated ones become severely skewed — which makes
buckets_count effectively meaningless and inflates P99 latency.

A parameter sweep indicates the routing graph's out-degree (max_degree) is the
sensitive factor, not the search width (ef). Both max_degree and ef_construction
of the routing graph are hard-coded (ivf_nearest_partition.cpp:164-165) and not
exposed to users, so this cannot be tuned without patching the source.

We are reporting the defect and the sensitivity — not prescribing a specific
value; the right default (likely a function of centroid count and/or data) is
for the maintainers to decide.

### Minimal reproducer

// Minimal reproducer: IVF centroid routing HGraph reaches only a fraction of
// the centroids, so most buckets are structurally unreachable.
//
// We ask route-only (disable_bucket_scan) for scan_buckets_count == buckets_count,
// i.e. "give me every bucket". route_buckets_only() fills id = -1 for slots the
// centroid-graph search could not reach, so counting non-negative ids per query
// directly measures reachability. Saturating far below buckets_count means the
// routing graph — not kmeans — causes the empty/oversized buckets.
//
// Build:
//   g++ -std=c++17 -O2 repro_issue2.cpp -o repro_issue2 \
//       -I<vsag>/include -L<vsag>/build/src -lvsag -Wl,-rpath,<vsag>/build/src -fopenmp -pthread
// Note: kmeans seeding uses std::random_device upstream (no public knob), but the
// defect reproduces on every run regardless of seed.
#include <vsag/vsag.h>

#include <cstdio>
#include <numeric>
#include <random>
#include <set>
#include <vector>

int main() {
    vsag::init();
    constexpr int64_t N = 200000, DIM = 768, BUCKETS = 1024, PROBE = 500;

    // clustered data: 1000 natural centers + gaussian noise
    std::vector<float> data(static_cast<size_t>(N) * DIM);
    std::vector<int64_t> ids(N);
    std::iota(ids.begin(), ids.end(), 0);
    std::mt19937 rng(1234);
    std::uniform_real_distribution<float> uni(-1.f, 1.f);
    std::normal_distribution<float> noise(0.f, 0.15f);
    std::vector<float> centers(1000 * DIM);
    for (auto& v : centers) v = uni(rng);
    std::uniform_int_distribution<int> pick(0, 999);
    for (int64_t i = 0; i < N; ++i) {
        const float* c = centers.data() + pick(rng) * DIM;
        for (int64_t d = 0; d < DIM; ++d) data[i * DIM + d] = c[d] + noise(rng);
    }

    auto base = vsag::Dataset::Make();
    base->NumElements(N)->Dim(DIM)->Ids(ids.data())->Float32Vectors(data.data())->Owner(false);

    auto index = vsag::Factory::CreateIndex("ivf", R"({
        "dtype":"float32","metric_type":"l2","dim":768,
        "index_param":{"buckets_count":1024,"base_quantization_type":"sq8",
                       "partition_strategy_type":"ivf","thread_count":16}})").value();
    index->Build(base);

    // route-only probe: ask for ALL 1024 buckets, count reachable ones
    auto q = vsag::Dataset::Make();
    q->NumElements(PROBE)->Dim(DIM)->Float32Vectors(data.data())->Owner(false);
    vsag::SearchRequest req;
    req.query_ = q;
    req.params_str_ = R"({"ivf":{"scan_buckets_count":1024,"disable_bucket_scan":true}})";
    auto r = index->SearchWithRequest(req);
    if (!r.has_value()) { fprintf(stderr, "route failed: %s\n", r.error().message.c_str()); return 1; }

    const int64_t* rid = r.value()->GetIds();
    int64_t sum = 0;
    std::set<int64_t> reachable;
    for (int64_t i = 0; i < PROBE; ++i)
        for (int64_t b = 0; b < BUCKETS; ++b) {
            int64_t v = rid[i * BUCKETS + b];
            if (v >= 0) { ++sum; reachable.insert(v); }
        }
    printf("asked 1024 buckets/query | avg reachable/query = %.1f | union over %ld queries = %zu / 1024\n",
           double(sum) / PROBE, (long)PROBE, reachable.size());
    printf("=> saturates far below 1024 => centroid routing graph cannot reach most centroids.\n");
    return 0;
}

### Environment

```shell
- OS: Linux (glibc), x86_64 
- vsag version: main @ 75946f3f
- compiler version: c++ (GCC) 9.2.1 20240328
```

### Expected behavior

Requesting all buckets from a 1024-bucket index should reach close to 1024
distinct buckets, and the bucket-size distribution should be reasonably
balanced (comparable to exhaustive nearest-centroid assignment).

### Actual behavior

routing config              reachable/query   non-empty buckets   CV      max bucket   build
ef=10  M=64 (upstream)            450                431          19.72    123,292      163s
ef=32  M=64                       436                439          15.90     99,050      213s
ef=128 M=64                       291                299          14.28     68,101      236s
ef=512 M=64                       320                328          16.86    104,280      318s
ef=10  M=128                      571                569          10.71     60,976      213s
ef=10  M=256                     1024               1019           0.396     1,084      273s
ef=10  M=512 efc=300              rejected by parameter validation
ef=10  M=512 efc=1024             919                915           3.065     18,949      321s
exhaustive (ground truth)        1024               1011           0.439     1,008      255s

Observations from this specific setup (1024 centroids, 768-dim uniform-random
synthetic data):
1. Search width ef is NOT the lever: at the default M=64, sweeping ef 10->512
   keeps reachable at ~291-450 and CV at 14-20.
2. Routing max_degree is highly sensitive: increasing it lifts reachability and
   reduces skew. In this setup M=256 happened to recover near-exhaustive balance
   (reachable 1024, CV 0.396 vs exhaustive 0.439) at +110s build.

Caveat on the numbers: 256 is only what worked for THIS configuration — it is
not a recommended constant. The relationship is not even monotonic here
(M=512/efc=1024 was worse, CV 3.065; M=512/efc=300 was rejected by validation),
which further suggests the right value is workload-dependent.

### References

- src/algorithm/ivf/ivf_nearest_partition.cpp:164-165  (max_degree=64, ef_construction=300, both hard-coded)
- src/algorithm/ivf/ivf_nearest_partition.cpp:34-40, 105-110  (routing ef_search = max(10, scan_buckets_count*1.2); no independent knob)
- src/impl/pruning_strategy.cpp:102-122  (heuristic edge selection, alpha=1.0; may aggressively cap effective out-degree under high-dim distance concentration)
- Introduced in #2494 

### Additional context

Rather than a fixed new default, the actionable ask is: expose the routing
graph's max_degree (and ef_construction) as configurable parameters, so callers
can tune reachability vs. build cost for their data. Whether to additionally
auto-scale the default with buckets_count (or otherwise) is a design decision we
defer to the maintainers. Today these are hard-coded, so varying them for the
sweep above required patching the source.

Why this is graph reachability, not a search-param issue: when requesting all
buckets, routing ef = max(10, 1024*1.2) = 1228 > 1024 nodes, so the early-stop in
basic_searcher.cpp:221-233 never fires and search degenerates into a full BFS —
hence "returned bucket count == nodes reachable via out-edges".

Caveat: the data here is synthetic (768-dim uniform-random centers), which may
be adversarial for the routing graph; real embeddings with lower intrinsic
dimension may behave better. Worth re-verifying on real data.

LHT129:
## Summary
- expose IVF routing HGraph maximum degree and construction breadth as partition strategy parameters
- preserve the existing 64/300 defaults and validate positive values
- document the parameters and add serialization, compatibility, and validation coverage

## Testing
- `make fmt`
- `git diff --check`
- `make debug EXTRA_DEFINED="-DFETCHCONTENT_FULLY_DISCONNECTED=ON"` *(blocked: the configured Boost mirror returns HTTP 403; no cached Boost source is available on the remote host)*

Fixes: #2659

### DIFF
--- a/src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
+++ b/src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
@@ -42,13 +42,17 @@ IVFPartitionStrategyParameters::FromJson(const JsonType& json) {
         this->partition_strategy_type = IVFPartitionStrategyType::GNO_IMI;
     }
     if (json.Contains(IVF_ROUTE_MAX_DEGREE_KEY)) {
-        this->route_max_degree = json[IVF_ROUTE_MAX_DEGREE_KEY].GetInt();
+        this->route_max_degree = static_cast<int32_t>(json[IVF_ROUTE_MAX_DEGREE_KEY].GetInt());
         CHECK_ARGUMENT(this->route_max_degree > 0, "route_max_degree must be positive");
     }
     if (json.Contains(IVF_ROUTE_EF_CONSTRUCTION_KEY)) {
-        this->route_ef_construction = json[IVF_ROUTE_EF_CONSTRUCTION_KEY].GetInt();
+        this->route_ef_construction =
+            static_cast<int32_t>(json[IVF_ROUTE_EF_CONSTRUCTION_KEY].GetInt());
         CHECK_ARGUMENT(this->route_ef_construction > 0, "route_ef_construction must be positive");
     }
+    CHECK_ARGUMENT(this->route_max_degree >= 4, "route_max_degree must be at least 4");
+    CHECK_ARGUMENT(this->route_ef_construction >= this->route_max_degree,
+                   "route_ef_construction must be no less than route_max_degree");
 
     this->gnoimi_param = std::make_shared<GNOIMIParameter>();
     if (this->partition_strategy_type == IVFPartitionStrategyType::GNO_IMI) {
--- a/src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
+++ b/src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
@@ -41,6 +41,14 @@ IVFPartitionStrategyParameters::FromJson(const JsonType& json) {
                IVF_PARTITION_STRATEGY_TYPE_GNO_IMI) {
         this->partition_strategy_type = IVFPartitionStrategyType::GNO_IMI;
     }
+    if (json.Contains(IVF_ROUTE_MAX_DEGREE_KEY)) {
+        this->route_max_degree = json[IVF_ROUTE_MAX_DEGREE_KEY].GetInt();
+        CHECK_ARGUMENT(this->route_max_degree > 0, "route_max_degree must be positive");
+    }
+    if (json.Contains(IVF_ROUTE_EF_CONSTRUCTION_KEY)) {
+        this->route_ef_construction = json[IVF_ROUTE_EF_CONSTRUCTION_KEY].GetInt();
+        CHECK_ARGUMENT(this->route_ef_construction > 0, "route_ef_construction must be positive");
+    }
 
     this->gnoimi_param = std::make_shared<GNOIMIParameter>();
     if (this->partition_strategy_type == IVFPartitionStrategyType::GNO_IMI) {
@@ -67,6 +75,8 @@ IVFPartitionStrategyParameters::ToJson() const {
     } else if (this->partition_strategy_type == IVFPartitionStrategyType::GNO_IMI) {
         json[IVF_PARTITION_STRATEGY_TYPE_KEY].SetString(IVF_PARTITION_STRATEGY_TYPE_GNO_IMI);
     }
+    json[IVF_ROUTE_MAX_DEGREE_KEY].SetInt(this->route_max_degree);
+    json[IVF_ROUTE_EF_CONSTRUCTION_KEY].SetInt(this->route_ef_construction);
     if (this->partition_strategy_type == IVFPartitionStrategyType::GNO_IMI) {
         json[IVF_PARTITION_STRATEGY_TYPE_GNO_IMI].SetJson(this->gnoimi_param->ToJson());
     }
@@ -77,6 +87,8 @@ bool
 IVFPartitionStrategyParameters::CheckCompatibility(const ParamPtr& other) const {
     PARAM_CAST_OR_RETURN(IVFPartitionStrategyParameters, p, other);
     CHECK_FIELD_EQ(*this, *p, partition_strategy_type);
+    CHECK_FIELD_EQ(*this, *p, route_max_degree);
+    CHECK_FIELD_EQ(*this, *p, route_ef_construction);
     CHECK_SUB_PARAM(*this, *p, gnoimi_param);
     return true;
 }
--- a/src/algorithm/ivf/ivf_partition_strategy_parameter.h
+++ b/src/algorithm/ivf/ivf_partition_strategy_parameter.h
@@ -51,6 +51,8 @@ public:
     IVFNearestPartitionTrainerType partition_train_type{
         IVFNearestPartitionTrainerType::KMeansTrainer};
     IVFPartitionStrategyType partition_strategy_type{IVFPartitionStrategyType::IVF};
+    int32_t route_max_degree{64};
+    int32_t route_ef_construction{300};
     GNOIMIParameterPtr gnoimi_param{nullptr};
 };
 
--- a/src/algorithm/ivf/ivf_partition_strategy_parameter_test.cpp
+++ b/src/algorithm/ivf/ivf_partition_strategy_parameter_test.cpp
@@ -87,4 +87,17 @@ TEST_CASE("IVF Partition Strategy Routing Parameters", "[ut][IVFPartitionStrateg
         "ivf_train_type": "kmeans",
         "route_ef_construction": 0
     })"));
+    REQUIRE_THROWS(param->FromString(R"(
+    {
+        "partition_strategy_type": "ivf",
+        "ivf_train_type": "kmeans",
+        "route_max_degree": 3
+    })"));
+    REQUIRE_THROWS(param->FromString(R"(
+    {
+        "partition_strategy_type": "ivf",
+        "ivf_train_type": "kmeans",
+        "route_max_degree": 512,
+        "route_ef_construction": 300
+    })"));
 }
--- a/src/algorithm/ivf/ivf_partition_strategy_parameter_test.cpp
+++ b/src/algorithm/ivf/ivf_partition_strategy_parameter_test.cpp
@@ -33,6 +33,8 @@ TEST_CASE("IVF Partition Strategy Parameters Test", "[ut][IVFPartitionStrategyPa
     REQUIRE(param->partition_train_type == vsag::IVFNearestPartitionTrainerType::RandomTrainer);
     REQUIRE(param->gnoimi_param->first_order_buckets_count == 200);
     REQUIRE(param->gnoimi_param->second_order_buckets_count == 50);
+    REQUIRE(param->route_max_degree == 64);
+    REQUIRE(param->route_ef_construction == 300);
 
     vsag::ParameterTest::TestToJson(param);
 }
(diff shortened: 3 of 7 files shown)


src/algorithm/ivf/ivf_partition_strategy_parameter.h
```diff
@@ -51,6 +51,8 @@ class IVFPartitionStrategyParameters : public Parameter {
     IVFNearestPartitionTrainerType partition_train_type{
         IVFNearestPartitionTrainerType::KMeansTrainer};
     IVFPartitionStrategyType partition_strategy_type{IVFPartitionStrategyType::IVF};
+    int32_t route_max_degree{64};
```
LHT129, line 54: [note] The newly added `route_max_degree` and `route_ef_construction` are declared as `int32_t`, but the downstream HGraph parameters they feed into use `uint64_t` (`GraphInterfaceParameter::max_degree_` and `HGraphParameter::ef_construction`). While the implicit conversion through RapidJSON `SetInt` is safe for positive values, using `uint64_t` here would be more consistent with the types they ultimately map to and would avoid a potential narrowing warning if the wiring ever changes to a direct assignment.

src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
```diff
@@ -41,6 +41,14 @@ IVFPartitionStrategyParameters::FromJson(const JsonType& json) {
...
         this->partition_strategy_type = IVFPartitionStrategyType::GNO_IMI;
     }
+    if (json.Contains(IVF_ROUTE_MAX_DEGREE_KEY)) {
+        this->route_max_degree = json[IVF_ROUTE_MAX_DEGREE_KEY].GetInt();
+        CHECK_ARGUMENT(this->route_max_degree > 0, "route_max_degree must be positive");
+    }
+    if (json.Contains(IVF_ROUTE_EF_CONSTRUCTION_KEY)) {
+        this->route_ef_construction = json[IVF_ROUTE_EF_CONSTRUCTION_KEY].GetInt();
+        CHECK_ARGUMENT(this->route_ef_construction > 0, "route_ef_construction must be positive");
+    }
```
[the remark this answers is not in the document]
LHT129: Fixed in bda1b782. Added cross-field validation after parsing: route_max_degree >= 4 and route_ef_construction >= route_max_degree, with route-specific error messages.

src/algorithm/ivf/ivf_partition_strategy_parameter_test.cpp
```diff
@@ -54,3 +56,35 @@ TEST_CASE("IVF Partition Strategy Parameters CheckCompatibility",
...
+        "partition_strategy_type": "ivf",
+        "ivf_train_type": "kmeans",
+        "route_max_degree": 0
+    })"));
+    REQUIRE_THROWS(param->FromString(R"(
+    {
+        "partition_strategy_type": "ivf",
+        "ivf_train_type": "kmeans",
+        "route_ef_construction": 0
+    })"));
```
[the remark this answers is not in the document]
LHT129: Fixed in bda1b782. Added regression tests for route_max_degree < 4 and route_max_degree > route_ef_construction (e.g. 512/300).

src/algorithm/ivf/ivf_partition_strategy_parameter.h
LHT129: Keeping int32_t intentionally: RapidJSON GetInt() returns int32_t, and parsing validation already prevents unsafe narrowing. The downstream uint64_t conversion through SetInt is safe for the validated positive range.

LHT129 (commented):
The updates in bda1b782 address the previous review feedback well:

- Added cross-field validation (`route_max_degree >= 4`, `route_ef_construction >= route_max_degree`) with route-specific error messages in `FromJson`.
- Added regression tests covering all boundary conditions: zero values, `route_max_degree < 4`, and `route_max_degree > route_ef_construction`.
- `CheckCompatibility` correctly includes the two new fields.
- `ToJson` / `FromJson` round-trip is covered by the existing `TestToJson` call.

No further issues found. LGTM.

LHT129 (commented):
The changes in bda1b782 look good. All previous review feedback has been addressed:

- Cross-field validation (`route_max_degree >= 4`, `route_ef_construction >= route_max_degree`) with route-specific error messages is now in place.
- Regression tests cover the invalid-but-positive cases (max_degree < 4, ef_construction < max_degree).
- The `int32_t` type choice is reasonable given the RapidJSON `GetInt()` return type and the validated positive range.

No new issues found. The implementation is clean, well-tested, and properly documented in both English and Chinese docs.

LHT129 (commented):
The changes in 67978b92 correctly address the previous review feedback:

- `GetInt()` → `GetInt32()` fixes the type mismatch for `int32_t` fields.
- Cross-field validation (`route_max_degree >= 4`, `route_ef_construction >= route_max_degree`) with route-specific error messages is properly placed after the optional parsing blocks, ensuring defaults are also validated.
- Regression tests cover all boundary conditions: `route_max_degree < 4` and `route_max_degree > route_ef_construction`.

No new issues found. The implementation is clean and well-tested.

LHT129 (commented):
Reviewed the latest commit 67978b92 which adds cross-field validation for routing HGraph parameters.

**Summary of changes (full PR):**
- Exposes `route_max_degree` (default 64) and `route_ef_construction` (default 300) as IVF partition strategy parameters
- Replaces hardcoded values in `ivf_nearest_partition.cpp` with the configurable parameters
- Adds validation: positive values, `route_max_degree >= 4`, `route_ef_construction >= route_max_degree`
- Includes serialization (`ToJson`) and compatibility checks (`CheckCompatibility`)
- Comprehensive unit tests covering valid values, edge cases, and validation failures
- Documentation updated in both EN and ZH

**Previous review feedback addressed:**
- Cross-field validation added (route_max_degree >= 4, route_ef_construction >= route_max_degree)
- Regression tests added for invalid-but-positive cases (max_degree=3, max_degree > ef_construction)

No blocking issues found. The implementation is clean, well-tested, and consistent with the existing codebase patterns.

src/algorithm/ivf/ivf_partition_strategy_parameter.cpp
```diff
@@ -41,6 +41,18 @@ IVFPartitionStrategyParameters::FromJson(const JsonType& json) {
                IVF_PARTITION_STRATEGY_TYPE_GNO_IMI) {
         this->partition_strategy_type = IVFPartitionStrategyType::GNO_IMI;
     }
+    if (json.Contains(IVF_ROUTE_MAX_DEGREE_KEY)) {
+        this->route_max_degree = json[IVF_ROUTE_MAX_DEGREE_KEY].static_cast<int32_t>(GetInt());
```
LHT129: [critical] The `static_cast` placement is syntactically invalid and will not compile.

`static_cast` is a C++ keyword/operator, not a member function. The current code:
```cpp
json[IVF_ROUTE_MAX_DEGREE_KEY].static_cast<int32_t>(GetInt());
```
attempts to call `static_cast` as a member of the JSON value, which is not valid C++.

The correct syntax, consistent with the existing pattern in `gno_imi_parameter.cpp:33`:
```cpp
static_cast<int32_t>(json[IVF_ROUTE_MAX_DEGREE_KEY].GetInt());
```

This applies to both `route_max_degree` (line 45) and `route_ef_construction` (line 49-50).

LHT129: Fixed in df7e990e. Corrected static_cast syntax.

LHT129 (commented):
Reviewed commit df7e990e. The changes look good:

- The `static_cast` syntax issue from the previous commit (97274e2) has been corrected.
- Cross-field validation (`route_max_degree >= 4`, `route_ef_construction >= route_max_degree`) is now in place with route-specific error messages.
- Regression tests cover all the boundary cases: zero values, max_degree < 4, and ef_construction < max_degree.
- The `CheckCompatibility` method correctly includes the two new fields.
- Serialization (`ToJson`) and deserialization (`FromJson`) are symmetric.
- Documentation (both EN and ZH) is updated with the new parameters.
- The `DEFAULT_MAP` entries are added for the new keys.

No new issues found. The earlier review comments from Copilot (cross-field validation, regression tests) have been addressed in this commit.

LHT129 (commented):
Reviewed commit df7e990e. The previously reported `static_cast` syntax issue has been fixed — the code now correctly uses `static_cast<int32_t>(json[...].GetInt())`. Validation logic (positive check, >= 4, and cross-field `route_ef_construction >= route_max_degree`) is complete with route-specific error messages. Test coverage includes normal values, zero, below-minimum, and cross-field violation cases. Documentation in both English and Chinese is updated. No new issues found.

wxyucs (approved):
lgtm

PR #2668, opened 2026-08-12
feat(pyramid): support path-aware AnalyzeIndexBySearch

### LINKED ISSUE #2661: [feat](other): support path-aware AnalyzeIndexBySearch for Pyramid
### Affected area

other

### Motivation

`Index::AnalyzeIndexBySearch` is a public API, but Pyramid does not override it and currently falls back to `UNSUPPORTED_INDEX_OPERATION`.

A `PyramidAnalyzer::AnalyzeIndexBySearch` implementation already exists, but it is not ready to expose directly:

- Query vectors are copied without their default or named-hierarchy paths.
- Per-query datasets are reconstructed with vectors only.
- Ground truth is calculated against every vector in the index instead of the vectors reachable through the query's hierarchy and path.
- Pyramid configurations whose root is `NO_INDEX` require a query path, so simply adding an override would either fail or produce incorrect recall metrics.

Pyramid users need the same query-driven recall, distance, latency, and quantization diagnostics available for HGraph, IVF, and SINDI, while preserving Pyramid's path-scoped search semantics.

### Proposed solution

Implement path-aware `AnalyzeIndexBySearch` support for Pyramid:

1. Add `Pyramid::AnalyzeIndexBySearch(const SearchRequest&)` and delegate to `PyramidAnalyzer`.
2. Preserve the query dataset's default paths and named-hierarchy paths when processing batched queries.
3. Select the hierarchy using the same search parameters as `Pyramid::KnnSearch`.
4. Calculate ground truth only over vectors eligible for the selected hierarchy and query path, including normal removal semantics.
5. Keep the existing query-analysis metrics:
   - `avg_distance_query`
   - `recall_query`
   - `time_cost_query`
   - quantization error and inversion ratio when reorder is enabled
6. Add regression coverage for:
   - single-hierarchy Pyramid with a `NO_INDEX` root;
   - path-specific ground-truth correctness;
   - named multi-hierarchy queries;
   - batched queries with different paths.
7. Update the analysis documentation to distinguish API support from any remaining `analyze_index` CLI path-input limitations.

### Alternatives considered

Adding only the missing Pyramid override was considered, but the current analyzer drops query paths and computes global ground truth, so this would expose incorrect behavior.

Restricting analysis to configurations with a built root was also considered, but it would exclude common Pyramid configurations and would not solve path-scoped recall correctness.

Keeping query-driven analysis unsupported and relying only on `GetStats()` does not provide workload-specific recall, latency, or quantization metrics.

none

### References

- `include/vsag/index.h:1049`
- `src/index/index_impl.h:89`
- `src/algorithm/inner_index_interface.h:83`
- `src/algorithm/pyramid/pyramid.h:241`
- `src/analyzer/analyzer.cpp:35`
- `src/analyzer/pyramid_analyzer.cpp:87`
- `src/analyzer/pyramid_analyzer.cpp:93`
- `src/analyzer/pyramid_analyzer.cpp:500`
- `src/analyzer/pyramid_analyzer.cpp:658`
- `src/algorithm/pyramid/pyramid.cpp:438`
- `src/algorithm/pyramid/pyramid.cpp:450`
- `include/vsag/dataset.h:327`
- `docs/docs/en/src/resources/analyze_index.md:18`
- `docs/docs/en/src/indexes/pyramid.md:304`

### Additional context

Issue #1765 introduced `PyramidAnalyzer` and described query-based analysis, but the public Pyramid index was never connected to that method.

Issue #2102 defined named-hierarchy analyzer semantics for static statistics. This issue focuses specifically on completing the missing path-aware query-analysis flow.

The existing `analyze_index` dense query file loader does not carry Pyramid hierarchy paths, so CLI support may require an additional path-input format even after the C++ API is enabled.

## Related issues

- [#1765 feat(pyramid): add PyramidAnalyzer for index statistics and diagnostics](https://github.com/antgroup/vsag/issues/1765)
- [#2102 [feat](multiPyramid): support named hierarchy stats for pyramid analyzer](https://github.com/antgroup/vsag/issues/2102)

---

jac0626:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2661

## What Changed

- Expose `Pyramid::AnalyzeIndexBySearch` for supported KNN requests and query batches.
- Preserve each query's default or named-hierarchy path and calculate ground truth only from live vectors reachable in the same Pyramid search scope.
- Follow Pyramid node semantics when collecting candidates: stop at `FLAT` or `GRAPH`, recurse through `NO_INDEX`, union overlapping path scopes, expand graph duplicate groups, and exclude mark-removed IDs.
- Use the most precise vectors or codes retained by the index for ground truth, and report recall, average distance, latency, and finite reorder quantization metrics.
- Add unit and functional coverage for path scoping, removals, graph duplicates, named hierarchies, batched queries, validation, raw-vector ground truth, and `topk == 1` quantization metrics.
- Update the English and Chinese analysis documentation, including the current CLI hierarchy-path limitation.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
make debug
# passed

clang-format 15 --dry-run --Werror <changed C++ files>
git diff --check
# passed

./build/tests/unittests '[ut][pyramid][analyzer]' -r compact
# All tests passed (18 assertions in 5 test cases)

./build/tests/functests '[ft][pyramid][analyzer]' -r compact
# All tests passed (60 assertions in 6 test cases)
```

## Compatibility Impact

- API/ABI compatibility: none; this implements an existing virtual API for Pyramid.
- Behavior changes: `AnalyzeIndexBySearch` now works for supported Pyramid KNN requests and uses path-scoped ground truth. Unsupported range, filtered, and iterator analysis requests are rejected explicitly.

- Performance impact: ground-truth cost is proportional to each query path's live candidate scope rather than the full index.
- Concurrency/thread-safety impact: analysis is read-only; the index must remain stable without concurrent `Add` or `Remove` operations during analysis.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/{indexes/pyramid.md,resources/analyze_index.md}`

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert this PR to restore the previous unsupported-operation behavior.

## Checklist

- [x] I have updated docs because behavior changed
- [x] My commit message follows project conventions

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -19,6 +19,7 @@
 
 #include <chrono>
 #include <exception>
+#include <limits>
 
 #include "algorithm/inner_index_interface.h"
 #include "analyzer/analyzer.h"
@@ -1703,6 +1704,25 @@ Pyramid::GetStats() const {
     auto analyzer = CreateAnalyzer(this, analyzer_param);
     JsonType stats = analyzer->GetStats();
     return stats.Dump(4);
+}
+
+std::string
+Pyramid::AnalyzeIndexBySearch(const SearchRequest& request) {
+    CHECK_ARGUMENT(request.mode_ == SearchMode::KNN_SEARCH,
+                   "Pyramid AnalyzeIndexBySearch only supports KNN search");
+    CHECK_ARGUMENT(not request.enable_filter_ && not request.enable_bitset_filter_ &&
+                       not request.enable_attribute_filter_ && not request.enable_iterator_search_,
+                   "Pyramid AnalyzeIndexBySearch does not support filtered or iterator search");
+    CHECK_ARGUMENT(request.topk_ > 0,
+                   fmt::format("topk({}) must be greater than 0", request.topk_));
+    CHECK_ARGUMENT(request.topk_ <= static_cast<int64_t>(std::numeric_limits<uint32_t>::max()),
+                   fmt::format("topk({}) exceeds the supported maximum", request.topk_));
+    CHECK_ARGUMENT(base_codes_->TotalCount() > 0,
+                   "Pyramid AnalyzeIndexBySearch requires a built index");
+    AnalyzerParam analyzer_param(allocator_);
+    analyzer_param.topk = request.topk_;
+    auto analyzer = CreateAnalyzer(this, analyzer_param);
+    return analyzer->AnalyzeIndexBySearch(request).Dump(4);
 }
 
 }  // namespace vsag
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -1710,8 +1710,10 @@ std::string
 Pyramid::AnalyzeIndexBySearch(const SearchRequest& request) {
     CHECK_ARGUMENT(request.mode_ == SearchMode::KNN_SEARCH,
                    "Pyramid AnalyzeIndexBySearch only supports KNN search");
-    CHECK_ARGUMENT(not request.enable_filter_ && not request.enable_bitset_filter_ &&
-                       not request.enable_attribute_filter_ && not request.enable_iterator_search_,
+    const bool is_supported_search =
+        not request.enable_filter_ && not request.enable_bitset_filter_ &&
+        not request.enable_attribute_filter_ && not request.enable_iterator_search_;
+    CHECK_ARGUMENT(is_supported_search,
                    "Pyramid AnalyzeIndexBySearch does not support filtered or iterator search");
     CHECK_ARGUMENT(request.topk_ > 0,
                    fmt::format("topk({}) must be greater than 0", request.topk_));
--- a/src/analyzer/pyramid_analyzer.cpp
+++ b/src/analyzer/pyramid_analyzer.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <limits>
 #include <random>
 
 #include "impl/heap/standard_heap.h"
@@ -99,36 +100,80 @@ PyramidAnalyzer::AnalyzeIndexBySearch(const SearchRequest& request) {
     }
 
     auto query = request.query_;
-    auto query_num = query->GetNumElements();
+    const auto query_num = query->GetNumElements();
     const float* query_vectors = query->GetFloat32Vectors();
-    if (query_vectors == nullptr || query_num == 0) {
+    if (query_num == 0) {
         return stats;
     }
+    CHECK_ARGUMENT(query_num > 0, fmt::format("query num({}) must be positive", query_num));
+    CHECK_ARGUMENT(static_cast<uint64_t>(query_num) <= std::numeric_limits<uint32_t>::max(),
+                   fmt::format("query num({}) exceeds the supported maximum", query_num));
+    CHECK_ARGUMENT(
+        query->GetDim() == dim_,
+        fmt::format("query.dim({}) must be equal to index.dim({})", query->GetDim(), dim_));
+    CHECK_ARGUMENT(query_vectors != nullptr, "query vectors is required");
+    const auto query_count = static_cast<uint32_t>(query_num);
+
+    auto parsed_param = PyramidSearchParameters::FromJson(request.params_str_);
+    CHECK_ARGUMENT(parsed_param.hierarchy_op == PyramidSearchParameters::HierarchyOp::SINGLE,
+                   "multi-hierarchy search (union/intersection) is not yet implemented");
+    const std::string hierarchy_name =
+        parsed_param.hierarchies.empty() ? "" : parsed_param.hierarchies[0];
+    auto h_iter = pyramid_->hierarchies_.find(hierarchy_name);
+    CHECK_ARGUMENT(h_iter != pyramid_->hierarchies_.end(),
+                   fmt::format("unknown hierarchy name: '{}'", hierarchy_name));
+
+    const auto* query_paths = query->GetPaths(hierarchy_name);
+    if (query_paths == nullptr) {
+        query_paths = query->GetPaths();
+    }
+    // NOLINTNEXTLINE(readability-simplify-boolean-expr)
+    CHECK_ARGUMENT(
+        query_paths != nullptr || h_iter->second->root->status_ != IndexNode::Status::NO_INDEX,
+        "query_path is required when level0 is not built");
 
     Vector<InnerIdType> query_ids(allocator_);
-    query_ids.resize(query_num);
+    query_ids.resize(query_count);
     std::iota(query_ids.begin(), query_ids.end(), 0);
 
     Vector<float> query_datas(allocator_);
-    query_datas.resize(static_cast<size_t>(query_num) * dim_);
-    std::memcpy(query_datas.data(), query_vectors, query_num * dim_ * sizeof(float));
+    query_datas.resize(static_cast<uint64_t>(query_count) * dim_);
+    std::memcpy(query_datas.data(),
+                query_vectors,
+                static_cast<uint64_t>(query_count) * dim_ * sizeof(float));
+
+    UnorderedSet<InnerIdType> deleted_ids(allocator_);
+    for (const auto id : pyramid_->label_table_->GetAllDeletedIds()) {
+        deleted_ids.insert(id);
+    }
 
     UnorderedMap<InnerIdType, DistHeapPtr> query_ground_truth(allocator_);
-    calculate_groundtruth(query_datas, query_ids, query_ground_truth, query_num);
+    calculate_groundtruth(query_datas,
+                          query_ids,
+                          query_ground_truth,
+                          query_count,
+                          hierarchy_name,
+                          query_paths,
+                          deleted_ids);
 
     UnorderedMap<InnerIdType, Vector<LabelType>> query_search_result(allocator_);
-    float query_time_ms = calculate_search_result(
-        query_datas, query_ids, query_search_result, request.params_str_, query_num);
+    float query_time_ms = calculate_search_result(query_datas,
+                                                  query_ids,
+                                                  query_search_result,
+                                                  request.params_str_,
+                                                  query_count,
+                                                  hierarchy_name,
+                                                  query_paths);
 
     stats["avg_distance_query"].SetFloat(
         get_avg_distance_from_groundtruth(query_ids, query_ground_truth));
-    stats["recall_query"].SetFloat(
-        get_search_recall(query_num, query_ids, query_ground_truth, query_search_result));
+    stats["recall_query"].SetFloat(get_search_recall(
+        query_count, query_datas, query_ids, query_ground_truth, query_search_result));
     stats["time_cost_query"].SetFloat(query_time_ms);
 
     if (pyramid_->use_reorder_) {
         auto [q_error, q_inversion] =
-            calculate_quantization_result(query_datas, query_ids, query_search_result, query_num);
+            calculate_quantization_result(query_datas, query_ids, query_search_result, query_count);
         stats["quantization_error_query"].SetFloat(q_error);
         stats["quantization_inversion_ratio_query"].SetFloat(q_inversion);
     }
@@ -514,32 +559,38 @@ void
 PyramidAnalyzer::calculate_groundtruth(const Vector<float>& sample_datas,
                                        const Vector<InnerIdType>& sample_ids,
                                        UnorderedMap<InnerIdType, DistHeapPtr>& ground_truth,
-                                       uint32_t sample_size) {
+                                       uint32_t sample_size,
+                                       const std::string& hierarchy_name,
+                                       const std::string* query_paths,
+                                       const UnorderedSet<InnerIdType>& deleted_ids) {
     if (not ground_truth.empty()) {
         return;
     }
 
-    Vector<float> distances_array(this->total_count_, allocator_);
-    Vector<InnerIdType> ids_array(this->total_count_, allocator_);
-    std::iota(ids_array.begin(), ids_array.end(), 0);
-
-    auto codes = pyramid_->has_precise_reorder() ? pyramid_->precise_codes_ : pyramid_->base_codes_;
+    auto codes = pyramid_->raw_vector_ != nullptr
+                     ? pyramid_->raw_vector_
+                     : (pyramid_->use_reorder_ ? pyramid_->precise_codes_ : pyramid_->base_codes_);
 
     for (uint32_t i = 0; i < sample_size; ++i) {
         if (i % 10 == 0) {
             logger::info("[calculate_groundtruth] Processing sample {} of {}", i, sample_size);
         }
 
+        const auto* query_path = query_paths == nullptr ? nullptr : query_paths + i;
+        auto ids_array = collect_search_scope_ids(hierarchy_name, query_path, deleted_ids);
+        Vector<float> distances_array(ids_array.size(), allocator_);
         auto comp = codes->FactoryComputer(sample_datas.data() + static_cast<size_t>(i) * dim_);
-        codes->Query(distances_array.data(), comp, ids_array.data(), this->total_count_);
+        if (not ids_array.empty()) {
+            codes->Query(distances_array.data(), comp, ids_array.data(), ids_array.size());
+        }
 
         DistHeapPtr gt = std::make_shared<StandardHeap<true, false>>(allocator_, -1);
-        for (uint32_t j = 0; j < this->total_count_; ++j) {
+        for (uint64_t j = 0; j < ids_array.size(); ++j) {
             float dist = distances_array[j];
             if (gt->Size() < topk_) {
-                gt->Push({dist, j});
+                gt->Push({dist, ids_array[j]});
             } else if (dist < gt->Top().first) {
-                gt->Push({dist, j});
+                gt->Push({dist, ids_array[j]});
                 gt->Pop();
             }
         }
@@ -547,6 +598,76 @@ PyramidAnalyzer::calculate_groundtruth(const Vector<float>& sample_datas,
     }
 
     logger::info("[calculate_groundtruth] Completed for {} samples", sample_size);
+}
+
+Vector<InnerIdType>
+PyramidAnalyzer::collect_search_scope_ids(const std::string& hierarchy_name,
+                                          const std::string* query_path,
+                                          const UnorderedSet<InnerIdType>& deleted_ids) {
+    Vector<InnerIdType> ids(allocator_);
+    UnorderedSet<InnerIdType> seen_ids(allocator_);
+    const auto& hierarchy = *pyramid_->hierarchies_.at(hierarchy_name);
+
+    if (query_path == nullptr) {
+        collect_searchable_node_ids(hierarchy.root.get(), deleted_ids, seen_ids, ids);
+        return ids;
+    }
+
+    for (const auto& one_path : Pyramid::parse_path(*query_path)) {
+        const IndexNode* node = hierarchy.root.get();
+        for (const auto& item : one_path) {
+            std::shared_lock lock(node->mutex_);
+            auto child = node->children_.find(item);
+            if (child == node->children_.end()) {
+                node = nullptr;
+                break;
+            }
+            node = child->second.get();
+        }
+        if (node != nullptr) {
+            collect_searchable_node_ids(node, deleted_ids, seen_ids, ids);
+        }
+    }
+    return ids;
+}
+
+void
+PyramidAnalyzer::collect_searchable_node_ids(const IndexNode* node,
+                                             const UnorderedSet<InnerIdType>& deleted_ids,
+                                             UnorderedSet<InnerIdType>& seen_ids,
+                                             Vector<InnerIdType>& ids) {
+    if (node == nullptr) {
+        return;
+    }
+
+    std::shared_lock lock(node->mutex_);
+    if (node->status_ != IndexNode::Status::NO_INDEX) {
+        Vector<InnerIdType> node_ids(allocator_);
+        if (node->status_ == IndexNode::Status::FLAT) {
+            node_ids = node->ids_;
+        } else if (node->graph_ != nullptr) {
+            node_ids = node->graph_->GetIds();
+        }
+        for (const auto id : node_ids) {
+            if (deleted_ids.find(id) == deleted_ids.end() && seen_ids.insert(id).second) {
+                ids.push_back(id);
+            }
+            if (node->graph_ == nullptr) {
+                continue;
+            }
+            for (const auto duplicate_id : node->graph_->GetDuplicateIds(id)) {
+                if (deleted_ids.find(duplicate_id) == deleted_ids.end() &&
+                    seen_ids.insert(duplicate_id).second) {
+                    ids.push_back(duplicate_id);
+                }
+            }
+        }
+        return;
+    }
+
+    for (const auto& [key, child] : node->children_) {
+        collect_searchable_node_ids(child.get(), deleted_ids, seen_ids, ids);
+    }
 }
 
 float
@@ -590,6 +711,8 @@ PyramidAnalyzer::calculate_quantization_result(
 
     float total_quantization_error = 0.0F;
     float total_quantization_inversion_count_rate = 0.0F;
+    uint32_t valid_sample_count = 0;
+    uint32_t valid_inversion_sample_count = 0;
 
     for (uint32_t i = 0; i < sample_size; ++i) {
         auto id = sample_ids[i];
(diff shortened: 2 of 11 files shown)


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1658,4 +1659,23 @@ Pyramid::GetStats() const {
     return stats.Dump(4);
 }
 
+std::string
+Pyramid::AnalyzeIndexBySearch(const SearchRequest& request) {
```
LHT129, line 1710: [note] The use of `not` keyword in `CHECK_ARGUMENT` is inconsistent with the rest of the VSAG codebase, which consistently uses `!` for logical negation. While `not` is a valid C++ alternative token, mixing styles within the same codebase can be surprising to readers.

Consider using `!` for consistency.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -535,6 +586,76 @@ PyramidAnalyzer::calculate_groundtruth(const Vector<float>& sample_datas,
...
+        return;
+    }
+
+    std::shared_lock lock(node->mutex_);
+    if (node->status_ != IndexNode::Status::NO_INDEX) {
+        Vector<InnerIdType> node_ids(allocator_);
+        if (node->status_ == IndexNode::Status::FLAT) {
+            node_ids = node->ids_;
+        } else if (node->graph_ != nullptr) {
+            node_ids = node->graph_->GetIds();
```
LHT129, line 647: [suggestion] `GetIds()` returns a copy of all graph IDs for every query in `collect_searchable_node_ids`. For batched queries with large graphs, this allocates and copies the full ID set per query. Since this is an analyzer path (not hot), the impact is limited, but for large-scale analysis with many queries it may add noticeable overhead. Consider whether the ID set could be computed once per node and reused across queries that share the same path scope.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -723,9 +883,24 @@ PyramidAnalyzer::get_search_recall(
...
         uint32_t hit_count = 0;
+        UnorderedSet<LabelType> seen_result_labels(allocator_);
+        const float distance_threshold = gt->Top().first;
+        const float tie_tolerance = 1e-6F * std::max(1.0F, std::abs(distance_threshold));
         for (const auto& label : sr) {
-            if (gt_set.find(label) != gt_set.end()) {
+            if (not seen_result_labels.insert(label).second) {
+                continue;
+            }
+            bool is_hit = gt_set.find(label) != gt_set.end();
```
LHT129, line 905: [suggestion] The tie-breaking logic in `get_search_recall` recalculates precise distance via `CalcDistanceById` for every result label not found in the GT set. For large `topk` values, this could be expensive since it decodes the vector for each miss. The tie tolerance (`1e-6F * max(1.0F, abs(threshold))`) is a reasonable approach for handling floating-point ties. However, consider whether the GT set should instead store distances alongside labels to avoid recomputation — the `DistHeapPtr` already holds `(distance, label)` pairs, so the distance could be extracted from the heap data rather than recalculated.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -584,30 +707,45 @@ PyramidAnalyzer::calculate_quantization_result(
         }
 
         const auto& result = search_result.at(id);
-        if (result.empty()) {
+        Vector<LabelType> unique_result(allocator_);
```
LHT129, line 722: [note] The `calculate_quantization_result` now deduplicates result labels via `UnorderedSet` before computing quantization metrics. This is a good fix that prevents double-counting duplicate labels in the error and inversion calculations. The `valid_sample_count` and `valid_inversion_sample_count` tracking also correctly handles edge cases where `result_size == 1` (no inversion pairs possible) and where `base_result` or `precise_result` are null. The per-sample error is now averaged by `result_size` rather than accumulated raw, which produces a more interpretable metric.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -535,6 +586,76 @@ PyramidAnalyzer::calculate_groundtruth(const Vector<float>& sample_datas,
...
+        if (node->status_ == IndexNode::Status::FLAT) {
+            node_ids = node->ids_;
+        } else if (node->graph_ != nullptr) {
+            node_ids = node->graph_->GetIds();
+        }
+        for (const auto id : node_ids) {
+            if (deleted_ids.find(id) == deleted_ids.end() && seen_ids.insert(id).second) {
+                ids.push_back(id);
+            }
+            if (node->graph_ == nullptr) {
```
LHT129, line 653: [suggestion] In `collect_searchable_node_ids`, when `node->status_ == FLAT`, `node->graph_` is always `nullptr`, so the `GetDuplicateIds` loop body (lines 644-649) is unconditionally skipped via the `continue` on line 642. This is correct behavior but the control flow is slightly confusing — the `continue` is inside a loop that iterates `node_ids`, and for FLAT nodes every iteration hits `continue`.

Consider restructuring to make the intent clearer, e.g. only enter the duplicate-expansion block when `node->status_ == GRAPH`:

```cpp
if (node->status_ == IndexNode::Status::GRAPH && node->graph_ != nullptr) {
    for (const auto duplicate_id : node->graph_->GetDuplicateIds(id)) {
        if (deleted_ids.find(duplicate_id) == deleted_ids.end() &&
            seen_ids.insert(duplicate_id).second) {
            ids.push_back(duplicate_id);
        }
    }
}
```

This avoids the `continue`-based skip and makes the GRAPH-only nature of duplicate expansion explicit.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -500,32 +545,38 @@ void
...
+        const auto* query_path = query_paths == nullptr ? nullptr : query_paths + i;
+        auto ids_array = collect_search_scope_ids(hierarchy_name, query_path, deleted_ids);
+        Vector<float> distances_array(ids_array.size(), allocator_);
         auto comp = codes->FactoryComputer(sample_datas.data() + static_cast<size_t>(i) * dim_);
-        codes->Query(distances_array.data(), comp, ids_array.data(), this->total_count_);
+        if (not ids_array.empty()) {
+            codes->Query(distances_array.data(), comp, ids_array.data(), ids_array.size());
+        }
 
         DistHeapPtr gt = std::make_shared<StandardHeap<true, false>>(allocator_, -1);
```
LHT129, line 585: [note] The `codes` selection in `calculate_groundtruth` changed from `has_precise_reorder() ? precise_codes_ : base_codes_` to `raw_vector_ != nullptr ? raw_vector_ : (use_reorder_ ? precise_codes_ : base_codes_)`. This now prefers `raw_vector_` (original uncompressed vectors) when available, which yields the most accurate ground truth distances. This aligns with the `decodable_codes()` pattern used elsewhere in Pyramid. The fallback chain correctly handles all configurations: raw vectors > precise reorder codes > base codes.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -723,9 +883,24 @@ PyramidAnalyzer::get_search_recall(
         }
 
         uint32_t hit_count = 0;
+        UnorderedSet<LabelType> seen_result_labels(allocator_);
+        const float distance_threshold = gt->Top().first;
+        const float tie_tolerance = 1e-6F * std::max(1.0F, std::abs(distance_threshold));
```
LHT129, line 900: [note] The tie-breaking logic in `get_search_recall` uses `gt->Top().first` as the distance threshold and recomputes precise distances via `CalcDistanceById` for result labels not found in the GT set. This correctly handles cases where quantization causes multiple distinct labels to share the same distance as the GT's farthest neighbor.

However, `distance_threshold` is captured once outside the loop (line 888 in the new file). If `gt` were modified between iterations this would be a bug, but since `gt` is read-only here it's fine. Still, consider adding a comment explaining that `gt->Top().first` is the distance of the k-th nearest ground-truth neighbor, which serves as the inclusion boundary for tie-breaking.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -668,10 +824,13 @@ PyramidAnalyzer::calculate_search_result(
 
         if (result != nullptr) {
             auto result_size = result->GetDim();
-            const auto* ids = result->GetIds();
             Vector<LabelType> result_labels(allocator_);
-            result_labels.resize(result_size);
-            std::memcpy(result_labels.data(), ids, result_size * sizeof(LabelType));
+            if (result_size > 0) {
```
LHT129, line 840: [suggestion] The `result_size > 0` guard and `CHECK_ARGUMENT(ids != nullptr, ...)` added in `calculate_search_result` are good defensive improvements. However, when `result_size == 0`, an empty `result_labels` vector is inserted into `search_result`. Downstream consumers (`get_search_recall`, `calculate_quantization_result`) already handle empty results, but this creates an entry that will be skipped. Consider whether it would be cleaner to skip inserting empty results entirely, which would also avoid the `search_result.at(id)` lookup returning an empty vector that then gets filtered out.

src/analyzer/pyramid_analyzer.cpp
LHT129, line 722: [note] In `calculate_quantization_result`, `unique_result.reserve(std::min<uint64_t>(result.size(), topk_))` mixes `size_t` (`result.size()`) and `uint32_t` (`topk_`) via an explicit `uint64_t` template parameter. While this is safe (both fit in `uint64_t`), it would be cleaner to use `std::min(result.size(), static_cast<size_t>(topk_))` to avoid the mixed-type `std::min` call. This is a minor style point and does not affect correctness.

LHT129 (commented):
Overall this is a well-structured PR that implements path-aware `AnalyzeIndexBySearch` for Pyramid with thorough test coverage. The core design — scoping ground truth to each query's hierarchy path, handling duplicates, excluding mark-removed IDs, and using the most precise available codes — is sound.

**Summary of findings:**

1. **[note] `pyramid.cpp:1663`** — `not` keyword usage is inconsistent with the rest of the codebase which uses `!`.
2. **[suggestion] `pyramid_analyzer.cpp:641`** — The `continue`-based skip for duplicate expansion in `collect_searchable_node_ids` could be restructured for clarity.
3. **[note] `pyramid_analyzer.cpp:888`** — The tie-breaking `distance_threshold` logic is correct but could benefit from a brief comment.
4. **[note] `pyramid_analyzer.cpp:710`** — Minor mixed-type `std::min` call in `calculate_quantization_result`.

**Positive observations:**
- Comprehensive test coverage: 5 unit tests + 6 functional tests covering path scoping, removals, graph duplicates, named hierarchies, batched queries, raw-vector ground truth, and topk=1 quantization metrics.
- Correct use of `raw_vector_` > `precise_codes_` > `base_codes_` priority chain for ground truth computation.
- Proper handling of `topk == 1` edge case in quantization metrics (inversion ratio is 0 when there is only one result).
- Good input validation in `Pyramid::AnalyzeIndexBySearch` with clear error messages.
- Documentation updates in both English and Chinese are thorough and accurate.

No blocking issues found. All comments are suggestions/notes.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -500,32 +545,38 @@ void
...
     if (not ground_truth.empty()) {
         return;
     }
 
-    Vector<float> distances_array(this->total_count_, allocator_);
-    Vector<InnerIdType> ids_array(this->total_count_, allocator_);
-    std::iota(ids_array.begin(), ids_array.end(), 0);
-
-    auto codes = pyramid_->has_precise_reorder() ? pyramid_->precise_codes_ : pyramid_->base_codes_;
+    auto codes = pyramid_->raw_vector_ != nullptr
```
LHT129: [critical] The `codes` selection in `calculate_groundtruth` (line 556-558) still uses `pyramid_->use_reorder_` directly instead of `pyramid_->has_precise_reorder()`. This was flagged in a previous review but has not been fixed in this push.

When `use_reorder_` is true but `base_codes_->SupportSplitCodeStorage()` is also true, `has_precise_reorder()` returns false and `precise_codes_` remains null. The current code would select `precise_codes_` and dereference a null pointer at line 568 (`codes->FactoryComputer(...)`) or line 570 (`codes->Query(...)`).

The fix is to use `pyramid_->decodable_codes()` which already implements the correct logic:

```cpp
auto codes = pyramid_->decodable_codes();
```

This helper (defined in `pyramid.h:414-417`) checks `raw_vector_ != nullptr`, then `has_precise_reorder()` (which guards against the split-code-storage case), then falls back to `base_codes_`.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -91,36 +92,80 @@ PyramidAnalyzer::AnalyzeIndexBySearch(const SearchRequest& request) {
...
 
     auto query = request.query_;
-    auto query_num = query->GetNumElements();
+    const auto query_num = query->GetNumElements();
     const float* query_vectors = query->GetFloat32Vectors();
-    if (query_vectors == nullptr || query_num == 0) {
+    if (query_num == 0) {
         return stats;
     }
+    CHECK_ARGUMENT(query_num > 0, fmt::format("query num({}) must be positive", query_num));
```
LHT129, line 108: [suggestion] The `CHECK_ARGUMENT(query_num > 0, ...)` on line 100 is dead code. Line 97-98 already returns early when `query_num == 0`:

```cpp
if (query_num == 0) {
    return stats;          // line 97-98: early return
}
CHECK_ARGUMENT(query_num > 0, ...);  // line 100: can never fail
```

Consider removing the redundant `CHECK_ARGUMENT` since the early return already handles the zero-query case.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -723,9 +883,24 @@ PyramidAnalyzer::get_search_recall(
...
+        const float tie_tolerance = 1e-6F * std::max(1.0F, std::abs(distance_threshold));
         for (const auto& label : sr) {
-            if (gt_set.find(label) != gt_set.end()) {
+            if (not seen_result_labels.insert(label).second) {
+                continue;
+            }
+            bool is_hit = gt_set.find(label) != gt_set.end();
+            if (not is_hit) {
+                const auto distance = pyramid_->CalcDistanceById(
+                    sample_datas.data() + static_cast<uint64_t>(i) * dim_, label, true);
```
LHT129, line 908: [suggestion] The tie-breaking logic in `get_search_recall` uses `CalcDistanceById(..., true)` (line 896) which selects `precise_codes_` for distance computation. However, `calculate_groundtruth` (line 556-558) selects `raw_vector_` as the distance source when it is available. When `store_raw_vector` is enabled, the GT heap distances are computed from raw vectors while the tie-breaking distances come from `precise_codes_`. These two code sources can produce slightly different distances, which could cause incorrect tie-breaking decisions — a result label might be incorrectly counted as a hit or miss due to the distance mismatch.

Consider using the same code source for both GT construction and tie-breaking. One approach: pass the codes pointer used in `calculate_groundtruth` to `get_search_recall`, or use `pyramid_->decodable_codes()` consistently in both places, and use that same codes instance (not `CalcDistanceById`) for tie-breaking distance computation.

LHT129 (commented):
This is a well-structured PR that implements path-aware `AnalyzeIndexBySearch` for Pyramid. The implementation correctly follows Pyramid node semantics for ground truth collection, handles edge cases (deleted vectors, graph duplicates, NO_INDEX nodes, named hierarchies), and includes comprehensive test coverage.

The existing inline comments from the previous review round have already identified the key areas for improvement:
- `GetIds()` copy overhead in `collect_searchable_node_ids` (per-query allocation)
- Control flow clarity in `collect_searchable_node_ids` for FLAT vs GRAPH nodes
- Tie-breaking distance source consistency between GT construction and recall calculation
- Use of `not` keyword vs `!` for consistency with the rest of the codebase

The `decodable_codes()` usage in `calculate_groundtruth` correctly resolves the earlier concern about `use_reorder_` vs `has_precise_reorder()`. The `Paths(hierarchy_name, ...)` behavior for default hierarchies is also correct per the `DatasetImpl` implementation.

No blocking issues found. The existing suggestions are non-blocking improvements that can be addressed at the author's discretion.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -549,6 +598,76 @@ PyramidAnalyzer::calculate_groundtruth(const Vector<float>& sample_datas,
...
+                                             const UnorderedSet<InnerIdType>& deleted_ids,
+                                             UnorderedSet<InnerIdType>& seen_ids,
+                                             Vector<InnerIdType>& ids) {
+    if (node == nullptr) {
+        return;
+    }
+
+    std::shared_lock lock(node->mutex_);
+    if (node->status_ != IndexNode::Status::NO_INDEX) {
+        Vector<InnerIdType> node_ids(allocator_);
```
LHT129, line 643: [suggestion] In `collect_searchable_node_ids`, when a node has `status_ == NO_INDEX` but its `children_` map is empty (a leaf NO_INDEX node), the function silently returns without collecting any IDs. While this is technically correct (there is nothing searchable), it means that if a query path resolves to a NO_INDEX leaf node, the ground-truth scope for that query will be empty, producing a recall of 0. This is consistent with the search behavior (a NO_INDEX leaf returns no results), but it may be surprising to users who expect the analyzer to report meaningful metrics. Consider logging a warning when `collect_searchable_node_ids` encounters a leaf NO_INDEX node to help users diagnose unexpected zero-recall results.

LHT129 (commented):
Thank you for this well-structured PR implementing path-aware `AnalyzeIndexBySearch` for Pyramid. The design correctly follows Pyramid search semantics (path scoping, NO_INDEX recursion, duplicate expansion, mark-removed exclusion) and the test coverage is thorough across unit and functional tests.

I left one additional inline comment about a potential diagnostic improvement for NO_INDEX leaf nodes. The other review comments from earlier rounds have already covered the main substantive points (`not` keyword consistency, `GetIds()` copy cost, tie-breaking distance source consistency, and the `continue`-based control flow in `collect_searchable_node_ids`).

Overall this is solid work — the code is clean, well-tested, and the documentation updates are comprehensive.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks for the note. The alternative token `not` is already used broadly in VSAG, including throughout this file and in other `CHECK_ARGUMENT` expressions, and the project style does not require `!`. We will keep the current form.

src/analyzer/pyramid_analyzer.cpp
jac0626: Good observation. Scope collection does copy graph IDs per query, but ground-truth distance computation must still scan the same scope. Caching would add memory and cache-key/lifetime complexity, so we will keep this PR minimal and revisit it only if profiling shows this copy is material.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. A result label missing from the GT set is not present in the retained top-k heap, so its distance cannot be reused from that heap. If profiling shows this path matters, the misses can be evaluated in one batched `CalcDistancesById` call in a follow-up. We will keep the current correctness-first implementation here.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks for confirming these edge cases. We will retain the current deduplication, per-sample normalization, and valid-sample accounting.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. The current flow is correct under the node invariant that a FLAT node has no graph, while GRAPH nodes may expand duplicate IDs. Since this is only a readability preference and the behavior is covered by tests, we will keep the implementation unchanged to minimize the patch.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. Follow-up CI against the latest main exposed that the handwritten fallback was incomplete for split-code storage. Commit `ed2feb8a` now uses Pyramid's `decodable_codes()` helper and adds regression coverage for split RaBitQ both with and without stored raw vectors.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. `gt` is a read-only max-heap here, so `Top()` is the farthest retained ground-truth item (the kth boundary when the heap is full). The value is captured once and the heap is never mutated in this function. We will keep the local code unchanged.

src/analyzer/pyramid_analyzer.cpp
jac0626: Keeping the empty entry is intentional. When ground truth is non-empty but search returns no labels, that query must contribute zero recall. Omitting the entry would hit the missing-map branch and exclude the failed query from the denominator, inflating recall.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. The explicit `uint64_t` template argument safely unifies `result.size()` and the `uint32_t` top-k without narrowing, and it follows this project's preference for `uint64_t` over `size_t` in changed code. We will keep it as-is.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -500,32 +545,38 @@ void
...
     }
 
-    Vector<float> distances_array(this->total_count_, allocator_);
-    Vector<InnerIdType> ids_array(this->total_count_, allocator_);
-    std::iota(ids_array.begin(), ids_array.end(), 0);
-
-    auto codes = pyramid_->has_precise_reorder() ? pyramid_->precise_codes_ : pyramid_->base_codes_;
+    auto codes = pyramid_->raw_vector_ != nullptr
+                     ? pyramid_->raw_vector_
+                     : (pyramid_->use_reorder_ ? pyramid_->precise_codes_ : pyramid_->base_codes_);
```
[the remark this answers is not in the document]
jac0626: Fixed in `ed2feb8a`: ground-truth code selection now delegates to `pyramid_->decodable_codes()`. The regression test covers split RaBitQ with no raw/standalone precise codes as well as the stored-raw-vector path, including an ASan run.

src/analyzer/pyramid_analyzer.cpp
jac0626: Fixed in `ed2feb8a` exactly as suggested by using `pyramid_->decodable_codes()`. The regression test explicitly verifies the split-code configuration where `use_reorder_` is true while both `raw_vector_` and `precise_codes_` are absent.

src/analyzer/pyramid_analyzer.cpp
jac0626: The check is still reachable for negative counts. `GetNumElements()` is `int64_t`: zero is intentionally accepted as an empty analysis and returns empty stats, while a negative value passes the zero branch and must be rejected before the later unsigned conversion. We will keep the validation.

src/analyzer/pyramid_analyzer.cpp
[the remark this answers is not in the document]
jac0626: Fixed in `ed2feb8a` by using Pyramid's existing `decodable_codes()` capability-based selection. Tests now cover both raw and no-raw split RaBitQ configurations.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -658,6 +811,9 @@ PyramidAnalyzer::calculate_search_result(
         auto query = Dataset::Make();
         query->Dim(dim_)->NumElements(1)->Owner(false)->Float32Vectors(
             sample_datas.data() + static_cast<size_t>(i) * dim_);
+        if (query_paths != nullptr) {
+            query->Paths(hierarchy_name, query_paths + i);
+        }
```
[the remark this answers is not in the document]
jac0626: This case is already handled by the Dataset contract. `DatasetImpl::Paths(const std::string&, ...)` forwards an empty hierarchy name directly to the default `Paths(...)` setter, and `GetPaths("")` likewise forwards to `GetPaths()`. Therefore the current call is equivalent and no branch is needed.

src/analyzer/pyramid_analyzer.cpp
jac0626: The two paths use the same source priority. `CalcDistanceById(..., true)` starts with base codes, selects standalone precise codes when available, and finally lets raw vectors override both. That is equivalent to `decodable_codes()` (`raw -> precise -> base`), so the proposed mismatch cannot occur.

src/analyzer/pyramid_analyzer.cpp
jac0626: This is intentional and mirrors real search semantics: a NO_INDEX leaf has no searchable descendants and returns no results. Empty-GT queries are skipped from the valid recall count; only an analysis with no valid scopes ultimately reports zero. A warning inside this per-query traversal would be noisy for a legal configuration, so we will not add one. If diagnostics are needed later, an aggregated empty-scope counter would be preferable.

LHT129 (approved):
LGTM

PR #2676, opened 2026-08-12
feat(ivf): add explicit bucket graph rebuild API

### LINKED ISSUE #2674: feat(ivf): add explicit bucket graph rebuild API
## Description

After Build() followed by incremental Add() calls, IVF indices with graph_build_threshold > 0 have stale bucket graphs. Buckets that grow beyond the threshold after Add fall back to flat search because their graph TotalCount() no longer matches the bucket size.

## Proposal

Add a public API RebuildIVFBucketGraphs(const DatasetPtr& base) that allows users to explicitly rebuild per-bucket graphs after Add operations.

### API Contract
- Caller supplies all original raw float32 vectors in insertion order
- Dataset must contain exactly GetNumElements() rows (including logically deleted vectors)
- Dimension must match index dimension
- Requires graph_build_threshold > 0 (graph mode enabled)

### Behavior
1. Validate inputs (threshold, null check, count match, dim match)
2. Clear stale bucket graphs (std::fill with nullptr)
3. Call existing build_bucket_graphs(base) to rebuild
4. Recalculate memory usage

### Validation
- Return error if graph mode disabled (graph_build_threshold == 0)
- Return error if dataset is null
- Return error if element count mismatch
- Return error if dimension mismatch

### Tests
- Build + Add + Rebuild restores graph-based search
- Disabled threshold rejected
- Count mismatch rejected

## Implementation
- include/vsag/index.h: public virtual API with unsupported default
- src/algorithm/inner_index_interface.h: internal interface default throwing
- src/index/index_impl.h: forwarding with CHECK_IMMUTABLE_INDEX + SAFE_CALL
- src/algorithm/ivf/ivf.{h,cpp}: IVF-specific implementation
- tests/test_ivf.cpp: 3 test cases

LHT129:
## Summary

Add RebuildIVFBucketGraphs() public API to allow users to explicitly rebuild per-bucket graphs after incremental Add() calls. This enables IVF indices with graph_build_threshold > 0 to restore graph-based search in long buckets that became stale after Add.

Closes #2674

## Changes

- include/vsag/index.h: Add public virtual RebuildIVFBucketGraphs with unsupported default
- src/algorithm/inner_index_interface.h: Add internal interface default that throws UNSUPPORTED_INDEX_OPERATION
- src/index/index_impl.h: Add forwarding override with CHECK_IMMUTABLE_INDEX + SAFE_CALL
- src/algorithm/ivf/ivf.{h,cpp}: Implement IVF-specific RebuildBucketGraphs with validation
- tests/test_ivf.cpp: Add 3 test cases for rebuild, disabled threshold, and count mismatch

## API Contract

The caller must supply the complete set of raw float32 vectors that were originally passed to Build() and any subsequent Add() calls, in insertion order. The dataset must contain exactly GetNumElements() rows and match the index dimension.

## Test Results

All 3 new tests pass. Existing IVF graph tests continue to pass (11/12, 1 pre-existing failure unrelated to this change).

### DIFF
--- a/include/vsag/index.h
+++ b/include/vsag/index.h
@@ -1092,6 +1092,27 @@ public:
             Error(ErrorType::UNSUPPORTED_INDEX_OPERATION, "Index does not support ImportCache"));
     }
 
+    /**
+     * @brief Rebuild per-bucket graphs for IVF indices after Add operations.
+     *
+     * When an IVF index is configured with graph_build_threshold > 0, large buckets
+     * get internal graphs for faster search. After Add() operations, these graphs
+     * become stale. This method explicitly rebuilds them using bucket-internal
+     * distance computation (no external dataset needed).
+     *
+     * Requirements:
+     * - Index must be IVF with graph_build_threshold > 0
+     * - Not safe for concurrent use with Add/Remove/Build/Serialize
+     *
+     * @return tl::expected<void, Error>; an Error is returned if the index does not
+     * support this operation or if graph_build_threshold was not configured.
+     */
+    virtual tl::expected<void, Error>
+    RebuildIVFBucketGraphs() {
+        return tl::unexpected(Error(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                                    "Index does not support RebuildIVFBucketGraphs"));
+    }
+
 public:
     virtual ~Index() = default;
 
--- a/src/algorithm/ivf/ivf.cpp
+++ b/src/algorithm/ivf/ivf.cpp
@@ -511,7 +511,7 @@ IVF::Build(const DatasetPtr& base) {
     // TODO(LHT): duplicate
     auto result = this->Add(base);
     if (graph_build_threshold_ > 0) {
-        this->build_bucket_graphs(base);
+        this->build_bucket_graphs();
     }
     this->cal_memory_usage();
     return result;
@@ -620,30 +620,6 @@ IVF::Add(const DatasetPtr& base) {
     return {};
 }
 
-static const void*
-get_ivf_graph_build_data(const DatasetPtr& dataset,
-                         DataTypes data_type,
-                         uint64_t index,
-                         uint64_t dim) {
-    if (data_type == DataTypes::DATA_TYPE_FLOAT) {
-        const auto* ptr = dataset->GetFloat32Vectors();
-        return ptr != nullptr ? ptr + index * dim : nullptr;
-    }
-    if (data_type == DataTypes::DATA_TYPE_INT8) {
-        const auto* ptr = dataset->GetInt8Vectors();
-        return ptr != nullptr ? ptr + index * dim : nullptr;
-    }
-    if (data_type == DataTypes::DATA_TYPE_FP16 || data_type == DataTypes::DATA_TYPE_BF16) {
-        const auto* ptr = dataset->GetFloat16Vectors();
-        return ptr != nullptr ? ptr + index * dim : nullptr;
-    }
-    if (data_type == DataTypes::DATA_TYPE_SPARSE) {
-        const auto* ptr = dataset->GetSparseVectors();
-        return ptr != nullptr ? ptr + index : nullptr;
-    }
-    throw VsagException(ErrorType::INVALID_ARGUMENT, "invalid data type for IVF graph build");
-}
-
 static GraphInterfaceParamPtr
 get_ivf_graph_param(const std::string& graph_param_string) {
     auto graph_json = JsonType::Parse(graph_param_string);
@@ -687,12 +663,51 @@ has_any_fresh_bucket_graph(const BucketInterfacePtr& bucket,
     return false;
 }
 
-void
-IVF::build_bucket_graphs(const DatasetPtr& base) {
-    if (graph_build_threshold_ <= 0) {
-        return;
+class PairwiseBucketDistanceProvider final : public DistanceProviderForGraph {
+public:
+    PairwiseBucketDistanceProvider(std::shared_ptr<BucketInterface> bucket,
+                                   BucketIdType bucket_id,
+                                   InnerIdType query_id)
+        : bucket_(std::move(bucket)), bucket_id_(bucket_id), query_id_(query_id) {
     }
-    if (common_param_.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
+
+    [[nodiscard]] float
+    QueryDistance(InnerIdType id, QueryContext* ctx = nullptr) const override {
+        return bucket_->ComputePairVectors(bucket_id_, query_id_, id);
+    }
+
+    void
+    BatchQueryDistance(float* distances,
+                       const InnerIdType* ids,
+                       InnerIdType count,
+                       QueryContext* ctx = nullptr) const override {
+        for (InnerIdType i = 0; i < count; ++i) {
+            distances[i] = bucket_->ComputePairVectors(bucket_id_, query_id_, ids[i]);
+        }
+    }
+
+    [[nodiscard]] float
+    PairwiseDistance(InnerIdType id1,
+                     InnerIdType id2,
+                     const ComputerInterfacePtr& computer = nullptr) const override {
+        return bucket_->ComputePairVectors(bucket_id_, id1, id2);
+    }
+
+    [[nodiscard]] ComputerInterfacePtr
+    FactoryComputerById(InnerIdType id) const override {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "PairwiseBucketDistanceProvider does not support FactoryComputerById");
+    }
+
+private:
+    std::shared_ptr<BucketInterface> bucket_;
+    InnerIdType query_id_;
+    BucketIdType bucket_id_;
+};
+
+void
+IVF::build_bucket_graphs() {
+    if (graph_build_threshold_ <= 0) {
         return;
     }
     if (graph_param_ == nullptr) {
@@ -704,10 +719,10 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
     const auto bucket_count = bucket_->bucket_count_;
     bucket_graphs_.resize(bucket_count);
 
-    for (BucketIdType b = 0; b < bucket_count; ++b) {
+    auto build_one_bucket = [&](BucketIdType b) {
         const auto bucket_size = bucket_->GetBucketSize(b);
         if (bucket_size < graph_build_threshold_) {
-            continue;
+            return;
         }
 
         const auto* inner_ids = bucket_->GetInnerIds(b);
@@ -719,13 +734,13 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
             }
         }
         if (valid_ids.size() < static_cast<uint64_t>(graph_build_threshold_)) {
-            continue;
+            return;
         }
 
         const auto effective_degree =
             std::min(max_degree, static_cast<int64_t>(valid_ids.size()) - 1);
         if (effective_degree <= 0) {
-            continue;
+            return;
         }
 
         auto graph = GraphInterface::MakeInstance(graph_param_, this->common_param_);
@@ -733,13 +748,6 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
         graph->SetTotalCount(bucket_size);
         graph->SetMaximumDegree(static_cast<uint32_t>(effective_degree));
 
-        auto make_computer = [&](InnerIdType local_id) {
-            const auto orig_idx = static_cast<uint64_t>(inner_ids[local_id]) / buckets_per_data_;
-            const auto* query =
-                get_ivf_graph_build_data(base, common_param_.data_type_, orig_idx, this->dim_);
-            CHECK_ARGUMENT(query != nullptr, "base dataset has no graph build vector data");
-            return bucket_->FactoryComputer(query);
-        };
         auto mutexes = std::make_shared<EmptyMutex>();
         BasicSearcher searcher(common_param_);
 
@@ -752,8 +760,7 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
             search_param.ep = entry;
             search_param.ef = std::min(ef_construction, node_pos);
             search_param.topk = static_cast<int64_t>(search_param.ef);
-            BucketDistanceProvider distance_provider(
-                bucket_, b, make_computer(node), make_computer, inner_ids, buckets_per_data_);
+            PairwiseBucketDistanceProvider distance_provider(bucket_, b, node);
             visited->Reset();
             auto candidates =
                 searcher.Search(graph, distance_provider, visited, search_param, nullptr, nullptr);
@@ -762,6 +769,21 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
         }
 
         bucket_graphs_[b] = std::move(graph);
+    };
+
+    if (this->thread_pool_ != nullptr) {
+        std::vector<std::future<void>> futures;
+        futures.reserve(bucket_count);
+        for (BucketIdType b = 0; b < bucket_count; ++b) {
+            futures.emplace_back(this->thread_pool_->GeneralEnqueue(build_one_bucket, b));
+        }
+        for (auto& future : futures) {
+            future.get();
+        }
+    } else {
+        for (BucketIdType b = 0; b < bucket_count; ++b) {
+            build_one_bucket(b);
+        }
     }
 }
 DatasetPtr
@@ -2401,4 +2423,29 @@ IVF::GetMemoryUsage() const {
     return memory;
 }
 
+void
+IVF::RebuildBucketGraphs() {
+    if (graph_build_threshold_ <= 0) {
+        throw VsagException(
+            ErrorType::UNSUPPORTED_INDEX_OPERATION,
+            "RebuildIVFBucketGraphs: index was not configured with graph_build_threshold > 0");
+    }
+    if (common_param_.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "RebuildIVFBucketGraphs: only supports float32 data type");
+    }
+
+    // Build new graphs in temp storage first (exception safety)
+    auto old_graphs = std::move(bucket_graphs_);
+    bucket_graphs_.resize(old_graphs.size(), nullptr);
+
+    try {
+        this->build_bucket_graphs();
+        this->cal_memory_usage();
+    } catch (...) {
+        // Restore old graphs on failure
+        bucket_graphs_ = std::move(old_graphs);
+        throw;
+    }
+}
 }  // namespace vsag
--- a/src/datacell/bucket_datacell.h
+++ b/src/datacell/bucket_datacell.h
@@ -67,6 +67,9 @@ public:
         return this->query_one_by_id(comp, bucket_id, offset_id);
     }
 
+    float
+    ComputePairVectors(BucketIdType bucket_id, InnerIdType id1, InnerIdType id2) override;
+
     ComputerInterfacePtr
     FactoryComputer(const void* query) override;
 
@@ -672,4 +675,46 @@ BucketDataCell<QuantTmpl, IOTmpl>::GetCodesById(BucketIdType bucket_id,
     this->datas_[bucket_id].Read(this->code_size_, offset_id * this->code_size_, data);
 }
 
+template <typename QuantTmpl, typename IOTmpl>
+float
+BucketDataCell<QuantTmpl, IOTmpl>::ComputePairVectors(BucketIdType bucket_id,
+                                                      InnerIdType id1,
+                                                      InnerIdType id2) {
+    if (bucket_id >= this->bucket_count_) {
+        throw VsagException(
+            ErrorType::INTERNAL_ERROR,
+            fmt::format("ComputePairVectors failed: bucket id ({}) is invalid", bucket_id));
+    }
+    if (id1 >= this->bucket_sizes_[bucket_id] || id2 >= this->bucket_sizes_[bucket_id]) {
+        throw VsagException(ErrorType::INTERNAL_ERROR,
+                            fmt::format("ComputePairVectors failed: offset id is invalid"));
+    }
+    if (this->inner_ids_[bucket_id][id1] == EMPTY_INNER_ID ||
+        this->inner_ids_[bucket_id][id2] == EMPTY_INNER_ID) {
+        throw VsagException(
+            ErrorType::INTERNAL_ERROR,
+            "ComputePairVectors failed: cannot compute distance for deleted vector");
+    }
+    std::shared_lock lock(this->bucket_mutexes_[bucket_id]);
+    constexpr size_t kStackThreshold = 256;
+    uint8_t stack_buf1[kStackThreshold];
+    uint8_t stack_buf2[kStackThreshold];
+
+    uint8_t *code1_ptr, *code2_ptr;
+    std::vector<uint8_t> heap_buf1, heap_buf2;
+
+    if (this->code_size_ <= kStackThreshold) {
+        code1_ptr = stack_buf1;
+        code2_ptr = stack_buf2;
+    } else {
+        heap_buf1.resize(this->code_size_);
+        heap_buf2.resize(this->code_size_);
+        code1_ptr = heap_buf1.data();
+        code2_ptr = heap_buf2.data();
+    }
+
+    this->datas_[bucket_id].Read(this->code_size_, id1 * this->code_size_, code1_ptr);
+    this->datas_[bucket_id].Read(this->code_size_, id2 * this->code_size_, code2_ptr);
+    return this->quantizer_->Compute(code1_ptr, code2_ptr);
+}
 }  // namespace vsag
--- a/tests/test_ivf.cpp
+++ b/tests/test_ivf.cpp
@@ -2272,6 +2272,55 @@ TEST_CASE("IVF GraphBucketSearcher Add After Build Falls Back To Flat", "[ft][iv
     REQUIRE(result.value()->GetIds()[0] == dataset->base_->GetIds()[base_count - 1]);
 }
 
+TEST_CASE("IVF RebuildBucketGraphs Restores Graph After Add", "[ft][ivf][graph]") {
+    auto dim = 32;
+    auto metric = "l2";
+    auto base_count = 1000;
+    auto buckets_count = 10;
+    auto threshold = 30;
+    auto param = GenerateIVFGraphBuildParametersString(metric, dim, buckets_count, threshold);
+    auto dataset = fixtures::IVFTestIndex::pool.GetDatasetAndCreate(dim, base_count, metric);
+
+    auto index = vsag::Factory::CreateIndex("ivf", param);
+    REQUIRE(index.has_value());
+    REQUIRE(index.value()->Build(dataset->base_).has_value());
+
+    auto add_count = 100;
+    auto added_dataset = fixtures::IVFTestIndex::pool.GetDatasetAndCreate(dim, add_count, metric);
+    REQUIRE(index.value()->Add(added_dataset->base_).has_value());
+
+    auto rebuild_result = index.value()->RebuildIVFBucketGraphs();
+    REQUIRE(rebuild_result.has_value());
+
+    auto query = fixtures::get_one_query(dataset->query_, 0);
+    auto search_param = GenerateIVFGraphSearchParametersString(buckets_count, 100);
+    auto result = index.value()->KnnSearch(query, 10, search_param);
+    REQUIRE(result.has_value());
+    REQUIRE(result.value()->GetDim() == 10);
+
+    // Verify graph search was actually used (check statistics)
+    auto statistics = vsag::JsonType::Parse(result.value()->GetStatistics());
+    REQUIRE(statistics["distance_evaluations_by_phase"]["approximate"].GetUint64() > 0);
+}
+
+TEST_CASE("IVF RebuildBucketGraphs Disabled Threshold Rejected", "[ft][ivf][graph]") {
+    auto dim = 32;
+    auto metric = "l2";
+    auto base_count = 100;
+    auto buckets_count = 10;
+    auto param = fixtures::IVFTestIndex::GenerateIVFBuildParametersString(
+        metric, dim, "fp32", buckets_count);
+    auto dataset = fixtures::IVFTestIndex::pool.GetDatasetAndCreate(dim, base_count, metric);
+
+    auto index = vsag::Factory::CreateIndex("ivf", param);
+    REQUIRE(index.has_value());
+    REQUIRE(index.value()->Build(dataset->base_).has_value());
+
+    auto rebuild_result = index.value()->RebuildIVFBucketGraphs();
+    REQUIRE_FALSE(rebuild_result.has_value());
+    REQUIRE(rebuild_result.error().type == vsag::ErrorType::UNSUPPORTED_INDEX_OPERATION);
+}
+
 TEST_CASE("IVF GraphBucketSearcher Serialization", "[ft][ivf][graph][serialize]") {
     auto dim = 32;
     auto metric = "l2";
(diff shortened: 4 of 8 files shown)


src/algorithm/ivf/ivf.cpp
```diff
@@ -764,6 +764,34 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
...
+                        this->total_elements_));
+    }
+    if (base->GetDim() != this->dim_) {
+        throw VsagException(
+            ErrorType::INVALID_ARGUMENT,
+            fmt::format(
+                "base dataset dim {} does not match index dim {}", base->GetDim(), this->dim_));
+    }
+    std::fill(bucket_graphs_.begin(), bucket_graphs_.end(), nullptr);
+    build_bucket_graphs(base);
```
LHT129: [suggestion] `RebuildBucketGraphs` does not check whether the index uses float32 data type before calling `build_bucket_graphs`. However, `build_bucket_graphs` silently returns when `common_param_.data_type_ != DataTypes::DATA_TYPE_FLOAT` (line 694-696). This means if the index was built with a non-float data type (e.g., fp16), the rebuild call will appear to succeed but no graphs will actually be rebuilt, which could mislead the caller.

Consider adding a data type check similar to the existing dimension check:

```cpp
if (common_param_.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
                        "RebuildBucketGraphs requires DATA_TYPE_FLOAT");
}
```

Alternatively, if non-float data types are never used with graph-enabled IVF in practice, at least add a comment noting this assumption.

LHT129 (commented):
I have reviewed this PR and the existing inline comments from LHT129 and Copilot already cover the key findings. Here is a summary:

**What this PR does well:**
- Clean API design following existing patterns (virtual default → InnerIndexInterface → IndexImpl → IVF)
- Proper validation: null check, element count check, dimension check, threshold check
- Good test coverage: happy path (rebuild restores graph search), disabled threshold rejection, count mismatch rejection
- Consistent error propagation via `CHECK_IMMUTABLE_INDEX` + `SAFE_CALL`

**Existing comments that should be addressed:**
1. **[LHT129] Data type check**: `build_bucket_graphs` silently returns when `data_type_ != DATA_TYPE_FLOAT`, which means `RebuildBucketGraphs` would appear to succeed but produce no graphs. Consider adding an explicit data type check in `RebuildBucketGraphs`.
2. **[Copilot] Doc mismatch**: The doc comment says `graph_build_threshold == 0` but the code checks `<= 0`. Update the doc to match.
3. **[Copilot] Insertion order validation**: The API contract requires insertion order but only validates count and dimension. Consider whether this is acceptable as a documented caller responsibility or if additional validation is warranted.
4. **[Copilot] Error message**: The error message references the internal method name `RebuildBucketGraphs` rather than the public API name `RebuildIVFBucketGraphs`.

No additional blocking issues found beyond what is already flagged.

include/vsag/index.h
```diff
@@ -1092,6 +1092,25 @@ class Index {
...
+     *
+     * The caller must supply the complete set of raw float32 vectors that
+     * were originally passed to Build() and any subsequent Add() calls, in
+     * insertion order.  The dataset must contain exactly @c GetNumElements()
+     * rows (including any logically deleted vectors) and match the index
+     * dimension.  Stale bucket graphs are discarded and rebuilt from scratch.
+     *
+     * @return tl::expected<void, Error>; an Error is returned if the index
+     * does not support this operation, the dataset is invalid, or graph
+     * construction is not enabled (graph_build_threshold == 0).
```
LHT129: [note] The doc comment says `graph_build_threshold == 0`, but the implementation in `RebuildBucketGraphs()` checks `graph_build_threshold_ <= 0`. The `<=` guard is the correct one (a negative threshold would also be invalid), so the doc should be updated to match:

```cpp
 * construction is not enabled (graph_build_threshold <= 0).
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -764,6 +736,18 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
         bucket_graphs_[b] = std::move(graph);
     }
 }
+
+void
+IVF::RebuildBucketGraphs() {
+    if (graph_build_threshold_ <= 0) {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "RebuildBucketGraphs requires graph_build_threshold > 0");
+    }
```
LHT129, line 787: [suggestion] The error message references the internal method name `RebuildBucketGraphs`, but this exception propagates to the user-facing `RebuildIVFBucketGraphs()` API. Consider using the public API name in the message so users can more easily identify the failing call:

```cpp
                    "RebuildIVFBucketGraphs requires graph_build_threshold > 0");
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -764,6 +736,18 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
         bucket_graphs_[b] = std::move(graph);
     }
 }
+
+void
+IVF::RebuildBucketGraphs() {
+    if (graph_build_threshold_ <= 0) {
```
LHT129: [suggestion] `RebuildBucketGraphs()` does not verify that the index stores raw float32 vectors before proceeding with graph construction. `build_bucket_graphs()` now uses `GetCodesById()` + `FactoryComputer()`, which only produces correct distances when the quantizer is FP32 (i.e., `has_raw_vector_` is true). For non-FP32 quantizers, `GetCodesById()` returns compressed codes that `FactoryComputer()` will misinterpret as raw floats, leading to corrupted graphs.

The old code had a `data_type_` guard in `build_bucket_graphs()` that was removed. Consider adding a check using the existing `has_raw_vector_` flag:

```cpp
if (not has_raw_vector_) {
                        "RebuildBucketGraphs requires FP32 quantizer");
}
```

src/algorithm/ivf/ivf.cpp
```diff
@@ -764,6 +736,18 @@ IVF::build_bucket_graphs(const DatasetPtr& base) {
...
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "RebuildBucketGraphs requires graph_build_threshold > 0");
+    }
+    std::fill(bucket_graphs_.begin(), bucket_graphs_.end(), nullptr);
+    build_bucket_graphs();
```
LHT129: [suggestion] `std::fill` clears all existing bucket graphs before `build_bucket_graphs()` runs. If `build_bucket_graphs()` throws (e.g., due to memory allocation failure during graph construction), the index is left in an inconsistent state: the old graphs are gone and no new ones were built. This would silently degrade search quality on subsequent queries.

Consider building the new graphs into a temporary vector first, then swapping with `bucket_graphs_` only on success:

```cpp
auto new_graphs = Vector<GraphInterfacePtr>(bucket_graphs_.size(), allocator_);
// ... build into new_graphs ...
bucket_graphs_.swap(new_graphs);
```

This ensures the index remains usable even if graph construction fails partway through.

include/vsag/index.h
```diff
@@ -1092,6 +1092,25 @@ class Index {
             Error(ErrorType::UNSUPPORTED_INDEX_OPERATION, "Index does not support ImportCache"));
     }
 
+    /**
+     * @brief Rebuild per-bucket graphs for IVF after incremental Add() calls.
+     *
```
LHT129, line 1097: [critical] The doc comment for RebuildIVFBucketGraphs() still describes the old API contract that required the caller to supply a dataset. The function signature is now parameterless. The entire paragraph about supplying vectors is misleading and should be updated.

include/vsag/index.h
```diff
@@ -1092,6 +1092,25 @@ class Index {
...
+     * The caller must supply the complete set of raw float32 vectors that
+     * were originally passed to Build() and any subsequent Add() calls, in
+     * insertion order.  The dataset must contain exactly @c GetNumElements()
+     * rows (including any logically deleted vectors) and match the index
+     * dimension.  Stale bucket graphs are discarded and rebuilt from scratch.
+     *
+     * @return tl::expected<void, Error>; an Error is returned if the index
```
LHT129: [suggestion] The `@return` clause mentions "the dataset is invalid" but the function no longer accepts a dataset parameter. Consider removing this phrase or replacing it with a description of the actual failure modes (e.g., graph construction not enabled).

LHT129 (commented):
Thanks for this update that simplifies the API by removing the dataset parameter. The key improvement is that `build_bucket_graphs()` now reads vectors internally from bucket codes, eliminating the need for callers to supply the full dataset.

A few issues to address:

1. **[critical] `include/vsag/index.h`**: The doc comment still describes the old API contract requiring the caller to supply vectors. Since the function is now parameterless, this documentation is misleading and should be updated.

2. **[critical] `src/algorithm/ivf/ivf.cpp`**: `build_bucket_graphs()` now calls `FactoryComputer()` with the output of `GetCodesById()`. For non-FP32 quantizers, `GetCodesById()` returns quantized codes, not raw float vectors. The previous code guarded against this with a `data_type_` check that has been removed. This needs either a guard or proper decode-before-compute logic.

3. **[suggestion] `include/vsag/index.h`**: The `@return` clause references "the dataset is invalid" which no longer applies.

4. **[suggestion] `src/algorithm/ivf/ivf.cpp`**: The error message in `RebuildBucketGraphs()` references the internal method name. Consider using the public API name `RebuildIVFBucketGraphs` for user-facing errors.

tests/test_ivf.cpp
```diff
@@ -2272,6 +2272,50 @@ TEST_CASE("IVF GraphBucketSearcher Add After Build Falls Back To Flat", "[ft][iv
...
+    auto index = vsag::Factory::CreateIndex("ivf", param);
+    REQUIRE(index.has_value());
+    REQUIRE(index.value()->Build(dataset->base_).has_value());
+
+    auto add_count = 100;
+    auto added_dataset = fixtures::IVFTestIndex::pool.GetDatasetAndCreate(dim, add_count, metric);
+    REQUIRE(index.value()->Add(added_dataset->base_).has_value());
+
+    auto rebuild_result = index.value()->RebuildIVFBucketGraphs();
+    REQUIRE(rebuild_result.has_value());
```
LHT129, line 2293: [suggestion] The test "IVF RebuildBucketGraphs Restores Graph After Add" only verifies that `RebuildIVFBucketGraphs()` and `KnnSearch()` return successfully, but does not verify that the rebuild actually restores graph-based search behavior. After rebuild, the search should use the bucket graph rather than falling back to brute-force. Consider adding a check that verifies the search path actually uses the rebuilt graph (e.g., by comparing recall against a known baseline, or by checking that the number of distance computations is consistent with graph-based search rather than brute-force scan).

Additionally, the test does not verify that the search results are correct (e.g., recall >= some threshold). A correctness check would catch cases where `ComputePairVectors` produces wrong distances due to quantizer mismatch.

src/algorithm/ivf/ivf.cpp
```diff
@@ -687,14 +687,53 @@ has_any_fresh_bucket_graph(const BucketInterfacePtr& bucket,
...
+                                   BucketIdType bucket_id,
+                                   InnerIdType query_id)
+        : bucket_(std::move(bucket)), bucket_id_(bucket_id), query_id_(query_id) {
+    }
+
+    [[nodiscard]] float
+    QueryDistance(InnerIdType id, QueryContext* ctx = nullptr) const override {
+        return bucket_->ComputePairVectors(bucket_id_, query_id_, id);
+    }
+
```
LHT129, line 678: [suggestion] `PairwiseBucketDistanceProvider::BatchQueryDistance` calls `ComputePairVectors` in a loop, and each call to `ComputePairVectors` allocates two `Vector<uint8_t>` buffers (`code1` and `code2`). For large batch sizes, this results in O(n) heap allocations per batch call.

Consider adding an overload of `ComputePairVectors` that accepts pre-allocated output buffers, or caching the code buffers inside `PairwiseBucketDistanceProvider` to reuse across calls:

```cpp
mutable Vector<uint8_t> code1_;
mutable Vector<uint8_t> code2_;
```

This would reduce allocation overhead during graph construction, especially for buckets with many vectors.

LHT129 (commented):
## Review Summary

This PR adds a `RebuildIVFBucketGraphs()` API that enables users to explicitly rebuild per-bucket HNSW graphs after incremental `Add()` calls. The latest iteration (76c0fe83) removes the dataset parameter requirement by using `ComputePairVectors` to compute distances directly from stored quantized codes, which is a cleaner design.

### Issues Identified

1. **[suggestion] Test coverage lacks correctness verification** — The rebuild test only checks that calls succeed, not that the rebuilt graph produces correct search results. Consider adding recall validation.

2. **[suggestion] Per-call allocations in hot path** — `ComputePairVectors` allocates two `Vector<uint8_t>` buffers per call. In `BatchQueryDistance`, this creates O(n) allocations. Consider reusing buffers in `PairwiseBucketDistanceProvider`.

### Existing concerns from other reviewers (not yet resolved)

- **Thread safety**: `RebuildBucketGraphs` clears and replaces `bucket_graphs_` without synchronization, which could race with concurrent searches.
- **Non-FP32 quantizers**: The removal of the `data_type_` guard in `build_bucket_graphs()` means non-FP32 quantizers may produce incorrect distances or out-of-bounds reads during graph construction.

Overall the implementation is solid and the API design is clean. The main concerns are around thread safety during rebuild and correctness guarantees for non-FP32 data types.

LHT129 (commented):
[note] After this refactor, `get_ivf_graph_build_data` (ivf.cpp:624-639) is no longer called anywhere — its only call site was in the `make_computer` lambda that was removed from `build_bucket_graphs`. Consider removing this dead code to keep the codebase clean.

src/datacell/bucket_datacell.h
```diff
@@ -672,4 +675,39 @@ BucketDataCell<QuantTmpl, IOTmpl>::GetCodesById(BucketIdType bucket_id,
...
+        heap_buf2.resize(this->code_size_);
+        code1_ptr = heap_buf1.data();
+        code2_ptr = heap_buf2.data();
+    }
+
+    this->datas_[bucket_id].Read(this->code_size_, id1 * this->code_size_, code1_ptr);
+    this->datas_[bucket_id].Read(this->code_size_, id2 * this->code_size_, code2_ptr);
+    return this->quantizer_->Compute(code1_ptr, code2_ptr);
+}
 }  // namespace vsag
```
LHT129, line 720: [note] `ComputePairVectors` reads from `datas_[bucket_id]` without acquiring any lock on `bucket_mutexes_[bucket_id]`. In contrast, `query_one_by_id` and `scan_bucket_by_id` both acquire `std::shared_lock` on the bucket mutex before reading.

While the `RebuildIVFBucketGraphs` API is documented as "not safe for concurrent use with Add/Remove/Build/Serialize", the initial `Build()` path also calls `build_bucket_graphs()` which internally uses `ComputePairVectors`. If this method is ever called from a context where concurrent reads/writes to the same bucket are possible, it would be a data race.

Consider either:
- Adding a `std::shared_lock` on `bucket_mutexes_[bucket_id]` for consistency with other read methods, or
- Adding a comment noting that the caller is responsible for ensuring exclusive access.

LHT129, line 720: [note] `ComputePairVectors` does not check whether `id1` or `id2` correspond to deleted vectors (marked with `EMPTY_INNER_ID`). The current caller `build_bucket_graphs()` pre-filters deleted entries via `valid_ids`, so this is safe for the current usage. However, since `ComputePairVectors` is now a public method on `BucketInterface`, a future caller could pass a deleted offset and get a meaningless distance computed from stale/zero-filled data.

Consider either:
- Adding an `EMPTY_INNER_ID` check similar to `query_one_by_id`, or
- Documenting that the caller is responsible for ensuring both IDs are valid.

src/algorithm/ivf/ivf.cpp
```diff
@@ -2401,4 +2432,32 @@ IVF::GetMemoryUsage() const {
...
+                            "RebuildIVFBucketGraphs: only supports float32 data type");
+    }
+
+    // Build new graphs in temp storage first (exception safety)
+    Vector<GraphInterfacePtr> new_graphs(bucket_graphs_.size(), nullptr, allocator_);
+
+    // Temporarily swap in new graphs for building
+    auto old_graphs = std::move(bucket_graphs_);
+    bucket_graphs_ = std::move(new_graphs);
+
```
LHT129, line 2441: [suggestion] The `new_graphs` vector is created, then immediately moved into `bucket_graphs_`, leaving `new_graphs` in a valid-but-empty state. The variable serves no purpose after the swap and its name is misleading since it never actually holds the new graphs (they are built directly into `bucket_graphs_`).

Consider simplifying to:

```cpp
auto old_graphs = std::move(bucket_graphs_);
bucket_graphs_.resize(old_graphs.size(), nullptr);
```

This avoids the intermediate `new_graphs` variable and makes the intent clearer: save old graphs, clear current, build new ones in place.

LHT129 (commented):
## Code Review Summary

This PR adds `RebuildIVFBucketGraphs()` public API to IVF indices, allowing explicit rebuild of per-bucket graphs after `Add()` operations. The implementation is clean, with proper exception safety via swap pattern and reasonable test coverage.

### Key Observations

**Strengths:**
- Exception safety: `RebuildBucketGraphs()` saves old graphs before clearing and restores on failure
- `PairwiseBucketDistanceProvider` cleanly separates distance computation from graph construction
- `ComputePairVectors` with stack buffer optimization (256-byte threshold) avoids heap allocation in the common case
- Tests cover the happy path, disabled threshold rejection, and basic search validation

**Issues already raised by reviewers (no need to repeat):**
- `data_type_` / `has_raw_vector_` guard needed before graph construction for non-FP32 quantizers
- Doc comment still references the old dataset parameter; needs updating to match parameterless signature
- `graph_build_threshold == 0` in docs should be `<= 0` to match implementation
- Error message references internal `RebuildBucketGraphs` instead of public `RebuildIVFBucketGraphs`
- Concurrency: `bucket_graphs_` mutation without synchronization during concurrent search
- `ComputePairVectors` lacks bucket mutex lock and `EMPTY_INNER_ID` check
- Test should verify correctness (recall) not just success
- `BatchQueryDistance` per-call buffer allocation overhead

Overall the PR is well-structured and the existing review comments comprehensively address the remaining concerns. No additional issues found beyond what has already been raised.

src/datacell/bucket_datacell.h
```diff
@@ -672,4 +675,39 @@ BucketDataCell<QuantTmpl, IOTmpl>::GetCodesById(BucketIdType bucket_id,
...
+        throw VsagException(
+            ErrorType::INTERNAL_ERROR,
+            fmt::format("ComputePairVectors failed: bucket id ({}) is invalid", bucket_id));
+    }
+    if (id1 >= this->bucket_sizes_[bucket_id] || id2 >= this->bucket_sizes_[bucket_id]) {
+        throw VsagException(ErrorType::INTERNAL_ERROR,
+                            fmt::format("ComputePairVectors failed: offset id is invalid"));
+    }
+    constexpr size_t kStackThreshold = 256;
+    uint8_t stack_buf1[kStackThreshold];
```
LHT129, line 700: [suggestion] `ComputePairVectors` performs two `Read` calls on `datas_[bucket_id]` without holding the per-bucket `shared_lock` that `query_one_by_id` (line 252) and other methods in this class use. If `RebuildBucketGraphs` is called concurrently with `Add` or `Remove` on the same bucket, this can lead to a data race on the underlying IO buffer.

Consider adding a `std::shared_lock lock(this->bucket_mutexes_[bucket_id]);` at the top of `ComputePairVectors` to match the locking pattern used by `query_one_by_id` and `scan_bucket_by_id`.

src/algorithm/ivf/ivf.cpp
```diff
@@ -511,7 +511,7 @@ IVF::Build(const DatasetPtr& base) {
     // TODO(LHT): duplicate
     auto result = this->Add(base);
     if (graph_build_threshold_ > 0) {
-        this->build_bucket_graphs(base);
+        this->build_bucket_graphs();
     }
     this->cal_memory_usage();
```
LHT129, line 516: [suggestion] `get_ivf_graph_build_data` (line 624) is no longer called after the switch to `PairwiseBucketDistanceProvider` which uses `ComputePairVectors` for bucket-internal distance computation. This function and its callers (`BucketDistanceProvider` which was also removed) are now dead code.

Consider removing `get_ivf_graph_build_data` to keep the codebase clean, or add a comment if it is intentionally kept for future use.

src/algorithm/ivf/ivf.cpp
```diff
@@ -687,14 +687,53 @@ has_any_fresh_bucket_graph(const BucketInterfacePtr& bucket,
     return false;
 }
 
+class PairwiseBucketDistanceProvider final : public DistanceProviderForGraph {
+public:
+    PairwiseBucketDistanceProvider(std::shared_ptr<BucketInterface> bucket,
+                                   BucketIdType bucket_id,
+                                   InnerIdType query_id)
+        : bucket_(std::move(bucket)), bucket_id_(bucket_id), query_id_(query_id) {
```
LHT129, line 671: [note] Member declaration order (`bucket_`, `query_id_`, `bucket_id_`) does not match the constructor initializer list order (`bucket_`, `bucket_id_`, `query_id_`). C++ initializes members in declaration order regardless of the initializer list order, so this is not a bug, but it can trigger `-Wreorder` compiler warnings. Consider reordering either the member declarations or the initializer list to match.

Current declaration order:
```cpp
std::shared_ptr<BucketInterface> bucket_;
InnerIdType query_id_;
BucketIdType bucket_id_;
```

Current initializer list:
```cpp
: bucket_(std::move(bucket)), bucket_id_(bucket_id), query_id_(query_id)
```

Suggestion: swap `query_id_` and `bucket_id_` in the member declarations, or swap them in the initializer list.

LHT129 (commented):
[suggestion] The static function `get_ivf_graph_build_data` (ivf.cpp lines 624-651) is no longer called anywhere after the refactor to `PairwiseBucketDistanceProvider`. It was previously used by `build_bucket_graphs(const DatasetPtr&)` to extract raw vectors from the dataset, but the new implementation uses `ComputePairVectors` instead. This is dead code and should be removed to avoid confusion and reduce maintenance burden.

src/algorithm/ivf/ivf.cpp
```diff
@@ -2401,4 +2422,29 @@ IVF::GetMemoryUsage() const {
...
 }
 
+void
+IVF::RebuildBucketGraphs() {
+    if (graph_build_threshold_ <= 0) {
+        throw VsagException(
+            ErrorType::UNSUPPORTED_INDEX_OPERATION,
+            "RebuildIVFBucketGraphs: index was not configured with graph_build_threshold > 0");
+    }
+    if (common_param_.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
```
LHT129, line 2433: [suggestion] `RebuildBucketGraphs()` restricts to `DATA_TYPE_FLOAT` only, but `build_bucket_graphs()` (called from both `Build()` and `RebuildBucketGraphs()`) no longer has any data type restriction — it works with any data type via `ComputePairVectors` which operates on encoded codes. This creates an inconsistency: `Build()` can construct bucket graphs for any data type, but `RebuildBucketGraphs()` refuses to rebuild them for non-FP32 indices.

If the intent is to restrict rebuild to FP32 only, the `Build()` path should also be guarded. Otherwise, the data type check in `RebuildBucketGraphs()` should be removed to match `build_bucket_graphs()` behavior.

If FP32-only is intentional, add the same guard in `Build()` before calling `build_bucket_graphs()`. If not, remove this check.

src/algorithm/ivf/ivf.cpp
```diff
@@ -687,14 +663,53 @@ has_any_fresh_bucket_graph(const BucketInterfacePtr& bucket,
...
+    }
+
+    [[nodiscard]] ComputerInterfacePtr
+    FactoryComputerById(InnerIdType id) const override {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "PairwiseBucketDistanceProvider does not support FactoryComputerById");
+    }
+
+private:
+    std::shared_ptr<BucketInterface> bucket_;
```
LHT129, line 703: [note] Member declaration order (`bucket_`, `query_id_`, `bucket_id_`) does not match the constructor initializer list order (`bucket_`, `bucket_id_`, `query_id_`). C++ initializes members in declaration order regardless of the initializer list order, so this is not a bug, but it can trigger `-Wreorder` compiler warnings.

Current declaration order (line 703-705):
```cpp
std::shared_ptr<BucketInterface> bucket_;
InnerIdType query_id_;
BucketIdType bucket_id_;
```

Current initializer list (line 671):
```cpp
: bucket_(std::move(bucket)), bucket_id_(bucket_id), query_id_(query_id)
```

Suggestion: swap `query_id_` and `bucket_id_` in the member declarations to match the initializer list order.

include/vsag/index.h
```diff
@@ -1092,6 +1092,27 @@ class Index {
...
+     * get internal graphs for faster search. After Add() operations, these graphs
+     * become stale. This method explicitly rebuilds them using bucket-internal
+     * distance computation (no external dataset needed).
+     *
+     * Requirements:
+     * - Index must be IVF with graph_build_threshold > 0
+     * - Not safe for concurrent use with Add/Remove/Build/Serialize
+     *
+     * @return tl::expected<void, Error>; an Error is returned if the index does not
+     * support this operation or if graph_build_threshold was not configured.
```
LHT129, line 1108: [note] The doc comment says "graph_build_threshold == 0" but the implementation in `RebuildBucketGraphs()` checks `graph_build_threshold_ <= 0`. The `<=` guard is correct (a negative threshold would also be invalid), so the doc should be updated to match:

```cpp
 * construction is not enabled (graph_build_threshold <= 0).
```

LHT129 (commented):
## Review Summary

This iteration removes the external dataset requirement from `RebuildBucketGraphs()`, replacing the old `BucketDistanceProvider` (which needed raw vectors) with `PairwiseBucketDistanceProvider` that computes distances from bucket-internal encoded codes via the new `ComputePairVectors()` method. This is a clean architectural improvement.

### Key observations:

1. **`build_bucket_graphs()` now works for any data type** — the old `DATA_TYPE_FLOAT` guard was removed, and `ComputePairVectors` operates on encoded codes regardless of quantizer type. This means `Build()` can now construct bucket graphs for non-FP32 indices too.

2. **`RebuildBucketGraphs()` still rejects non-FP32** — this creates an inconsistency where `Build()` succeeds but `RebuildBucketGraphs()` throws for the same index. Either the FP32 guard should be removed from `RebuildBucketGraphs()`, or added back to `build_bucket_graphs()`.

3. **Member declaration order** in `PairwiseBucketDistanceProvider` (`bucket_`, `query_id_`, `bucket_id_`) does not match the constructor initializer list order (`bucket_`, `bucket_id_`, `query_id_`). This triggers `-Wreorder` warnings.

4. **Doc comment** in `index.h` still references the old API contract (requiring external dataset). The comment should be updated to reflect that no external dataset is needed — distances are computed from bucket-internal encoded codes.

The test coverage is adequate for the happy path and rejection case. Consider adding a test for the non-FP32 rejection path once the FP32 restriction decision is finalized.

include/vsag/index.h
```diff
@@ -1092,6 +1092,27 @@ class Index {
             Error(ErrorType::UNSUPPORTED_INDEX_OPERATION, "Index does not support ImportCache"));
     }
 
+    /**
+     * @brief Rebuild per-bucket graphs for IVF indices after Add operations.
+     *
```
LHT129, line 1097: [suggestion] The doc comment still describes the old API contract that required an external dataset passed by the caller. The current implementation no longer takes a dataset parameter — distances are computed from bucket-internal encoded codes via `ComputePairVectors()`. The comment should be updated to reflect this.

Current text:
```
 * The caller must supply the complete set of raw float32 vectors that
 * were originally passed to Build() and any subsequent Add() calls, in
 * insertion order.  The dataset must contain exactly @c GetNumElements()
 * rows (including any logically deleted vectors) and match the index
 * dimension.  Stale bucket graphs are discarded and rebuilt from scratch.
```

Suggested replacement:
```
 * This method rebuilds bucket graphs using bucket-internal distance
 * computation (no external dataset needed). Stale bucket graphs are
 * discarded and rebuilt from scratch.
```

LHT129 (commented):
## Code Review for PR #2676

### [critical] TOCTOU race in `ComputePairVectors` (`src/datacell/bucket_datacell.h:699`)

`ComputePairVectors` reads `bucket_sizes_` and `inner_ids_` **before** acquiring the `shared_lock`, which is inconsistent with other read paths like `query_one_by_id` (line 252) that take the lock first. This creates a TOCTOU race: a concurrent writer (e.g., `Add`) could resize the bucket or modify `inner_ids_` between the bounds check and the lock acquisition, leading to out-of-bounds access or operating on stale data.

Move the `shared_lock` acquisition before the bounds checks:

```cpp
std::shared_lock lock(this->bucket_mutexes_[bucket_id]);
if (bucket_id >= this->bucket_count_) { ... }
if (id1 >= this->bucket_sizes_[bucket_id] || id2 >= this->bucket_sizes_[bucket_id]) { ... }
if (this->inner_ids_[bucket_id][id1] == EMPTY_INNER_ID || ...) { ... }
```

### [suggestion] Member initializer order in `PairwiseBucketDistanceProvider` (`src/algorithm/ivf/ivf.cpp:670`)

The member declaration order is `bucket_`, `query_id_`, `bucket_id_`, but the initializer list orders them as `bucket_`, `bucket_id_`, `query_id_`. While functionally harmless here (no inter-member dependencies), this triggers `-Wreorder` warnings. Match the initializer order to the declaration order:

```cpp
: bucket_(std::move(bucket)), query_id_(query_id), bucket_id_(bucket_id)
```

wxyucs (approved):
lgtm

LHT129 (commented):
## Review Summary

This commit (e4db9ffa) makes significant improvements over the previous iteration:

**Improvements in this commit:**
- Removed the now-unused `get_ivf_graph_build_data` helper function
- Replaced `BucketDistanceProvider` with `PairwiseBucketDistanceProvider` that uses bucket-internal `ComputePairVectors` for distance computation, eliminating the need for an external dataset
- Added proper `std::shared_lock` in `ComputePairVectors` for thread-safe reads
- Added `EMPTY_INNER_ID` (deleted vector) validation in `ComputePairVectors`
- Added float32 data type validation in `RebuildBucketGraphs()`
- Improved exception safety in `RebuildBucketGraphs()` using move semantics instead of `std::fill`
- Updated error messages to reference the public API name
- Parallelized `build_bucket_graphs()` via thread pool for better performance
- Updated the doc comment to reflect the parameterless API

**Remaining minor issues (already noted in existing inline comments):**
- Doc comment says `graph_build_threshold == 0` but implementation uses `<= 0`
- `PairwiseBucketDistanceProvider` member declaration order does not match initializer list order
- `ComputePairVectors` reads `bucket_sizes_`/`inner_ids_` before acquiring the shared lock
- PR description is stale: mentions 3 test cases (only 2 present) and describes the old dataset-based API contract

The core implementation is solid. The switch to bucket-internal distance computation is the right design choice for a parameterless rebuild API.

tests/test_ivf.cpp
```diff
@@ -2272,6 +2272,55 @@ TEST_CASE("IVF GraphBucketSearcher Add After Build Falls Back To Flat", "[ft][iv
     REQUIRE(result.value()->GetIds()[0] == dataset->base_->GetIds()[base_count - 1]);
 }
 
+TEST_CASE("IVF RebuildBucketGraphs Restores Graph After Add", "[ft][ivf][graph]") {
```
LHT129, line 2275: [note] The PR description mentions 3 test cases but only 2 are present. The count mismatch test was likely removed when the API changed to parameterless form. Please update the PR description.

include/vsag/index.h
LHT129, line 1097: [note] The PR description still describes the old API contract: "The caller must supply the complete set of raw float32 vectors that were originally passed to Build()... The dataset must contain exactly GetNumElements() rows." The current API is parameterless and uses bucket-internal distance computation. Please update the PR description to reflect the current implementation.

PR #2679, opened 2026-08-13
feat(build): qualify third-party overrides by dependency pin

### LINKED ISSUE #2677: [feat](build): qualify third-party override variables by dependency pin
### Affected area

build

### Motivation

Every CMake third-party download override is currently keyed only by dependency name, for
example `VSAG_THIRDPARTY_FMT`. A machine or CI image therefore cannot safely export archives
for multiple VSAG release lines at the same time: after switching branches, the same variable
can point to an archive for a different dependency pin and fail hash verification or select the
wrong candidate.

The goal is to let one environment export a persistent set of branch-compatible overrides, such
as `VSAG_THIRDPARTY_FMT_10_2_1`, so switching among VSAG 0.16, 0.17, 0.18, 1.0, and main selects
the archive matching that branch's own pinned identifier. A pinned identifier may be a release
version, tag, or commit hash.

The current main inventory has 16 external download overrides; the 0.x branches have 15 because
they do not contain the httplib override. These are environment variables read by dependency
download CMake files, not internal CMake variables such as `VSAG_THIRDPARTY_C_FLAGS`.

### Proposed solution

Add a **pinned variable** for every discovered `ENV{VSAG_THIRDPARTY_<DEPENDENCY>}` download
override:

```
VSAG_THIRDPARTY_<DEPENDENCY>_<PIN_SUFFIX>
```

Derive `PIN_SUFFIX` from the exact identifier pinned by the branch being built:

1. For numeric releases, strip one leading `v`/`V` and any dependency-name decoration directly
   before the numeric part, then uppercase ASCII and replace non-alphanumeric runs with one
   underscore. For example, `10.2.1` and `v10.2.1` become `10_2_1`, `hdf5_1.14.4` becomes
   `1_14_4`, and `yaml-cpp-0.9.0` becomes `0_9_0`.
2. For other tags or refs, use `TAG_<NORMALIZED>_H<12_HEX>`. `NORMALIZED` applies the same
   uppercase and separator rule, and `12_HEX` is the first 12 uppercase hexadecimal characters
   of SHA-256 over the exact, case-sensitive pinned identifier. This keeps prerelease, build,
   date, and unusual tag spellings deterministic and distinguishable.
3. For a full commit hash, use `COMMIT_<12_HEX>`, where `12_HEX` is the first 12 uppercase
   hexadecimal characters of the hash. For example, the verified main CPUINFO pin produces
   `COMMIT_CA678952A9A8`.
4. Check generated names for different pins of the same dependency across the five required
   branches. If a 12-character hash prefix collides, extend both colliding prefixes one character
   at a time until they are unique, up to the full hash or digest. Fail configuration rather than
   choose an ambiguous name if uniqueness cannot be established.

Examples from main:

| Existing variable | Main pin | Pinned variable | Current source |
| --- | --- | --- | --- |
| `VSAG_THIRDPARTY_ANTLR4` | 4.13.2 | `VSAG_THIRDPARTY_ANTLR4_4_13_2` | `extern/antlr4/antlr4.cmake:17`, `extern/antlr4/antlr4.cmake:22` |
| `VSAG_THIRDPARTY_ARGPARSE` | v3.1 | `VSAG_THIRDPARTY_ARGPARSE_3_1` | `extern/argparse/argparse.cmake:4`, `extern/argparse/argparse.cmake:9` |
| `VSAG_THIRDPARTY_BOOST` | 1.67.0 | `VSAG_THIRDPARTY_BOOST_1_67_0` | `extern/boost/boost.cmake:16`, `extern/boost/boost.cmake:21` |
| `VSAG_THIRDPARTY_CATCH2` | v3.7.1 | `VSAG_THIRDPARTY_CATCH2_3_7_1` | `extern/catch2/catch2.cmake:3`, `extern/catch2/catch2.cmake:8` |
| `VSAG_THIRDPARTY_CPUINFO` | ca678952a9a8eaa6de112d154e8e104b22f9ab3f | `VSAG_THIRDPARTY_CPUINFO_COMMIT_CA678952A9A8` | `extern/cpuinfo/cpuinfo.cmake:4`, `extern/cpuinfo/cpuinfo.cmake:9` |
| `VSAG_THIRDPARTY_FMT` | 10.2.1 | `VSAG_THIRDPARTY_FMT_10_2_1` | `extern/fmt/fmt.cmake:70`, `extern/fmt/fmt.cmake:75` |
| `VSAG_THIRDPARTY_HDF5` | hdf5_1.14.4 | `VSAG_THIRDPARTY_HDF5_1_14_4` | `extern/hdf5/hdf5.cmake:11`, `extern/hdf5/hdf5.cmake:16` |
| `VSAG_THIRDPARTY_HTTPLIB` | v0.35.0 | `VSAG_THIRDPARTY_HTTPLIB_0_35_0` | `extern/httplib/httplib.cmake:20`, `extern/httplib/httplib.cmake:24` |
| `VSAG_THIRDPARTY_JSON` | v3.11.3 | `VSAG_THIRDPARTY_JSON_3_11_3` | `extern/json/json.cmake:5`, `extern/json/json.cmake:10` |
| `VSAG_THIRDPARTY_OPENBLAS` | v0.3.24 | `VSAG_THIRDPARTY_OPENBLAS_0_3_24` | `extern/openblas/openblas.cmake:175`, `extern/openblas/openblas.cmake:178` |
| `VSAG_THIRDPARTY_PYBIND11` | v2.11.1 | `VSAG_THIRDPARTY_PYBIND11_2_11_1` | `extern/pybind11/pybind11.cmake:4`, `extern/pybind11/pybind11.cmake:9` |
| `VSAG_THIRDPARTY_ROARINGBITMAP` | v3.0.1 | `VSAG_THIRDPARTY_ROARINGBITMAP_3_0_1` | `extern/roaringbitmap/roaringbitmap.cmake:4`, `extern/roaringbitmap/roaringbitmap.cmake:9` |
| `VSAG_THIRDPARTY_TABULATE` | 3a58301067bbc03da89ae5a51b3e05b7da719d38 | `VSAG_THIRDPARTY_TABULATE_COMMIT_3A58301067BB` | `extern/tabulate/tabulate.cmake:4`, `extern/tabulate/tabulate.cmake:9` |
| `VSAG_THIRDPARTY_THREAD_POOL` | 3507796e172d36555b47d6191f170823d9f6b12c | `VSAG_THIRDPARTY_THREAD_POOL_COMMIT_3507796E172D` | `extern/thread_pool/thread_pool.cmake:5`, `extern/thread_pool/thread_pool.cmake:10` |
| `VSAG_THIRDPARTY_TSL` | v1.4.0 | `VSAG_THIRDPARTY_TSL_1_4_0` | `extern/tsl/tsl.cmake:5`, `extern/tsl/tsl.cmake:8` |
| `VSAG_THIRDPARTY_YAML_CPP` | yaml-cpp-0.9.0 | `VSAG_THIRDPARTY_YAML_CPP_0_9_0` | `extern/yaml-cpp/yaml-cpp.cmake:8`, `extern/yaml-cpp/yaml-cpp.cmake:13` |

#### Lookup precedence and compatibility

For each dependency, use this order:

1. The pinned variable for the branch's pinned identifier.
2. The existing unversioned `VSAG_THIRDPARTY_<DEPENDENCY>` variable as a temporary compatibility
   fallback.
3. The existing upstream/mirror URL list.

If both pinned and legacy variables are set, the pinned variable wins. A pinned variable for any
other identifier is ignored. Preserve the existing `URL_HASH` check for every selected archive.

Keep the legacy fallback in all requested backports so existing offline-build environments do
not break. When it is used, emit a CMake deprecation warning naming the expected pinned variable.
Do not remove the fallback in these backports; removal from a future main release requires a
separate compatibility decision and release-note documentation.

Diagnostics must identify the dependency, pinned identifier, selected variable name, and whether
the pinned, legacy, or default source won. Do not print the variable value because override URLs
may contain credentials.

#### Remove Alibaba Cloud OSS caches

Remove every Alibaba Cloud OSS cache or mirror URL for third-party dependency archives. Default
downloads must use only the authoritative upstream URL(s); users can continue to supply local or
offline mirrors through the pinned variable, or temporarily through the deprecated unversioned
legacy variable. Apply this requirement independently on `main`, `1.0`, `0.18`, `0.17`, and
`0.16`, using each branch's own third-party URL inventory rather than copying another branch's
list.

#### Required branches and branch-owned pins

Implement independently on each required branch using that branch's own dependency pins:

| Branch | Ref inspected | Override inventory | Pin differences relevant to this request |
| --- | --- | --- | --- |
| `main` | `2b6cda30611eba576a6d2f213e46fb1e4ea5dcc9` | 16 variables | httplib present; OpenBLAS 0.3.24; yaml-cpp 0.9.0 |
| `1.0` | `93fdbb2a81dc8bc81eda5c9e315a848200681b58` | 16 variables | httplib present; OpenBLAS 0.3.23; yaml-cpp 0.9.0 |
| `0.18` | `0c6456a3c58bfb157459cba0b0428f5805760f41` | 15 variables | no httplib; OpenBLAS 0.3.23; yaml-cpp 0.8.0 |
| `0.17` | `f9660693776f083fbb0602a29db14d246b91d10a` | 15 variables | no httplib; OpenBLAS 0.3.23; yaml-cpp 0.8.0 |
| `0.16` | `e62058be44a69e79f7c1b0c0ae6a5ee3f5926203` | 15 variables | no httplib; OpenBLAS 0.3.23; yaml-cpp 0.8.0 |

Each branch must inventory its own dependency files and derive suffixes from its own pinned
identifiers. Do not copy main's pins into a backport. In particular, main expects
`VSAG_THIRDPARTY_OPENBLAS_0_3_24`, while 1.0 and all three 0.x branches expect
`VSAG_THIRDPARTY_OPENBLAS_0_3_23`; main/1.0 expect `VSAG_THIRDPARTY_YAML_CPP_0_9_0`, while
0.16–0.18 expect `VSAG_THIRDPARTY_YAML_CPP_0_8_0`.

### Alternatives considered

- Keep one unversioned variable per dependency and rewrite it when switching branches. This does
  not meet the persistent multi-release CI-image use case and is easy to get wrong.
- Prefix variables with the VSAG release line instead of the dependency pin. This duplicates
  identical archives across release lines and becomes stale when a dependency pin changes.
- Remove the unversioned variables immediately. This would unnecessarily break existing offline
  and mirrored builds, especially in stable release backports.
- Use only CMake's download cache. This is less explicit, is not a stable user-facing mapping,
  and does not support centrally managed archive/path overrides as directly.

none

This is a build-time environment-variable extension. Public C++, Python, and Node.js APIs and ABI
are unchanged. The legacy fallback preserves current build behavior.

### References

- Current main override and pin inventory: the 16 `extern/*/*.cmake` path:line pairs in the table
  above.
- OpenBLAS main pin and override: `extern/openblas/openblas.cmake:175`,
  `extern/openblas/openblas.cmake:178`.
- yaml-cpp main pin and override: `extern/yaml-cpp/yaml-cpp.cmake:8`,
  `extern/yaml-cpp/yaml-cpp.cmake:13`.
- httplib exists on main/1.0: `extern/httplib/httplib.cmake:20`,
  `extern/httplib/httplib.cmake:24`.
- Prior backport of unversioned overrides: #2308.
- Broader third-party dependency roadmap: #1949.

#### Acceptance criteria

- [ ] All 16 variables discovered on main/1.0 and all 15 variables discovered on each 0.x branch
  implement pinned lookup; internal non-download variables such as `VSAG_THIRDPARTY_C_FLAGS` are
  not treated as archive overrides.
- [ ] A checked-in CMake helper or equivalent single implementation defines suffix generation and
  lookup precedence consistently within each branch. Tests cover numeric releases, leading `v`,
  dependency-decorated tags, other tags/refs, 12-character commit suffixes, and deterministic
  collision extension or rejection.
- [ ] With pinned variables for multiple identifiers exported simultaneously, configuring after
  switching branches selects only the archive matching that branch's pin. Verify at least the
  OpenBLAS 0.3.24/0.3.23 and yaml-cpp 0.9.0/0.8.0 branch differences.
- [ ] When both pinned and legacy variables are set to different valid candidates, the pinned
  variable is selected and the diagnostic reports that precedence without exposing the value.
- [ ] When only a legacy variable is set, it remains usable, retains hash verification, and emits
  a deprecation warning naming the pinned variable expected by that branch.
- [ ] Negative/mismatch coverage proves that a pinned variable for another identifier is ignored
  and that a pinned variable pointing to an archive with the wrong content fails the existing hash
  check rather than silently building it.
- [ ] Configure-time diagnostics report dependency, pin, selected variable name/source category,
  and fallback behavior without printing credentials embedded in URLs.
- [ ] The branch-owned third-party URL inventory contains no Alibaba Cloud OSS cache or mirror
  URLs and retains authoritative upstream URLs plus pinned and deprecated legacy overrides.
- [ ] English and Chinese canonical user documentation under `docs/docs/{en,zh}/src/` explains
  suffix generation, precedence, compatibility status, and branch-owned pins.
- [ ] Documentation includes a copy-pasteable environment example exporting both
  `VSAG_THIRDPARTY_OPENBLAS_0_3_23` and `VSAG_THIRDPARTY_OPENBLAS_0_3_24` (plus yaml-cpp
  0.8.0/0.9.0) and demonstrates switching branches without editing exports.
- [ ] Configure/build verification passes independently on `0.16`, `0.17`, `0.18`, `1.0`, and
  `main`, using each branch's own complete override inventory and pinned identifiers.

## Related issues

Focused searches across open and closed issues for “third-party override environment variables”,
“VSAG_THIRDPARTY versioned CMake”, “dependency archive overrides release branches”, and “version
third party download environment variable” found no duplicate.

- #2308 — [feat](build): backport VSAG_THIRDPARTY_* download overrides to 0.15-0.18 release lines
  (closed). This established the unversioned override baseline but did not qualify variables by
  dependency pin.
- #1949 — Roadmap: reduce bundled third-party dependencies in core VSAG (open). This is related
  build/dependency context, not a duplicate of pin-specific overrides.

---

wxyucs:
## What changed

- resolve each third-party archive override through a centralized pinned-variable helper
- prefer pin-qualified variables while retaining deprecated unversioned fallbacks
- remove VSAG-maintained Alibaba Cloud OSS archive mirrors and keep authoritative upstream URLs
- add focused CMake inventory tests and synchronized English/Chinese offline-build documentation

## Validation

- `make test-cmake`
- `make fmt` (clang-format 15)
- `git diff --check`
- representative configure with pinned and legacy variables set together
- authoritative ANTLR4 upstream download and hash verification with overrides unset
- `make debug`

Fixes: #2677


LHT129 (commented):
test

cmake/VSAGThirdPartyOverride.cmake
```diff
@@ -0,0 +1,185 @@
...
+
+    _vsag_normalize_thirdparty_component ("${dependency}" dependency_component)
+    if (NOT dependency_component)
+        message (FATAL_ERROR "A pinned variable requires a non-empty dependency name")
+    endif ()
+    if ("${pin}" STREQUAL "")
+        message (FATAL_ERROR "A pinned variable requires a non-empty pin")
+    endif ()
+
+    _vsag_classify_thirdparty_pin (
```
LHT129, line 93: [suggestion] The version-detection regex `([0-9]+(\.[0-9]+)+)$` requires at least one dot, so a single-component numeric pin (e.g. `"10"`) would not be classified as VERSION and would fall through to TAG/SHA-256 hashing. This is probably intentional since VSAG pins are always multi-component, but a brief comment noting this constraint would help future maintainers.

Additionally, the regex captures the *last* dotted sequence. For a hypothetical pin like `libfoo-1.2.3-beta.4`, it would match only `4`. While VSAG controls all pin values, the docs could clarify that pins should end with the version number.

cmake/VSAGThirdPartyOverride.cmake
```diff
@@ -0,0 +1,185 @@
...
+    endif ()
+    if (NOT hash_length MATCHES "^[0-9]+$")
+        message (FATAL_ERROR "Pinned-variable hash length must be an integer")
+    endif ()
+
+    _vsag_normalize_thirdparty_component ("${dependency}" dependency_component)
+    if (NOT dependency_component)
+        message (FATAL_ERROR "A pinned variable requires a non-empty dependency name")
+    endif ()
+    if ("${pin}" STREQUAL "")
```
LHT129, line 89: [note] The `_vsag_classify_thirdparty_pin` function takes a `dependency` parameter on line 89 but only uses it to compute `dependency_component` for the VERSION-branch comparison on line 105 (`prefix_component STREQUAL dependency_component`). In the COMMIT and TAG branches (lines 114-128), `dependency` is unused — the classification is based solely on the pin string. This is fine for the current use case, but it means the function signature implies a stronger coupling to the dependency name than actually exists. Consider whether the `dependency` parameter should be removed from this function and the VERSION comparison moved to the caller, or alternatively add a comment clarifying that `dependency` is only used for version-prefix disambiguation.

cmake/VSAGThirdPartyOverride.cmake
```diff
@@ -0,0 +1,185 @@
...
+    if (ARG_UNPARSED_ARGUMENTS)
+        message (FATAL_ERROR
+                 "Unexpected third-party override arguments: ${ARG_UNPARSED_ARGUMENTS}")
+    endif ()
+
+    _vsag_normalize_thirdparty_component ("${dependency}" dependency_component)
+    set (legacy_variable "VSAG_THIRDPARTY_${dependency_component}")
+    vsag_thirdparty_pinned_variable (
+        "${dependency}" "${pin}" pinned_variable COLLISION_PINS ${ARG_COLLISION_PINS})
+    set (resolved_urls "${${urls_variable}}")
```
LHT129, line 161: [note] The `vsag_resolve_thirdparty_override` function dereferences `${urls_variable}` with `set(resolved_urls "${${urls_variable}}")`. If the caller misspells the URL variable name, CMake will silently expand to an empty string, and the function will only have the override URL without any fallback URLs. This would cause downloads to fail with a confusing error. Consider adding a guard like `if(NOT DEFINED ${urls_variable})` with `message(FATAL_ERROR ...)` to catch typos at configure time.

cmake/VSAGThirdPartyOverride.cmake
```diff
@@ -0,0 +1,185 @@
...
+                    math (EXPR hash_length "${hash_length} + 1")
+                else ()
+                    message (FATAL_ERROR
+                             "Pins '${pin}' and '${collision_pin}' cannot have unique pinned "
+                             "variables for dependency '${dependency}'")
+                endif ()
+            endwhile ()
+        endforeach ()
+
+        string (SUBSTRING "${pin_digest}" 0 ${hash_length} pin_prefix)
```
LHT129, line 133: [suggestion] The `hash_length` variable is mutated inside the collision-resolution `while` loop. After the loop exits, the extended `hash_length` is used for the final `pin_prefix` extraction. This means if collision resolution extended the hash from 12 to 13, the final variable name uses 13 hex characters. This is correct behavior, but the variable name `hash_length` is slightly misleading since it no longer represents the configured/default hash length. Consider using a separate local variable like `effective_hash_length` after the collision loop, or adding a brief comment.

LHT129 (commented):
Overall this is a well-designed PR that cleanly centralizes the third-party override logic. The pin-qualified variable naming scheme is a genuine improvement for multi-branch CI environments. The CMake helper tests are thorough and cover the key scenarios (version/tag/commit classification, collision resolution, credential-safe diagnostics, legacy fallback).

A few observations from the inline comments:
1. The version-detection regex has some edge-case behavior worth documenting.
2. The `urls_variable` dereference could benefit from a definedness guard to catch typos at configure time.
3. The `hash_length` mutation pattern is correct but slightly confusing — a comment or separate variable would help readability.

The documentation updates (both English and Chinese) are comprehensive and well-synchronized. The `test-cmake` Makefile target is a nice touch for CI integration.

No blocking issues found.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -770,14 +620,254 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
...
+            precise_flatten = this->high_precise_codes_;
+        }
+        if (create_new_raw_vector_) {
+            precise_flatten = this->raw_vector_;
+        }
+        computer = precise_flatten->FactoryComputer(get_data(request.query_));
+    }
+    for (const auto& pair : label_to_inner_id) {
+        float dist = 0.0F;
+        const auto inner_id = pair.second;
```
LHT129: [critical] MCI eligibility detection uses request.filter_ (lines 703, 707) while the actual search and mci_result construction use ft from create_search_filter (lines 635, 700). When create_search_filter wraps or augments the raw filter (e.g. via use_extra_info_filter), the routing decision may be based on a different filter than the one used for search, potentially causing incorrect route selection and inconsistent statistics. The seed-source detection (has_valid_id_source, has_bitset_source) and the mci_result construction should use the same filter instance. Consider running seed-source detection on ft instead of request.filter_.

LHT129 (commented):
[note] Review of `cmake/VSAGThirdPartyOverride.cmake`:

**Line 33** (`_vsag_classify_thirdparty_pin`): The regex `([0-9]+(\\.[0-9]+)+)$` requires at least one dot, so single-component numeric strings (e.g. `20240101` date-based versions) fall through to TAG classification. This is fine for the current dependency set, but worth noting for future maintainers.

**Line 133** (`vsag_thirdparty_pinned_variable`): The `hash_length` variable is mutated inside the collision-resolution `while` loop. After the loop exits, the mutated value is used for the final `pin_prefix` extraction. This is intentional but could surprise readers — consider adding a brief comment noting that `hash_length` may be extended by the collision loop.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: Judged independently under the low-cost review policy: the current dependency pins are controlled inputs and use dotted versions; non-matching pins intentionally fall back to collision-resistant tag/commit naming. The hypothetical forms do not create a correctness, compatibility, security, data-loss, or significant maintainability risk here, so I am leaving the code and docs unchanged.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: Judged independently under the low-cost review policy: `dependency` is required for the version-prefix disambiguation performed by this classifier, while the other branches naturally do not need it. Keeping that input at the classification boundary is coherent and presents no concrete correctness or significant maintainability risk, so I am not renaming/refactoring or adding a comment.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: Judged independently under the low-cost review policy: `hash_length` is deliberately the active prefix length, including any collision-driven extension, and its final value must be used to form the variable name. This is correct and localized, with no concrete correctness or significant maintainability risk, so I am avoiding a polish-only rename/comment.

src/algorithm/hgraph/hgraph_search.cpp
wxyucs: This thread is not attached to code in the live PR diff. Its `original_commit_id` is d57e23a993bb59ba76680421e5fa33ec6235eb35 (`refactor(hgraph): centralize regular search routing`, parent c849a07b...), which is not an ancestor of the current PR head. The current PR has one commit (93e9e211) directly above merge-base/base 499146a82dea856b939ff0dd301531e025115134, and `git diff 499146a8...93e9e211 -- src/algorithm/hgraph/hgraph_search.cpp` is empty; GitHub's live PR file list also contains no HGraph path. This is therefore an obsolete/misattributed review attachment, not Issue #2677 scope, and no HGraph code is changed here.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: This is a reachable maintainability/correctness issue: a typo in any future call silently removes every default mirror and defers failure to FetchContent/download handling. I added a narrow `DEFINED` guard with a focused negative CMake fixture assertion in commit 93e9e211; `make test-cmake`, `make fmt` with clang-format 15, and `git diff --check` pass.

LHT129, line 161: [note] The response to the earlier review thread states that a `DEFINED` guard was added to `vsag_resolve_thirdparty_override` to catch typos in the `urls_variable` argument. However, the current code at line 158 still reads:

```cmake
set (resolved_urls "${${urls_variable}}")
```

without any preceding `if(NOT DEFINED ${urls_variable})` guard. The negative CMake fixture assertion mentioned in the response also does not appear in `tests/cmake/thirdparty_override_test.cmake`. If the intent is to address this, the guard and test still need to be added; if the intent changed, the earlier response should be updated to avoid confusion for future reviewers.

wxyucs: Correction: the final amended commit containing the guard and focused test is c97f734a (the earlier 93e9e211 reference was the immediately preceding amended head).

LHT129 (commented):
Reviewed commit c97f734a. This update addresses the previously flagged issues:

- Added `DEFINED` guard on `${urls_variable}` in `vsag_resolve_thirdparty_override` (line 217), preventing silent fallback to an empty URL list when the variable name is misspelled.
- Added `TEST_UNDEFINED_URLS` test case in `thirdparty_override_test.cmake` (lines 1188-1199) that verifies the guard correctly rejects undefined URL variables.

The remaining earlier review threads (version regex dot requirement, `hash_length` mutation in collision loop, `dependency` parameter usage) were already judged acceptable under the low-cost review policy. No new issues found in this commit.

The test coverage is thorough: pinned-variable name generation (version/commit/tag), collision prefix extension, pinned-vs-legacy precedence, credential-safe diagnostics, mismatched-pin isolation, and undefined-URL rejection are all verified.

cmake/VSAGThirdPartyOverride.cmake
```diff
@@ -0,0 +1,188 @@
...
+    set (${component_output} "${tag_component}" PARENT_SCOPE)
+    set (${digest_output} "${tag_digest}" PARENT_SCOPE)
+endfunction ()
+
+function (vsag_thirdparty_pinned_variable dependency pin output)
+    set (one_value_args HASH_LENGTH)
+    set (multi_value_args COLLISION_PINS)
+    cmake_parse_arguments (ARG "" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
+    if (ARG_UNPARSED_ARGUMENTS)
```
LHT129, line 72: [note] `string(SHA256 ...)` was introduced in CMake 3.19, but the project declares `cmake_minimum_required(VERSION 3.18)`. This would cause a configure-time error on CMake 3.18 if a TAG-type pin is ever used. In practice this path is not exercised by any current dependency (all pins are VERSION or COMMIT), and all CI/developer environments likely use CMake >= 3.19. Consider either bumping the minimum to 3.19 or guarding the SHA256 call with a CMake version check.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: The live head now contains the intended fix. In commit 9b72b113, `cmake/VSAGThirdPartyOverride.cmake` has the `if (NOT DEFINED ${urls_variable})` fail-fast guard at line 153, and `tests/cmake/thirdparty_override_test.cmake` has the focused `missing_urls` negative assertion at lines 136–147. `make test-cmake` passes. The omission was limited to an earlier intermediate amended head; thank you for catching the mismatch.

cmake/VSAGThirdPartyOverride.cmake
wxyucs: No compatibility change is needed: `string(SHA256 ...)` predates the project's CMake 3.18 baseline. The official CMake 3.18 documentation lists SHA256 under `string(<HASH> ...)`, and even the official CMake 3.0 documentation lists `MD5|SHA1|SHA224|SHA256|SHA384|SHA512`: https://cmake.org/cmake/help/v3.18/command/string.html#hashing and https://cmake.org/cmake/help/v3.0/command/string.html. The 3.19 addition was `string(JSON ...)`, not SHA256, so bumping or guarding would create unnecessary churn.

LHT129 (commented):
Overall this is a well-structured PR that cleanly centralizes third-party override logic. The CMake module is well-designed with clear separation of concerns between normalization, classification, pinned-variable generation, and override resolution. The test coverage is thorough, covering version/commit/tag pin types, collision resolution, credential-safe diagnostics, and legacy fallback behavior.

A few observations already noted in inline comments:
1. The `SHA256` call requires CMake 3.19 while the project declares 3.18 minimum — though this path is not exercised by any current dependency pin.
2. The `dependency` parameter in `_vsag_classify_thirdparty_pin` is only used for version-prefix disambiguation and is unused in COMMIT/TAG branches.
3. The version regex requires at least one dot, so single-component numeric pins would not be classified as VERSION.

None of these are blocking for the current set of dependencies. The removal of Alibaba Cloud OSS mirrors and the switch to authoritative upstream URLs is a good hygiene improvement.

tests/cmake/thirdparty_override_test.cmake
```diff
@@ -0,0 +1,169 @@
...
+assert_matches ("${output}" "TEST_SELECTION=pinned" "pinned URL precedence")
+assert_not_matches ("${output}" "password" "credential-safe diagnostics")
+
+run_fixture ("" "legacy-archive" output)
+assert_matches ("${output}" "source=legacy" "legacy fallback source")
+assert_matches ("${output}" "use pinned variable" "deprecation guidance")
+assert_matches ("${output}" "VSAG_THIRDPARTY_FMT_10_2_1" "expected pinned variable")
+assert_matches ("${output}" "TEST_SELECTION=legacy" "legacy URL precedence")
+
+set (ENV{VSAG_THIRDPARTY_FMT_9_0_0} "mismatched-archive")
```
LHT129, line 117: [note] The environment variable `VSAG_THIRDPARTY_FMT_9_0_0` set on line 117 is never cleaned up after the test assertion. While it does not affect the current test suite (subsequent tests use different variable names), leaving it set in the parent CMake process could cause confusion if additional tests are added later that iterate over `VSAG_THIRDPARTY_FMT_*` variables. Consider adding `unset(ENV{VSAG_THIRDPARTY_FMT_9_0_0})` after the assertion block (after line 121).

Similarly, `ENV{VSAG_THIRDPARTY_FMT}` set on line 123 for the credential-safe legacy test is not explicitly cleaned up after that test block completes.

wxyucs: Judged independently under the low-cost review policy: these variables are intentionally scoped to this one `cmake -P` process, and each fixture invocation explicitly unsets/re-establishes the pinned and legacy FMT variables before use. The later hash fixture sets the pinned variable, so the legacy value is ignored by the tested precedence. There is no current correctness or significant maintainability risk, and future tests that enumerate environment names would need their own isolation regardless, so I am avoiding cleanup-only churn.

LHT129 (approved):
LGTM

PR #2681, opened 2026-08-13
fix: support debug builds with GCC 15

wxyucs:
## What changed

- Compile bundled OpenBLAS as GNU C17 when using GCC 15 or newer.
- Keep OpenBLAS' legacy incompatible-pointer diagnostic as a warning under GCC 15.
- Add direct standard-library includes required by GCC 15/libstdc++ 15.
- Replace the attributed `std::fclose` function-pointer deleter with a named deleter.

## Why

GCC 15 defaults C compilation to GNU C23. In C23, an empty function parameter list means
that the function accepts no arguments, while OpenBLAS' f2c-generated LAPACK sources depend
on the pre-C23 unspecified-arguments behavior. GCC 15 also exposes stricter diagnostics and
header self-containment issues that prevented `make debug` from completing.

The OpenBLAS compatibility flags are scoped to GNU C compiler version 15 or newer, so older
GCC versions and Clang retain their existing behavior.

## Validation

- `CCACHE_DIR=/tmp/vsag-ccache make debug`
- `git diff --check`

The GCC 15.2 build completed both `vsag` and `vsag_static` targets successfully.

### DIFF
--- a/extern/openblas/openblas.cmake
+++ b/extern/openblas/openblas.cmake
@@ -184,6 +184,13 @@ if (NOT OPENBLAS_FOUND)
     # (which implies -ffast-math) breaks CPU detection. Replace with -O2.
     string (REPLACE "-Ofast" "-O2" _openblas_c_flags "${VSAG_THIRDPARTY_C_FLAGS}")
     string (REPLACE "-ffast-math" "" _openblas_c_flags "${_openblas_c_flags}")
+    # GCC 15 defaults to GNU C23, where an empty parameter list means that a function takes
+    # no arguments. OpenBLAS' f2c-generated LAPACK sources use empty parameter lists with the
+    # pre-C23 "unspecified arguments" meaning, so compile the bundled release as GNU C17.
+    if (CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 15)
+        string (APPEND _openblas_c_flags
+                " -std=gnu17 -Wno-error=incompatible-pointer-types")
+    endif ()
     string (REPLACE "-Ofast" "-O2" _openblas_cxx_flags "${VSAG_THIRDPARTY_CXX_FLAGS}")
     string (REPLACE "-ffast-math" "" _openblas_cxx_flags "${_openblas_cxx_flags}")
     string (STRIP "${_openblas_c_flags}" _openblas_c_flags)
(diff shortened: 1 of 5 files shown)


LHT129 (commented):
All changes look correct and well-scoped:

- **`extern/openblas/openblas.cmake`**: The GCC 15 detection is precise (`CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 15`), and the flags (`-std=gnu17 -Wno-error=incompatible-pointer-types`) are appropriate for the f2c-generated LAPACK sources. The comment clearly explains the rationale.

- **`include/vsag/attribute.h`**, **`src/attr/attribute.cpp`**, **`src/attr/expression.h`**: The added includes (`<cstdint>`, `<type_traits>`) are necessary for GCC 15/libstdc++ 15 which no longer provides these headers transitively. Each file directly uses the symbols from these headers (`uint64_t`, `std::is_same_v`), so the includes are justified.

- **`src/storage/tlv_section.cpp`**: Replacing `decltype(&std::fclose)` with a named `FileCloser` functor is the correct fix for GCC 15 stricter diagnostics on attributed function-pointer deleters. The `FileCloser` struct is stateless, trivial, and placed in the anonymous namespace — no ABI or linkage concerns.

No issues found. LGTM.

LHT129 (commented):
All changes look good. This is a clean, well-scoped fix for GCC 15 compatibility.

1. **`extern/openblas/openblas.cmake`** — Correctly scoped to GCC 15+ only. The `-std=gnu17` flag addresses the C23 empty-parameter-list semantic change in OpenBLAS f2c-generated LAPACK sources. The `-Wno-error=incompatible-pointer-types` keeps the legacy warning non-fatal. Both flags are appropriate and well-documented with inline comments.

2. **`include/vsag/attribute.h`** — Adding `#include <cstdint>` fixes header self-containment for GCC 15/libstdc++ 15. The file uses `uint64_t` etc. which require this header.

3. **`src/attr/attribute.cpp`** — Adding `#include <type_traits>` is needed for `std::is_same_v` and `std::decay_t` usage in this file.

4. **`src/attr/expression.h`** — Adding `#include <cstdint>` fixes header self-containment, same rationale as `attribute.h`.

5. **`src/storage/tlv_section.cpp`** — Replacing `decltype(&std::fclose)` with a named `FileCloser` functor is the right approach. GCC 15 attributes on `std::fclose` make taking its address as a function pointer problematic, and a stateless functor deleter is more idiomatic (no function pointer storage overhead in `unique_ptr`).

No issues found. All changes are minimal, correct, and properly scoped.

LHT129 (approved):
LGTM

PR #2683, opened 2026-08-14
feat(build): qualify third-party overrides for 0.18

wxyucs:
## What changed

- resolve all 15 VSAG 0.18 third-party archive overrides through a central pin-qualified helper

## Validation

- `make test-cmake`
- `cmake --build build-validation-http-0.18 --target fmt cpuinfo roaring --parallel 2`
- `git diff --check origin/0.18`
- zero-diff check for `git diff origin/0.18...HEAD -- docs/`

Backport of the approach merged in #2679 for the 0.18-owned dependency pins and URL inventory.

Fixes: #2677


extern/spdlog/spdlog.cmake
```diff
@@ -5,9 +5,6 @@ set(install_dir ${CMAKE_CURRENT_BINARY_DIR}/${name}/install)
 ExternalProject_Add(
     ${name}
     URL https://github.com/gabime/spdlog/archive/refs/tags/v1.12.0.tar.gz
-        # this url is maintained by the vsag project, if it's broken, please try
-        #  the latest commit or contact the vsag project
-        http://vsagcache.oss-rg-china-mainland.aliyuncs.com/spdlog/v1.12.0.tar.gz
     URL_HASH MD5=6b4446526264c1d1276105482adc18d1
```
LHT129, line 8: [suggestion] `spdlog.cmake` and `cpr.cmake` only had their OSS cache URLs removed but were not integrated into the `vsag_resolve_thirdparty_override` mechanism. This means users cannot override these two dependencies via the new pinned-variable system. If this is intentional (e.g., because they use `ExternalProject_Add` with a single `URL` argument rather than a URL list), consider documenting this exception in the offline build guide so users know which dependencies support overrides and which do not.

Currently the test asserts exactly 15 overrides, and spdlog/cpr are not among them — so this appears intentional but undocumented.

extern/cpr/fix_curl.patch
```diff
@@ -2,11 +2,10 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
 index 3f6e1d02..2e5a568f 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -297,8 +297,11 @@ else()
+@@ -297,8 +297,10 @@ else()
```
LHT129, line 5: [note] The `fix_curl.patch` change removes the Alibaba Cloud OSS cache URL line that was previously added in the patch hunk. The surrounding context line count changed from `@@ -297,8 +297,11 @@` to `@@ -297,8 +297,10 @@`, which is correct since one line was removed. However, the patch file itself is not covered by the `vsag_resolve_thirdparty_override` mechanism — the curl download inside cpr remains a hardcoded upstream URL with no user-overridable path. This is fine since curl is a transitive dependency of cpr, but worth noting for completeness.

LHT129 (commented):
Overall this is a well-structured PR that cleanly centralizes third-party archive override logic. The implementation is thorough with strong test coverage.

**Summary of findings:**

1. **[suggestion] spdlog and cpr not integrated into override mechanism** — `spdlog.cmake` and `cpr.cmake` only had OSS cache URLs removed but do not call `vsag_resolve_thirdparty_override`. This means users cannot override these two dependencies via the new pinned-variable system. The test correctly asserts exactly 15 overrides (excluding spdlog/cpr), so this appears intentional, but the offline build documentation should mention which dependencies are covered and which are not.

2. **[note] curl transitive dependency** — The `fix_curl.patch` change correctly removes the OSS URL, but curl (a transitive dependency of cpr) remains hardcoded with no override path. This is acceptable for a transitive dependency.

**Positive observations:**
- The `VSAGThirdPartyOverride.cmake` module is well-designed with clear separation between pin classification, variable name generation, and URL resolution
- Credential-safe diagnostics are properly implemented — the test verifies that URLs containing `user:password@` are never leaked into configure output
- Collision detection for commit/tag hash prefixes is correctly handled with automatic prefix extension
- The test suite is comprehensive: pin name generation, precedence (pinned > legacy > default), mismatched-pin rejection, credential safety, undefined URL rejection, and hash verification
- Both English and Chinese documentation are synchronized and thorough
- The `Makefile` `test-cmake` target provides a convenient way to run the CMake-level tests
- All 15 extern cmake files are consistently migrated to the new pattern

No blocking issues found.

LHT129 (commented):
Overall this is a well-structured PR that cleanly centralizes third-party archive override logic for the 0.18 branch. The new `VSAGThirdPartyOverride.cmake` module is well-designed with proper input validation, collision detection, and credential-safe diagnostics. The test coverage is thorough, covering pinned/legacy/default precedence, credential safety, hash verification, and inventory count assertions.

Two inline comments were left for consideration (spdlog/cpr override gap and fix_curl.patch context), neither blocking.

extern/spdlog/spdlog.cmake
wxyucs: Confirmed intentional. Issue #2677 scopes this release to the 15 pre-existing VSAG_THIRDPARTY_* overrides, and the guide’s exact 15-row inventory is the authoritative supported set. spdlog and CPR had no existing override to qualify; adding or documenting a new compatibility surface is outside this backport.

extern/cpr/fix_curl.patch
wxyucs: Agreed. Curl is a transitive CPR download and was not part of the branch’s 15 existing VSAG_THIRDPARTY_* inventory. This change removes its Alibaba OSS default while leaving any new transitive override design to a separate compatibility decision.

PR #2684, opened 2026-08-14
feat(build): qualify third-party overrides for 0.17

wxyucs:
## What changed

- resolve all 15 VSAG 0.17 third-party archive overrides through a central pin-qualified helper

## Validation

- `make test-cmake`
- `cmake --build build-validation-http-0.17 --target fmt cpuinfo roaring --parallel 2`
- `git diff --check origin/0.17`
- zero-diff check for `git diff origin/0.17...HEAD -- docs/`

Backport of the approach merged in #2679 for the 0.17-owned dependency pins and URL inventory.

Fixes: #2677


extern/spdlog/spdlog.cmake
```diff
@@ -5,9 +5,6 @@ set(install_dir ${CMAKE_CURRENT_BINARY_DIR}/${name}/install)
 ExternalProject_Add(
```
LHT129, line 5: [note] `spdlog.cmake` had its Alibaba Cloud OSS cache URL removed but did not gain a `vsag_resolve_thirdparty_override` call. This is because spdlog uses `ExternalProject_Add` with an inline `URL` (not a `set(urls ...)` list), so there is no `urls` variable to pass to the resolver. As a result, spdlog loses the ability for offline/mirror overrides that the other 15 dependencies now have.

This is likely intentional since spdlog has no version-specific pin to qualify, but it is worth confirming whether spdlog should also be brought under the override mechanism, perhaps by refactoring it to use a `set(spdlog_urls ...)` pattern like the others.

extern/cpr/cpr.cmake
```diff
@@ -2,7 +2,6 @@ include(FetchContent)
 FetchContent_Declare(
         cpr
```
LHT129, line 3: [note] `cpr.cmake` similarly had its OSS cache URL removed without gaining a `vsag_resolve_thirdparty_override` call. cpr uses `FetchContent_Declare` with an inline `URL` (not a url list variable), so it cannot use the new override mechanism as-is. Same consideration as spdlog: is this intentional, or should cpr be refactored to use the url-list pattern?

extern/cpr/fix_curl.patch
```diff
@@ -2,11 +2,10 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
 index 3f6e1d02..2e5a568f 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -297,8 +297,11 @@ else()
+@@ -297,8 +297,10 @@ else()
```
LHT129, line 5: [note] `fix_curl.patch` removes the OSS cache URL from the vendored curl `FetchContent_Declare` inside the cpr patch. This is consistent with the overall OSS removal, but note that this vendored curl dependency also has no override mechanism. If cpr is not being brought under the override system, this is fine. If cpr is refactored later, this vendored curl URL would need separate handling.

tests/cmake/thirdparty_override_test.cmake
```diff
@@ -0,0 +1,188 @@
...
+endif ()
+set (output "${stdout}\n${stderr}")
+assert_matches ("${output}" "URL variable 'missing_urls' is not defined"
+                "undefined URL variable rejection")
+
+set (hash_fixture "${CMAKE_CURRENT_BINARY_DIR}/thirdparty-override-hash-fixture")
+set (download_fixture "${CMAKE_CURRENT_BINARY_DIR}/thirdparty-override-download")
+file (WRITE "${hash_fixture}" "wrong archive content")
+set (ENV{VSAG_THIRDPARTY_FMT_10_2_1} "file://${hash_fixture}")
+execute_process (
```
LHT129, line 172: [suggestion] In `thirdparty_override_test.cmake`, the hash fixture test uses `CMAKE_CURRENT_BINARY_DIR` for `hash_fixture` and `download_fixture` paths. When this script is invoked via `cmake -P` (script mode), `CMAKE_CURRENT_BINARY_DIR` may be empty or undefined, which could cause the `file(WRITE ...)` and `file(REMOVE ...)` calls to operate on unexpected paths.

Consider using `CMAKE_CURRENT_LIST_DIR` or an explicit temporary directory (e.g. `${CMAKE_BINARY_DIR}/Testing/Temporary` when available, or a hardcoded `/tmp` fallback) for these fixture paths to ensure predictable behavior in script mode.

LHT129 (commented):
Reviewed the changes. The PR is well-structured and the implementation is solid:

- The `VSAGThirdPartyOverride.cmake` module is cleanly designed with proper input validation, collision-resistant hash prefix generation, and credential-safe diagnostics.
- All 15 third-party dependency files are consistently migrated to use `vsag_resolve_thirdparty_override`, with Alibaba Cloud OSS cache URLs properly removed while preserving authoritative upstream URLs and `URL_HASH` values.
- The test suite (`thirdparty_override_test.cmake` + `thirdparty_override_fixture.cmake`) provides thorough coverage: pinned/legacy/default precedence, credential-safe output, mismatched pin rejection, hash verification, and collision prefix extension.
- Documentation (EN + ZH) is clear and complete with lookup order, naming rules, inventory table, and practical examples.

No blocking issues found.

extern/spdlog/spdlog.cmake
wxyucs: Confirmed intentional. Issue #2677 scopes this backport to the 15 existing VSAG_THIRDPARTY_* archive overrides on 0.x. spdlog has no existing override, so adding one would expand the compatibility surface beyond this change; only its Alibaba OSS default was removed.

tests/cmake/thirdparty_override_test.cmake
wxyucs: Thanks. In `cmake -P` script mode, CMAKE_CURRENT_BINARY_DIR is defined as the invocation working directory here; `make test-cmake` exercises both writes and cleanup successfully on every backport. I’m retaining the focused fixture structure from #2679.

extern/cpr/cpr.cmake
wxyucs: Confirmed intentional for the same scope reason: CPR is not one of the 15 pre-existing VSAG_THIRDPARTY_* overrides on this release branch. Its Alibaba OSS default is removed, while adding a new public override is left to a separate compatibility decision.

extern/cpr/fix_curl.patch
wxyucs: Agreed. This patch only removes the branch-owned Alibaba OSS URL. A future CPR override design would need to consider its nested curl download separately, but adding that new mechanism is outside Issue #2677’s verified 15-variable inventory.

PR #2685, opened 2026-08-14
feat(build): qualify third-party overrides for 0.16

wxyucs:
## What changed

- resolve all 15 VSAG 0.16 third-party archive overrides through a central pin-qualified helper

## Validation

- `make test-cmake`
- `cmake --build build-validation-http-0.16 --target fmt cpuinfo roaring --parallel 2`
- `git diff --check origin/0.16`
- zero-diff check for `git diff origin/0.16...HEAD -- docs/`

Backport of the approach merged in #2679 for the 0.16-owned dependency pins and URL inventory.

Fixes: #2677


extern/cpr/fix_curl.patch
```diff
@@ -2,11 +2,10 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
...
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -297,8 +297,11 @@ else()
+@@ -297,8 +297,10 @@ else()
      endif()
      FetchContent_Declare(curl
                           URL                    https://github.com/curl/curl/releases/download/curl-8_10_1/curl-8.10.1.tar.xz
-+                                                https://vsagcache.oss-rg-china-mainland.aliyuncs.com/curl/curl-8.10.1.tar.xz
                           URL_HASH               SHA256=73a4b0e99596a09fa5924a4fb7e4b995a85fda0d18a2c02ab9cf134bebce04ee # the file hash for curl-8.10.1.tar.xz
 -                         USES_TERMINAL_DOWNLOAD TRUE)   # <---- This is needed only for Ninja to show download progress
```
LHT129, line 10: [note] The `fix_curl.patch` change adds `INACTIVITY_TIMEOUT` and `TIMEOUT` to the curl `FetchContent_Declare` call. While this is a sensible improvement (prevents hanging downloads), it is unrelated to the stated goal of "qualify third-party overrides for 0.16". Consider splitting unrelated changes into a separate PR for cleaner history and easier bisection.

extern/spdlog/spdlog.cmake
```diff
@@ -5,9 +5,6 @@ set(install_dir ${CMAKE_CURRENT_BINARY_DIR}/${name}/install)
 ExternalProject_Add(
     ${name}
     URL https://github.com/gabime/spdlog/archive/refs/tags/v1.12.0.tar.gz
-        # this url is maintained by the vsag project, if it's broken, please try
-        #  the latest commit or contact the vsag project
-        http://vsagcache.oss-rg-china-mainland.aliyuncs.com/spdlog/v1.12.0.tar.gz
     URL_HASH MD5=6b4446526264c1d1276105482adc18d1
```
LHT129, line 8: [note] `spdlog` and `cpr` dependencies have their OSS cache URLs removed but are not integrated with `vsag_resolve_thirdparty_override`. This is because they use `ExternalProject_Add` / `FetchContent_Declare` with a direct `URL` argument rather than a URL list variable, so there is no list to pass to the resolver.

This means offline/mirror builds cannot override these two dependencies via the new pinned-variable mechanism. If offline build support for spdlog (v1.12.0) and cpr (1.11.2) is desired in the future, consider refactoring them to use a URL list variable pattern similar to the other 15 dependencies.

extern/cpr/fix_curl.patch
```diff
@@ -2,11 +2,10 @@ diff --git a/CMakeLists.txt b/CMakeLists.txt
 index 3f6e1d02..2e5a568f 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -297,8 +297,11 @@ else()
+@@ -297,8 +297,10 @@ else()
```
LHT129, line 5: [note] The curl OSS cache URL (`vsagcache.oss-rg-china-mainland.aliyuncs.com/curl/curl-8.10.1.tar.xz`) is removed from `fix_curl.patch` along with other OSS URLs. Curl is a transitive dependency of cpr and is fetched via `FetchContent_Declare` with a direct `URL` argument inside the patched cpr `CMakeLists.txt`, so it similarly cannot benefit from the new `vsag_resolve_thirdparty_override` mechanism. This is consistent with the cpr/spdlog treatment but worth noting for offline build completeness.

LHT129 (commented):
Thanks for this well-structured PR. The centralized `VSAGThirdPartyOverride.cmake` module is clean and the test coverage is thorough.

- New `cmake/VSAGThirdPartyOverride.cmake` with pin-qualified variable generation, collision resolution, and override resolution with proper precedence (pinned > legacy > default)
- 15 extern cmake files migrated from ad-hoc OSS URL + unversioned env var patterns to `vsag_resolve_thirdparty_override`
- Removal of Alibaba Cloud OSS cache URLs across all dependency files
- `fix_curl.patch` updated to remove OSS URL and add timeout settings
- Comprehensive CMake test suite covering variable naming, precedence, credential-safe diagnostics, hash verification, and inventory validation
- Synchronized English and Chinese offline-build documentation

**One note posted as inline comment** regarding the `fix_curl.patch` timeout additions being unrelated to the override qualification goal.

The core logic in `_vsag_classify_thirdparty_pin` (VERSION/COMMIT/TAG classification), collision resolution with prefix extension, and the three-tier resolution in `vsag_resolve_thirdparty_override` all look correct. The test coverage for credential-safe diagnostics is a nice touch.

extern/spdlog/spdlog.cmake
wxyucs: Confirmed. Issue #2677 scopes these backports to the 15 existing VSAG_THIRDPARTY_* overrides. spdlog and CPR had no override to qualify, so their Alibaba OSS defaults are removed without adding new compatibility surface; a future override design would be separate.

extern/cpr/fix_curl.patch
wxyucs: The timeout lines are pre-existing in the 0.16 patch. This commit only removes the added OSS URL line and adjusts the hunk count from +11 to +10 so the patch remains valid; it does not introduce a timeout behavior change.

extern/cpr/fix_curl.patch
wxyucs: Agreed. Curl is a transitive CPR download outside the branch’s verified 15-variable override inventory. This change removes the Alibaba OSS URL; any new transitive override mechanism requires a separate compatibility decision.

PR #2693, opened 2026-08-15
perf(sindi): batch posting normalization for incremental add

### LINKED ISSUE #2616: [improve](sindi): avoid repeated posting-list sorting during incremental add
### Kind of improvement

perf

### Affected area

sindi

### Summary

`SINDI::Add` calls `SortByValue()` for every affected window after each Add. Repeated
single-document additions can therefore repeatedly sort a growing posting list.

PR #2542 intentionally does not optimize incremental Add workloads. Track this performance
improvement separately.

### Expected benefit

Avoid potentially quadratic posting-list normalization work during incremental ingestion while
preserving correct value-aware term pruning.

### Benchmark evidence

A review probe reported 1,426 ms for 2,000 single-document Adds versus 9 ms for one batch in a
debug build. These results should be reproduced with a dedicated release-mode benchmark before
selecting an implementation.

### References

- `src/algorithm/sindi/sindi.cpp:369`
- `src/datacell/sparse_term_datacell.cpp:45`
- https://github.com/antgroup/vsag/pull/2542#discussion_r3689278733

### Additional context

Potential approaches include tracking dirty posting suffixes, incrementally merging sorted
postings, or deferring normalization. Any solution must ensure pruning is disabled or scans the
complete posting list while a window is not normalized.

## Related issues

_None found._

---

CharlesXu-HQ:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2616

## What Changed

- Keep each mutable SINDI posting list as a sorted prefix plus a bounded sorted dirty suffix, so incremental `Add` no longer re-sorts the full posting list after every batch.
- Merge the two sorted runs during scan, pruning, and heap insertion. The already-normalized path retains a compact loop to avoid paying merge-on-read overhead after build/serialization.
- Normalize dirty postings at bounded thresholds, full-window/build boundaries, compaction, and serialization.
- Apply the lifecycle to both SINDI and SINDI V2, with deterministic boundary, randomized, filtering, quantization, serialization, and restore coverage.

## Test Evidence

- [x] `clang-format 15.0.7 --dry-run --Werror` on all 10 changed C++ files
- [x] `git diff --check`
- [x] Linux x86-64 complete current-head `unittests` suite
- [x] Linux x86-64 focused SINDI suites at current head; macOS ARM64 focused pre-rebase evidence
- [x] Controlled Release A/B benchmarks with order reversal, load recording, and exact-result hashes

Final commit: `fc11f8426f47b2085466c7876511d6ad1cbc616a` (tree `3e8755002d5af77ca9158b612e9fddab4eb8842d`, rebased on `90bcba1953f5a992b3450a550185bd672a56dac4`)

```text
# Linux x86-64, final formatted source, focused filters
All tests passed (188626 assertions in 97 test cases)

# Linux x86-64, final formatted source
./build-linux-debug/tests/unittests
All tests passed (85398093 assertions in 948 test cases)

# macOS ARM64, pre-rebase evidence at 3463288, focused filters
All tests passed (177635 assertions in 75 test cases)
```

The target worktree's ten changed files were checked individually with SHA-256 against the final commit tree before the final runs.

The latest-main rebase was rebuilt in a fresh target clone and verified under a constrained container (CPU 12-15, 4-CPU quota, 6 GiB memory, nice +10, network disabled). The focused binary SHA-256 was `ff1a19c7a66d859483fdcfde73d244a713b3a64ee32439b448aeaa123197c4eb`; CPU 12-15 averaged 98.66%-100% idle before and 99.33%-99.67% idle after the full suite, while other host services remained low-usage.

## Compatibility Impact

- API/ABI compatibility: no public API change.
- Serialized compatibility: no wire-version or byte-layout change. Baseline and current produced identical serialized hashes and byte counts for FP32, FP16, and SQ8 in every measured process.
- Behavior: mutable postings may temporarily contain two individually sorted runs. Query paths merge them exactly; persistence boundaries restore canonical order before writing.

- Add already takes the global write lock. Serialization takes the same exclusive lock while normalizing pending postings; query locking is unchanged.
- Deferred normalization moves about 0.84-1.10 ms of work to serialization in this workload.

### Controlled A/B setup

- Host: AMD Ryzen 7 9700X, Linux x86-64.
- Baseline: `e6dabf2911634be66cd71bfcfeb1e586a5774521`; current: `3463288b4ebc9c80a3615bc42b7ac113fb2e9653`.
- Same compiler, dependencies, Release flags, dataset, and harness for both versions.
- Dataset per process: 30,000 initial + 9,000 incremental sparse vectors, vocabulary 1,024, 32 terms/document, window size 10,000.
- Isolation: CPU 15 only, one-CPU quota, 6 GiB limit, `nice +10`, `OPENBLAS_NUM_THREADS=1`, `OMP_NUM_THREADS=1`, network disabled. Other host services were recorded and left untouched.
- Order-balanced main matrix: forward and reverse version order, 152 independent processes across 10 encoding/batch configurations. Each FP32 cell has 10 samples/version; each FP16/SQ8 cell has 6 samples/version.
- Query-normalized control: `Serialize` before timed queries, 120 independent processes at batch 256, 20 samples/version/encoding.
- CPU15 pre-run idle checks were 99.67%-100% for the final matrices.
- Each process ran 100 KNN top-100 and 100 range-limit-100 queries. Values below are medians.

### Incremental Add results

| Encoding | Batch | Baseline vec/s | Current vec/s | Throughput gain | Add p95 reduction |
|---|---:|---:|---:|---:|---:|
| FP32 | 1 | 5,726 | 69,189 | +1108.39% | 91.67% |
| FP32 | 16 | 22,948 | 136,707 | +495.72% | 76.57% |
| FP32 | 256 | 110,787 | 449,234 | +305.49% | 76.56% |
| FP32 | 4,096 | 514,146 | 859,982 | +67.26% | 29.44% |
| FP16 | 1 | 7,596 | 70,428 | +827.20% | 88.09% |
| FP16 | 256 | 131,847 | 488,480 | +270.49% | 75.05% |
| FP16 | 4,096 | 579,931 | 935,604 | +61.33% | 27.46% |
| SQ8 | 1 | 7,894 | 65,132 | +725.04% | 86.72% |
| SQ8 | 256 | 131,774 | 481,836 | +265.65% | 74.57% |
| SQ8 | 4,096 | 597,519 | 956,571 | +60.09% | 26.30% |

### Accuracy, query RT, serialization, and memory

- All 152 main-matrix processes and all 120 normalized-control processes passed exact A/B consistency checks for KNN hashes, range hashes, serialized hashes, and serialized byte counts.
- FP32: KNN `c8c02d5c436e7561`, range `a8791bd19b32cc97`, serialized `ec785c19cea62734`, 10,398,147 bytes.
- FP16: KNN `eba60050ecc56435`, range `1ffdf5c7871487ff`, serialized `658e543a8b68023a`, 7,905,646 bytes.
- SQ8: KNN `7224cdf82ecda77b`, range `ab7ebf6bbfe94ba2`, serialized `5beb0893264ad66d`, 6,662,690 bytes.
- With a pending dirty run, the largest measured KNN p95 regression was 3.54% (SQ8/batch 256); the largest range p95 regression was 1.79% (FP32/batch 256).
- After normalization, FP16/FP32/SQ8 KNN p95 latency regressions were 0.05%, 0.25%, and 2.57%; all range p95 changes were within +/-0.40%.
- Point-in-time RSS stayed within 0.28%; anonymous RSS increased 0.23%-0.29%, and `mallinfo2` heap increased 0.13%-0.21%. Baseline `GetMemoryUsage()` does not account for all post-window incremental vectors, so it is not used for the memory conclusion.

## Documentation Impact

- [x] No docs update needed

No user-facing parameter, API, serialized format, or workflow changes.

## Risk and Rollback

- Risk level: medium; search and pruning must merge two sorted posting runs exactly until normalization.
- Mitigations: bounded dirty suffixes, canonicalization at persistence/build boundaries, exact-result A/B hashes, randomized lifecycle coverage, and full non-daily regression testing.
- Rollback: revert this commit to restore eager full-list sorting after each `Add`.

## Checklist

- [x] I have updated docs if behavior/workflow changed (not applicable; no user-facing change)

### DIFF
--- a/src/algorithm/sindi/sindi.cpp
+++ b/src/algorithm/sindi/sindi.cpp
@@ -435,12 +435,7 @@ SINDI::Add(const DatasetPtr& base) {
         throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
                             "SINDI DMQ rerank does not support incremental Add");
     }
-    const auto previous_window_count = mutable_term_datacell_->GetWindowCount();
-    auto failed_ids = this->add(base, true);
-    if (mutable_term_datacell_->GetWindowCount() != previous_window_count) {
-        this->cal_memory_usage();
-    }
-    return failed_ids;
+    return this->add(base, true);
 }
 
 std::vector<int64_t>
@@ -468,9 +463,9 @@ SINDI::add(const DatasetPtr& base, bool sort_affected_windows) {
     Vector<uint32_t> pruned_ids(allocator_);
     Vector<float> pruned_vals(allocator_);
     Vector<uint32_t> remapped_ids(allocator_);
-    const auto first_affected_window = cur_element_count_ / window_size_;
-    // This remains -1 when every input vector is rejected, so post-insert loops are no-ops.
-    int64_t last_affected_window = -1;
+    const auto first_affected_window =
+        static_cast<uint32_t>(cur_element_count_.load(std::memory_order_relaxed) / window_size_);
+    std::optional<uint32_t> last_affected_window;
     std::vector<SparseVector> rerank_vectors;
     if (use_reorder_) {
         rerank_vectors.reserve(data_num);
@@ -525,7 +520,8 @@ SINDI::add(const DatasetPtr& base, bool sort_affected_windows) {
             rerank_vectors.push_back(sparse_vectors[i]);
         }
         metadata_build.RecordSuccess(static_cast<uint32_t>(position));
-        last_affected_window = cur_element_count_ / window_size_;
+        last_affected_window = static_cast<uint32_t>(
+            cur_element_count_.load(std::memory_order_relaxed) / window_size_);
         cur_element_count_++;
     }
     if (not rerank_vectors.empty()) {
@@ -535,9 +531,24 @@ SINDI::add(const DatasetPtr& base, bool sort_affected_windows) {
     metadata_filter_.CommitBuild(std::move(metadata_build),
                                  first_inner_id,
                                  static_cast<uint32_t>(cur_element_count_.load()));
-    if (sort_affected_windows) {
-        for (int64_t window = first_affected_window; window <= last_affected_window; ++window) {
-            mutable_term_datacell_->SortByValue(static_cast<uint32_t>(window));
+    if (sort_affected_windows && last_affected_window.has_value()) {
+        bool posting_state_changed = false;
+        const auto current_count =
+            static_cast<uint64_t>(cur_element_count_.load(std::memory_order_relaxed));
+        for (uint32_t window = first_affected_window;; ++window) {
+            const auto window_end =
+                (static_cast<uint64_t>(window) + 1) * static_cast<uint64_t>(window_size_);
+            if (window_end <= current_count) {
+                posting_state_changed |= mutable_term_datacell_->NormalizeDirtyPostings(window);
+            } else {
+                posting_state_changed |= mutable_term_datacell_->FinalizeInsertBatch(window);
+            }
+            if (window == last_affected_window.value()) {
+                break;
+            }
+        }
+        if (posting_state_changed) {
+            this->cal_memory_usage();
         }
     }
     return failed_ids;
@@ -1198,7 +1209,7 @@ SINDI::UseTermListsHeapInsert(const SINDISearchParameter& search_param,
 }
 
 void
-SINDI::cal_memory_usage() {
+SINDI::cal_memory_usage() const {
     auto memory = sizeof(SINDI);
     if (term_datacell_ != nullptr) {
         memory += term_datacell_->GetMemoryUsage();
@@ -1219,8 +1230,16 @@ SINDI::cal_memory_usage() {
 }
 
 void
+SINDI::normalize_dirty_postings_for_serialization() const {
+    if (mutable_term_datacell_ != nullptr && mutable_term_datacell_->NormalizeDirtyPostings()) {
+        this->cal_memory_usage();
+    }
+}
+
+void
 SINDI::Serialize(StreamWriter& writer) const {
-    std::shared_lock rlock(this->global_mutex_);
+    std::scoped_lock wlock(this->global_mutex_);
+    this->normalize_dirty_postings_for_serialization();
 
     if (cur_element_count_ == 0) {
         const auto cur_element_count = cur_element_count_.load();
@@ -1390,7 +1409,8 @@ SINDI::serialize_windows(StreamWriter& writer) const {
 
 void
 SINDI::serialize_streaming_body(StreamWriter& writer) const {
-    std::shared_lock rlock(this->global_mutex_);
+    std::scoped_lock wlock(this->global_mutex_);
+    this->normalize_dirty_postings_for_serialization();
 
     auto windows_tag = static_cast<uint32_t>(StreamSerializationTag::SINDI_WINDOWS);
     auto label_tag = static_cast<uint32_t>(StreamSerializationTag::LABEL_TABLE);
--- a/src/algorithm/sindi_v2/sindi_v2.cpp
+++ b/src/algorithm/sindi_v2/sindi_v2.cpp
@@ -528,8 +528,8 @@ SINDIV2::Add(const DatasetPtr& base) {
     Vector<uint32_t> pruned_ids(allocator_);
     Vector<float> pruned_vals(allocator_);
     Vector<uint32_t> remapped_ids(allocator_);
-    const auto first_affected_window = cur_element_count_ / window_size_;
-    int64_t last_affected_window = -1;
+    const auto first_affected_window = static_cast<uint32_t>(cur_element_count_ / window_size_);
+    std::optional<uint32_t> last_affected_window;
     std::vector<SparseVector> dmq_rerank_vectors;
     if (use_reorder_ && rerank_type_ == SPARSE_RERANK_TYPE_DMQ8) {
         dmq_rerank_vectors.reserve(data_num);
@@ -600,7 +600,7 @@ SINDIV2::Add(const DatasetPtr& base) {
         }
 
         metadata_build.RecordSuccess(static_cast<uint32_t>(position));
-        last_affected_window = cur_element_count_ / window_size_;
+        last_affected_window = static_cast<uint32_t>(cur_element_count_ / window_size_);
         cur_element_count_++;
     }
 
@@ -614,10 +614,25 @@ SINDIV2::Add(const DatasetPtr& base) {
     metadata_filter_.CommitBuild(
         std::move(metadata_build), first_inner_id, static_cast<uint32_t>(cur_element_count_));
 
-    for (int64_t window = first_affected_window; window <= last_affected_window; ++window) {
-        mutable_term_datacell->SortByValue(static_cast<uint32_t>(window));
+    if (last_affected_window.has_value()) {
+        bool posting_state_changed = false;
+        const auto current_count = static_cast<uint64_t>(cur_element_count_);
+        for (uint32_t window = first_affected_window;; ++window) {
+            const auto window_end =
+                (static_cast<uint64_t>(window) + 1) * static_cast<uint64_t>(window_size_);
+            if (window_end <= current_count) {
+                posting_state_changed |= mutable_term_datacell->NormalizeDirtyPostings(window);
+            } else {
+                posting_state_changed |= mutable_term_datacell->FinalizeInsertBatch(window);
+            }
+            if (window == last_affected_window.value()) {
+                break;
+            }
+        }
+        if (posting_state_changed) {
+            this->cal_memory_usage();
+        }
     }
-    this->cal_memory_usage();
     return failed_ids;
 }
 
@@ -1108,7 +1123,7 @@ SINDIV2::UseTermListsHeapInsert(const SINDIV2SearchParameter& search_param) cons
 }
 
 void
-SINDIV2::cal_memory_usage() {
+SINDIV2::cal_memory_usage() const {
     auto memory = sizeof(SINDIV2);
     if (term_datacell_ != nullptr) {
         memory += term_datacell_->GetMemoryUsage();
@@ -1492,9 +1507,8 @@ SINDIV2::Serialize(StreamWriter& writer) const {
 
     if (term_datacell_ != nullptr &&
         std::dynamic_pointer_cast<MutableSindiTermDataCell>(term_datacell_) != nullptr) {
-        const auto mutable_datacell = this->get_mutable_term_datacell();
-        for (uint32_t window = 0; window < mutable_datacell->GetWindowCount(); ++window) {
-            mutable_datacell->SortByValue(window);
+        if (this->get_mutable_term_datacell()->NormalizeDirtyPostings()) {
+            this->cal_memory_usage();
         }
     }
 
--- a/src/datacell/mutable_sindi_term_datacell.cpp
+++ b/src/datacell/mutable_sindi_term_datacell.cpp
@@ -29,7 +29,6 @@ namespace vsag {
 void
 MutableSindiTermDataCell::Finalize() {
     for (auto& window : windows_) {
-        this->SortByValue(window);
         this->Compact(window);
     }
 }
@@ -210,6 +209,70 @@ MutableSindiTermDataCell::GetTermDictCount() const {
     return term_dict_count;
 }
 
+MutableSindiTermDataCell::PostingRunSelection
+MutableSindiTermDataCell::SelectPostingRuns(const MutableSINDIWindow& window,
+                                            uint32_t term,
+                                            uint32_t retained_count) const {
+    const auto posting_count = window.term_sizes_[term];
+    retained_count = std::min(retained_count, posting_count);
+    if (window.postings_sorted_) {
+        return {retained_count, posting_count, 0};
+    }
+    const auto dirty = window.dirty_posting_prefixes_.find(term);
+    if (dirty == window.dirty_posting_prefixes_.end()) {
+        return {retained_count, posting_count, 0};
+    }
+
+    const auto prefix_count = dirty->second;
+    const auto suffix_count = posting_count - prefix_count;
+    if (retained_count == 0) {
+        return {0, prefix_count, 0};
+    }
+    if (retained_count == posting_count) {
+        return {prefix_count, prefix_count, suffix_count};
+    }
+
+    uint32_t lower = retained_count > prefix_count ? retained_count - prefix_count : 0;
+    uint32_t upper = std::min(retained_count, suffix_count);
+    while (lower <= upper) {
+        const auto take_suffix = lower + (upper - lower) / 2;
+        const auto take_prefix = retained_count - take_suffix;
+        if (take_prefix > 0 && take_suffix < suffix_count &&
+            PostingEntryBefore(window, term, prefix_count + take_suffix, take_prefix - 1)) {
+            lower = take_suffix + 1;
+            continue;
+        }
+        if (take_suffix > 0 && take_prefix < prefix_count &&
+            PostingEntryBefore(window, term, take_prefix, prefix_count + take_suffix - 1)) {
+            upper = take_suffix - 1;
+            continue;
+        }
+        return {take_prefix, prefix_count, take_suffix};
+    }
+
+    CHECK_ARGUMENT(false, "failed to partition sorted posting runs");
+}
+
+void
+MutableSindiTermDataCell::ScanPostingRange(uint32_t term_iterator,
+                                           const SparseTermComputerPtr& computer,
+                                           const uint16_t* term_ids,
+                                           const uint8_t* term_data,
+                                           uint32_t term_count,
+                                           float* dists) const {
+    if (term_count == 0) {
+        return;
+    }
+    if (sparse_value_quant_type_ == SparseValueQuantizationType::SQ8) {
+        computer->ScanForAccumulateSQ8(term_iterator, term_ids, term_data, term_count, dists);
+    } else if (sparse_value_quant_type_ == SparseValueQuantizationType::FP16) {
+        computer->ScanForAccumulateFP16Bytes(term_iterator, term_ids, term_data, term_count, dists);
+    } else {
+        computer->ScanForAccumulateFloatBytes(
+            term_iterator, term_ids, term_data, term_count, dists);
+    }
+}
+
 void
 MutableSindiTermDataCell::QueryWindow(float* dists,
                                       uint32_t window_id,
@@ -237,25 +300,19 @@ MutableSindiTermDataCell::QueryWindow(float* dists,
         const auto posting_count = window.term_sizes_[term];
         const auto term_size = computer->GetTermScanCount(posting_count);
         query_context.has_untracked_approximate_evaluations |= term_size > 0;
-
-        if (sparse_value_quant_type_ == SparseValueQuantizationType::SQ8) {
-            computer->ScanForAccumulateSQ8(it,
-                                           window.term_ids_[term]->data(),
-                                           window.term_datas_[term]->data(),
-                                           term_size,
-                                           dists);
-        } else if (sparse_value_quant_type_ == SparseValueQuantizationType::FP16) {
-            computer->ScanForAccumulateFP16Bytes(it,
-                                                 window.term_ids_[term]->data(),
-                                                 window.term_datas_[term]->data(),
-                                                 term_size,
-                                                 dists);
-        } else {
-            computer->ScanForAccumulateFloatBytes(it,
-                                                  window.term_ids_[term]->data(),
-                                                  window.term_datas_[term]->data(),
-                                                  term_size,
-                                                  dists);
+        const auto selected = this->SelectPostingRuns(window, term, term_size);
+        const auto code_size = this->GetTermValueCodeSize();
+        const auto* term_ids = window.term_ids_[term]->data();
+        const auto* term_data = window.term_datas_[term]->data();
+        this->ScanPostingRange(it, computer, term_ids, term_data, selected.prefix_count, dists);
+        if (selected.suffix_count > 0) {
+            this->ScanPostingRange(
+                it,
+                computer,
+                term_ids + selected.suffix_offset,
+                term_data + static_cast<uint64_t>(selected.suffix_offset) * code_size,
+                selected.suffix_count,
+                dists);
         }
     }
     computer->ResetTerm();
@@ -499,61 +556,110 @@ MutableSindiTermDataCell::InsertHeapByTermLists(const MutableSINDIWindow& window
             continue;
         }
 
-        uint32_t i = 0;
         const auto posting_count = window.term_sizes_[term];
         const auto term_size = computer->GetTermScanCount(posting_count);
-        auto& one_term_ids = *window.term_ids_[term];
-        if constexpr (mode == InnerSearchMode::KNN_SEARCH) {
-            if (heap.size() < n_candidate) {
-                for (; i < term_size; i++) {
-                    id = one_term_ids[i];
-                    const bool heap_filled = fill_heap_initial<type>(id,
-                                                                     dists[id],
-                                                                     cur_heap_top,
-                                                                     heap,
-                                                                     offset_id,
-                                                                     n_candidate,
-                                                                     filter,
-                                                                     param.distance_threshold,
-                                                                     param.enable_reorder);
-                    if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
-                        if (filter_callback_remaining != nullptr and
-                            *filter_callback_remaining == 0) {
-                            computer->ResetTerm();
-                            return true;
+        if (window.postings_sorted_) {
+            uint32_t i = 0;
+            const auto& term_ids = *window.term_ids_[term];
+            if constexpr (mode == InnerSearchMode::KNN_SEARCH) {
+                if (heap.size() < n_candidate) {
+                    for (; i < term_size; ++i) {
+                        id = term_ids[i];
+                        const bool heap_filled = fill_heap_initial<type>(id,
+                                                                         dists[id],
+                                                                         cur_heap_top,
+                                                                         heap,
+                                                                         offset_id,
+                                                                         n_candidate,
+                                                                         filter,
+                                                                         param.distance_threshold,
+                                                                         param.enable_reorder);
+                        if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
+                            if (filter_callback_remaining != nullptr &&
+                                *filter_callback_remaining == 0) {
+                                computer->ResetTerm();
+                                return true;
+                            }
+                        }
+                        if (heap_filled) {
+                            ++i;
+                            break;
                         }
                     }
-                    if (heap_filled) {
-                        i++;
-                        break;
+                }
+            }
+            for (; i < term_size; ++i) {
+                id = term_ids[i];
+                if constexpr (mode == InnerSearchMode::RANGE_SEARCH) {
+                    if (candidate_tracker != nullptr && !candidate_tracker->MarkOne(id)) {
+                        continue;
+                    }
+                }
+                insert_candidate_into_heap<mode, type>(id,
+                                                       dists[id],
+                                                       cur_heap_top,
+                                                       heap,
+                                                       offset_id,
+                                                       n_candidate,
+                                                       radius,
+                                                       filter,
+                                                       param.distance_threshold,
+                                                       param.enable_reorder);
+                if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
+                    if (filter_callback_remaining != nullptr && *filter_callback_remaining == 0) {
+                        computer->ResetTerm();
+                        return true;
                     }
                 }
             }
+            continue;
         }
-
-        for (; i < term_size; i++) {
-            id = one_term_ids[i];
-            if constexpr (mode == InnerSearchMode::RANGE_SEARCH) {
-                if (candidate_tracker != nullptr && not candidate_tracker->MarkOne(id)) {
-                    continue;
+        const auto selected = this->SelectPostingRuns(window, term, term_size);
+        bool heap_initialized = heap.size() >= n_candidate;
+        const auto completed =
+            this->ForEachSelectedPosting(window, term, selected, [&](uint16_t selected_id) {
+                id = selected_id;
+                if constexpr (mode == InnerSearchMode::RANGE_SEARCH) {
+                    if (candidate_tracker != nullptr && !candidate_tracker->MarkOne(id)) {
+                        return true;
+                    }
                 }
-            }
-            insert_candidate_into_heap<mode, type>(id,
-                                                   dists[id],
-                                                   cur_heap_top,
-                                                   heap,
-                                                   offset_id,
-                                                   n_candidate,
-                                                   radius,
-                                                   filter,
-                                                   param.distance_threshold,
-                                                   param.enable_reorder);
-            if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
-                if (filter_callback_remaining != nullptr and *filter_callback_remaining == 0) {
-                    computer->ResetTerm();
-                    return true;
+                if constexpr (mode == InnerSearchMode::KNN_SEARCH) {
+                    if (!heap_initialized) {
+                        heap_initialized = fill_heap_initial<type>(id,
+                                                                   dists[id],
+                                                                   cur_heap_top,
+                                                                   heap,
+                                                                   offset_id,
+                                                                   n_candidate,
+                                                                   filter,
+                                                                   param.distance_threshold,
+                                                                   param.enable_reorder);
+                        if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
+                            return filter_callback_remaining == nullptr ||
+                                   *filter_callback_remaining != 0;
+                        }
+                        return true;
+                    }
                 }
-            }
+                insert_candidate_into_heap<mode, type>(id,
+                                                       dists[id],
+                                                       cur_heap_top,
+                                                       heap,
+                                                       offset_id,
+                                                       n_candidate,
+                                                       radius,
+                                                       filter,
+                                                       param.distance_threshold,
+                                                       param.enable_reorder);
+                if constexpr (type == InnerSearchType::WITH_FILTER_LIMIT) {
+                    return filter_callback_remaining == nullptr || *filter_callback_remaining != 0;
+                }
+                return true;
+            });
+        if (!completed) {
+            computer->ResetTerm();
+            return true;
         }
     }
     computer->ResetTerm();
@@ -654,6 +760,7 @@ MutableSindiTermDataCell::InsertVector(const SparseVector& sparse_base, uint32_t
             window->term_datas_[term] = std::make_unique<Vector<uint8_t>>(allocator_);
         }
 
+        const auto old_term_size = window->term_sizes_[term];
         window->term_ids_[term]->push_back(window_local_id);
 
         auto& data_vec = *window->term_datas_[term];
@@ -663,6 +770,11 @@ MutableSindiTermDataCell::InsertVector(const SparseVector& sparse_base, uint32_t
             val, sparse_value_quant_type_, quantization_params_.get(), data_vec.data() + old_size);
 
         window->term_sizes_[term] += 1;
+        window->dirty_posting_prefixes_.try_emplace(term, old_term_size);
+        if (window->pending_posting_flags_[term] == 0) {
+            window->pending_posting_flags_[term] = 1;
+            window->pending_posting_terms_.push_back(term);
+        }
     }
     if (sparse_base.len_ > 0) {
         window->postings_sorted_ = false;
@@ -677,7 +789,7 @@ MutableSindiTermDataCell::SortByValue(uint32_t window_id) {
 }
 
 void
-MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) const {
+MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) {
     if (window.postings_sorted_) {
         return;
     }
@@ -702,7 +814,219 @@ MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) const {
                                                      sorted_ids,
                                                      sorted_data);
     }
+    window.dirty_posting_prefixes_.clear();
+    window.dirty_posting_prefixes_.rehash(0);
+    for (uint32_t term : window.pending_posting_terms_) {
+        window.pending_posting_flags_[term] = 0;
+    }
+    window.pending_posting_terms_.clear();
     window.postings_sorted_ = true;
+    this->ReleaseNormalizationScratch();
+}
+
+bool
+MutableSindiTermDataCell::PostingEntryBefore(const MutableSINDIWindow& window,
+                                             uint32_t term,
+                                             uint32_t left,
+                                             uint32_t right) const {
+    const auto& ids = *window.term_ids_[term];
+    const auto& data = *window.term_datas_[term];
+    const auto compare = [&ids, &data, left, right](auto code_type) {
+        using CodeType = decltype(code_type);
+        CodeType left_code{};
+        CodeType right_code{};
+        std::memcpy(&left_code,
+                    data.data() + static_cast<uint64_t>(left) * sizeof(CodeType),
+                    sizeof(CodeType));
+        std::memcpy(&right_code,
+                    data.data() + static_cast<uint64_t>(right) * sizeof(CodeType),
+                    sizeof(CodeType));
+        if (left_code != right_code) {
+            return left_code > right_code;
+        }
+        return ids[left] < ids[right];
+    };
+
+    switch (sparse_value_quant_type_) {
+        case SparseValueQuantizationType::SQ8:
+            return compare(uint8_t{});
+        case SparseValueQuantizationType::FP16:
+            return compare(uint16_t{});
+        case SparseValueQuantizationType::FP32:
+            return compare(float{});
+        default:
+            CHECK_ARGUMENT(false, "unknown sparse value quantization type");
+    }
+}
+
+void
+MutableSindiTermDataCell::NormalizePosting(MutableSINDIWindow& window, uint32_t term) {
+    const auto dirty = window.dirty_posting_prefixes_.find(term);
+    if (dirty == window.dirty_posting_prefixes_.end()) {
+        return;
+    }
+
+    const auto prefix_count = dirty->second;
+    const auto posting_count = window.term_sizes_[term];
+    const auto code_size = this->GetTermValueCodeSize();
+    auto& ids = *window.term_ids_[term];
+    auto& data = *window.term_datas_[term];
+    normalization_ids_scratch_.resize(posting_count);
+    normalization_data_scratch_.resize(static_cast<uint64_t>(posting_count) * code_size);
+
+    uint32_t prefix = 0;
+    uint32_t suffix = prefix_count;
+    uint32_t output = 0;
+    while (prefix < prefix_count && suffix < posting_count) {
+        const auto source = PostingEntryBefore(window, term, prefix, suffix) ? prefix++ : suffix++;
+        normalization_ids_scratch_[output] = ids[source];
+        std::memcpy(normalization_data_scratch_.data() + static_cast<uint64_t>(output) * code_size,
+                    data.data() + static_cast<uint64_t>(source) * code_size,
+                    code_size);
+        ++output;
+    }
+    while (prefix < prefix_count) {
+        normalization_ids_scratch_[output] = ids[prefix];
+        std::memcpy(normalization_data_scratch_.data() + static_cast<uint64_t>(output) * code_size,
+                    data.data() + static_cast<uint64_t>(prefix) * code_size,
+                    code_size);
+        ++prefix;
+        ++output;
+    }
+    while (suffix < posting_count) {
+        normalization_ids_scratch_[output] = ids[suffix];
+        std::memcpy(normalization_data_scratch_.data() + static_cast<uint64_t>(output) * code_size,
+                    data.data() + static_cast<uint64_t>(suffix) * code_size,
+                    code_size);
+        ++suffix;
+        ++output;
+    }
+
+    std::copy(normalization_ids_scratch_.begin(), normalization_ids_scratch_.end(), ids.begin());
+    std::copy(normalization_data_scratch_.begin(), normalization_data_scratch_.end(), data.begin());
+    window.dirty_posting_prefixes_.erase(term);
+}
+
+bool
+MutableSindiTermDataCell::FinalizeInsertBatch(uint32_t window_id) {
+    CHECK_ARGUMENT(window_id < windows_.size(), "mutable SINDI window id out of range");
+    return this->FinalizeInsertBatch(windows_[window_id], true);
+}
+
+bool
+MutableSindiTermDataCell::FinalizeInsertBatch(MutableSINDIWindow& window, bool release_scratch) {
+    if (window.dirty_posting_prefixes_.empty()) {
+        return false;
+    }
+
+    struct PostingSortScratch {
+        explicit PostingSortScratch(Allocator* allocator)
+            : order(allocator), sorted_ids(allocator), sorted_data(allocator) {
+        }
+
+        Vector<uint32_t> order;
+        Vector<uint16_t> sorted_ids;
+        Vector<uint8_t> sorted_data;
+    };
+    std::optional<PostingSortScratch> scratch;
+    bool posting_state_changed = false;
+    const auto code_size = this->GetTermValueCodeSize();
+    for (uint32_t term : window.pending_posting_terms_) {
+        window.pending_posting_flags_[term] = 0;
+        const auto dirty = window.dirty_posting_prefixes_.find(term);
+        if (dirty == window.dirty_posting_prefixes_.end()) {
+            continue;
+        }
+        if (!scratch.has_value()) {
+            scratch.emplace(allocator_);
+        }
+        auto& sort_scratch = scratch.value();
+        posting_state_changed = true;
+        const auto prefix_count = dirty->second;
+        const auto suffix_count = window.term_sizes_[term] - prefix_count;
+        auto& ids = *window.term_ids_[term];
+        auto& data = *window.term_datas_[term];
+        sindi_datacell_utils::SortPostingListByValue(
+            ids.data() + prefix_count,
+            data.data() + static_cast<uint64_t>(prefix_count) * code_size,
+            suffix_count,
+            sparse_value_quant_type_,
+            sort_scratch.order,
+            sort_scratch.sorted_ids,
+            sort_scratch.sorted_data);
+
+        if (prefix_count == 0) {
+            window.dirty_posting_prefixes_.erase(dirty);
+            continue;
+        }
+        const auto normalize_threshold =
+            std::clamp(prefix_count, MIN_DIRTY_POSTING_SIZE, MAX_DIRTY_POSTING_SIZE);
+        if (suffix_count >= normalize_threshold) {
+            this->NormalizePosting(window, term);
+        }
+    }
+    window.pending_posting_terms_.clear();
+    if (window.dirty_posting_prefixes_.empty()) {
+        window.dirty_posting_prefixes_.rehash(0);
+        window.postings_sorted_ = true;
+    } else {
+        window.postings_sorted_ = false;
+    }
+    if (release_scratch) {
+        this->ReleaseNormalizationScratch();
+    }
+    return posting_state_changed;
+}
+
+void
+MutableSindiTermDataCell::ReleaseNormalizationScratch() {
+    Vector<uint16_t> empty_ids(allocator_);
+    Vector<uint8_t> empty_data(allocator_);
+    normalization_ids_scratch_.swap(empty_ids);
+    normalization_data_scratch_.swap(empty_data);
+}
+
+bool
+MutableSindiTermDataCell::NormalizeDirtyPostings(uint32_t window_id) {
+    CHECK_ARGUMENT(window_id < windows_.size(), "mutable SINDI window id out of range");
+    return this->NormalizeDirtyPostings(windows_[window_id], true);
+}
+
+bool
+MutableSindiTermDataCell::NormalizeDirtyPostings() {
+    bool posting_state_changed = false;
+    for (auto& window : windows_) {
+        posting_state_changed |= this->NormalizeDirtyPostings(window, false);
+    }
+    this->ReleaseNormalizationScratch();
+    return posting_state_changed;
+}
+
+bool
+MutableSindiTermDataCell::NormalizeDirtyPostings(MutableSINDIWindow& window, bool release_scratch) {
+    const auto finalized = this->FinalizeInsertBatch(window, false);
+    if (window.dirty_posting_prefixes_.empty()) {
+        if (release_scratch) {
+            this->ReleaseNormalizationScratch();
+        }
+        return finalized;
+    }
+
+    Vector<uint32_t> dirty_terms(allocator_);
+    dirty_terms.reserve(window.dirty_posting_prefixes_.size());
+    for (const auto& item : window.dirty_posting_prefixes_) {
+        dirty_terms.push_back(item.first);
+    }
+    for (uint32_t term : dirty_terms) {
+        this->NormalizePosting(window, term);
+    }
+    window.dirty_posting_prefixes_.clear();
+    window.dirty_posting_prefixes_.rehash(0);
+    window.postings_sorted_ = true;
+    if (release_scratch) {
+        this->ReleaseNormalizationScratch();
+    }
+    return true;
 }
 
 void
@@ -728,14 +1052,19 @@ MutableSindiTermDataCell::ResizeTermList(MutableSINDIWindow& window,
     Vector<std::unique_ptr<Vector<uint16_t>>> new_ids(new_capacity, allocator_);
     Vector<std::unique_ptr<Vector<uint8_t>>> new_datas(new_capacity, allocator_);
     Vector<uint32_t> new_sizes(new_capacity, 0, allocator_);
+    Vector<uint8_t> new_pending_flags(new_capacity, 0, allocator_);
 
     std::move(window.term_ids_.begin(), window.term_ids_.end(), new_ids.begin());
     std::move(window.term_datas_.begin(), window.term_datas_.end(), new_datas.begin());
     std::copy(window.term_sizes_.begin(), window.term_sizes_.end(), new_sizes.begin());
+    std::copy(window.pending_posting_flags_.begin(),
+              window.pending_posting_flags_.end(),
+              new_pending_flags.begin());
 
     window.term_ids_.swap(new_ids);
     window.term_datas_.swap(new_datas);
     window.term_sizes_.swap(new_sizes);
+    window.pending_posting_flags_.swap(new_pending_flags);
     window.term_capacity_ = new_capacity;
 }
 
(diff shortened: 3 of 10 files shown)


src/algorithm/sindi/sindi.cpp
```diff
@@ -1570,7 +1575,17 @@ SINDI::cal_memory_usage() {
...
+
+    bool normalized_window = false;
+    if (immutable_data_ == nullptr) {
+        for (const auto& window : window_term_list_) {
+            normalized_window |= window->NormalizeDirtyPostings();
+        }
+    }
+    if (normalized_window) {
+        const_cast<SINDI*>(this)->cal_memory_usage();
+    }
```
LHT129, line 1181: [suggestion] The `const_cast` in `Serialize` and `serialize_streaming_body` to call `cal_memory_usage()` after `NormalizeDirtyPostings()` is a code smell. Since serialization already holds an exclusive lock (`scoped_lock`), the cast is safe in practice, but it breaks the logical const contract of the method.

Consider one of these alternatives:
1. Declare `memory_usage_` (or the cached value) as `mutable` so `cal_memory_usage()` can be called from const context.
2. Factor the normalization+recalc into a separate non-const helper that both `Serialize` and `serialize_streaming_body` call before acquiring their respective locks.

This would eliminate the need for `const_cast` entirely.

src/algorithm/sindi/sindi.cpp
```diff
@@ -1570,7 +1575,17 @@ SINDI::cal_memory_usage() {
 
 void
 SINDI::Serialize(StreamWriter& writer) const {
-    std::shared_lock rlock(this->global_mutex_);
+    std::scoped_lock wlock(this->global_mutex_);
```
LHT129, line 1186: [note] `Serialize` and `serialize_streaming_body` now acquire an exclusive lock (`scoped_lock`) instead of the previous shared lock (`shared_lock`). This means serialization will block concurrent search operations while it normalizes dirty postings and writes.

This is a necessary trade-off to ensure postings are fully normalized before serialization, and the PR description already documents that "Serialization now takes the same exclusive lock while normalizing pending postings before writing." Just confirming this is an intentional design choice with no further action needed.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -135,18 +201,22 @@ SparseTermDataCell::query_impl(float* global_dists,
...
+        const auto selected = SelectPostingRuns(term, term_count);
+        const auto code_size = GetSparseValueCodeSize(sparse_value_quant_type_);
+        ScanPostingRange(it,
+                         computer,
+                         term_ids,
+                         term_data,
+                         selected.prefix_count,
+                         global_dists,
+                         evaluation_tracker);
+        ScanPostingRange(it,
```
LHT129: [suggestion] When a term has no dirty postings, `SelectPostingRuns` returns `{term_count, posting_count, 0}` (suffix_count=0). The second `ScanPostingRange` call with `term_count=0` returns immediately, but the extra function call and branch are on the hot query path for every term in every query.

Consider guarding the second call:
```cpp
if (selected.suffix_count > 0) {
    ScanPostingRange(it, computer,
                     term_ids + selected.suffix_offset,
                     term_data + static_cast<uint64_t>(selected.suffix_offset) * code_size,
                     selected.suffix_count, global_dists, evaluation_tracker);
}
```
This avoids the overhead for the common case (fully normalized postings) while keeping the logic identical.

LHT129 (commented):
Overall this is a well-engineered PR with thorough test coverage and detailed performance benchmarks. The core idea — deferring full posting-list normalization by maintaining a sorted prefix plus a bounded sorted dirty suffix — is sound and delivers significant incremental Add throughput improvements (up to 10x for batch-size-1).

**Summary of findings:**

1. **[suggestion] `const_cast` in Serialize methods** — The `const_cast<SINDI*>(this)->cal_memory_usage()` calls break the logical const contract. Consider making the cached memory usage `mutable` or refactoring the normalization+recalc into a pre-serialization helper.

2. **[note] Serialize lock upgrade** — Changing from `shared_lock` to `scoped_lock` in serialization is a necessary trade-off to normalize dirty postings before writing. The PR description already documents this; no action needed.

3. **[suggestion] Redundant `ScanPostingRange` call on hot path** — When postings are fully normalized (the common case), the second `ScanPostingRange` call has `term_count=0` and returns immediately. A guard on `suffix_count > 0` would avoid the function call overhead.

**What is done well:**
- The binary-search partition in `SelectPostingRuns` is correct and well-tested across all three quantization types.
- `ForEachSelectedPosting` elegantly merges two sorted runs in heap-insertion order without materializing a merged array.
- Test coverage is comprehensive: deterministic correctness, serialization round-trips, filter paths, randomized stress tests (20 seeds), and all three quantization encodings.
- The normalization threshold clamping (`MIN_DIRTY_POSTING_SIZE`=32, `MAX_DIRTY_POSTING_SIZE`=256) provides a good balance between deferring work and bounding the dirty suffix size.
- Performance data is thorough and reproducible, with identical A/B result hashes confirming correctness.

LHT129 (commented):
Overall this is a well-engineered optimization with thorough test coverage. The performance improvements are impressive (up to 10x for single-vector incremental Add).

**Suggestions:**

1. **`src/algorithm/sindi/sindi.cpp:1587` and `:1737`** — `[suggestion]` The `Serialize` and `serialize_streaming_body` methods are marked `const` but modify internal state by calling `NormalizeDirtyPostings()` (which mutates posting lists) and `cal_memory_usage()` via `const_cast`. This violates logical const-correctness — callers of a `const` method do not expect side effects that change the object's observable state. Consider making these methods non-const, or marking the affected mutable state as `mutable` with documentation that serialization is a normalization boundary.

2. **`src/datacell/sparse_term_datacell.cpp:143`** — `[note]` The binary search in `SelectPostingRuns` correctly partitions two sorted runs. The fallthrough `CHECK_ARGUMENT(false, ...)` is an unreachable defensive guard — consistent with the codebase convention. The four uncovered lines noted in the PR description correspond to this and similar defensive paths.

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: Updated. current_memory_usage_ is now treated as a mutable cache and cal_memory_usage() is const, so serialization can refresh the cache after normalizing pending postings without using const_cast. Serialization remains an exclusive normalization boundary.

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: Confirmed. The exclusive lock is intentional because serialization normalizes pending postings before writing.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: Updated. The suffix scan is now skipped when suffix_count is zero.

LHT129 (commented):
This is a well-executed optimization that defers full posting-list re-sorting during incremental `Add` by maintaining a sorted prefix plus a bounded sorted dirty suffix. The core algorithm (`SelectPostingRuns` binary search, `ForEachSelectedPosting` merge, `NormalizePosting` two-way merge) is correct and well-tested across all three quantization types with deterministic, randomized, and lifecycle coverage.

**Summary of observations (no blocking issues):**

1. The `const_cast` removal (via `mutable` on `current_memory_usage_` and `const` on `cal_memory_usage()`) cleanly resolves the const-correctness concern in serialization paths.

2. The exclusive lock in `Serialize` / `serialize_streaming_body` is a necessary trade-off to guarantee postings are normalized before writing, and is clearly documented in the PR description.

3. The `query_impl` suffix scan guard (`if (selected.suffix_count > 0)`) noted in a prior review comment has been addressed — this avoids an unnecessary function call on the hot path for fully normalized postings.

4. The `Compact()` method calls `FinalizeInsertBatch()` defensively but does not call `NormalizeDirtyPostings()`. This is safe because all current callers normalize before compacting, but future callers should be aware of this invariant.

5. Test coverage is thorough: 98.28% changed-line coverage, all three quantization types, KNN/range/filter paths, serialization round-trips, randomized differential testing with 20 seeds, and wire format version retention.

The performance improvements (up to 10x for single-vector incremental adds) are well-supported by controlled A/B benchmarks with identical KNN/range/serialized hashes confirming result correctness.

LHT129 (commented):
I have reviewed this PR and the changes look solid. The deferred normalization strategy for incremental posting batches is well-designed and the test coverage is comprehensive.

**Summary of changes:**
- `SparseTermDataCell`: Introduces two-run posting lists with deferred normalization via `FinalizeInsertBatch()` and `NormalizeDirtyPostings()`. Query and heap-insert paths correctly handle split runs through `SelectPostingRuns()` and `ForEachSelectedPosting()`.
- `SINDI::Add()`: Replaces eager `SortByValue()` with batched `FinalizeInsertBatch()` and defers full normalization until a window reaches capacity.
- `SINDI::Build()`: Normalizes dirty postings before compacting.
- Serialization (`Serialize` / `serialize_streaming_body`): Upgraded to exclusive lock and normalizes dirty postings before writing to ensure canonical output.
- `cal_memory_usage()`: Made `const` with `mutable` memory-usage atomic.
- Tests: Updated existing test, added serialization normalization test, and added comprehensive `SelectPostingRuns` and two-run heap equivalence tests.

**Key observations:**

1. The binary search in `SelectPostingRuns` correctly finds the optimal partition between sorted prefix and suffix runs. The monotonicity property holds because as more entries are taken from the suffix, the boundary entries move in consistent directions.

2. The `ForEachSelectedPosting` template correctly merges two sorted runs in value-descending order, maintaining the same traversal semantics as the original single-run iteration.

3. The lock upgrade in serialization from shared to exclusive is necessary since normalization mutates internal state. This is an intentional trade-off: serialization now blocks concurrent reads but produces canonical output.

4. The `term >= term_ids_.size() || term_sizes_[term] == 0` guard added to `InsertHeapByTermLists` fixes a missing null-pointer check that was already present in `query_impl`.

No blocking issues found. The existing review comments have been properly addressed (mutable atomic, exclusive lock rationale, suffix_count zero-guard).

src/algorithm/sindi/sindi.cpp
```diff
@@ -1568,9 +1573,24 @@ SINDI::cal_memory_usage() {
     this->current_memory_usage_.store(memory);
 }
 
+void
```
LHT129, line 1177: [suggestion] `normalize_dirty_postings_for_serialization` is declared `const` but calls `window->NormalizeDirtyPostings()` which is a non-const method that mutates posting list state. While this works because `const auto&` binds to a `const shared_ptr&` (and `operator->` on a const shared_ptr returns a non-const raw pointer), the logical const contract is misleading — a reader would not expect a `const` serialization helper to modify index internals.

Consider either:
1. Removing the `const` qualifier from `normalize_dirty_postings_for_serialization` and `Serialize`/`serialize_streaming_body`, since serialization already takes an exclusive lock and is effectively a mutable operation.
2. Adding a comment explicitly documenting that this "const" method may mutate internal posting state for canonicalization, and why that is safe under the exclusive lock.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -431,10 +495,172 @@ SparseTermDataCell::InsertVector(const SparseVector& sparse_base, uint16_t base_
...
+        }
+        const auto normalize_threshold =
+            std::clamp(prefix_count, MIN_DIRTY_POSTING_SIZE, MAX_DIRTY_POSTING_SIZE);
+        if (suffix_count >= normalize_threshold) {
+            NormalizePosting(term);
+        }
+    }
+    pending_posting_terms_.clear();
+    if (dirty_posting_prefixes_.empty()) {
+        dirty_posting_prefixes_.rehash(0);
```
LHT129: [suggestion] `FinalizeInsertBatch` unconditionally returns `true` after processing `pending_posting_terms_`, even when every pending term is missing from `dirty_posting_prefixes_` (the `continue` branch). In that case no posting state actually changed, but the caller in `SINDI::Add()` will still invoke `cal_memory_usage()`. Consider tracking whether any term was actually processed and returning that instead.

src/algorithm/sindi/sindi.cpp
```diff
@@ -1568,9 +1573,24 @@ SINDI::cal_memory_usage() {
...
+SINDI::normalize_dirty_postings_for_serialization() const {
+    // Serialization canonicalizes dirty posting runs without changing the logical index contents.
```
LHT129: [note] `normalize_dirty_postings_for_serialization()` is declared `const` but mutates posting list internals through `window->NormalizeDirtyPostings()`. While this is safe under the exclusive lock that the caller (`Serialize` / `serialize_streaming_body`) already holds, a reader skimming the header would reasonably expect a `const` method not to modify index state. A brief comment explaining that this is a logical const method that canonicalizes internal representation without changing the logical index contents would help future readers.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -431,10 +495,172 @@ SparseTermDataCell::InsertVector(const SparseVector& sparse_base, uint16_t base_
...
+bool
+SparseTermDataCell::FinalizeInsertBatch() {
+    if (dirty_posting_prefixes_.empty()) {
+        return false;
+    }
+
+    Vector<uint32_t> order(allocator_);
+    Vector<uint16_t> sorted_ids(allocator_);
+    Vector<uint8_t> sorted_data(allocator_);
+    const auto code_size = GetSparseValueCodeSize(sparse_value_quant_type_);
```
LHT129: [note] `FinalizeInsertBatch` allocates `order`, `sorted_ids`, and `sorted_data` scratch vectors on every invocation, even when `dirty_posting_prefixes_` is empty (which returns early at line 594) or when every pending term is missing from the dirty map. The vectors are default-constructed so the allocation is deferred until first use, but the objects themselves are still created on the stack. This is negligible for correctness but worth noting for hot-path callers like `SINDI::Add()`.

LHT129 (commented):
I have completed a thorough review of this PR. The implementation is well-structured and the test coverage is comprehensive (976 additions across 7 files with extensive unit tests covering all three quantization types, serialization round-trips, randomized differential testing, and heap equivalence).

**Summary of findings:**

The core approach — maintaining each mutable SINDI posting list as a sorted prefix plus a bounded sorted dirty suffix — is sound. The binary search in `SelectPostingRuns` correctly partitions the two sorted runs, and the merge logic in both `ForEachSelectedPosting` (for heap insertion) and `NormalizePosting` (for full normalization) is correct.

**Previous review comments addressed:**
- The `const_cast` in serialization has been eliminated by making `current_memory_usage_` mutable and `cal_memory_usage()` const.
- The unconditional second `ScanPostingRange` call in `query_impl` is now guarded by `if (selected.suffix_count > 0)`.
- The exclusive lock for serialization is confirmed as intentional and documented.

**Remaining notes (non-blocking, already raised by LHT129):**
- `normalize_dirty_postings_for_serialization()` is declared `const` but mutates posting list internals — the existing comment explains the logical const semantics.
- Scratch vector allocations in `FinalizeInsertBatch` are created on the stack before the early-return check — negligible overhead.

The performance data in the PR description shows impressive improvements (up to 10x for single-vector adds) with no regression in search accuracy or serialization compatibility. The controlled A/B methodology is rigorous.

No blocking issues found. The PR looks ready to merge from a code review perspective.

LHT129 (commented):
I have completed a thorough review of this PR. The existing inline comments from the previous review round have already covered the main areas of concern, and the author has addressed several of them (e.g., making `current_memory_usage_` mutable, guarding the suffix scan).

**Summary of my review:**

**Correctness**: The core algorithm — `SelectPostingRuns` binary search, `NormalizePosting` two-way merge, and `ForEachSelectedPosting` merge-traversal — all appear correct. The `PostingEntryBefore` comparator correctly implements descending-by-value with ascending-by-id tiebreaking, consistent with the existing `SortPostingListByValue` ordering.

**Concurrency**: The lock upgrade from `shared_lock` to `scoped_lock` in `Serialize`/`serialize_streaming_body` is intentional and necessary for the normalization-before-write guarantee. Search paths continue to use shared locking, and the dirty-posting data structures are only mutated under the exclusive lock.

**Memory**: The scratch vectors (`normalization_ids_scratch_`, `normalization_data_scratch_`) are properly released via `ReleaseNormalizationScratch()` at appropriate boundaries. `GetMemoryUsage()` accounts for the new data structures.

**Test coverage**: The new tests cover all three quantization types (FP32/FP16/SQ8), dirty-run lifecycle (thresholds, boundaries, explicit normalization), serialization round-trips (both ordinary and streaming), heap equivalence (KNN/range, pure/filtered), and randomized differential testing. This is thorough.

**One observation** (not blocking): In `Compact()`, `FinalizeInsertBatch()` is called at the start, but `Build()` already calls `NormalizeDirtyPostings()` (which internally calls `FinalizeInsertBatch()`) immediately before `Compact()`. The second call is a no-op (dirty map is already empty), so this is harmless defensive coding.

No new blocking issues found. The existing review thread covers the remaining suggestions.

LHT129 (commented):
I have reviewed the latest commit (25e45f80) which addresses the earlier review feedback from LHT129. The changes look good:

1. **`const_cast` removal** - `cal_memory_usage()` is now `const` and `current_memory_usage_` is `mutable std::atomic<uint64_t>`. The serialization paths no longer need `const_cast`. Clean solution.

2. **Suffix scan guard** - The second `ScanPostingRange` call in `query_impl` is now guarded by `if (selected.suffix_count > 0)`, avoiding the unnecessary function call for the common fully-normalized case.

3. **`normalize_dirty_postings_for_serialization()` extraction** - The normalization logic is now factored into a dedicated helper, reducing duplication between `Serialize` and `serialize_streaming_body`.

4. **Test coverage** - Comprehensive tests cover all three quantization types (FP32/FP16/SQ8), dirty-run thresholds and boundaries, KNN/range search with pure and filtered paths, serialization round-trips, randomized differential testing, and wire format version retention.

The core algorithm (binary-search partition of two sorted runs via `SelectPostingRuns`, merge-on-the-fly via `ForEachSelectedPosting`, and bounded normalization via `FinalizeInsertBatch`/`NormalizeDirtyPostings`) is sound. The performance data (3-10x Add throughput improvement) is compelling and the A/B testing confirms result-identical output.

No blocking issues found in the latest revision.

LHT129 (commented):
## Code Review Summary

I have completed a thorough review of this PR. The overall quality is high — the dirty posting mechanism is well-designed, the test coverage is comprehensive (98.28% changed-line coverage), and the existing review comments from LHT129 have addressed the main substantive issues.

### Issues already identified and addressed by the author:
- **const_cast elimination**: `current_memory_usage_` is now `mutable`, and `cal_memory_usage()` is `const`. ✅
- **ScanPostingRange guard**: The second `ScanPostingRange` call is now guarded by `if (selected.suffix_count > 0)`. ✅
- **Lock upgrade documentation**: The exclusive lock in `Serialize`/`serialize_streaming_body` is acknowledged as intentional. ✅

### Remaining items for consideration:

**[suggestion] `normalize_dirty_postings_for_serialization()` const-correctness**
The method is declared `const` in both the header (`sindi.h:360`) and implementation (`sindi.cpp:1578`), but it calls `window->NormalizeDirtyPostings()` which is a non-const mutating method. While this works because `const shared_ptr<T>&` yields a non-const `T*` via `operator->`, the `const` qualifier is misleading to readers. The header comment already notes "The caller must hold global_mutex_" which partially addresses this. Consider either removing `const` from this method and its callers (`Serialize`, `serialize_streaming_body`), or adding a brief comment explaining the logical const pattern.

**[note] `FinalizeInsertBatch` always returns `true` after processing**
As LHT129 noted, when every pending term is missing from `dirty_posting_prefixes_` (the `continue` branch at line 603), no posting state actually changed but the method still returns `true`. The caller in `SINDI::Add()` uses this return value to decide whether to call `cal_memory_usage()`. This is a minor inefficiency — `cal_memory_usage()` is cheap, so the impact is negligible.

**[note] Redundant `ReleaseNormalizationScratch()` in `NormalizeDirtyPostings`**
When `FinalizeInsertBatch()` already emptied `dirty_posting_prefixes_`, `NormalizeDirtyPostings` calls `ReleaseNormalizationScratch()` again (line 649) even though `FinalizeInsertBatch` already released it (line 633). The double-release is harmless (swap with already-empty vectors is a no-op), but slightly redundant.

### What this PR does well:
- Clean separation of sorted prefix + unsorted dirty suffix with bounded thresholds (MIN=32, MAX=256)
- Binary partition (`SelectPostingRuns`) correctly handles all edge cases including fully-dirty and fully-clean terms
- `ForEachSelectedPosting` template enables efficient merged iteration for heap insertion
- Memory accounting covers all new data structures
- Tests cover all three quantization types, serialization round-trips, filter correctness, and randomized stress scenarios
- Backward compatible wire format (version 1 retained)

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: The const qualifier is retained because Serialize and serialize_streaming_body override const methods in InnerIndexInterface. The documentation now states that this helper only canonicalizes the internal posting representation under the caller-held exclusive lock and does not change the logical index contents.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: FinalizeInsertBatch() now returns true only when at least one dirty term is actually processed. SINDI::Add() therefore recalculates memory usage only when posting state changes.

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: The const qualifier is retained because Serialize and serialize_streaming_body override const methods in InnerIndexInterface. The documentation now states that this helper only canonicalizes the internal posting representation under the caller-held exclusive lock and does not change the logical index contents.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: The sorting scratch vectors are now initialized only after the first valid dirty term is found. No scratch vectors are constructed when no pending term can be processed, and the same buffers are reused for the remaining terms.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -112,6 +113,72 @@ SparseTermDataCell::Query(float* global_dists,
...
+    uint32_t upper = std::min(retained_count, suffix_count);
+    while (lower <= upper) {
+        const auto take_suffix = lower + (upper - lower) / 2;
+        const auto take_prefix = retained_count - take_suffix;
+        if (take_prefix > 0 && take_suffix < suffix_count &&
+            PostingEntryBefore(term, prefix_count + take_suffix, take_prefix - 1)) {
+            lower = take_suffix + 1;
+            continue;
+        }
+        if (take_suffix > 0 && take_prefix < prefix_count &&
```
LHT129: [suggestion] The `return {};` is dead code because `CHECK_ARGUMENT(false, ...)` always throws `VsagException`. If the macro were ever changed to a non-throwing variant, returning `{0, 0, 0}` would silently produce incorrect query results (zero prefix and zero suffix means no postings would be scanned). Consider removing the return statement or replacing it with `__builtin_unreachable()`.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -112,6 +113,72 @@ SparseTermDataCell::Query(float* global_dists,
...
+            PostingEntryBefore(term, take_prefix, prefix_count + take_suffix - 1)) {
+            upper = take_suffix - 1;
```
LHT129: [note] Same dead-code-after-CHECK_ARGUMENT pattern as in `SelectPostingRuns` (flagged in an earlier review). The `return false;` after the switch is unreachable because `CHECK_ARGUMENT(false, ...)` always throws. If the macro were ever changed to a non-throwing variant, returning `false` would silently invert the comparison result for unrecognized quantization types.

Consider removing the return statement or replacing it with `__builtin_unreachable()` for consistency with the `SelectPostingRuns` fix.

LHT129 (commented):
I have completed a thorough review of this PR. The overall code quality is high, with well-structured changes and comprehensive test coverage.

1. **`sparse_term_datacell.h/cpp`** — Core of the optimization. The two-run posting list design (sorted prefix + bounded dirty suffix) with binary-search partition (`SelectPostingRuns`) and two-pointer merge (`NormalizePosting`, `ForEachSelectedPosting`) is correct and well-implemented. The `FinalizeInsertBatch` → `NormalizeDirtyPostings` lifecycle is properly integrated into `SINDI::Add`, `SINDI::Build`, and serialization paths.

2. **`sindi.cpp/h`** — The `cal_memory_usage()` const-correctness fix and `normalize_dirty_postings_for_serialization()` helper are clean. The exclusive lock upgrade in serialization is intentional and documented.

3. **`inner_index_interface.h`** — Making `current_memory_usage_` mutable is appropriate since it is a cached value, not logical state.

4. **Tests** — Excellent coverage: lifecycle tests (partial/full window, serialization boundaries), `SelectPostingRuns` correctness across all three quantization types, heap equivalence for KNN/range/filter paths, and a 20-seed randomized stress test. The test for serialization round-trip with dirty postings is particularly valuable.

**Existing review comments from LHT129 have been addressed:**
- `const_cast` eliminated via mutable `current_memory_usage_`
- Unnecessary `ScanPostingRange` call now guarded by `suffix_count > 0`
- `FinalizeInsertBatch` return value now accurately reflects whether changes occurred
- Scratch vector allocation is now lazy (only after first dirty term found)

**Remaining open items (already flagged by LHT129):**
- Dead code after `CHECK_ARGUMENT(false, ...)` in `SelectPostingRuns` and `PostingEntryBefore` — minor, non-blocking

The performance data in the PR description is thorough and convincing, with 1.6x–10x Add throughput improvements across all quantization types and batch sizes. The A/B verification confirms identical KNN/range/serialization results.

No new blocking issues found. The PR is in good shape.

src/algorithm/sindi/sindi.cpp
```diff
@@ -404,10 +404,14 @@ SINDI::Add(const DatasetPtr& base) {
     if (window_term_list_.size() > populated_window_count) {
         window_term_list_.resize(populated_window_count);
     }
+    bool posting_state_changed = false;
     for (int64_t window = first_affected_window; window <= last_affected_window; ++window) {
-        window_term_list_[window]->SortByValue();
+        posting_state_changed |= window_term_list_[window]->FinalizeInsertBatch();
```
LHT129: [note] In SINDI::Add(), when a window becomes full (total_count_ == window_size_), FinalizeInsertBatch() is called twice: first directly, then again inside NormalizeDirtyPostings(). The second call is a fast no-op since pending_posting_terms_ was already cleared, but it still performs the empty check and empty iteration. This is not a correctness issue, just a minor redundancy.

LHT129 (commented):
Thanks for this well-crafted optimization. The two-run posting list design with lazy normalization is a clean approach that delivers substantial incremental Add throughput improvements (up to 10x for single-vector batches) while preserving exact query semantics.

**Summary of findings from this review pass:**

The code quality is high. Most issues raised in earlier review rounds have been addressed:
- `const_cast` removed via `mutable` on `current_memory_usage_` and const `cal_memory_usage()`
- Suffix scan guarded by `suffix_count > 0` on the hot query path
- `FinalizeInsertBatch` return value now accurately reflects whether state changed
- Scratch vectors lazily initialized only when needed

**One minor note added:**
- `FinalizeInsertBatch()` is called twice when a window becomes full in `SINDI::Add()` (once directly, once inside `NormalizeDirtyPostings()`). The second call is a fast no-op but represents minor redundancy.

**Unaddressed from prior reviews (LHT129, Aug 18):**
- Dead code after `CHECK_ARGUMENT(false, ...)` in `SelectPostingRuns` (line 152) and `PostingEntryBefore` (line 536). These `return` statements are unreachable but harmless.

The test coverage is excellent — spanning all three quantization types, dirty-run thresholds, boundaries, serialization round-trips, randomized differential testing, and heap equivalence verification. The A/B benchmark methodology is rigorous and the results are compelling.

LHT129 (commented):
## Code Review Summary

I have completed a thorough review of this PR (commits `25e45f80` and `d92c1973`). The deferred posting-list normalization strategy is well-designed, and the implementation quality is high. The test coverage is comprehensive across all three quantization types (FP32, FP16, SQ8) with deterministic, randomized, lifecycle, serialization, and heap-equivalence coverage.

### What was already addressed (from LHT129's prior review)

- `const_cast` eliminated by making `current_memory_usage_` `mutable` and `cal_memory_usage()` `const`
- `FinalizeInsertBatch()` now correctly returns `posting_state_changed` (only `true` when actual work was done)
- Scratch vectors (`order`, `sorted_ids`, `sorted_data`) are lazily initialized inside `FinalizeInsertBatch`
- Suffix scan in `query_impl` is guarded by `suffix_count > 0`
- Documentation added for `normalize_dirty_postings_for_serialization` explaining the logical const contract

### Remaining open items (from LHT129's latest review round)

1. **[suggestion] Dead code after `CHECK_ARGUMENT` in `SelectPostingRuns`** (`sparse_term_datacell.cpp:144`): The `return {};` is unreachable because `CHECK_ARGUMENT(false, ...)` always throws. Consider replacing with `__builtin_unreachable()` or removing it.

2. **[note] Dead code after `CHECK_ARGUMENT` in `PostingEntryBefore`** (`sparse_term_datacell.cpp:146`): Same pattern — `return false;` after the switch-default `CHECK_ARGUMENT`. Consider the same treatment for consistency.

3. **[note] Redundant `FinalizeInsertBatch` call in `SINDI::Add()`** (`sindi.cpp:409-411`): When a window becomes full (`total_count_ == window_size_`), `FinalizeInsertBatch()` is called first at line 409, then `NormalizeDirtyPostings()` at line 411 calls `FinalizeInsertBatch()` again internally. The second call is a fast no-op (empty `pending_posting_terms_`), but the redundancy is worth noting.

### Additional observations

- The `SelectPostingRuns` binary search correctly partitions two sorted runs to find the top-k boundary. The edge cases (`retained_count == 0`, `retained_count == posting_count`, no dirty postings) are all handled correctly.
- The `ForEachSelectedPosting` merge loop correctly interleaves prefix and suffix entries in sorted order, matching the semantics of the old single-run scan.
- The exclusive lock change in `Serialize` and `serialize_streaming_body` is an intentional and necessary trade-off to ensure postings are fully normalized before writing. This is well-documented in the PR description.
- `NormalizeDirtyPostings` return value semantics are correct: returns `true` when it actually performed normalization work, `false` when there was nothing to normalize (including when `FinalizeInsertBatch` already handled everything).

### Verdict

This is a well-executed optimization with impressive performance results (up to 10x throughput improvement for single-vector incremental Add). The remaining open items are all non-blocking `[note]`/`[suggestion]` level. No critical issues found.

src/algorithm/sindi/sindi.cpp
```diff
@@ -404,10 +404,14 @@ SINDI::Add(const DatasetPtr& base) {
     if (window_term_list_.size() > populated_window_count) {
         window_term_list_.resize(populated_window_count);
     }
+    bool posting_state_changed = false;
     for (int64_t window = first_affected_window; window <= last_affected_window; ++window) {
```
LHT129: [note] The loop variable `window` is declared as `int64_t` (line 408) but is used to index `window_term_list_` which expects `size_t` (unsigned). While `first_affected_window` is derived from `cur_element_count_ / window_size_` and will be non-negative in practice, using a signed type for a container index is inconsistent with the rest of the codebase and could trigger signed/unsigned conversion warnings on some compilers.

Consider using `uint64_t` or `size_t` for the loop variable, and casting `first_affected_window` accordingly.

LHT129 (commented):
Overall this is a well-engineered optimization with thorough test coverage (98.28% changed-line coverage), comprehensive A/B benchmarks showing 1.57x–10.11x Add throughput improvements with no regression in search accuracy, and careful attention to serialization compatibility (wire version unchanged).

The core idea — deferring full posting-list normalization by maintaining a sorted prefix + bounded sorted dirty suffix — is sound and the merge-during-query approach preserves correctness. The lifecycle management (normalize at window-full, build boundary, compaction, and serialization) covers all necessary synchronization points.

Most of the detailed feedback has already been addressed in the existing review threads. The remaining open items are minor notes/suggestions that do not block merge.

LHT129 (commented):
I completed a thorough review of all changed files in this PR. The existing reviews from LHT129 have already covered the key issues, and the author has addressed the important ones:

1. **const_cast removal** — `cal_memory_usage()` is now `const` and `current_memory_usage_` is `mutable`. Clean solution.
2. **suffix_count guard in query_impl** — The hot-path `ScanPostingRange` call is now guarded by `if (selected.suffix_count > 0)`. Good.
3. **Exclusive lock in serialization** — Confirmed as intentional design choice, documented in the PR description.

The core algorithm (`SelectPostingRuns` binary search + `ForEachSelectedPosting` merge) is correct and well-tested across all three quantization types with randomized coverage. The `FinalizeInsertBatch` threshold heuristic (`clamp(prefix_count, 32, 256)`) is reasonable.

The remaining open `[note]` items (dead code after `CHECK_ARGUMENT`, double `FinalizeInsertBatch` call in `Add`, `int64_t` loop variable) are minor and non-blocking.

No new issues found. The test coverage is comprehensive (deterministic lifecycle, serialization, filtering, boundary, randomized, and three-quantization regression tests).

LHT129 (commented):
Reviewed the latest commit (d92c197). All three issues from the previous review round have been addressed:

1. **const_cast removed** — `cal_memory_usage()` is now `const`, `current_memory_usage_` is `mutable`, and `normalize_dirty_postings_for_serialization` is `const` with clear documentation explaining why physical mutation is safe under the caller-held exclusive lock.

2. **suffix_count > 0 guard added** — `query_impl` now guards the suffix scan with `if (selected.suffix_count > 0)`.

3. **Lazy scratch allocation** — `FinalizeInsertBatch` uses `std::optional<Vector<>>` for on-demand allocation of sort scratch buffers.

The implementation is solid:
- Two-run posting representation (sorted prefix + bounded dirty suffix) is well-designed
- Binary search partitioning in `SelectPostingRuns` correctly handles all edge cases
- Merge-on-read during queries (`ForEachSelectedPosting`, `InsertHeapByTermLists`) is correct
- Lock discipline is sound: exclusive lock for serialization (intentional, documented), shared lock for queries (unchanged)
- Tests cover all three quantization types, dirty posting lifecycle, and serialization round-trip
- No correctness issues found

One minor observation (not blocking): `FinalizeInsertBatch` returns `posting_state_changed = true` even when a term only had its suffix sorted without a merge (prefix_count == 0 case). This triggers a harmless `cal_memory_usage()` recalculation since the posting list size hasn't changed. This is on the write path, not the hot query path, so the impact is negligible.

LHT129 (commented):
This is a well-structured optimization with comprehensive test coverage and thorough benchmarking. The core idea of deferring full posting-list normalization by maintaining a sorted prefix + bounded sorted dirty suffix is sound.

The latest commit (d92c197) addresses several important review items: `const_cast` removal via `mutable` cache, lazy scratch-vector initialization with `std::optional`, correct `FinalizeInsertBatch` return value, and suffix-scan guard. The remaining open review threads (dead code after CHECK_ARGUMENT, int64_t loop variable, double FinalizeInsertBatch call in Add) are all `[note]` level and do not block merging.

The test coverage is excellent: deterministic lifecycle tests, serialization round-trip tests with all three quantization types, randomized differential tests, and heap-equivalence verification for both KNN and range search with filters. The A/B benchmarks show substantial incremental Add throughput improvements (1.6x-10x) with no regression in search accuracy or serialized output.

LHT129 (commented):
## Review Summary

This PR implements batch posting normalization for incremental SINDI Add, deferring full-list re-sorting and instead maintaining a sorted prefix plus a bounded dirty suffix. The merge-on-read strategy during query/pruning is sound, and normalization boundaries are correctly placed at full-window, build, compaction, and serialization points.

### What was already reviewed and addressed

Previous review rounds by @LHT129 covered several items, all of which have been resolved or acknowledged:

1. **`const_cast` in serialization** — Resolved by making `current_memory_usage_` `mutable` and `cal_memory_usage()` `const`.
2. **Exclusive lock in `Serialize`** — Confirmed as intentional design choice for normalization-before-write.
3. **Redundant `ScanPostingRange` call when no dirty postings** — Fixed with `if (selected.suffix_count > 0)` guard.
4. **Dead code after `CHECK_ARGUMENT`** in `PostingEntryBefore` — Noted as a defensive pattern consistent with the rest of the codebase.
5. **Double `FinalizeInsertBatch` call in `SINDI::Add()`** — Minor redundancy; the second call is a fast no-op.
6. **`int64_t` loop variable for container indexing** — Style note; non-negative in practice.

### Positive observations

- **Algorithm correctness**: The binary-search partition in `SelectPostingRuns` correctly finds the split point between two sorted runs. The merge logic in `ForEachSelectedPosting` and `NormalizePosting` is consistent and correct.
- **Test coverage**: Comprehensive — covers all three quantization types (FP32/FP16/SQ8), randomized stress tests with 20 seeds, serialization round-trips, heap equivalence (KNN + range, pure + filtered), and lifecycle boundaries.
- **Performance**: The benchmark data shows 1.6x–10x Add throughput improvement with no search regression, which is impressive.
- **Wire compatibility**: SINDI posting list format version remains 1; serialized bytes are unchanged.
- **Memory**: No material heap increase; scratch buffers are released after use via `ReleaseNormalizationScratch`.

No new blocking or substantive issues found. The implementation is solid.

LHT129 (commented):
This PR has already received a thorough review from LHT129 with 18 inline comments covering correctness, performance, code style, and maintainability concerns. The author has been responsive and addressed many of the initial findings in the second commit (d92c1973).

The remaining 4 unresolved comments are all `[suggestion]`/`[note]` level:
- Dead code after `CHECK_ARGUMENT` in `SelectPostingRuns` and `PostingEntryBefore` — minor style issue, safe as-is since `CHECK_ARGUMENT` always throws.
- Double `FinalizeInsertBatch()` call when a window becomes full in `SINDI::Add()` — the second call is a fast no-op, no correctness impact.
- `int64_t` loop variable for indexing `window_term_list_` — minor type inconsistency, safe in practice since the index is derived from `cur_element_count_ / window_size_`.

I have no additional blocking concerns to add. The PR is well-structured with comprehensive test coverage (unit tests, randomized stress tests, serialization round-trip tests across all three quantization types), thorough performance benchmarks, and clear documentation of the design decisions.

LHT129 (commented):
I have completed a thorough review of this PR. The changes have already been well-reviewed by LHT129, and the author has addressed the key feedback items (const_cast removal, ScanPostingRange guard, exclusive lock justification).

**Summary of my review:**

The core approach — maintaining a sorted prefix plus a bounded dirty suffix to defer full-list re-sorting during incremental Add — is sound and well-motivated. The binary search in `SelectPostingRuns` correctly partitions two sorted runs, and the merge logic in `ForEachSelectedPosting` and `NormalizePosting` is correct.

The test coverage is excellent: the PR adds deterministic lifecycle tests, serialization round-trip tests with wire format version verification, filtering boundary tests, randomized differential tests across all three quantization types (FP32/FP16/SQ8), and heap equivalence tests for both KNN and range search with pure and filtered paths.

The performance data is thorough and convincing, with 1.6x-10x Add throughput improvements depending on batch size, and search latency within normal variation.

**Remaining items from prior review (all [note] level):**

1. Dead code after `CHECK_ARGUMENT(false, ...)` in both `SelectPostingRuns` (`return {};`) and `PostingEntryBefore` (`return false;`) — these are unreachable but harmless.
2. Double `FinalizeInsertBatch()` call in `SINDI::Add()` when a window becomes full — the second call is a fast no-op.
3. `int64_t` loop variable for window indexing — minor type inconsistency.

None of these are blocking. The PR is in good shape for merging once the author decides on the remaining [note] items.

LHT129 (commented):
I have completed a thorough review of this PR. The existing review comments from LHT129 have already covered the key points:

1. The `const_cast` issue has been resolved by making `current_memory_usage_` mutable and `cal_memory_usage()` const.
2. The hot-path `suffix_count > 0` guard has been added to `query_impl`.
3. Dead-code-after-`CHECK_ARGUMENT` patterns in `SelectPostingRuns` and `PostingEntryBefore` have been noted.
4. The redundant `FinalizeInsertBatch` call in `Add()` when a window becomes full has been noted.
5. The signed loop variable for window indexing has been noted.

The PR is well-structured with comprehensive test coverage (deterministic lifecycle, serialization, filtering, boundary, randomized, and three-quantization regression tests). The performance data is thorough and convincing. The design of maintaining a sorted prefix + bounded sorted dirty suffix is sound, and the merge-during-pruning approach correctly preserves existing query semantics.

No new blocking issues found. The remaining open [note]-level comments are non-blocking and can be addressed at the author's discretion.

LHT129 (commented):
This is a follow-up review after the second commit (d92c197). The previous review by @LHT129 already covered the key issues thoroughly, and most actionable suggestions have been addressed in this revision.

**Summary of resolved issues from the first review round:**
- `const_cast` removed: `current_memory_usage_` is now `mutable`, `cal_memory_usage()` is `const` — clean.
- `suffix_count == 0` guard added in `query_impl` to skip the second `ScanPostingRange` — addressed.
- `FinalizeInsertBatch` now returns `true` only when at least one dirty term is processed — addressed.
- Scratch vector allocation in `FinalizeInsertBatch` is now deferred until the first valid dirty term — addressed.
- Serialization exclusive lock is confirmed as intentional and documented.

**Remaining open notes from @LHT129 (non-blocking):**
- Dead code after `CHECK_ARGUMENT` in `SelectPostingRuns` (`return {};`) and `PostingEntryBefore` (`return false;`).
- `FinalizeInsertBatch` called twice when a window becomes full in `SINDI::Add()` (once directly, once inside `NormalizeDirtyPostings`).
- `int64_t window` loop variable vs `size_t` container index.

**My assessment:**
The core algorithm — maintaining a sorted prefix + bounded dirty suffix, merging on-the-fly during queries, and normalizing at boundaries — is sound. The two-run merge logic in `SelectPostingRuns`, `ForEachSelectedPosting`, and `NormalizePosting` is correct. Test coverage is extensive (deterministic lifecycle, serialization round-trips, randomized differential tests across all three quantization types). The performance data in the PR description is thorough and shows significant incremental Add throughput improvements with no regression in search accuracy.

No new issues found beyond what has already been noted. The remaining notes are minor and do not block merge.

LHT129 (commented):
I have completed a thorough review of the latest commit (d92c197). The previous round of review comments has been well addressed:

1. The `const_cast` in serialization paths has been eliminated by making `cal_memory_usage()` const and `current_memory_usage_` mutable, with normalization factored into a dedicated `normalize_dirty_postings_for_serialization()` helper.
2. The hot-path `ScanPostingRange` call is now guarded by `if (selected.suffix_count > 0)`, avoiding the no-op function call for fully normalized postings.
3. The exclusive lock in serialization is an intentional design choice documented in the PR description.

The remaining open notes (dead code after `CHECK_ARGUMENT`, redundant `FinalizeInsertBatch` call in the full-window path, and `int64_t` loop variable) are minor and non-blocking.

The core algorithm — binary-search partitioning of two sorted posting runs, merge-based normalization, and bounded dirty-suffix thresholds — is correct and well-tested with comprehensive coverage across all three quantization types, randomized seeds, and lifecycle boundaries. No new issues found.

LHT129 (commented):
This is the third review round for this PR. The second commit (d92c1973) addressed 5 out of 9 issues from the previous review rounds:

**Resolved:**
1. `const_cast` eliminated — `current_memory_usage_` made `mutable`, `cal_memory_usage()` made `const` ✅
2. `ScanPostingRange` suffix guard added (`if (selected.suffix_count > 0)`) ✅
3. `FinalizeInsertBatch` now returns `posting_state_changed` (only `true` when at least one dirty term is processed) ✅
4. Scratch vectors (`order`, `sorted_ids`, `sorted_data`) are now lazily allocated via `std::optional` ✅
5. `normalize_dirty_postings_for_serialization` const documentation improved ✅

**Remaining (all `[note]` level, previously flagged by LHT129):**
- `PostingEntryBefore` dead code after `CHECK_ARGUMENT` (sparse_term_datacell.cpp:538)
- `SelectPostingRuns` dead code after `CHECK_ARGUMENT` (sparse_term_datacell.cpp:152-153)
- `FinalizeInsertBatch` called twice in `SINDI::Add()` — once directly, then inside `NormalizeDirtyPostings()` (sindi.cpp:409/411)
- `int64_t` loop variable for container indexing (sindi.cpp:408)

These are all minor style issues that do not affect correctness or performance. No new issues found in this review round.

**Overall assessment:** The implementation is solid. The core algorithm (deferred batch posting normalization with bounded dirty suffix) is correctly implemented with proper binary search partitioning, two-way merge, and threshold-triggered normalization. Test coverage is thorough across FP32/FP16/SQ8 quantization types, including randomized differential tests. Serialization correctly normalizes dirty state before writing. Memory accounting covers all new data structures.

LGTM.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -431,10 +496,179 @@ SparseTermDataCell::InsertVector(const SparseVector& sparse_base, uint16_t base_
...
+    return posting_state_changed;
+}
+
+void
+SparseTermDataCell::ReleaseNormalizationScratch() {
+    Vector<uint16_t> empty_ids(allocator_);
+    Vector<uint8_t> empty_data(allocator_);
+    normalization_ids_scratch_.swap(empty_ids);
+    normalization_data_scratch_.swap(empty_data);
+}
```
LHT129: [note] In `NormalizeDirtyPostings()`, `FinalizeInsertBatch()` is called first, which may have already normalized some terms (when `suffix_count >= normalize_threshold`). The subsequent loop iterates over the remaining entries in `dirty_posting_prefixes_` and calls `NormalizePosting()` for each. This is correct, but there is a subtle redundancy: `FinalizeInsertBatch()` already calls `ReleaseNormalizationScratch()` at its end, and then `NormalizeDirtyPostings()` calls it again after the loop. The second call is a harmless no-op since the scratch vectors were already swapped to empty, but it adds a minor code path that could confuse future readers.

Consider restructuring to avoid the double-release pattern, e.g. by having `FinalizeInsertBatch` not release scratch when called from `NormalizeDirtyPostings`, or by documenting that the double-release is intentional.

LHT129 (commented):
[note] In `Compact()`, the newly created `compact_pending_flags` vector is zero-initialized without copying the existing `pending_posting_flags_` values. While `FinalizeInsertBatch()` is called at the start of `Compact()` and clears flags for terms in `pending_posting_terms_`, there is a subtle edge case: if any terms exist in `dirty_posting_prefixes_` but were not tracked in `pending_posting_terms_` (the "unprocessed insert batch" scenario), their corresponding flags would be silently dropped after compaction. The current callers are safe because `SINDI::Build()` calls `NormalizeDirtyPostings()` before `Compact()`, but future callers of `Compact()` may not be aware of this precondition. Consider either copying the existing flags into `compact_pending_flags`, or documenting that `Compact()` expects dirty postings to already be normalized.

src/datacell/sparse_term_datacell.cpp
```diff
@@ -502,6 +748,7 @@ SparseTermDataCell::Compact() {
     Vector<std::unique_ptr<Vector<uint16_t>>> compact_ids(compact_term_capacity, allocator_);
     Vector<std::unique_ptr<Vector<uint8_t>>> compact_datas(compact_term_capacity, allocator_);
     Vector<uint32_t> compact_sizes(compact_term_capacity, 0, allocator_);
+    Vector<uint8_t> compact_pending_flags(compact_term_capacity, 0, allocator_);
     for (uint32_t i = 0; i < compact_term_capacity; ++i) {
         compact_sizes[i] = term_sizes_[i];
         if (term_sizes_[i] != 0) {
```
LHT129: [note] In `Compact()`, the newly created `compact_pending_flags` vector is zero-initialized without copying the existing `pending_posting_flags_` values. While `FinalizeInsertBatch()` is called at the start and clears flags for terms in `pending_posting_terms_`, there is a subtle edge case: if any terms exist in `dirty_posting_prefixes_` but were not tracked in `pending_posting_terms_` (the "unprocessed insert batch" scenario), their flags would be silently dropped after compaction. Current callers are safe because `SINDI::Build()` calls `NormalizeDirtyPostings()` before `Compact()`, but future callers may not be aware of this precondition. Consider either copying the existing flags or documenting the precondition.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: Removed the unreachable fallback return; the invalid partition branch terminates through CHECK_ARGUMENT.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: Removed the unreachable fallback return; the invalid quantization branch terminates through CHECK_ARGUMENT.

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: Full windows now call NormalizeDirtyPostings once; partial windows call FinalizeInsertBatch once.

src/algorithm/sindi/sindi.cpp
CharlesXu-HQ: Affected window tracking now uses std::optional<uint32_t> and uint32_t loop indices.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: NormalizeDirtyPostings finalizes without releasing scratch, then releases it exactly once at the end.

src/datacell/sparse_term_datacell.cpp
CharlesXu-HQ: Compact now normalizes dirty postings internally before shrinking storage.

LHT129 (commented):
## Review Summary

I reviewed all 10 changed files (779 additions, 106 deletions) and the 26 existing inline review comments have all been addressed by the author.

### What this PR does
Defers full sorting of posting lists during incremental `Add()` by maintaining a sorted prefix + bounded unsorted dirty suffix. Normalization (two-run merge) happens lazily at:
- Bounded thresholds (32-256 dirty entries)
- Build boundaries (`FinalizeInsertBatch`)
- Compaction (`Compact`)
- Serialization (`Serialize`)

### Assessment
- **Correctness**: The two-run merge logic in `SelectPostingRuns`/`NormalizePosting` is sound. The test coverage is thorough — equivalence tests against eager normalization across all quantization types (FP32/FP16/SQ8), both KNN and range search, with and without filters.
- **Lock safety**: `Serialize` upgrades from `shared_lock` to `scoped_lock` (exclusive) to normalize dirty postings before writing. This is intentional and documented.
- **Memory ordering**: `cur_element_count_` reads in `SINDI::add()` use `memory_order_relaxed` which is appropriate since the caller holds the exclusive lock.
- **No new issues found.** The existing 26 review comments have been resolved. The code is consistent with project patterns and ready to merge.

src/algorithm/sindi_v2/sindi_v2.cpp
```diff
@@ -593,8 +593,20 @@ SINDIV2::Add(const DatasetPtr& base) {
...
+                mutable_term_datacell->NormalizeDirtyPostings(window);
+            } else {
+                mutable_term_datacell->FinalizeInsertBatch(window);
+            }
+            if (window == last_affected_window.value()) {
+                break;
+            }
+        }
     }
     this->cal_memory_usage();
```
LHT129: [suggestion] In SINDI v2 `Add()`, `cal_memory_usage()` is called unconditionally at line 611 after the window normalization loop, regardless of whether any posting state actually changed. In contrast, SINDI v1 conditionally calls `cal_memory_usage()` only when `posting_state_changed` is true (tracked via the return values of `NormalizeDirtyPostings` and `FinalizeInsertBatch`).

This means SINDI v2 will recalculate memory usage on every `Add()` call even when no new postings were added (e.g., all vectors were rejected as duplicates). Consider tracking the return values from `NormalizeDirtyPostings` and `FinalizeInsertBatch` and only calling `cal_memory_usage()` when at least one window was actually modified.

src/datacell/mutable_sindi_term_datacell.cpp
```diff
@@ -210,6 +209,74 @@ MutableSindiTermDataCell::GetTermDictCount() const {
...
+    if (retained_count == 0) {
+        return {0, prefix_count, 0};
+    }
+    if (retained_count == posting_count) {
+        return {prefix_count, prefix_count, suffix_count};
+    }
+
+    uint32_t lower = retained_count > prefix_count ? retained_count - prefix_count : 0;
+    uint32_t upper = std::min(retained_count, suffix_count);
+    while (lower <= upper) {
```
LHT129, line 237: [note] The binary search in `SelectPostingRuns` partitions `retained_count` between the sorted prefix and the dirty suffix. When `retained_count == posting_count`, the method returns `{prefix_count, prefix_count, suffix_count}` — meaning the suffix scan starts at offset `prefix_count` and covers `suffix_count` elements.

This is correct, but there is a subtlety: the `ForEachSelectedPosting` template in the header uses `suffix` starting at `selected.suffix_offset` (= `prefix_count`) and iterates up to `suffix_end` (= `prefix_count + suffix_count` = `posting_count`). The `prefix` variable starts at 0 and iterates up to `prefix_count`. So the two runs cover `[0, prefix_count)` and `[prefix_count, posting_count)` respectively — a perfect partition of the full posting list.

This is correct but worth noting for future readers since the naming `suffix_offset` = `prefix_count` could be confusing at first glance.

src/datacell/mutable_sindi_term_datacell.cpp
```diff
@@ -703,7 +825,219 @@ MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) const {
...
+    const auto finalized = this->FinalizeInsertBatch(window, false);
+    if (window.dirty_posting_prefixes_.empty()) {
+        if (release_scratch) {
+            this->ReleaseNormalizationScratch();
+        }
+        return finalized;
+    }
+
+    Vector<uint32_t> dirty_terms(allocator_);
+    dirty_terms.reserve(window.dirty_posting_prefixes_.size());
```
LHT129, line 1016: [note] `NormalizeDirtyPostings(window, release_scratch)` collects dirty term IDs into a temporary `Vector<uint32_t>` before iterating and calling `NormalizePosting()` for each. This is necessary because `NormalizePosting()` calls `window.dirty_posting_prefixes_.erase(term)` which would invalidate iterators if iterating the map directly.

The approach is correct, but the temporary vector allocation is on the hot path for full-window normalization during `SINDI::Add()`. Since `FinalizeInsertBatch` is called first (with `release_scratch=false`) and may have already normalized some terms, the remaining dirty terms count is bounded. This is fine in practice but worth noting for memory-sensitive deployments.

src/datacell/mutable_sindi_term_datacell.cpp
```diff
@@ -703,7 +825,219 @@ MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) const {
...
+        window.pending_posting_flags_[term] = 0;
+    }
+    window.pending_posting_terms_.clear();
+    window.postings_sorted_ = true;
+    this->ReleaseNormalizationScratch();
+}
+
+bool
+MutableSindiTermDataCell::PostingEntryBefore(const MutableSINDIWindow& window,
+                                             uint32_t term,
```
LHT129, line 829: [note] `PostingEntryBefore` compares two posting entries by their quantized values (descending) with document ID as tiebreaker (ascending). The comparison uses `left_code > right_code` for the primary key and `ids[left] < ids[right]` for the tiebreaker.

This is consistent with `SortPostingListByValue` which sorts by value descending then by ID ascending. The merge logic in `ForEachSelectedPosting` and `NormalizePosting` both use this comparator, ensuring the two-run merge produces the same total order as a full sort.

One subtlety: the `compare` lambda captures `left` and `right` by value (indices), then uses `memcpy` to read the quantized values from `data`. This is correct because the indices are stable for the duration of the comparison. The `memcpy` approach avoids alignment issues with packed FP16/SQ8 data.

src/datacell/mutable_sindi_term_datacell.cpp
```diff
@@ -703,7 +825,219 @@ MutableSindiTermDataCell::SortByValue(MutableSINDIWindow& window) const {
...
+            sort_scratch.sorted_ids,
+            sort_scratch.sorted_data);
+
+        if (prefix_count == 0) {
+            window.dirty_posting_prefixes_.erase(dirty);
+            continue;
+        }
+        const auto normalize_threshold =
+            std::clamp(prefix_count, MIN_DIRTY_POSTING_SIZE, MAX_DIRTY_POSTING_SIZE);
+        if (suffix_count >= normalize_threshold) {
```
LHT129, line 964: [note] The normalization threshold in `FinalizeInsertBatch` is `std::clamp(prefix_count, MIN_DIRTY_POSTING_SIZE, MAX_DIRTY_POSTING_SIZE)` where MIN=32 and MAX=256. This means:

- For small posting lists (prefix_count < 32): threshold = 32, so normalization happens when suffix_count >= 32
- For medium posting lists (32 <= prefix_count <= 256): threshold = prefix_count, so normalization happens when suffix_count >= prefix_count (i.e., dirty suffix is at least as large as the sorted prefix)
- For large posting lists (prefix_count > 256): threshold = 256, so normalization happens when suffix_count >= 256

This is a well-designed adaptive threshold: it eagerly normalizes small lists (where the merge overhead is negligible), uses a proportional threshold for medium lists, and caps the dirty suffix at 256 for large lists (bounding the worst-case merge cost during search). The choice of 32 and 256 as bounds is reasonable for typical sparse vector workloads.

LHT129 (commented):
## Code Review Summary

This PR implements deferred posting normalization for SINDI incremental adds, a well-designed performance optimization. Instead of re-sorting the full posting list after each batch insert, it maintains a sorted prefix plus a bounded sorted dirty suffix, merging them on-the-fly during search.

### Architecture
The core idea is clean: `InsertVector` appends to posting lists and marks them dirty via `dirty_posting_prefixes_` (recording the prefix/suffix boundary) and `pending_posting_terms_` (tracking which terms need finalization). `FinalizeInsertBatch` sorts the dirty suffix and conditionally normalizes based on an adaptive threshold (clamped between 32 and 256). `NormalizeDirtyPostings` fully merges the two runs. During search, `SelectPostingRuns` uses binary search to partition the retained count between prefix and suffix, and `ForEachSelectedPosting` merges them on-the-fly.

### What I checked
- **Correctness**: The two-run merge logic in `ForEachSelectedPosting` and `NormalizePosting` correctly implements the same total order as `SortByValue`. The binary search in `SelectPostingRuns` correctly finds the partition point. The `PostingEntryBefore` comparator is consistent with the sort order.
- **Test coverage**: The new `two-run search matches eager normalization` test covers FP32/FP16/SQ8 quantization, KNN and range search, and filter callbacks. The `bounds dirty posting runs` test covers the threshold clamping behavior. The existing SINDI/SINDIV2 integration tests verify the end-to-end flow including serialization.
- **Serialization**: Correctly normalizes dirty postings before writing, with exclusive lock acquisition. The `postings_sorted` version flag in the serialized format allows deserialization to skip re-sorting for versioned data.
- **Memory**: Scratch buffers (`normalization_ids_scratch_`, `normalization_data_scratch_`) are properly released via `ReleaseNormalizationScratch()` and accounted for in `GetMemoryUsage()`.

### Minor observations (posted as inline comments)
1. SINDI v2 `Add()` unconditionally calls `cal_memory_usage()` while SINDI v1 conditionally calls it only when posting state changes — a minor inconsistency.
2. The `SelectPostingRuns` suffix_offset = prefix_count naming could be clarified for future readers.
3. The temporary vector allocation in `NormalizeDirtyPostings` for collecting dirty terms is correct (needed because `NormalizePosting` erases from the map) but worth noting for memory-sensitive deployments.
4. The adaptive normalization threshold design (clamp between 32 and 256) is well-reasoned.

### Verdict
The implementation is correct, well-tested, and the design is sound. The deferred normalization approach is a genuine performance optimization that reduces the cost of incremental adds while keeping search overhead bounded. No blocking issues found.

src/algorithm/sindi_v2/sindi_v2.cpp
CharlesXu-HQ: Updated SINDI v2 to track posting-state changes and recalculate memory usage only when normalization or finalization modifies a window.

PR #2699, opened 2026-08-17
fix: include cstdint in binaryset header

### LINKED ISSUE #2698: [bug](build): make the binaryset header self-contained
### Affected area

build

### Interface

cpp

### VSAG version

v1.0.0-100-g2044122f

### Summary

`include/vsag/binaryset.h` uses `uint64_t` without directly including `<cstdint>`. Translation units that include this public header without another header providing `uint64_t` fail to compile.

### Minimal reproducer

```cpp
#include "vsag/binaryset.h"
```

Compile with:

```shell
c++ -std=c++17 -Iinclude -fsyntax-only reproducer.cpp
```

### Environment

```shell
OS: Ubuntu 26.04
VSAG version: v1.0.0-100-g2044122f
Compiler version: c++ (Ubuntu 15.2.0-16ubuntu1) 15.2.0
```

### Expected behavior

Every public header should include the standard headers required for the types it uses. A translation unit containing only `#include "vsag/binaryset.h"` should compile successfully.

### Actual behavior

Compilation fails because `uint64_t` is undefined:

```text
include/vsag/binaryset.h:32:5: error: ‘uint64_t’ does not name a type
note: ‘uint64_t’ is defined in header ‘<cstdint>’
```

The same error causes `make debug` to fail while compiling `tests/fixtures/framework/test_reader.cpp`.

### References

- `include/vsag/binaryset.h:17`
- `include/vsag/binaryset.h:33`
- `tests/fixtures/framework/test_reader.h:22`

### Additional context

CMake successfully enables ccache and compilation proceeds normally until the missing standard header is encountered. This is a source-level header self-containment issue, not a ccache or filesystem-permission problem.

## Related issues

_None found._

---

wxyucs:
## What changed

- Include `<cstdint>` directly from the public `binaryset.h` header.

## Why

`binaryset.h` declares a `uint64_t` member but did not include the standard header that defines it. Translation units without an incidental `<cstdint>` include failed to compile, including the debug build's `fixtures_framework` target under GCC 15.

## Impact

The public header is now self-contained. There is no API or ABI change.

## Validation

- Compiled a translation unit containing only `#include "vsag/binaryset.h"` with `c++ -std=c++17 -Iinclude -fsyntax-only`.
- Ran `make debug` successfully to 100%.
- Ran `git diff --check` successfully.

Fixes: #2698

### DIFF
--- a/include/vsag/binaryset.h
+++ b/include/vsag/binaryset.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <algorithm>
+#include <cstdint>
 #include <memory>
 #include <set>
 #include <string>


LHT129 (approved):
LGTM

Issue #2702, opened 2026-08-17
[improve](other): eliminate transformquantizer hot-path allocations


LHT129:
## Kind of improvement
perf

## Affected area
other

## Summary
TransformQuantizer currently allocates transformer metadata and temporary scratch storage for each transformed vector. In multi-stage chains, it also copies every intermediate result back to the previous buffer. This increases allocator pressure and reference-count work in training, encoding, and query preprocessing.

The change uses optional caller-provided metadata output and caller-owned ping-pong scratch buffers. Query scratch remains private to each Computer instance; PCA keeps its existing four-byte metadata layout for code compatibility.

## Expected benefit
- Eliminate empty TransformerMeta heap allocations for FHT, ROM, and MRLE on every transformed vector.
- Reuse scratch across TrainImpl and EncodeBatchImpl loops, and retain scratch per query Computer without shared mutable quantizer state.
- Avoid an intermediate copy for each transform-chain stage while preserving encoded-code and distance-recovery behavior.

## Benchmark evidence
No before/after benchmark has been captured yet. Before merging, measure allocations and per-vector latency using a fixed 100,000-vector, 768-dimensional dataset with the rom,pca,fp32 chain, fixed allocator, BLAS, and thread configuration. Report encode throughput, SetQuery latency, recall, and peak memory using benchs/ or tools/eval/.

## References
- src/quantization/transform_quantization/transform_quantizer.h:249
- src/impl/transform/vector_transformer.h:53
- src/quantization/computer.h:105
- Related PR: https://github.com/antgroup/vsag/pull/2703

## Related issues
- #2163 closed: TransformQuantizer ProcessQueryImpl memory leak

## Additional context
The scope is limited to TQ wrapper allocations. Per-vector temporary allocation inside a concrete quantizer is out of scope unless it is directly caused by the wrapper.

_Drafted with AI assistant: Codex: GPT-5_

PR #2703, opened 2026-08-17
perf: eliminate TQ transform allocations

LHT129:
Summary: remove per-vector transform metadata allocation and reuse caller-owned scratch buffers. Query scratch is private to each computer; PCA metadata code layout remains unchanged.\n\nValidation: targeted TransformQuantizer and transformer tests pass after rebuilding against main; changed transformer implementations pass clang-tidy-15.\n\nRefs #2702

### DIFF
--- a/src/impl/transform/mrle_transformer.h
+++ b/src/impl/transform/mrle_transformer.h
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <cstring>
+
 #include "metric_type.h"
 #include "simd/normalize.h"
 #include "vector_transformer.h"
@@ -33,14 +35,14 @@ public:
 
     ~MRLETransformer() override = default;
 
-    TransformerMetaPtr
-    Transform(const float* original_vec, float* transformed_vec) const override {
-        auto meta = std::make_shared<MRLETMeta>();
+    void
+    Transform(const float* original_vec,
+              float* transformed_vec,
+              uint8_t* /*meta*/ = nullptr) const override {
         memcpy(transformed_vec, original_vec, this->output_dim_ * sizeof(float));
         if constexpr (metric == MetricType::METRIC_TYPE_COSINE) {
             Normalize(transformed_vec, transformed_vec, this->output_dim_);
         }
-        return meta;
     }
 
     void
--- a/src/impl/transform/pca_transformer.cpp
+++ b/src/impl/transform/pca_transformer.cpp
@@ -17,6 +17,7 @@
 
 #include <fmt/format.h>
 
+#include <cstring>
 #include <random>
 
 #include "impl/blas/blas_function.h"
@@ -62,9 +63,8 @@ PCATransformer::Train(const float* data, uint64_t count) {
     PerformEigenDecomposition(covariance_matrix.data());
 }
 
-TransformerMetaPtr
-PCATransformer::Transform(const float* input_vec, float* output_vec) const {
-    auto meta = std::make_shared<PCAMeta>();
+void
+PCATransformer::Transform(const float* input_vec, float* output_vec, uint8_t* meta) const {
     vsag::Vector<float> centralized_vec(allocator_);
     centralized_vec.resize(input_dim_, 0.0F);
 
@@ -89,7 +89,12 @@ PCATransformer::Transform(const float* input_vec, float* output_vec) const {
                         output_vec,
                         1);
 
-    return meta;
+    if (meta != nullptr) {
+        // Preserve the existing serialized code layout and value. The metadata is
+        // currently not consumed by RecoveryDistance.
+        const float residual_norm = 0.0F;
+        std::memcpy(meta, &residual_norm, sizeof(residual_norm));
+    }
 }
 
 void
--- a/src/impl/transform/vector_transformer.h
+++ b/src/impl/transform/vector_transformer.h
@@ -50,10 +50,8 @@ public:
 
     virtual ~VectorTransformer() = default;
 
-    virtual TransformerMetaPtr
-    Transform(const float* input_vec, float* output_vec) const {
-        return nullptr;
-    };
+    virtual void
+    Transform(const float* input_vec, float* output_vec, uint8_t* meta = nullptr) const = 0;
 
     virtual void
     Serialize(StreamWriter& writer) const = 0;
--- a/src/quantization/transform_quantization/transform_quantizer.h
+++ b/src/quantization/transform_quantization/transform_quantizer.h
@@ -114,13 +114,23 @@ public:
     void
     TransformBaseVector(const float* input, Vector<float>& output) const;
 
+    bool
+    EncodeOneWithScratch(const float* data,
+                         uint8_t* codes,
+                         Vector<float>& primary_scratch,
+                         Vector<float>& secondary_scratch) const;
+
 public:
     VectorTransformerPtr
     MakeTransformerInstance(std::string transform_str,
                             const VectorTransformerParameter& param) const;
 
-    void
-    ExecuteChainTransform(float* prev_data, const uint32_t* meta_offsets, uint8_t* codes) const;
+    const float*
+    ExecuteChainTransform(const float* input,
+                          const uint32_t* meta_offsets,
+                          uint8_t* codes,
+                          Vector<float>& primary_scratch,
+                          Vector<float>& secondary_scratch) const;
 
     float
     ExecuteChainDistanceRecovery(float quantize_dist,
@@ -245,11 +255,13 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
     // 2. execute transform on original data
     const uint64_t transformed_dim = this->GetTransformedDim();
     Vector<float> transformed_data(transformed_dim * count, 0, this->allocator_);
-    Vector<float> transformed_vector(this->allocator_);
+    Vector<float> primary_scratch(this->allocator_);
+    Vector<float> secondary_scratch(this->allocator_);
     for (uint64_t i = 0; i < count; ++i) {
-        this->TransformBaseVector(data + i * this->dim_, transformed_vector);
+        const auto* transformed = ExecuteChainTransform(
+            data + i * this->dim_, nullptr, nullptr, primary_scratch, secondary_scratch);
         memcpy(transformed_data.data() + i * transformed_dim,
-               transformed_vector.data(),
+               transformed,
                transformed_dim * sizeof(float));
     }
 
@@ -259,43 +271,75 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
 }
 
 template <typename QuantTmpl, MetricType metric>
-void
-TransformQuantizer<QuantTmpl, metric>::ExecuteChainTransform(float* prev_data,
-                                                             const uint32_t* meta_offsets,
-                                                             uint8_t* codes) const {
-    Vector<float> next_data(this->dim_, 0, this->allocator_);
-
-    for (uint32_t i = 0; i < this->transform_chain_.size(); i++) {
-        auto vector_transformer = this->transform_chain_[i];
-        auto meta = vector_transformer->Transform(prev_data, next_data.data());
-        if (codes != nullptr) {
-            meta->EncodeMeta(codes + meta_offsets[i]);
-        }
-
-        memcpy(prev_data, next_data.data(), vector_transformer->GetOutputDim() * sizeof(float));
+const float*
+TransformQuantizer<QuantTmpl, metric>::ExecuteChainTransform(
+    const float* input,
+    const uint32_t* meta_offsets,
+    uint8_t* codes,
+    Vector<float>& primary_scratch,
+    Vector<float>& secondary_scratch) const {
+    if (primary_scratch.size() < this->dim_) {
+        primary_scratch.resize(this->dim_, 0.0F);
     }
+
+    const float* current = input;
+    float* next = primary_scratch.data();
+    for (uint32_t i = 0; i < this->transform_chain_.size(); i++) {
+        const auto& vector_transformer = this->transform_chain_[i];
+        uint8_t* meta = nullptr;
+        if (codes != nullptr && meta_offsets != nullptr && vector_transformer->GetMetaSize() > 0) {
+            meta = codes + meta_offsets[i];
+        }
+        vector_transformer->Transform(current, next, meta);
+        current = next;
+
+        if (i + 1 < this->transform_chain_.size()) {
+            if (next == primary_scratch.data()) {
+                if (secondary_scratch.size() < this->dim_) {
+                    secondary_scratch.resize(this->dim_, 0.0F);
+                }
+                next = secondary_scratch.data();
+            } else {
+                next = primary_scratch.data();
+            }
+        }
+    }
+    return current;
 }
 
 template <typename QuantTmpl, MetricType metric>
 void
 TransformQuantizer<QuantTmpl, metric>::TransformBaseVector(const float* input,
                                                            Vector<float>& output) const {
-    output.assign(input, input + this->dim_);
-    ExecuteChainTransform(output.data(), nullptr, nullptr);
-    output.resize(this->GetTransformedDim());
+    Vector<float> secondary_scratch(this->allocator_);
+    const auto* transformed =
+        ExecuteChainTransform(input, nullptr, nullptr, output, secondary_scratch);
+    if (transformed == output.data()) {
+        output.resize(this->GetTransformedDim());
+    } else {
+        output.assign(transformed, transformed + this->GetTransformedDim());
+    }
+}
+
+template <typename QuantTmpl, MetricType metric>
+bool
+TransformQuantizer<QuantTmpl, metric>::EncodeOneWithScratch(
+    const float* data,
+    uint8_t* codes,
+    Vector<float>& primary_scratch,
+    Vector<float>& secondary_scratch) const {
+    const auto* transformed = ExecuteChainTransform(
+        data, base_meta_offsets_.data(), codes, primary_scratch, secondary_scratch);
+    return quantizer_->EncodeOne(transformed, codes);
 }
 
 template <typename QuantTmpl, MetricType metric>
 bool
 TransformQuantizer<QuantTmpl, metric>::EncodeOneImpl(const float* data, uint8_t* codes) const {
-    // 1. execute transform
-    Vector<float> data_buffer(this->dim_, 0, this->allocator_);
-    data_buffer.assign(data, data + this->dim_);
-    ExecuteChainTransform(data_buffer.data(), base_meta_offsets_.data(), codes);
-
-    // 2. execute quantize
-    return quantizer_->EncodeOne(data_buffer.data(), codes);
-};
+    Vector<float> primary_scratch(this->allocator_);
+    Vector<float> secondary_scratch(this->allocator_);
+    return EncodeOneWithScratch(data, codes, primary_scratch, secondary_scratch);
+}
 
 template <typename QuantTmpl, MetricType metric>
 void
@@ -313,14 +357,15 @@ TransformQuantizer<QuantTmpl, metric>::ProcessQueryImpl(
     }
 
     // 1. execute transform
-    Vector<float> data_buffer(this->dim_, 0, this->allocator_);
-    data_buffer.assign(query, query + this->dim_);
-    ExecuteChainTransform(
-        data_buffer.data(), query_meta_offsets_.data(), computer.inner_computer_->buf_);
+    const auto* transformed = ExecuteChainTransform(query,
+                                                    query_meta_offsets_.data(),
+                                                    computer.inner_computer_->buf_,
+                                                    computer.primary_scratch_,
+                                                    computer.secondary_scratch_);
 
     // 2. execute quantize
-    // note that only when computer.buf_ == nullptr, quantizer_ will allocate data to buf_
-    quantizer_->ProcessQuery(data_buffer.data(), *computer.inner_computer_);
+    // note that only when computer.inner_computer_->buf_ == nullptr, quantizer_ will allocate its query buffer
+    quantizer_->ProcessQuery(transformed, *computer.inner_computer_);
 };
 
 template <typename QuantTmpl, MetricType metric>
@@ -332,7 +377,7 @@ TransformQuantizer<QuantTmpl, metric>::ExecuteChainDistanceRecovery(float quanti
                                                                     const uint8_t* codes_2) const {
     auto dist = quantize_dist;
     for (uint32_t i = 0; i < this->transform_chain_.size(); i++) {
-        auto vector_transformer = this->transform_chain_[i];
+        const auto& vector_transformer = this->transform_chain_[i];
         const auto* meta_1 = codes_1 + meta_offsets_1[i];
         const auto* meta_2 = codes_2 + meta_offsets_2[i];
 
@@ -395,8 +440,15 @@ bool
 TransformQuantizer<QuantTmpl, metric>::EncodeBatchImpl(const float* data,
                                                        uint8_t* codes,
                                                        uint64_t count) const {
+    Vector<float> primary_scratch(this->allocator_);
+    Vector<float> secondary_scratch(this->allocator_);
     for (uint64_t i = 0; i < count; ++i) {
-        EncodeOneImpl(data + i * this->dim_, codes + i * this->code_size_);
+        if (not EncodeOneWithScratch(data + i * this->dim_,
+                                     codes + i * this->code_size_,
+                                     primary_scratch,
+                                     secondary_scratch)) {
+            return false;
+        }
     }
     return true;
 }
--- a/src/impl/transform/fht_kac_rotate_transformer.cpp
+++ b/src/impl/transform/fht_kac_rotate_transformer.cpp
@@ -61,9 +61,8 @@ FhtKacRotator::Train(const float* data, uint64_t count) {
     this->Train();
 }
 
-TransformerMetaPtr
-FhtKacRotator::Transform(const float* data, float* rotated_vec) const {
-    auto meta = std::make_shared<FHTMeta>();
+void
+FhtKacRotator::Transform(const float* data, float* rotated_vec, uint8_t* /*meta*/) const {
     auto dim = static_cast<uint64_t>(this->input_dim_);
     std::memcpy(rotated_vec, data, sizeof(float) * dim);
     if (trunc_dim_ == dim) {
@@ -72,7 +71,7 @@ FhtKacRotator::Transform(const float* data, float* rotated_vec) const {
             FHTRotate(rotated_vec, trunc_dim_);
             VecRescale(rotated_vec, trunc_dim_, fac_);
         }
-        return meta;
+        return;
     }
 
     uint64_t start = dim - trunc_dim_;
@@ -90,8 +89,6 @@ FhtKacRotator::Transform(const float* data, float* rotated_vec) const {
     }
     VecRescale(rotated_vec, dim, 0.25F);
     //origin vec(x,y), after kacs_walk_generic() -> (x+y, x-y),should be resize by sqrt(0.5) for each KacsWalk() to make the len of vector consistent
-
-    return meta;
 }
 void
 FhtKacRotator::InverseTransform(float const* data, float* rotated_vec) const {
--- a/src/impl/transform/fht_kac_rotate_transformer.h
+++ b/src/impl/transform/fht_kac_rotate_transformer.h
@@ -28,8 +28,8 @@ public:
 
     ~FhtKacRotator() override = default;
 
-    TransformerMetaPtr
-    Transform(const float* data, float* rotated_vec) const override;
+    void
+    Transform(const float* data, float* rotated_vec, uint8_t* meta = nullptr) const override;
 
     void
     InverseTransform(const float* data, float* rotated_vec) const override;
--- a/src/impl/transform/mrle_transformer_test.cpp
+++ b/src/impl/transform/mrle_transformer_test.cpp
@@ -35,8 +35,7 @@ TEST_CASE("MRLETransformer L2 Transform Test", "[ut][MRLETransformer]") {
         input[i] = static_cast<float>(i + 1);
     }
 
-    auto meta = transformer.Transform(input.data(), output.data());
-    REQUIRE(meta != nullptr);
+    transformer.Transform(input.data(), output.data());
 
     for (int64_t i = 0; i < dim; ++i) {
         REQUIRE(std::abs(output[i] - input[i]) < EPSILON);
@@ -55,8 +54,7 @@ TEST_CASE("MRLETransformer Cosine Transform Test", "[ut][MRLETransformer]") {
         input[i] = static_cast<float>(i + 1);
     }
 
-    auto meta = transformer.Transform(input.data(), output.data());
-    REQUIRE(meta != nullptr);
+    transformer.Transform(input.data(), output.data());
 
     float sum = 0.0F;
     for (int64_t i = 0; i < dim; ++i) {
--- a/src/impl/transform/pca_transformer.h
+++ b/src/impl/transform/pca_transformer.h
@@ -62,8 +62,8 @@ public:
     void
     Deserialize(StreamReader& reader) override;
 
-    TransformerMetaPtr
-    Transform(const float* input_vec, float* output_vec) const override;
+    void
+    Transform(const float* input_vec, float* output_vec, uint8_t* meta = nullptr) const override;
 
     void
     InverseTransform(const float* input_vec, float* output_vec) const override;
--- a/src/impl/transform/random_orthogonal_transformer.cpp
+++ b/src/impl/transform/random_orthogonal_transformer.cpp
@@ -55,9 +55,10 @@ RandomOrthogonalMatrix::CopyOrthogonalMatrix(float* out_matrix) const {
               out_matrix);
 }
 
-TransformerMetaPtr
-RandomOrthogonalMatrix::Transform(const float* original_vec, float* transformed_vec) const {
-    auto meta = std::make_shared<ROMMeta>();
+void
+RandomOrthogonalMatrix::Transform(const float* original_vec,
+                                  float* transformed_vec,
+                                  uint8_t* /*meta*/) const {
     // perform matrix-vector multiplication: y = Q * x
     auto dim = static_cast<int32_t>(this->input_dim_);
     BlasFunction::Sgemv(BlasFunction::RowMajor,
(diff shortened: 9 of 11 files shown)


src/impl/transform/pca_transformer.cpp
```diff
@@ -89,7 +89,12 @@ PCATransformer::Transform(const float* input_vec, float* output_vec) const {
                         output_vec,
                         1);
 
-    return meta;
+    if (meta != nullptr) {
+        // Preserve the existing serialized code layout and value. The metadata is
+        // currently not consumed by RecoveryDistance.
+        const float residual_norm = 0.0F;
```
LHT129, line 95: [note] The `residual_norm` is hardcoded to `0.0F` here. The comment indicates this preserves the existing serialized code layout, and the metadata is currently not consumed by `RecoveryDistance`. If `RecoveryDistance` is expected to consume this field in the future, consider adding a TODO or tracking issue reference so the placeholder is not forgotten.

```cpp
const float residual_norm = 0.0F;
std::memcpy(meta, &residual_norm, sizeof(residual_norm));
```

src/impl/transform/pca_transformer.cpp
```diff
@@ -62,9 +63,8 @@ PCATransformer::Train(const float* data, uint64_t count) {
     PerformEigenDecomposition(covariance_matrix.data());
 }
 
-TransformerMetaPtr
-PCATransformer::Transform(const float* input_vec, float* output_vec) const {
-    auto meta = std::make_shared<PCAMeta>();
+void
+PCATransformer::Transform(const float* input_vec, float* output_vec, uint8_t* meta) const {
     vsag::Vector<float> centralized_vec(allocator_);
     centralized_vec.resize(input_dim_, 0.0F);
```
LHT129, line 69: [suggestion] `PCATransformer::Transform` still allocates a `Vector<float> centralized_vec` on every call, which undermines the goal of eliminating hot-path allocations. Since the caller (`ExecuteChainTransform`) now provides scratch buffers, consider accepting an additional scratch buffer parameter or using a pre-allocated member buffer to avoid this per-call allocation.

```cpp
// Current code still allocates per call:
vsag::Vector<float> centralized_vec(allocator_);
centralized_vec.resize(input_dim_, 0.0F);
```

LHT129 (commented):
Thanks for this PR. The change to eliminate per-vector `make_shared` allocations in the transform hot path is a solid performance improvement. The ping-pong scratch buffer pattern in `ExecuteChainTransform` is clean and correctly handles the input/output pointer swapping.

I have reviewed the full diff and the existing inline comments from Copilot and the author already cover the notable points:

1. **`TransformBaseVector` still allocates `secondary_scratch` per call** — this path is used by `BottomQuantizerAccessor::PrepareBottomInput`, so the allocation overhead remains there. Consider plumbing a caller-owned buffer.

2. **`PCATransformer::Transform` still allocates `centralized_vec` per call** — this undermines the goal of eliminating hot-path allocations for PCA-based configurations. A pre-allocated member buffer or an additional scratch parameter would help.

3. **`meta_offsets` null-safety in `ExecuteChainTransform`** — when `codes != nullptr` but `meta_offsets` is null, the pointer arithmetic `codes + meta_offsets[i]` would dereference null. In practice the current callers always pass valid offsets, but a defensive check or assertion would make the contract clearer.

4. **`VectorTransformer::Transform` empty default** — making this pure-virtual would catch missing overrides at compile time rather than silently producing uninitialized output.

5. **Missing `<cstring>` include in `mrle_transformer.h`** — currently relies on transitive includes.

6. **PCA `residual_norm` hardcoded to 0.0F** — the comment explains this preserves the existing serialized layout and the field is not consumed by `RecoveryDistance`. A TODO or tracking issue reference would help ensure this placeholder is not forgotten if `RecoveryDistance` starts consuming it.

No blocking issues found. The core change — replacing `TransformerMetaPtr` returns with caller-owned `uint8_t*` scratch — is correct and well-executed.

src/impl/transform/pca_transformer.cpp
LHT129: The zero value is the legacy PCAMeta value and RecoveryDistance does not consume it. The existing compatibility comment records that invariant. Adding an operational TODO would imply planned behavior that this PR does not introduce, so I am leaving the stable layout unchanged.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,42 +265,64 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
+    float* next = primary_scratch.data();
     for (uint32_t i = 0; i < this->transform_chain_.size(); i++) {
-        auto vector_transformer = this->transform_chain_[i];
-        auto meta = vector_transformer->Transform(prev_data, next_data.data());
-        if (codes != nullptr) {
-            meta->EncodeMeta(codes + meta_offsets[i]);
+        const auto& vector_transformer = this->transform_chain_[i];
+        auto* meta = codes == nullptr ? nullptr : codes + meta_offsets[i];
+        vector_transformer->Transform(current, next, meta);
+        current = next;
```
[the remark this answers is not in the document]
LHT129: Fixed in 8ce20ddb. The code now asserts the offsets contract and only passes a metadata pointer when the transformer has nonzero metadata size.

src/impl/transform/vector_transformer.h
```diff
@@ -50,10 +50,8 @@ class VectorTransformer {
 
     virtual ~VectorTransformer() = default;
 
-    virtual TransformerMetaPtr
-    Transform(const float* input_vec, float* output_vec) const {
-        return nullptr;
-    };
+    virtual void
+    Transform(const float* input_vec, float* output_vec, uint8_t* meta = nullptr) const {};
```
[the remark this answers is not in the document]
LHT129: Fixed in 8ce20ddb. Transform is now pure virtual, so missing derived implementations fail at compile time.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,42 +265,64 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
 template <typename QuantTmpl, MetricType metric>
 void
 TransformQuantizer<QuantTmpl, metric>::TransformBaseVector(const float* input,
                                                            Vector<float>& output) const {
-    output.assign(input, input + this->dim_);
-    ExecuteChainTransform(output.data(), nullptr, nullptr);
-    output.resize(this->GetTransformedDim());
+    Vector<float> secondary_scratch(this->allocator_);
+    const auto* transformed =
+        ExecuteChainTransform(input, nullptr, nullptr, output, secondary_scratch);
```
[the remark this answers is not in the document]
LHT129: Not changed in this PR. The only caller is the generic BottomQuantizerAccessor API, which currently accepts exactly one caller scratch buffer. Threading a scratch pair through that generic accessor is a broader API and lifecycle change. The hot TrainImpl, EncodeBatchImpl, and query Computer paths already reuse both scratch buffers; I will keep the accessor extension as follow-up work.

src/impl/transform/mrle_transformer.h
```diff
@@ -33,14 +33,14 @@ class MRLETransformer : public VectorTransformer {
...
 
-    TransformerMetaPtr
-    Transform(const float* original_vec, float* transformed_vec) const override {
-        auto meta = std::make_shared<MRLETMeta>();
+    void
+    Transform(const float* original_vec,
+              float* transformed_vec,
+              uint8_t* /*meta*/ = nullptr) const override {
         memcpy(transformed_vec, original_vec, this->output_dim_ * sizeof(float));
         if constexpr (metric == MetricType::METRIC_TYPE_COSINE) {
```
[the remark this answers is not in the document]
LHT129: Fixed in 8ce20ddb by adding the direct cstring include.

src/impl/transform/pca_transformer.cpp
LHT129: Not changed in this PR. A transformer member scratch buffer would introduce a data race for concurrent quantizer use, while extending every transformer call with additional scratch changes the polymorphic API beyond this TQ wrapper allocation fix. PCA internal workspace optimization should be handled separately with an explicit concurrency-safe scratch design.

LHT129 (commented):
I have reviewed this PR. The changes look well-structured and correctly eliminate per-vector `shared_ptr` allocations in the TransformQuantizer hot path by:

1. **Interface change**: `VectorTransformer::Transform()` now takes a raw `uint8_t* meta` pointer instead of returning a `shared_ptr`, and is pure virtual — forcing all derived classes to explicitly implement it.
2. **Scratch buffer reuse**: `Computer<TransformQuantizer>` now owns `primary_scratch_` and `secondary_scratch_` members, which are reused across `ProcessQuery` calls.
3. **Ping-pong buffering**: `ExecuteChainTransform` uses caller-provided scratch buffers with ping-pong swapping between transform chain steps, eliminating internal `Vector<float>` allocations.
4. **Batch path optimization**: `EncodeBatchImpl` allocates scratch once outside the loop instead of per-vector, and now correctly checks `EncodeOne` return values (a silent bug fix from the old code which ignored failures).

The previously raised comments (Copilot and LHT129) have been addressed in the updated commit 8ce20ddb: `<cstring>` include added to `mrle_transformer.h`, `meta_offsets` null guard with assert added, and `Transform` is now pure virtual.

No blocking issues found. The `TransformBaseVector` secondary scratch allocation noted by Copilot is acknowledged as follow-up work requiring broader API changes to the `BottomQuantizerAccessor` interface.

LHT129 (commented):
Overall this is a well-structured change that eliminates per-vector heap allocations in the transform-quantizer hot paths by reusing caller-owned scratch buffers. The approach of passing scratch `Vector<float>&` references through `ExecuteChainTransform` and storing them in the `Computer` specialization is clean and avoids the previous `std::make_shared` allocations on every transform call.

Key design decisions that look correct:
- The ping-pong double-buffering in `ExecuteChainTransform` correctly handles transform chains of any length without extra allocations.
- The `Computer<TransformQuantizer>` specialization stores both scratch buffers as members, so the query hot path (`ProcessQueryImpl`) has zero per-query allocations.
- `EncodeBatchImpl` now reuses scratch buffers across the entire batch instead of allocating per-vector inside `EncodeOneImpl`.
- The `Transform` signature change from `TransformerMetaPtr` return to `void` with an `uint8_t* meta` out-parameter is consistent across all transformer implementations.
- PCA metadata write is guarded by a null check, preserving backward compatibility for callers that do not need metadata.

The existing review comments have covered the notable points (PCA `centralized_vec` internal allocation, `TransformBaseVector` secondary scratch, base-class pure-virtual consideration), and the author has addressed or acknowledged each one. No blocking issues remain.

LHT129 (commented):
Overall this is a clean and well-structured change. The core idea — replacing per-vector `std::make_shared<Meta>` allocations with caller-owned `uint8_t*` scratch buffers and reusing ping-pong scratch across the transform chain — is sound and directly addresses the hot-path allocation problem.

**Summary of review findings:**

1. The `VectorTransformer::Transform` signature change to pure virtual (`= 0`) is the correct design choice; every derived transformer must implement it, and the empty default was a latent footgun. (Addressed in the latest commit.)

2. The `ExecuteChainTransform` refactor correctly uses caller-provided `primary_scratch`/`secondary_scratch` for ping-pong buffering. The `assert` + `if` guard on `meta_offsets` is reasonable defense-in-depth. (Addressed in the latest commit.)

3. The `Computer<TransformQuantizer>` specialization now owns `primary_scratch_` and `secondary_scratch_` as members, which are reused across `ProcessQueryImpl` calls — this covers the query hot path well.

4. `EncodeBatchImpl` now reuses scratch buffers across the batch loop and properly propagates `EncodeOne` failures, which is a nice correctness improvement over the previous code.

5. The `TransformBaseVector` path still allocates a fresh `secondary_scratch` per call. This is a known limitation acknowledged by the author as follow-up work requiring a broader API change to `BottomQuantizerAccessor`.

The PR description mentions validation against TransformQuantizer and transformer tests, and clang-tidy-15 passing. No blocking issues remain.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -395,8 +431,18 @@ bool
 TransformQuantizer<QuantTmpl, metric>::EncodeBatchImpl(const float* data,
                                                        uint8_t* codes,
                                                        uint64_t count) const {
+    Vector<float> primary_scratch(this->allocator_);
+    Vector<float> secondary_scratch(this->allocator_);
     for (uint64_t i = 0; i < count; ++i) {
-        EncodeOneImpl(data + i * this->dim_, codes + i * this->code_size_);
+        auto* output_codes = codes + i * this->code_size_;
+        const auto* transformed = ExecuteChainTransform(data + i * this->dim_,
+                                                        base_meta_offsets_.data(),
```
LHT129: [note] `EncodeBatchImpl` now duplicates the transform + encode logic from `EncodeOneImpl` instead of calling it. While reusing scratch buffers across the batch loop is a valid optimization, the duplicated logic means any future bug fix or behavioral change in `EncodeOneImpl` will not automatically propagate to the batch path.

Consider extracting a shared private helper (e.g. `EncodeOneTransformed`) that both `EncodeOneImpl` and `EncodeBatchImpl` can call, so the core encode logic stays in one place while the batch loop handles scratch reuse.

LHT129 (commented):
## Review Summary

This PR eliminates per-vector `TransformerMeta` heap allocations by changing `VectorTransformer::Transform()` from returning `shared_ptr<TransformerMeta>` to accepting a caller-owned `uint8_t*` scratch buffer, and refactoring `ExecuteChainTransform` to use ping-pong scratch buffers.

### Changes reviewed
- **11 files**, +115/-63 lines
- Core changes in `transform_quantizer.h` (`ExecuteChainTransform`, `EncodeOneImpl`, `EncodeBatchImpl`, `ProcessQueryImpl`)
- Interface change in `vector_transformer.h` (Transform now pure virtual with `uint8_t* meta` parameter)
- All transformer implementations updated (PCA, FHT, ROM, MRLE)
- `Computer<TransformQuantizer>` now owns `primary_scratch_` and `secondary_scratch_` members

### Issues addressed in commit `8ce20ddb`
- `VectorTransformer::Transform` is now pure virtual (`= 0`) — prevents silent no-op if a derived class forgets to override
- Added `<cstring>` include to `mrle_transformer.h` — fixes reliance on transitive includes
- Added `meta_offsets` null guard with `assert` + runtime check in `ExecuteChainTransform` — prevents null dereference

### Remaining notes (non-blocking)
- `TransformBaseVector` still allocates a fresh `secondary_scratch` per call; author has acknowledged this as follow-up work
- `EncodeBatchImpl` duplicates transform+encode logic from `EncodeOneImpl`; author has self-noted this for potential future refactoring
- PCA `residual_norm` is hardcoded to `0.0F` to preserve serialized code layout; author has noted this is intentional

### Verdict
The core optimization is sound: scratch buffers are now caller-owned and reused across calls in the hot path (`ProcessQueryImpl`, `EncodeOneImpl`, `EncodeBatchImpl`). All previously raised concerns have been addressed or acknowledged. No blocking issues found.

LHT129 (commented):
I have reviewed this PR and the changes look solid overall. The core idea — eliminating per-vector `TransformerMetaPtr` heap allocations by passing caller-owned scratch buffers through `ExecuteChainTransform` — is well executed.

**Summary of what was reviewed:**

1. **`vector_transformer.h`**: `Transform` signature changed from returning `TransformerMetaPtr` to `void` with an optional `uint8_t* meta` out-parameter. Making it pure virtual (`= 0`) is the right call — it prevents silent no-ops from derived classes that forget to override.

2. **`transform_quantizer.h`**: The main logic change. `ExecuteChainTransform` now takes two caller-owned `Vector<float>&` scratch buffers and ping-pongs between them, avoiding the per-iteration `Vector<float> next_data(...)` allocation. The `meta` pointer is now only computed when `codes != nullptr` AND the transformer actually has metadata (`GetMetaSize() > 0`), which is correct. The `assert(meta_offsets != nullptr)` guard in debug builds is a reasonable safety net.

3. **`computer.h`**: `primary_scratch_` and `secondary_scratch_` are added as `Computer` members, which is the right place — the `Computer` object is per-query and reused across `SetQuery` calls, so the scratch buffers are amortized over the query lifetime.

4. **Transformer implementations** (FHT, PCA, ROM, MRLE): All updated consistently to the new `Transform` signature. PCA correctly handles the `meta != nullptr` case by writing the residual norm placeholder.

5. **`EncodeBatchImpl`**: Now reuses scratch buffers across the batch loop instead of calling `EncodeOneImpl` (which would allocate fresh scratch each time). The trade-off between code deduplication and allocation avoidance has been noted in existing review comments.

The existing review thread has already covered the main discussion points (scratch buffer plumbing, `TransformBaseVector` secondary scratch, code deduplication in `EncodeBatchImpl`), and the author has addressed the actionable items. No blocking issues remain from my review.

LHT129 (commented):
## Review Summary

This PR eliminates per-vector `TransformerMetaPtr` allocations by changing the `Transform` interface to accept a caller-owned `uint8_t* meta` scratch buffer and reusing ping-pong scratch buffers (`primary_scratch` / `secondary_scratch`) across the transform chain. The approach is sound and well-scoped.

### What was reviewed

- **Interface change** (`vector_transformer.h`): `Transform` now returns `void` and is pure-virtual (`= 0`), which is the correct design choice.
- **Transformer implementations** (FHT, ROM, MRLE, PCA): All correctly updated to the new signature. PCA preserves backward-compatible metadata layout.
- **TransformQuantizer** (`transform_quantizer.h`): `ExecuteChainTransform` now accepts caller-owned scratch buffers, eliminating the per-call `Vector<float>` allocation. The ping-pong buffer strategy between `primary_scratch` and `secondary_scratch` is correct.
- **Computer** (`computer.h`): Scratch buffers are now members of the TQ `Computer` specialization, reused across `ProcessQuery` calls — this is the key hot-path win.
- **Tests** (`mrle_transformer_test.cpp`): Updated to match the new void-return signature.

### Existing review comments

Several inline comments from Copilot and the author (LHT129) have already been addressed or discussed:
- `VectorTransformer::Transform` pure-virtual → resolved (`= 0`)
- Missing `<cstring>` include in `mrle_transformer.h` → resolved
- PCA `residual_norm` hardcoded to `0.0F` → acknowledged with rationale
- `TransformBaseVector` secondary scratch allocation → author explained concurrency-safe design choice
- `EncodeBatchImpl` logic duplication with `EncodeOneImpl` → noted for future refactoring

No blocking issues remain. The changes are consistent with the stated goal and the existing codebase style.

LHT129 (commented):
Thanks for this PR. The approach of eliminating per-vector allocations by reusing caller-owned scratch buffers is sound and well-executed.

I reviewed the full diff across all 11 changed files. The key changes look correct:

- **`VectorTransformer::Transform`** is now pure virtual (`= 0`), which is the right design choice — no silent no-ops.
- **`ExecuteChainTransform`** correctly implements ping-pong buffering between `primary_scratch` and `secondary_scratch`, and the `meta_offsets` null-safety with `GetMetaSize() > 0` guard is appropriate.
- **`Computer<TransformQuantizer>`** now owns `primary_scratch_` and `secondary_scratch_` as members, which is the right place for query-path scratch reuse.
- **PCA `Transform`** preserves the existing serialized code layout by writing `residual_norm = 0.0F` when `meta != nullptr`, with a clear comment explaining the current state.

One observation (non-blocking): `EncodeBatchImpl` now inlines the transform+encode logic that was previously delegated to `EncodeOneImpl`. This is a reasonable trade-off for scratch buffer reuse, but it means future changes to the encode path need to be kept in sync across both methods. A follow-up could extract a shared helper if this becomes a maintenance concern.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,42 +266,70 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
+    }
 
+    const float* current = input;
+    float* next = primary_scratch.data();
     for (uint32_t i = 0; i < this->transform_chain_.size(); i++) {
-        auto vector_transformer = this->transform_chain_[i];
-        auto meta = vector_transformer->Transform(prev_data, next_data.data());
+        const auto& vector_transformer = this->transform_chain_[i];
+        uint8_t* meta = nullptr;
         if (codes != nullptr) {
```
LHT129: [suggestion] The `assert(meta_offsets != nullptr)` provides a debug-only guard. In release builds (NDEBUG), if a future caller passes `codes != nullptr` with `meta_offsets == nullptr`, the `if (meta_offsets != nullptr && ...)` check will silently skip metadata encoding. Consider a stronger check for release builds too. Low-risk given current callers all pass valid pointers.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: Acknowledged. The duplication is intentional here so EncodeBatchImpl can retain its batch-owned scratch pair without changing the public EncodeOneImpl lifecycle. Extracting a helper that carries both transformed data and scratch ownership is a follow-up refactor; no behavioral divergence exists today.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: Acknowledged. The release guard deliberately avoids dereferencing a null pointer, while the assertion documents the internal caller contract in debug builds. All current calls with codes pass the corresponding offsets; turning this internal contract violation into a runtime exception is outside this performance-focused PR.

src/impl/transform/pca_transformer.cpp
LHT129, line 69: [suggestion] `PCATransformer::Transform` still allocates a `Vector<float> centralized_vec` on every call, which undermines the goal of eliminating per-vector allocations in the hot path. This is called from `ExecuteChainTransform` for every query and every base vector during encoding.

The `centralized_vec` buffer is sized to `input_dim_` and used as scratch for the centralized data before the BLAS gemv. Consider accepting a caller-owned scratch buffer (or reusing one of the ping-pong buffers already plumbed through `ExecuteChainTransform`) to avoid this allocation.

```cpp
// Current code still allocates:
vsag::Vector<float> centralized_vec(allocator_);
centralized_vec.resize(input_dim_, 0.0F);
```

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,42 +266,70 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
 TransformQuantizer<QuantTmpl, metric>::TransformBaseVector(const float* input,
                                                            Vector<float>& output) const {
-    output.assign(input, input + this->dim_);
-    ExecuteChainTransform(output.data(), nullptr, nullptr);
-    output.resize(this->GetTransformedDim());
+    Vector<float> secondary_scratch(this->allocator_);
+    const auto* transformed =
+        ExecuteChainTransform(input, nullptr, nullptr, output, secondary_scratch);
+    if (transformed == output.data()) {
+        output.resize(this->GetTransformedDim());
```
LHT129, line 318: [suggestion] `EncodeOneImpl` creates fresh `primary_scratch` and `secondary_scratch` vectors on every call, while `ProcessQueryImpl` and `EncodeBatchImpl` reuse caller-owned scratch buffers. If `EncodeOneImpl` is called in a hot path (e.g., incremental index insertion), this per-call allocation pair defeats the purpose of eliminating hot-path allocations.

Consider either:
1. Accepting scratch buffers as parameters (similar to how `ProcessQueryImpl` uses `computer.primary_scratch_`), or
2. Adding a note that `EncodeOneImpl` is not performance-critical and the allocation is acceptable.

```cpp
Vector<float> primary_scratch(this->allocator_);
Vector<float> secondary_scratch(this->allocator_);
```

LHT129 (commented):
Overall this is a well-structured performance improvement that eliminates per-vector `shared_ptr` allocations in the transform chain by switching to caller-owned scratch buffers. The ping-pong buffer approach in `ExecuteChainTransform` is clean and avoids the per-iteration `Vector<float>` allocation that existed before.

Key observations:
- The `VectorTransformer::Transform` signature change to `void` with `uint8_t* meta` is a breaking API change, but all in-tree transformers have been updated consistently.
- The `Computer<TransformQuantizer>` now owns scratch buffers as members, which is the right place for query-path reuse.
- `EncodeBatchImpl` inlining the loop (instead of calling `EncodeOneImpl`) avoids the extra scratch allocation pair per element — good tradeoff even with the code duplication.
- The `meta_offsets` null-safety guard in `ExecuteChainTransform` is reasonable for internal callers.

Two suggestions were left as inline comments regarding remaining allocation opportunities in `PCATransformer::Transform` and `EncodeOneImpl`.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -395,8 +431,18 @@ bool
 TransformQuantizer<QuantTmpl, metric>::EncodeBatchImpl(const float* data,
                                                        uint8_t* codes,
                                                        uint64_t count) const {
+    Vector<float> primary_scratch(this->allocator_);
+    Vector<float> secondary_scratch(this->allocator_);
     for (uint64_t i = 0; i < count; ++i) {
-        EncodeOneImpl(data + i * this->dim_, codes + i * this->code_size_);
+        auto* output_codes = codes + i * this->code_size_;
```
LHT129: [suggestion] `EncodeBatchImpl` now inlines the transform+encode logic (creating scratch buffers, calling `ExecuteChainTransform`, then `quantizer_->EncodeOne`) instead of delegating to `EncodeOneImpl`. This duplicates the logic between single-encode and batch-encode paths. If the encoding logic changes in the future (e.g., additional pre/post processing), both paths need to be updated independently, which is a maintenance risk.

Consider having `EncodeBatchImpl` call `EncodeOneImpl` in the loop and check the return value, so there is a single source of truth for the encode path. The scratch allocation concern in `EncodeOneImpl` (noted in a separate comment) can be addressed by accepting scratch buffers as parameters or using thread-local storage.

LHT129 (commented):
## Review Summary

This PR eliminates per-vector `TransformerMeta` heap allocations by changing the `Transform()` interface from returning `shared_ptr<Meta>` to accepting a caller-owned `uint8_t*` scratch buffer, and by introducing ping-pong scratch buffers in `ExecuteChainTransform` and the `Computer<TransformQuantizer>` specialization.

### What was reviewed

- **Interface change**: `VectorTransformer::Transform` is now pure virtual (`= 0`) with a `uint8_t* meta = nullptr` parameter — all derived transformers updated consistently.
- **Scratch buffer management**: `ExecuteChainTransform` uses caller-provided `primary_scratch`/`secondary_scratch` with correct ping-pong swapping between transform steps.
- **Query path**: `ProcessQueryImpl` reuses `computer.primary_scratch_`/`computer.secondary_scratch_` — no per-query allocations.
- **Encode paths**: `EncodeOneImpl` and `EncodeBatchImpl` both create local scratch buffers and pass them to `ExecuteChainTransform`.
- **PCA metadata**: `PCATransformer::Transform` writes a `residual_norm = 0.0F` placeholder into the caller-provided meta buffer, preserving the existing serialized code layout.

### Issues already addressed (commit `8ce20ddb`)

- `VectorTransformer::Transform` made pure virtual (was empty default impl)
- `<cstring>` include added to `mrle_transformer.h`
- `meta_offsets` null-pointer guard with `assert` + runtime check added in `ExecuteChainTransform`

### Remaining discussion points (acknowledged by author)

- `TransformBaseVector` creates a fresh `secondary_scratch` per call — author notes this is a generic API constraint (used by `BottomQuantizerAccessor`) and out of scope for this PR.
- `PCATransformer::Transform` still allocates `centralized_vec` internally — author notes a member scratch buffer would introduce data races under concurrent use.
- `EncodeBatchImpl` inlines transform+encode instead of calling `EncodeOneImpl` — author confirms this is intentional to retain batch-owned scratch buffers.

### Verdict

No blocking issues found. The core design (caller-owned scratch buffers, ping-pong in `ExecuteChainTransform`, `Computer`-owned scratch for the query path) is sound and directly eliminates the targeted hot-path allocations. The remaining allocation sites have been discussed and are either constrained by API design or deferred to follow-up work.

src/impl/transform/pca_transformer.cpp
```diff
@@ -62,9 +63,8 @@ PCATransformer::Train(const float* data, uint64_t count) {
...
```
LHT129, line 70: [suggestion] `PCATransformer::Transform` still allocates a `Vector<float> centralized_vec` on every call (line 70). Since this PR aims to eliminate per-vector transform allocations, this remaining allocation inside the PCA transform hot path partially defeats the purpose.

The `centralized_vec` buffer has a fixed size of `input_dim_`. Consider accepting a caller-owned scratch buffer for this purpose, or pre-allocating it as a member (with appropriate synchronization if needed for thread safety).

```cpp
// Current code still allocates:
vsag::Vector<float> centralized_vec(allocator_);
centralized_vec.resize(input_dim_, 0.0F);
```

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,42 +266,70 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
-    output.assign(input, input + this->dim_);
-    ExecuteChainTransform(output.data(), nullptr, nullptr);
-    output.resize(this->GetTransformedDim());
+    Vector<float> secondary_scratch(this->allocator_);
+    const auto* transformed =
+        ExecuteChainTransform(input, nullptr, nullptr, output, secondary_scratch);
+    if (transformed == output.data()) {
+        output.resize(this->GetTransformedDim());
+    } else {
+        output.assign(transformed, transformed + this->GetTransformedDim());
```
LHT129, line 320: [note] The `assert(meta_offsets != nullptr)` on line 318 will abort in debug builds if `codes != nullptr` but `meta_offsets` is null. However, the subsequent `if (meta_offsets != nullptr && ...)` guard on line 319 handles this gracefully in release builds. This mixed assert+guard pattern is fine defensively, but if the invariant is truly that `codes != nullptr` implies `meta_offsets != nullptr`, consider making the assert unconditional (remove the `if` guard) to catch violations early. If null `meta_offsets` is a valid scenario, the assert should be removed to avoid debug-only crashes.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: [note] `EncodeBatchImpl` now propagates `EncodeOne` failures (returns `false`), whereas the original implementation always returned `true` regardless of individual encode results. This is a correctness improvement, but callers that previously assumed `EncodeBatchImpl` never fails may now encounter error paths that were previously silently ignored. Ensure all callers handle the `false` return appropriately.

```cpp
// New behavior: returns false on first encode failure
if (not quantizer_->EncodeOne(transformed, output_codes)) {
    return false;
}
```

src/impl/transform/pca_transformer.cpp
LHT129: Not changed in this PR. A member workspace would be shared by concurrent EncodeOne and ProcessQuery calls and would introduce a data race. Supplying PCA workspace safely needs a third live float buffer when PCA follows another transform, so it requires a broader transformer scratch-interface design. This PR removes the TQ wrapper allocations; PCA-internal workspace is deferred.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: Not changed. EncodeOne is a stateless public API that may be called concurrently, so storing reusable scratch on the quantizer would be unsafe. Its local RAII scratch preserves that contract; callers with a batch workload use EncodeBatchImpl, which reuses the scratch pair.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: Fixed in 6ecd5394. EncodeOneWithScratch now centralizes transform plus inner encode; EncodeOneImpl owns local scratch while EncodeBatchImpl supplies its reusable pair.

src/impl/transform/pca_transformer.cpp
LHT129: Same rationale as the earlier PCA workspace thread: a concurrency-safe removal of this allocation needs an explicit broader scratch design, including a third buffer for a PCA stage after another transform. It is intentionally deferred from this TQ wrapper-focused PR.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: Fixed in 6ecd5394. The debug-only assertion was removed; the single runtime condition consistently treats absent offsets as absent metadata in every build mode.

src/quantization/transform_quantization/transform_quantizer.h
LHT129: No change required. Quantizer::EncodeBatch is documented to return false on failure and its base implementation already propagates EncodeOneImpl failures. Keeping that behavior for TQ restores the expected Quantizer contract; the existing void FlattenDataCell batch API has historically ignored all quantizer return values and is outside this PR scope.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -313,14 +357,15 @@ TransformQuantizer<QuantTmpl, metric>::ProcessQueryImpl(
...
+    const auto* transformed = ExecuteChainTransform(query,
+                                                    query_meta_offsets_.data(),
+                                                    computer.inner_computer_->buf_,
+                                                    computer.primary_scratch_,
+                                                    computer.secondary_scratch_);
 
     // 2. execute quantize
     // note that only when computer.buf_ == nullptr, quantizer_ will allocate data to buf_
-    quantizer_->ProcessQuery(data_buffer.data(), *computer.inner_computer_);
+    quantizer_->ProcessQuery(transformed, *computer.inner_computer_);
```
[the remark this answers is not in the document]
LHT129: Fixed in 97816f79. The comment now correctly refers to computer.inner_computer_->buf_ and clarifies that the inner quantizer owns the query buffer.

src/quantization/transform_quantization/transform_quantizer.h
```diff
@@ -259,43 +271,75 @@ TransformQuantizer<QuantTmpl, metric>::TrainImpl(const float* data, uint64_t cou
...
+const float*
+TransformQuantizer<QuantTmpl, metric>::ExecuteChainTransform(
+    const float* input,
+    const uint32_t* meta_offsets,
+    uint8_t* codes,
+    Vector<float>& primary_scratch,
+    Vector<float>& secondary_scratch) const {
+    if (primary_scratch.size() < this->dim_) {
+        primary_scratch.resize(this->dim_, 0.0F);
+    }
```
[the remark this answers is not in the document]
LHT129: The supported TQ chain configurations are non-expanding: PCA is configured for dimensionality reduction and FHT/ROM/MRLE preserve dimension. Therefore the original input dimension is the scratch capacity upper bound for this PR. Supporting expanding transforms would need explicit parameter validation and a broader sizing contract; it is outside the current TQ optimization scope.

wxyucs (approved):
lgtm

PR #2707, opened 2026-08-18
feat(ivf): support RaBitQ x+y split storage and search

### LINKED ISSUE #2443: [feat](ivf): support RaBitQ x+y split storage and search
### Affected area

ivf

### Motivation

IVF currently supports standard RaBitQ as its bucket quantizer, but it does not support the
RaBitQ x+y split layout used by HGraph.

With x+y split, each vector is represented by:

- an x-bit filter record used during the initial scan;
- a y-bit supplement record fetched only for candidates that require more accurate ranking;
- a final x+y-bit distance used for the result ranking.

This would benefit large IVF indexes where scanning full multi-bit RaBitQ codes for every vector
in the selected buckets causes unnecessary memory traffic or disk I/O. It would also allow IVF to
keep compact filter records in memory while storing colder supplement records separately.

### Proposed solution

Add an IVF bucket-storage implementation that supports RaBitQ x+y split while preserving the
existing bucket and inner-ID semantics.

The external configuration should reuse the HGraph split terminology where possible. For
example:

```json
{
    "dtype": "float32",
    "metric_type": "l2",
    "dim": 128,
    "index_param": {
        "buckets_count": 1024,
        "base_quantization_type": "rabitq",
        "precise_quantization_type": "rabitq",
        "use_reorder": true,
        "rabitq_bits_per_dim_query": 32,
        "rabitq_bits_per_dim_base": 3,
        "rabitq_bits_per_dim_precise": 5,
        "rabitq_error_rate": 1.9,
        "ivf_train_type": "kmeans",
        "train_sample_count": 100000
    }
}
```

In split mode:

1. Map `rabitq_bits_per_dim_base` to the x filter bits.
2. Map `rabitq_bits_per_dim_precise` to the y supplement bits.
3. Store filter and supplement records separately for every IVF bucket.
4. Scan only the filter records during the initial bucket scan.
5. Use the filter distance and lower bound to select reorder candidates.
6. Fetch the supplement record only for selected candidates.
7. Compute the final distance from all x+y bits without rescanning data unnecessarily.

The implementation will likely require either a split-aware `BucketInterface` implementation or
an extension of `BucketDataCell` that can coordinate two bucket-local code stores.

The implementation must also define and test:

- Build and Add behavior;
- concurrent Add behavior;
- serialization and deserialization;
- memory usage reporting;
- L2, IP, and cosine support;
- residual IVF compatibility;
- supported homogeneous and hybrid I/O combinations;
- clear validation errors for unsupported combinations.

Existing IVF RaBitQ configurations that do not request split mode must retain their current
behavior.

### Alternatives considered

1. Continue using standard RaBitQ bucket codes and store a separate complete precise-code copy
   for reorder. This is already compatible with the current IVF architecture, but duplicates code
   data and cannot avoid reading the full base code during bucket scans.

2. Reuse `RaBitQSplitDataCell` directly. This datacell implements `FlattenInterface` and uses
   global inner IDs, while IVF requires bucket-local storage, offsets, scans, and inner-ID
   mappings. A bucket-aware abstraction is therefore still needed.

3. Store only x-bit RaBitQ codes without a supplement. This reduces scan cost but loses the
   higher-precision x+y distance required for final ranking.

additive only (no break)

### References

- `src/datacell/bucket_interface.cpp:40`
- `src/datacell/bucket_interface.cpp:71`
- `src/datacell/bucket_datacell.h:29`
- `src/datacell/bucket_datacell.h:237`
- `src/algorithm/ivf/ivf.cpp:239`
- `src/algorithm/ivf/ivf.cpp:299`
- `src/algorithm/ivf/ivf.cpp:316`
- `src/datacell/rabitq_split_datacell.h:133`
- `src/quantization/rabitq_quantization/rabitq_quantizer.cpp:831`
- `tests/test_ivf.cpp:592`
- `docs/docs/en/src/quantization/rabitq_split.md:1`

### Additional context

Suggested acceptance criteria:

- Existing standard IVF + RaBitQ tests continue to pass.
- Add functional tests for Build, Add, search, and serialization with split RaBitQ.
- Verify that initial bucket scans access only x-bit filter records.
- Verify that y-bit supplement records are fetched only for selected reorder candidates.
- Compare recall, QPS, build time, memory usage, and index size against standard multi-bit IVF +
  RaBitQ.
- Document IVF configuration and any I/O or residual-mode limitations in both English and Chinese
  user documentation.

## Related issues

- [#2301 — IVF rejects rabitq base quantization](https://github.com/antgroup/vsag/issues/2301)
  established standard IVF + RaBitQ support and is now closed.
- [#2096 — Discuss split RabitQ storage refactor for disk-based xbit reorder](https://github.com/antgroup/vsag/issues/2096)
  discusses the split-storage abstraction, primarily for HGraph.

---

LightWant:
## Performance Results

This PR adds RaBitQ x+y split storage and search support to IVF flat buckets. When split parameters are not configured, IVF continues to use the existing RaBitQ implementation. The bucket graph path does not use the split searcher.

Test environment: GIST1M, 1,000,000 960-dimensional float32 L2 vectors, 1,000 queries, 1,024 IVF buckets, and 100,000 training vectors. The index was built with 32 threads and 12 fast-encode rounds. Search used `nprobe=32`, `factor=10`, top-10, and index-internal `parallelism=1`.

### Traditional IVF RaBitQ Before This PR vs. Final Residual 1+7

| Version | Layout | Build Time | Build TPS | Index Memory | Index File | Single-thread QPS | Avg | Recall@10 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Before this PR | Traditional 8-bit RaBitQ | 149.369 s | 6,694.851 | 1,284,595,232 B | 946,484,662 B | 16.784 | 59.576 ms | 0.9166 |
| Final PR | Residual 1-bit filter + 7-bit supplement | 144.253 s | 6,932.263 | 1,340,849,360 B | 1,026,371,655 B | 576.11 | 1.733 ms | 0.9184 |

Compared with traditional 8-bit RaBitQ, the final residual 1+7 implementation reduces build time by approximately 3.4%, improves build TPS by approximately 3.5%, delivers approximately 34.3x higher single-thread search throughput, and improves Recall@10 by 0.0018.

### Five-round Pre-submission Search Regression

The final-format GIST1M residual 1+7 index was reused. After warm-up, five runs were executed on a fixed CPU core, and the median is reported below:

| QPS | Avg | P50 | P95 | Recall@10 | Scanned/query | Reranked/query | x-bit IP hits | Fallbacks |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 576.11 | 1.733 ms | 1.652 ms | 2.475 ms | 0.9184 | 60,274.60 | 100 | 100 | 0 |

### IVF Layering

- The dedicated `RaBitQSplitBucketSearcher` is selected only when an IVF flat bucket uses RaBitQ split storage.
- The regular `FlatBucketSearcher` retains its existing search decisions. Graph buckets only forward the required workspace parameter; they do not use split scanning or change the graph search strategy.
- The IVF searcher owns filtering, thresholds, reasoning, and the candidate heap. Bucket/DataCell owns split-block scanning, version validation, and supplement reads.

### Retained 500+ QPS Path

- Filter codes are packed in 32-vector FastScan blocks. The 1+7, 2+6, and 3+5 layouts share the HighAcc query LUT and bit-plane weighting.
- Supplements use compact 5/6/7-bit layouts with generic, AVX2, and AVX512 runtime dispatch.
- Candidate heap entries carry the quantized x-bit inner product together with the source bucket, offset, and version. The normal reranking path reads only ybits and does not read xbits again.
- Residual candidates use original-query full factors. The normal path does not create bucket-specific `q-c` computers; only stale or invalid lanes enter the full fallback path.
- Optimized build, Add/Update/Delete, relocation, Merge, and Serialize/Deserialize all maintain the same packed-layout and version semantics.

### Removed Experimental Code

- The optional lower-bound heap search strategy and its YAML parameter.
- The low-precision uint8 FastScan query LUT and duplicate lower-bound/filter-IP interfaces.
- The dense whole-bucket filter-IP cache, the old centroid-IP conversion, and the old residual-computer factory.
- The split-unrelated SIMD candidate fast path in the generic `FlatBucketSearcher`.
- The corresponding experimental documentation, performance tables, and duplicate tests.

The cleanup commit removed 2,496 lines and added 191 lines, reducing the PR from 14,246 added lines to 11,940 added lines. The remaining changes primarily cover the split Bucket/DataCell implementation, cross-architecture SIMD, packed layouts, concurrency and serialization correctness, and tests.

## Tests and Checks

- clang-format 15 dry run and `git diff --check`: passed.
- Full Release build, including the library, unittests, functests, and eval_performance: passed.
- Focused split/IVF/RaBitQ tests: 43 cases and 3,669,967 assertions passed.
- Dedicated split searcher and parameter tests: 3 cases and 145 assertions passed.
- GIST1M residual 1+7: median of five runs was 576.11 QPS with Recall@10 of 0.9184.

## Benchmark Configuration

```yaml
global:
  num_threads_searching: 1

residual_1_7_candidate:
  datapath: /root/data/gist1M/gist-960-euclidean.hdf5
  type: search
  index_name: ivf
  index_path: /tmp/vsag-pr-2443-consolidated-gist/residual-1-7-v4.index
  create_params: >-
    {"dtype":"float32","metric_type":"l2","dim":960,"index_param":{
      "buckets_count":1024,"base_quantization_type":"rabitq",
      "precise_quantization_type":"rabitq","rabitq_bits_per_dim_query":32,
      "rabitq_bits_per_dim_base":1,"rabitq_bits_per_dim_precise":7,
      "rabitq_error_rate":1.9,"use_reorder":true,"use_residual":true,
      "partition_strategy_type":"ivf","ivf_train_type":"kmeans",
      "train_sample_count":100000,"thread_count":32,
      "fast_encode_rabitq":true,"fast_encode_rabitq_rounds":12}}
  search_params: >-
    {"ivf":{"scan_buckets_count":32,"factor":10.0,"parallelism":1}}
  search_mode: knn
  topk: 10
  search_query_count: 1000
  delete_index_after_search: false
```

Fixes: #2443


LightWant:
Maintainers: GitHub rejected the fork author's attempt to set upstream labels. Please add `kind/feature` and `version/1.1`, matching linked issue #2443.

src/algorithm/ivf/ivf.cpp
```diff
@@ -403,23 +462,191 @@ IVF::CheckAndMappingExternalParam(const JsonType& external_param,
...
+                    static_cast<InnerIdType>(candidate_count),
+                    &ctx);
+            }
+        }
+        for (uint64_t i = 0; i < candidate_count; ++i) {
+            if (ctx.reasoning_ctx != nullptr) {
+                ctx.reasoning_ctx->RecordReorder(ids[i], hints[i], distances[i]);
+            }
+            if (distance_threshold.has_value() and
+                (not std::isfinite(distances[i]) or distances[i] > *distance_threshold)) {
```
LHT129, line 614: [critical] `std::isfinite` is used directly in `SplitBucketReorder::Reorder` but the same PR replaces `std::isfinite` with `IsFiniteFloatBits` in `flat_bucket_searcher.cpp` and `searcher_utils.h` to avoid `-Ofast` compatibility issues. This location was missed and will break under `-Ofast` builds.

The `searcher_utils.h` change explicitly documents the motivation: `-Ofast` eliminates `std::isfinite`/`std::isnan` checks because they are considered "finite math only" operations. The `SplitBucketReorder` class uses `std::isfinite(distances[i])` directly in the distance threshold check, which would be eliminated under `-Ofast`.

Suggested fix: Replace with `IsFiniteFloatBits(distances[i])` or use the same IEEE-754 bit-check pattern from `searcher_utils.h`.

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2591 @@
...
+#include "rabitq_split_datacell.h"
+#include "simd/fp32_simd.h"
+#include "simd/normalize.h"
+
+namespace vsag {
+
+namespace {
+
+bool
+IsFiniteFloat(float value) {
```
LHT129, line 33: [suggestion] `IsFiniteFloatBits` (or `IsFiniteFloat`) is duplicated in at least 4 files:
- `src/datacell/rabitq_split_bucket_datacell.cpp:33` (as `IsFiniteFloat`)
- `src/algorithm/ivf/rabitq_split_bucket_searcher.cpp:33` (as `IsFiniteFloatBits`)
- `src/algorithm/ivf/flat_bucket_searcher.cpp:31` (as `IsFiniteFloatBits`)
- `src/quantization/rabitq_quantization/rabitq_quantizer.cpp:61` (as `IsFiniteFloatBits`)

Each is a copy of the same IEEE-754 bit-check pattern. Consider centralizing this in a shared utility header (e.g., `simd/fp32_simd.h` or a new `utils/float_utils.h`) to avoid duplication and ensure consistent usage across the codebase. The `searcher_utils.h` already has a similar `is_finite_distance_bits` helper that could serve as the canonical version.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -20,17 +20,60 @@
...
 #include "simd/rabitq_simd.h"
 #include "typing.h"
 #include "utils/util_functions.h"
 
 namespace vsag {
 
+namespace {
+
+uint64_t&
+FastScanFallbackRandomState() {
```
LHT129, line 39: [note] The `FastScanFallbackRandomState()` uses `thread_local` state initialized with `std::random_device`. This is a splittable PRNG (likely SplitMix64 variant) used for stochastic rounding in the FastScan32 LUT quantization.

Two observations:
1. The `thread_local` initialization via lambda with `std::random_device` is correct and ensures each thread gets an independent random stream.
2. The PRNG passes the `random_state` pointer through the call chain (`PrepareFastScan32Query` → `FastScanUniformRandom`), allowing deterministic replay when an explicit state is provided. This is a good design for reproducibility.

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2591 @@
...
+        const auto bucket_id = static_cast<BucketIdType>(location >> LOCATION_SPLIT_BIT);
+        const auto offset_id = static_cast<InnerIdType>(location & kLocationMask);
+        if (bucket_id < 0 or bucket_id >= this->bucket_count_) {
+            return false;
+        }
+
+        std::shared_lock bucket_lock(this->bucket_mutexes_[bucket_id]);
+        std::lock_guard codes_lock(this->codes_insert_mutex_);
+        {
+            // Writers publish locations while holding the corresponding bucket lock and follow the
```
LHT129, line 727: [note] The lock ordering in `query_current_by_inner_id` is carefully documented: bucket lock → codes lock → locations lock (re-check). The comment on lines 1067-1069 explains that writers publish locations while holding the bucket lock and follow bucket → codes → locations order. Taking the same locks before rechecking ensures a consistent snapshot of lane + supplement without lock inversion.

This is correct but subtle. Consider adding a static assertion or a centralized lock-ordering document if more lock-taking paths are added in the future.

LHT129 (commented):
## Overall Review

This is a substantial and well-engineered PR that adds RaBitQ x+y split storage and search to the IVF index. The architecture is sound and the implementation is thorough, with strong attention to parameter validation, concurrency, and SIMD optimization.

### Summary of findings

**Issues identified (4 inline comments):**

1. **[critical] `std::isfinite` in `SplitBucketReorder::Reorder`** (`src/algorithm/ivf/ivf.cpp:614`): The PR correctly replaced `std::isfinite`/`std::isnan` with IEEE-754 bit-based checks in `searcher_utils.h` and `flat_bucket_searcher.cpp` to avoid `-Ofast` elimination. However, `std::isfinite` is still used in `SplitBucketReorder::Reorder`, which will silently produce incorrect results under `-Ofast`. This should use the same `IsFiniteFloatBits` helper.

2. **[suggestion] `IsFiniteFloatBits` duplicated across 4 files** (`rabitq_split_bucket_datacell.cpp:33`, `rabitq_split_bucket_searcher.cpp:33`, `flat_bucket_searcher.cpp:31`, `rabitq_quantizer.cpp:61`): The same IEEE-754 bit-checking helper is defined as an anonymous-namespace function in four separate translation units. This should be extracted into a shared utility header (e.g., `simd/fp32_simd.h` or a new `fp32_utils.h`) to avoid drift and ensure consistency.

3. **[note] `FastScanFallbackRandomState` thread_local PRNG** (`rabitq_quantizer.cpp:39`): The SplitMix64 PRNG uses `thread_local` state for reproducible random quantization in FastScan32. This is fine for the current use case but is not seedable per-call — if reproducibility across runs is ever needed, consider making the seed a parameter.

4. **[note] Lock ordering in `query_current_by_inner_id`** (`rabitq_split_bucket_datacell.cpp:1067`): The lock ordering (bucket → codes → locations) is carefully documented and consistent with the writer path. This is correct but subtle — consider a centralized lock-ordering document if more lock-taking paths are added.

### What is done well

- **Parameter validation**: Comprehensive checks in `ivf.cpp` ensure split storage is only enabled with compatible settings (both quantizers are rabitq, `use_reorder=true`, `buckets_per_data=1`, no bucket graphs). This prevents silent misconfiguration.
- **Concurrency design**: Per-bucket mutexes + version counters + lock ordering discipline provide correct concurrent read/write without global locks. The version-based cache invalidation for filter inner products is well thought out.
- **SIMD coverage**: FastScan32 byte-LUT scanning and packed supplement kernels are implemented for AVX2, AVX512, SSE, NEON, and SVE — full architecture coverage.
- **Residual L2 encoding**: The residual query transform with per-lane factors and `residual_adjustment` correctly handles the L2SQR residual bias, which is easy to get wrong.
- **Two search strategies**: Candidate reorder (scan x-bits, reorder top candidates with y-bits) and heap search (lower-bound pruning) give users a choice between throughput and precision.
- **Build lifecycle**: `IVFBucketOptimizedBuildSession` RAII wrapper ensures proper cleanup on error paths.
- **Documentation**: Both EN and ZH docs are updated for the new parameters.

### Recommendation

The PR is in good shape. The one critical issue (missing `-Ofast` fix in `SplitBucketReorder`) should be addressed before merge. The duplicated `IsFiniteFloatBits` helper is a maintenance concern worth fixing but not a correctness issue. The other notes are informational.

src/algorithm/ivf/rabitq_split_bucket_searcher.cpp
```diff
@@ -0,0 +1,353 @@
...
+                continue;
+            }
+
+            float distance_limit = cur_heap_top;
+            if (param.distance_threshold.has_value()) {
+                distance_limit = std::min(distance_limit, param.distance_threshold.value());
+            }
+            const float lower_bound = lower_bounds[j];
+            const bool has_usable_lower_bound =
+                std::isfinite(lower_bound) and lower_bound < std::numeric_limits<float>::max();
```
LHT129: [critical] `std::isfinite` is used in the RaBitQ heap-search path (lines 179 and 205) but the same PR introduces `IsFiniteFloatBits` to avoid `-Ofast` compatibility issues elsewhere. Under `-Ofast`, `std::isfinite` is eliminated as a "finite math only" operation, causing the lower-bound filter (`not(lower_bound < distance_limit)`) and the exact-distance validity check to be silently removed.

Line 179: `std::isfinite(lower_bound) and lower_bound < std::numeric_limits<float>::max()`
Line 205: `not std::isfinite(exact_distance)`

Suggested fix: Replace both with `IsFiniteFloatBits(...)` (already defined in the anonymous namespace of this file at line 33).

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2591 @@
...
+                this->residual_adjustment(query_centroid_adjustment, bucket_id, offset_id);
+        }
+        return true;
+    }
+    return false;
+}
+
+void
+RaBitQSplitBucketDataCell::query_non_residual_by_inner_ids(
+    float* result_dists,
```
LHT129, line 780: [suggestion] `query_non_residual_by_inner_ids` (~200 lines) and `query_residual_by_inner_ids` (~200 lines) share the same structure: sort entries by bucket, group by bucket, check staleness, classify into reusable/fallback, batch query with filter inner products, then handle stale entries via `query_current_by_inner_id`. The only differences are residual-specific handling (transformed centroids, original-query filter inner product conversion, `full_add` computation).

Consider extracting the common grouping/staleness/batching logic into a shared helper to reduce the ~400 lines of near-duplicate code. This would make bug fixes and future optimizations apply to both paths automatically.

src/impl/heap/auxiliary_standard_heap.cpp
```diff
@@ -0,0 +1,122 @@
...
+                            -1,
+                            std::numeric_limits<InnerIdType>::max(),
+                            std::numeric_limits<uint64_t>::max());
+}
+
+template <bool max_heap, bool fixed_size>
+void
+AuxiliaryStandardHeap<max_heap, fixed_size>::PushWithAuxiliary(float dist,
+                                                               InnerIdType id,
+                                                               float auxiliary,
```
LHT129, line 55: [note] `Push(dist, id)` delegates to `PushWithAuxiliary` passing `-1` for `source_bucket_id`. Since `BucketIdType` is unsigned (`uint32_t` in `bucket_datacell.h`), `-1` wraps to `UINT32_MAX`, which serves as a sentinel. This is intentional but subtle — a named constant like `kUnknownBucketId` would make the intent clearer and avoid confusion about the sign mismatch.

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2591 @@
+// Copyright 2024-present the vsag project
```
LHT129, line 1: [note] The new `rabitq_split_bucket_datacell.cpp` (2591 lines) and `rabitq_split_bucket_searcher.cpp` (353 lines) introduce ~2900 lines of new core logic without a dedicated unit test file. The existing test changes in `bucket_datacell_test.cpp` (+1213 lines) and `flat_bucket_searcher_test.cpp` (+799 lines) cover the generic bucket/searcher interfaces, but the split-bucket-specific code paths (filter inner product caching, version-aware staleness checks, routed bucket computer management, residual vs non-residual query dispatch) would benefit from focused tests that exercise the split-storage edge cases directly.

LHT129 (commented):
[critical] `std::isfinite` regression in `flat_bucket_searcher.cpp`

This commit removed the local `IsFiniteFloatBits` implementation from `src/algorithm/ivf/flat_bucket_searcher.cpp` and replaced it with `std::isfinite` at line 89 (`(not std::isfinite(dist[j])`). However, the same PR introduces `IsFiniteFloatBits` in `rabitq_split_bucket_searcher.cpp` specifically to avoid `-Ofast` compatibility issues — `-Ofast` eliminates `std::isfinite`/`std::isnan` checks because they are treated as "finite math only" operations.

This is inconsistent: `rabitq_split_bucket_searcher.cpp` correctly retains `IsFiniteFloatBits`, but `flat_bucket_searcher.cpp` regresses to `std::isfinite`. The distance threshold check in `FlatBucketSearcher::Search` will be silently removed under `-Ofast` builds.

Suggested fix: Restore `IsFiniteFloatBits` (or use the centralized version from `searcher_utils.h`) and replace `std::isfinite(dist[j])` with `IsFiniteFloatBits(dist[j])`.

LHT129 (commented):
[suggestion] SIMD fast path removed from `FlatBucketSearcher::Search`

This commit removes the SIMD-accelerated candidate push path (~30 lines) from `FlatBucketSearcher::Search` that used `FP32LessThan32Mask` for batch distance comparison when no filters or reasoning context are active. This path was a general optimization for the flat bucket searcher — it is not specific to RaBitQ split storage.

The removed code provided a meaningful throughput improvement for the common case of unfiltered KNN search by processing 32 distances at a time with a single SIMD comparison, avoiding per-element scalar checks. Without it, every candidate goes through the slower per-element loop even when no filters are configured.

If this removal is intentional (e.g., the SIMD path is being moved elsewhere or the performance impact was measured and deemed acceptable), please clarify in the commit message. Otherwise, consider restoring this optimization for the non-split flat bucket searcher path.

LHT129 (commented):
[note] `FilterInnerProductMode::CACHED` enum value is now dead code

The `FilterInnerProductMode` enum in `src/datacell/rabitq_split_bucket_datacell.h` still contains `CACHED` as a member, but this commit removes all code paths that use it:
- `claim_filter_inner_product_cache`
- `get_cached_filter_inner_product`
- `publish_candidate_scan_version`
- `publish_heap_scan_version`
- `candidate_scan_version_matches`
- `heap_scan_version_matches`
- All `routed_filter_cache_*` fields in `SplitBucketComputer`

The `CACHED` mode is no longer reachable. Consider removing the enum value and the associated `SplitBucketComputer` fields (`routed_filter_inner_products_`, `routed_filter_cache_claimed_`, `routed_filter_cache_versions_`, `routed_candidate_scan_versions_`, `routed_heap_scan_versions_`) if they are truly no longer needed, to keep the codebase clean and avoid confusion for future readers.

src/algorithm/ivf/flat_bucket_searcher.cpp
```diff
@@ -33,15 +33,26 @@ FlatBucketSearcher::Search(BucketIdType bucket_id,
                            BucketIdType buckets_per_data,
                            DistHeapPtr& heap,
                            Vector<float>& dist,
+                           Vector<InnerIdType>& scanned_inner_ids,
```
LightWant, line 37: 这里由“const auto* ids = bucket->GetInnerIds(bucket_id);”变成参数“Vector<InnerIdType>& scanned_inner_ids,”是否合理，

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2180 @@
...
+    } else {
+        std::lock_guard lock(this->locations_mutex_);
+        for (InnerIdType i = 0; i < id_count; ++i) {
+            CHECK_ARGUMENT(inner_ids[i] < this->locations_.size(),
+                           "invalid inner id for RaBitQ split bucket");
+            const uint64_t location = this->locations_[inner_ids[i]];
+            CHECK_ARGUMENT(location != INVALID_LOCATION,
+                           "invalid inner id for RaBitQ split bucket");
+            entries.push_back({static_cast<BucketIdType>(location >> LOCATION_SPLIT_BIT),
+                               static_cast<InnerIdType>(location & 0xFFFFFFFFULL),
```
LHT129, line 825: [suggestion] In `query_non_residual_by_inner_ids` and `query_residual_by_inner_ids`, the offset is extracted from the packed location using the hardcoded mask `0xFFFFFFFFULL`. However, `query_current_by_inner_id` uses the named constant `kLocationMask` (defined as `(1ULL << LOCATION_SPLIT_BIT) - 1ULL`). Since `LOCATION_SPLIT_BIT = 32`, both are equivalent, but the hardcoded literal is fragile: if `LOCATION_SPLIT_BIT` were ever changed, the hardcoded `0xFFFFFFFFULL` would silently become incorrect while `kLocationMask` would update automatically. Suggested fix: Replace `0xFFFFFFFFULL` with `kLocationMask` in both locations.

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2180 @@
...
+    }
+
+    for (InnerIdType i = 0; i < entry_count; ++i) {
+        if (source_stale[i] == 0U) {
+            result_dists[entries[i].result_offset] = sorted_dists[i] - adjustments[i];
+        }
+    }
+    for (uint64_t i = 0; i < stale_ids.size(); ++i) {
+        float stale_dist = std::numeric_limits<float>::max();
+        this->query_current_by_inner_id(stale_dist, computer, stale_ids[i], ctx);
```
LHT129, line 967: **suggestion:** `query_current_by_inner_id` returns `bool` (false when the inner_id is invalid or the location is stale after retries), but both call sites discard the return value. When the function fails, `stale_dist` remains `std::numeric_limits<float>::max()`, which is then silently written into `result_dists`. This means a failed lookup produces a max-float distance that will be treated as a valid result downstream.

Consider either:
- Checking the return value and skipping the entry (or using a sentinel that callers can detect), or
- Documenting why silently using `max()` is intentional (e.g., it naturally falls out of the top-k heap).

src/datacell/rabitq_split_bucket_datacell.cpp
```diff
@@ -0,0 +1,2180 @@
...
+    }
+
+    for (InnerIdType i = 0; i < entry_count; ++i) {
+        if (source_stale[i] == 0U) {
+            result_dists[entries[i].result_offset] = sorted_dists[i];
+        }
+    }
+    for (uint64_t i = 0; i < stale_ids.size(); ++i) {
+        float stale_dist = std::numeric_limits<float>::max();
+        this->query_current_by_inner_id(stale_dist, computer, stale_ids[i], ctx);
```
LHT129, line 1240: **suggestion:** Same issue as above — `query_current_by_inner_id` return value is discarded. If the lookup fails, `stale_dist` remains `max()` and is silently written to `result_dists`.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -1252,22 +1325,792 @@ RaBitQuantizer<metric>::MergeSplitCode(const uint8_t* one_bit_code,
...
+
+template <MetricType metric>
+typename RaBitQuantizer<metric>::FastScan32Layout
+RaBitQuantizer<metric>::GetFastScan32Layout() const {
+    return FastScan32Layout{FilterPlanesSize() * FASTSCAN_BATCH_SIZE,
+                            OneBitRecordNormOffset(),
+                            GetOneBitCodeSize() - OneBitRecordNormOffset()};
+}
+
+template <MetricType metric>
```
LHT129, line 1368: [suggestion] `ComputeFastScan32ResidualFactors` uses `std::fabs(filter_error)` to guard against near-zero errors, but the same PR introduces `EffectiveRaBitQError` which handles this case differently: it returns ±1.0 based on the sign of the error rather than taking the absolute value. The `ComputeDistWithSplitCode` path (line 2485) was updated to use `EffectiveRaBitQError`, but this residual factor path was not.

For consistency and to avoid subtle numerical differences between the FastScan32 residual path and the exact split-code path, consider using `EffectiveRaBitQError` here as well, or document why `std::fabs` is the correct choice for the residual factor computation.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -369,16 +412,49 @@ RaBitQuantizer<metric>::RaBitQFloatSQIPByPlanes(const float* query,
...
+        const float filter_center = 0.5F * static_cast<float>((1U << filter_bits) - 1U);
+        const float supplement_ip =
+            RaBitQFloatSupplementCodeIP(query, supplement_code, this->dim_, supplement_bits);
+        return static_cast<float>(1U << supplement_bits) *
+                   (centered_filter_ip + filter_center * query_sum) +
+               supplement_ip;
+    }
+
+    float result = 0.0F;
+    for (uint64_t d = 0; d < this->dim_; ++d) {
```
LHT129, line 428: [note] The fallback path in `RaBitQFloatSQIPByPlanes` for `filter_bits > 3` (or when the 2/3-bit centered path doesn't apply) uses a scalar O(dim * total_bits) loop that reconstructs the full integer code per dimension. This is significantly slower than the SIMD-optimized paths above it.

If this fallback is expected to be hit in production (e.g., for filter_bits=4), consider adding a SIMD-accelerated path. If it's only a correctness fallback for unusual configurations, consider adding a comment to document that expectation.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -1252,22 +1325,792 @@ RaBitQuantizer<metric>::MergeSplitCode(const uint8_t* one_bit_code,
...
+                const float value = query[dim];
+                if (value < 0.0F) {
+                    group_lower += value;
+                } else {
+                    group_upper += value;
+                }
+            }
+        }
+        lower = std::min(lower, group_lower);
+        upper = std::max(upper, group_upper);
```
LHT129, line 1592: [suggestion] `PrepareFastScan32HighAccQuery` computes `deltas[0] = range / kHighAccQuantizedMax` at line 1592, then checks `if (range <= epsilon)` at line 1593 and returns early with a zeroed lookup table. The `deltas[0]` value computed before the early return is unused in that case, but it's a wasted division. Consider moving the range check before the `deltas[0]` computation to avoid the unnecessary work.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -2127,10 +3076,91 @@ RaBitQuantizer<metric>::RecoverOrderSQ(const uint8_t* output, uint8_t* input) co
...
+    if (this->dim_ < this->original_dim_) {
+        std::copy(pca_data.data() + this->dim_,
+                  pca_data.data() + this->original_dim_,
+                  transformed_query + this->dim_);
+    }
+}
+
+template <MetricType metric>
+void
+RaBitQuantizer<metric>::ProcessTransformedResidualQuery(const float* transformed_query,
```
LHT129, line 3100: [note] `ProcessQueryImpl` now always transforms the query via `TransformResidualQuery` + `ProcessTransformedResidualQuery` when `metric == METRIC_TYPE_L2SQR`, regardless of whether residual mode is actually in use. The `ProcessTransformedResidualQuery` handles both residual and non-residual cases internally, so this is functionally correct, but it adds a PCA + random orthogonal matrix transform on every L2 query even when the index is not using residual encoding.

If the transform is cheap enough this may be fine, but consider adding a flag to skip the transform when residual mode is not active (e.g., when `use_residual_ == false`).

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -369,16 +412,49 @@ RaBitQuantizer<metric>::RaBitQFloatSQIPByPlanes(const float* query,
...
 }
 
 template <MetricType metric>
 float
 RaBitQuantizer<metric>::RaBitQFloatSQIPBySplitCode(const float* query,
                                                    const uint8_t* one_bit_code,
                                                    const uint8_t* supplement_code) const {
-    return RaBitQFloatSplitCodeIP(query, one_bit_code, supplement_code, this->dim_, ReorderBits());
+    const float filter_ip = RaBitQFloatBinaryIP(query, one_bit_code, this->dim_, 0.0F) *
+                            static_cast<float>(1U << ReorderBits());
```
LHT129, line 455: [note] `RaBitQFloatSQIPBySplitCode` changed from using `RaBitQFloatSplitCodeIP` (which reads unpacked supplement planes) to using `RaBitQFloatBinaryIP + RaBitQFloatPackedSupplementCodeIP` (which reads from the packed supplement format). This is paired with the corresponding changes in `SplitCode`/`MergeSplitCode` which now use `RaBitQPackSupplementPlanes`/`RaBitQUnpackSupplementPlanes` for format conversion.

This is correct but worth calling out: the supplement code format in the split storage path is now packed (using the compact chunk layout from `rabitq_packed_supplement.h`) rather than the traditional bit-plane format. Any code that reads supplement codes directly from the split storage must use the packed format accessors.

src/simd/kernels/rabitq_packed_supplement.h
```diff
@@ -0,0 +1,203 @@
...
+            packed[lane] =
+                static_cast<uint8_t>((values[lane] & 0x0FU) | ((values[16U + lane] & 0x0FU) << 4U));
+            packed[16U + lane] = static_cast<uint8_t>((values[32U + lane] & 0x0FU) |
+                                                      ((values[48U + lane] & 0x0FU) << 4U));
+        }
+        std::memset(packed + 32U, 0, 8U);
+        for (uint64_t lane = 0; lane < RaBitQPackedSupplementLayout::CHUNK_DIM; ++lane) {
+            packed[32U + lane / 8U] |=
+                static_cast<uint8_t>(((values[lane] >> 4U) & 1U) << (lane % 8U));
+        }
```
LHT129, line 93: [note] `RaBitQPackCompactSupplementChunk` for bits=6 uses the common 6/7-bit path but only zeroes the 7th bit plane (`packed + 48U`) when `bits == 7U`. For bits=6, bytes 48-55 are not explicitly written or zeroed. This is correct as long as the caller has already zeroed the output buffer (which `RaBitQPackScalarSupplementCode` does via `std::memset(packed_supplement, 0, layout.PackedSize())`). However, if this function is ever called directly without a prior memset, the bits=6 case would leave stale data in bytes 48-55.

Consider adding a `std::memset(packed + 48U, 0, 8U)` for the bits=6 case as well, or documenting the precondition that the output buffer must be pre-zeroed.

LHT129 (changes requested):
## Summary

This is a large PR (+11,940/-300, 60 files) that implements RaBitQ x+y split storage and search for IVF indexes. The core idea is to split RaBitQ codes into filter bits (x) and supplement/reorder bits (y), store them separately with different I/O strategies, and use a two-phase search: fast filter-bit scan to collect candidates, then supplement-bit reorder for precise distances.

## What I Reviewed

- **Core datacell**: `rabitq_split_bucket_datacell.cpp/h` (2570 lines) — split storage, FastScan32, residual/non-residual query dispatch, version-aware staleness checks
- **Searcher**: `rabitq_split_bucket_searcher.cpp/h` (290 lines) — two-phase search with filter inner product collection
- **Quantizer**: `rabitq_quantizer.cpp/h` (+1200 lines) — split code pack/unpack, FastScan32 packing, FastEncodeRaBitQ, residual query transform
- **SIMD kernels**: `rabitq_packed_supplement.h` (new), `generic.cpp`, `avx2.cpp`, `avx512.cpp`, etc. — packed supplement format, FastScan32 LUT dispatch
- **Infrastructure**: `auxiliary_standard_heap`, `bucket_interface.h`, `flatten_interface.h`, `reorder.h`, `bucket_reorder.cpp`, `flatten_reorder.cpp`
- **Tests**: `bucket_datacell_test.cpp` (+1213), `flatten_datacell_test.cpp` (+350), `ivf_parameter_test.cpp` (+86), `rabitq_quantizer_test.cpp` (+650), etc.
- **Docs**: English and Chinese documentation for IVF and RaBitQ split storage

## Key Findings

### Critical (already noted in existing review comments)
- **`std::isfinite` under `-Ofast`**: Used in `ivf.cpp:614` (SplitBucketReorder) and `rabitq_split_bucket_searcher.cpp:179,205`. Will be eliminated by `-Ofast`, silently removing lower-bound and validity filters. The PR already introduces `IsFiniteFloatBits` elsewhere for this exact reason.

### Suggestions (already noted)
- `IsFiniteFloat` duplicated across 4 files — centralize in a shared utility
- ~400 lines of near-duplicate code between `query_non_residual_by_inner_ids` and `query_residual_by_inner_ids`
- Hardcoded `0xFFFFFFFFULL` instead of `kLocationMask`
- Return value of `query_current_by_inner_id` discarded at two call sites

### New observations from this review
1. **`ComputeFastScan32ResidualFactors` uses `std::fabs` instead of `EffectiveRaBitQError`** — The `ComputeDistWithSplitCode` path was updated to use `EffectiveRaBitQError` (which returns ±1.0 for near-zero errors based on sign), but the residual factor path uses `std::fabs`. This inconsistency could produce subtle numerical differences between the FastScan32 and exact paths.

2. **`RaBitQFloatSQIPByPlanes` fallback is O(dim × bits) scalar** — The fallback path for `filter_bits > 3` uses a scalar loop reconstructing full codes per dimension. If hit in production, this would be significantly slower than SIMD paths.

3. **`PrepareFastScan32HighAccQuery` computes `deltas[0]` before range check** — A wasted division when range is below epsilon.

4. **`ProcessQueryImpl` always transforms for L2SQR** — Adds PCA + ROM transform overhead on every L2 query even when residual mode is not active.

5. **Supplement code format change** — The split storage now uses packed supplement format (compact chunks) instead of traditional bit-planes. This is correctly paired with `SplitCode`/`MergeSplitCode` changes but is a significant format change worth documenting.

6. **`RaBitQPackCompactSupplementChunk` bits=6 precondition** — The bits=6 case relies on the caller pre-zeroing the output buffer. Consider making this explicit or adding the memset defensively.

## Overall Assessment

The architecture is well-thought-out: the two-phase search with filter inner product caching, version-aware staleness detection, and FastScan32 batch processing is a solid design. The test coverage is reasonable for a feature of this size. The main concerns are the `-Ofast` compatibility issues (already identified) and the numerical consistency between the FastScan32 residual path and the exact split-code path.

src/datacell/flatten_interface.h
```diff
@@ -178,6 +218,173 @@ class FlattenInterface {
...
+    PrefetchSupplement(InnerIdType id) {
+        this->Prefetch(id);
+    }
+
+    virtual void
+    DiscardFilterCodes() {
+    }
+
+    [[nodiscard]] virtual bool
+    HasFilterCodes() const {
```
LHT129, line 378: [suggestion] `HasFilterCodes()` defaults to `true` in the base class, but every non-split `FlattenInterface` implementation (FP32, SQ8, standard RaBitQ, etc.) has no filter codes. A default of `true` means callers that check `HasFilterCodes()` will incorrectly believe filter codes are available. Consider defaulting to `false` and only overriding to `true` in `RaBitQSplitDataCell`.

src/datacell/flatten_interface.h
```diff
@@ -104,9 +104,49 @@ class FlattenInterface {
         this->Query(result_dists, computer, idx, id_count, ctx);
     }
 
+    virtual void
+    QueryWithFilterInnerProduct(float* result_dists,
```
LHT129, line 108: [suggestion] `QueryWithFilterInnerProduct` (singular) silently ignores the `filter_inner_products` parameter and delegates to `Query()`. This is confusing because the plural form `QueryWithFilterInnerProducts` has completely different semantics (supplement-only query). The singular form appears to exist only for API symmetry but its behavior is identical to `Query()`. Consider either removing it or making the naming clearly distinct (e.g. `QueryWithOptionalFilterInnerProduct`), since a caller passing filter inner products would expect them to be used.

src/datacell/bucket_interface.h
```diff
@@ -41,13 +43,69 @@ class BucketInterface {
...
                  const BucketIdType& bucket_id,
                  const InnerIdType& offset_id) = 0;
 
+    [[nodiscard]] virtual bool
+    SupportSplitCodeStorage() const {
+        return false;
+    }
+
+    virtual void
+    QueryWithCandidateFilterInnerProductBySource(float* result_dists,
```
LHT129, line 80: [suggestion] The default implementation of `QueryWithCandidateFilterInnerProductBySource` discards `source_bucket_ids`, `source_offset_ids`, and `source_versions` and delegates to `QueryWithDistanceHintByInnerId`, which throws `UNSUPPORTED_INDEX_OPERATION`. This means any non-split bucket implementation that reaches this path will always throw. Consider providing a fallback that at least computes distances via `QueryOneById` for each inner_id, so callers that do not need source provenance can still get correct results without implementing the full split-code interface.

src/algorithm/ivf/ivf.cpp
```diff
@@ -2459,7 +2749,32 @@ IVF::CalDistanceById(const float* query,
             }
         }
     }
-    if (this->use_reorder_ && calculate_precise_distance && reorder_codes_ != nullptr) {
+    if (this->bucket_->SupportSplitCodeStorage() and calculate_precise_distance) {
+        auto computer = this->bucket_->FactoryComputer(query);
```
LHT129, line 2753: [suggestion] `CalDistanceById` (batch, line 2753) and `CalcDistanceById` (single, line 2834) both call `FactoryComputer(query)` unconditionally on the split-storage path, which allocates a new computer on every invocation. For the batch variant this is a single allocation per call, but `CalcDistanceById` may be called in a tight loop (e.g. per-ID validation). Consider accepting an optional `preset_computer` parameter (as `SplitBucketReorder::Reorder` already does at line 584) so callers that already hold a computer can reuse it, or document that this path is not intended for hot loops.

src/datacell/rabitq_split_bucket_datacell.h
```diff
@@ -0,0 +1,390 @@
...
+            routed_bucket_ids_.reserve(bucket_computer_capacity);
+            routed_bucket_norm_sqrs_.reserve(bucket_computer_capacity);
+        }
+
+        ComputerInterfacePtr inner_{nullptr};
+        ComputerInterfacePtr fastscan_{nullptr};
+        Vector<float> raw_query_;
+        Vector<float> query_centroid_adjustments_;
+        Vector<float> transformed_query_;
+        mutable Vector<float> residual_query_scratch_;
```
LHT129, line 209: [suggestion] `residual_query_scratch_` is declared `mutable` but is used as a scratch buffer inside `ResetComputerFromResidualQuery` (called from `ScanBucketById` and `QueryWithCandidateFilterInnerProductBySource`). If multiple threads share the same `SplitBucketComputer` instance and call methods that trigger residual query reset concurrently, this could cause a data race on `residual_query_scratch_`. The existing `bucket_computers_mutex_` protects `bucket_ids_`/`bucket_inner_computers_`/`bucket_fastscan_computers_` but does not cover `residual_query_scratch_`. Consider either protecting it with a separate mutex or documenting that `SplitBucketComputer` is not thread-safe and must be used from a single thread at a time.

src/algorithm/ivf/ivf.cpp
LightWant: Fixed in `d59e2ab5`. The threshold check now calls `vsag::IsFiniteFloatBits` from the new `src/utils/float_utils.h`, which tests the IEEE-754 exponent bits instead of `std::isfinite`, so the guard survives `-Ofast`.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Fixed in `d59e2ab5`. The copies in `rabitq_split_bucket_datacell.cpp`, `rabitq_split_bucket_searcher.cpp`, `rabitq_quantizer.cpp` and `simd/generic.cpp` are gone; every call site now uses the single `vsag::IsFiniteFloatBits` declared in the new `src/utils/float_utils.h`, which is also what the two `std::isfinite` call sites use.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: Thanks for confirming. The `thread_local` splitmix state stays as it is: it keeps the per-thread streams independent without threading a seed through the API, and the explicit `random_state` argument still allows deterministic replay when a caller supplies one.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Thanks. The bucket -> codes -> locations order is documented at the recheck inside `query_current_by_inner_id`, and the final revision adds no new lock-taking path, so the note needs no code change.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Fair point, and agreed that the two functions share the whole grouping/staleness/batching skeleton. The residual variant also carries transformed centroids, original-query filter inner products and the `full_add` path through every stage, so a shared helper would need roughly a dozen parameters or a visitor, which we judged harder to follow than the duplication. Keeping them separate also lets the residual path be tuned without touching the non-residual one. We would rather land that extraction as its own change, with the split benchmarks available to show it is behaviour preserving.

src/impl/heap/auxiliary_standard_heap.cpp
LightWant: `BucketIdType` is `int32_t` (`src/basic_types.h`), not `uint32_t`, so `-1` does not wrap here - it is the canonical sentinel for this field: `distance_heap.h` already defaults `source_bucket_id` to `-1`, and `distance_heap_test.cpp` asserts the value. I did try `std::numeric_limits<BucketIdType>::max()` while checking this, which made that test fail with `2147483647 == -1`, so the literal stays. Thanks for prompting the check.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Fixed in `d59e2ab5`: both sites use the named `LOCATION_MASK` constant, now shared with `query_current_by_inner_id`, so it follows `LOCATION_SPLIT_BIT` automatically.

src/datacell/flatten_interface.h
LightWant: Agreed and fixed in `d59e2ab5`. The base implementation now returns `false` and only `RaBitQSplitDataCell`, which really stores filter codes, overrides it to `true`. The single caller in the split legacy-load path therefore throws only when the serialized data genuinely carries no filter codes.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Documented in `d59e2ab5`. Keeping `FLT_MAX` is intentional: `query_current_by_inner_id` seeds `result_dist` with `FLT_MAX` and only replaces it on success, and the caller passes that through for a lane that was deleted or kept moving across retries. The batch interface promises one distance per requested inner id, `FLT_MAX` can never win top-k selection, and the alternative (`-1`) is already taken by `CalcDistancesById` for ids that are not in the index at all. A comment at both call sites now records that contract.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: Same as the other call site: documented in `d59e2ab5` rather than changed, because the `FLT_MAX` fallback is the intended contract of `query_current_by_inner_id` for a lane that disappeared. The comment now says so.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: Documented in `d59e2ab5`, deliberately not changed. `ComputeFastScan32ResidualFactors` divides `base_norm` by `filter_error` to build a *magnitude* scale that is stored in the FastScan block, so the sign of the stored error has to be dropped. `EffectiveRaBitQError` preserves that sign (it only substitutes +/-1 near zero), which is what the exact split-code path needs; using it here would flip the sign of `f_scale` for negative errors. A comment now records that the difference is intentional.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: This does not add work for non-residual L2. `TransformResidualQuery` performs the same PCA plus ROM transform that `TransformFusedQuery` already performed on every L2 query before this PR; the new entry point just reuses that transformed query and `ProcessTransformedResidualQuery` dispatches to the non-residual computation when residual mode is off. Skipping it would change the query encoding, not only its cost, so there is nothing to gate here.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: Confirmed, and it is the intended format. Split storage now keeps supplements in the packed chunk layout, and all readers go through the packed accessors (`RaBitQFloatPackedSupplementCodeIP`, `RaBitQUnpackSupplementPlanes`), with `SplitCode`/`MergeSplitCode` converting at the boundary.

src/simd/kernels/rabitq_packed_supplement.h
LightWant: Checked carefully - this one has to stay as it is. A compact chunk spans `ChunkBytes() == bits * 8` bytes, so for `bits == 6` the chunk ends at `packed[48]` and `packed[48..55]` does not exist; the suggested `memset(packed + 48, 0, 8)` would write past the chunk (and past `PackedSize()` for a single-chunk layout). There are no stale bytes: `RaBitQPackScalarSupplementCode` zeroes the whole `PackedSize()` buffer first and the 6-bit path writes every byte of the chunk. `d59e2ab5` adds a comment documenting that precondition instead.

src/datacell/flatten_interface.h
LightWant: The singular entry point is the generic "query one lane with optional external filter codes" call that the split fallback batch needs; the overload in `rabitq_split_datacell.h` does consume `filter_inner_products`, while the base implementation can only ignore it because a datacell without filter codes has nothing to apply it to. Agreed that it reads as a near-duplicate of `Query()`; we are happy to fold or rename it in a follow-up, but would rather not churn the datacell interface inside this PR.

src/datacell/bucket_interface.h
LightWant: This is deliberately fail-closed. The method carries source bucket/offset/version provenance, and only `SplitBucketReorder::Reorder` produces candidates with that provenance (it is the reorder used when the bucket advertises split storage), so a non-split bucket reaching this default is a mis-selection bug. A fallback that recomputed per inner id would silently publish distances without the version check the split path relies on, which is worse than the exception.

src/algorithm/ivf/ivf.cpp
LightWant: `CalcDistancesById` builds one computer for the whole call, not one per id, so the batch path is already fine. The single-id `CalcDistanceById` does allocate, but it is a public entry point that only has the raw query pointer, and the split path has to build the bucket-specific residual computer from it anyway. We would rather keep that signature than add a `preset_computer` parameter for a path that is not a hot loop; happy to revisit if a caller shows up in one.

src/datacell/rabitq_split_bucket_datacell.h
LightWant: The mutex does cover it. `residual_query_scratch_` is written only inside `append_scan_computer`, and the only caller of `append_scan_computer` is `get_scan_computers`, which holds `bucket_computers_mutex_` around the whole lookup-or-append sequence (line 127). `prepare_scan_computers` uses a local `residual_scratch` rather than the member. Two threads sharing one computer therefore serialize on that mutex, so no extra guard or documented restriction is needed.

src/algorithm/ivf/flat_bucket_searcher.cpp
LightWant: Good question. `ScanBucketById` reports how many lanes it actually scanned and fills `scanned_inner_ids` with the inner ids of exactly those lanes (see `BucketDataCell::scan_bucket_by_id`), which can be fewer than `GetInnerIds(bucket_id).size()`. The distances in `dist` are produced for the scanned lanes only, so `ids[j]` has to index the scanned list - using the full inner-id vector would pair a distance with the wrong id. Passing the caller-owned buffer also avoids a second allocation per bucket scan. So yes, the parameter is the right shape.

src/datacell/rabitq_split_bucket_datacell.cpp
LightWant: The split-specific paths are covered indirectly but fairly broadly: `bucket_datacell_test.cpp` exercises the split bucket contract including candidate filter inner products, source provenance and its rejection case, stale-version fallback and the optimized-build path, `flat_bucket_searcher_test.cpp` covers the split searcher wiring, and `rabitq_quantizer_test.cpp` covers split-code packing, the finite guard and the FastScan32 residual factors. A dedicated `rabitq_split_bucket_datacell_test.cpp` is a reasonable follow-up; we did not want to grow this PR further while it is still being reviewed.

src/algorithm/ivf/rabitq_split_bucket_searcher.cpp
LightWant: Addressed. Both the lower-bound filter and the exact-distance validity check in `rabitq_split_bucket_searcher.cpp` now go through the shared `IsFiniteFloatBits` (`d59e2ab5`), so `-Ofast` cannot fold them away.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: Correct. `RaBitQFloatSQIPByPlanes` is the generic fallback for `filter_bits > 3`, while the split FastScan32 path this PR optimizes is limited to `FASTSCAN_MAX_FILTER_BITS == 3` (the 1+7, 2+6 and 3+5 layouts), so the scalar branch is a correctness fallback rather than a production path for this feature. We will add a SIMD kernel if a wider filter ever becomes configurable.

src/quantization/rabitq_quantization/rabitq_quantizer.cpp
LightWant: Already the case in the current revision: `PrepareFastScan32HighAccQuery` tests `range <= std::numeric_limits<float>::epsilon()` and returns with a zeroed lookup table before computing `deltas[0]`, so the division is not wasted.

LightWant:
## Update: conflicts resolved, CI fixes, review follow-ups

Pushed `a08ee70e` on top of `feat/issue-2443`; `mergeable` is `MERGEABLE` again.

### Merge conflicts

Merged `origin/main` (`7dc90ad5`) into the branch. Three files conflicted:

- `src/algorithm/ivf/ivf.cpp` - kept the split-code branch of `CalcDistancesById` ahead of the reorder branch, adopted main's validity-filtered reorder query and its `use_reorder_` guard, and aligned the single-id `CalcDistanceById` guard with main.
- `src/algorithm/ivf/ivf_build.cpp` - adopted main's unconditional `SUPPORT_CAL_DISTANCE_BY_ID` / `SUPPORT_BATCH_CALC_DISTANCE_BY_ID` advertisement, since stored distances no longer require FP32 storage (main's new `[distance_contract]` tests assert this for `sq8` and `pqfs`).
- `tests/test_ivf.cpp` - kept both the RaBitQ split-storage fixture test and main's distance-contract tests.

### CI

- **Format Check** - reformatted with clang-format-15; passes.
- **SIMD scoped coverage** - the gate failed with `Unmeasured changed file` for `src/simd/pqfs_simd.h` and `src/simd/rabitq_simd.h`. Both headers only carry prototypes, aliases and macros, so they can never produce gcov line records. `compare_simd_coverage.py` now classifies a changed scoped *header* with no detectable function body as declaration-only, keeps it out of the error list and reports it in the summary. Changed translation units are still never exempt and the detection errs towards "defines code", so the gate stays fail-closed. `tests/scripts/simd_coverage_test.py` covers the new behaviour.
- **Run TSAN Tests** - TSan reported a race between OpenBLAS zero-scaling its output buffer from its persistent worker threads and the caller reading that buffer after `cblas_sgemm` returned (in `NearestCentroidAssign` during KMeans training). OpenBLAS rejoins its workers with an internal barrier that TSan cannot observe, and neither side is code this PR touches (`nearest_centroid_assign.cpp` comes from #2816). The workflow now suppresses `libopenblas` race reports exactly like the existing `libomp` / `libmkl_intel_thread` entries.

### Review follow-ups

All 23 review threads have been answered and resolved. Must-fix items:

- `std::isfinite` under `-Ofast` in `SplitBucketReorder::Reorder` and `FlatBucketSearcher::Search` - replaced with a shared `vsag::IsFiniteFloatBits` in the new `src/utils/float_utils.h`, and the four duplicate copies of the helper are gone. Linux Release builds really do use `-Ofast` (i.e. `-ffast-math`), so those guards were being compiled out.
- `FlattenInterface::HasFilterCodes()` now defaults to `false`; only `RaBitQSplitDataCell`, which really stores filter codes, overrides it to `true`.

Also applied: the named `LOCATION_MASK` constant instead of the hardcoded `0xFFFFFFFFULL`, plus comments recording the intentional `FLT_MAX` fallback for stale lanes, the magnitude-only `std::fabs` residual factor, and the pre-zeroed chunk precondition of `RaBitQPackCompactSupplementChunk`.

Two suggestions were checked and deliberately not applied:

- The `-1` source-bucket sentinel in `AuxiliaryStandardHeap` stays: `BucketIdType` is `int32_t`, not `uint32_t`, so it does not wrap, and `distance_heap_test.cpp` asserts `-1`.
- The suggested `memset(packed + 48, 0, 8)` for `bits == 6` would write past `ChunkBytes() == 48`; a comment documents the precondition instead.

### Local validation

Release build with `-Ofast` (library, unittests, functests) plus `[distance_contract]` (403 assertions / 12 cases), `[rabitq_split]` (19473 assertions / 26 cases) and `[ivf]~[daily]` (912016 assertions / 61 cases) pass, as does `tests/scripts/simd_coverage_test.py`.

LightWant (commented):
Submitting this review so that the thread replies (which GitHub attached to the pending review) become visible.

All 23 review threads are answered and resolved. The must-fix items are the two `std::isfinite` `-Ofast` regressions and the `HasFilterCodes()` default; the remaining suggestions are either applied or explained in the replies. The conflict resolution, CI fixes and local validation are summarised in the PR comment above.

LightWant:
## All checks are green

`3825ca22` passes every job, including the ones that had never actually run on this PR before because they were gated behind the Format Check / ASan jobs:

- Format Check, DCO, Coverage Configuration, Validate documentation, issue link
- ASan Build X86 and ASan Build macOS arm64, Python Build X86, Test Python X86
- Build with TSAN, Run TSAN Tests
- SIMD scoped coverage
- Test X86 Unittests, Test X86 Functests, Test Example X86, Check Compatibility
- CircleCI `pull-request-check`

One more fix was needed to get there: the ASan job's changed-source lint step (clang-tidy 15 with warnings as errors) reported 16 findings in `src/datacell/rabitq_split_bucket_datacell.cpp` and `src/quantization/rabitq_quantization/rabitq_quantizer.cpp` - naming, a DeMorgan pattern inside `CHECK_ARGUMENT`, one member function that can be static, and one discarded `[[nodiscard]]` result. The ASan build itself was already fine; only lint failed, and that step had been skipped while Format Check was red. The commit message of `3825ca22` has the breakdown.

`mergeable` is back to `true`. The only thing still open is the review state: LHT129's earlier `CHANGES_REQUESTED` stands, and every review thread is answered and resolved, so a re-review would be appreciated.

LHT129 (approved):
LGTM

PR #2709, opened 2026-08-18
fix: correct sindi n_candidate error message and doc

### LINKED ISSUE #2708: [bug](sindi): n_candidate error message uses wrong amplification factor
### Affected area

sindi

### Interface

cpp

### VSAG version

v0.18.10 (also reproducible on current main)

### Summary

In `src/algorithm/sindi/sindi.cpp`, the `n_candidate` limit check uses `SPARSE_AMPLIFICATION_FACTOR * k` (500×), but the error message printed to the user uses `AMPLIFICATION_FACTOR * k` (100×). This makes the error message contradict the actual enforcement logic and misleads users when tuning the parameter.

**Code location:**

```cpp
CHECK_ARGUMENT(search_param.n_candidate <= SPARSE_AMPLIFICATION_FACTOR * k,
               fmt::format("n_candidate ({}) should be less than {} * k ({})",
                           search_param.n_candidate,
                           AMPLIFICATION_FACTOR,   // <- should be SPARSE_AMPLIFICATION_FACTOR
                           k));

### Minimal reproducer

With topk = 10 and n_candidate = 5000000, the error says:

n_candidate (5000000) should be less than 100 * k (10)

### Environment

```shell
-
```

### Expected behavior

The error message should read less than 500 * k to match the actual SPARSE_AMPLIFICATION_FACTOR * k check.

### Actual behavior

n_candidate (5000000) should be less than 100 * k (10)

### References

_No response_

### Additional context

`docs/docs/en/src/indexes/sindi.md` and `docs/docs/zh/src/indexes/sindi.md` state that `n_candidate = 0` defaults to `SPARSE_AMPLIFICATION_FACTOR · topk` (500×). The actual implementation uses `ef = max(n_candidate, k)`, so the default candidate heap size is `topk`, not `500 * topk`.

biaozy:
# fix: correct sindi n_candidate error message and doc

Fixes: #2708

## Description

The `SINDI::KnnSearch` parameter check validates `n_candidate` against `SPARSE_AMPLIFICATION_FACTOR * k` (500x), but the error message printed `AMPLIFICATION_FACTOR * k` (100x). This mismatch misleads users when tuning the parameter.

Additionally, the SINDI documentation stated that `n_candidate = 0` defaults to `SPARSE_AMPLIFICATION_FACTOR * topk` (500x), which does not match the actual implementation. The code uses `ef = max(n_candidate, k)`, so the default candidate heap size is `topk`.

## Changes

- `src/algorithm/sindi/sindi.cpp`: use `SPARSE_AMPLIFICATION_FACTOR` in the error message.
- `docs/docs/en/src/indexes/sindi.md`: correct the description of `n_candidate` default behavior and allowed range.
- `docs/docs/zh/src/indexes/sindi.md`: sync the Chinese documentation.
- `tests/test_sindi.cpp`: add a test verifying that the error message matches the actual limit (`500 * k`).

## Testing

- Built with `make release` in the `vsaglib/vsag:ubuntu` Docker development image.
- Ran the new SINDI test: the boundary case `n_candidate = 500 * k` succeeds, and `n_candidate = 500 * k + 1` fails with a message containing `"should be less than 500 * k"`.
- Ran `make fmt` and `make lint`; no new issues introduced.

## Checklist

- [x] Code follows the project coding style.
- [x] Added/updated unit tests for the bug fix.
- [x] Updated both English and Chinese documentation.
- [x] Verified the build and tests pass.

### DIFF
--- a/tests/test_sindi.cpp
+++ b/tests/test_sindi.cpp
@@ -201,6 +201,31 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::SINDITestIndex,
 }
 
 TEST_CASE_PERSISTENT_FIXTURE(fixtures::SINDITestIndex,
+                             "n_candidate Limit Matches Error Message",
+                             "[ft][search][sindi]") {
+    fixtures::SINDIParam param;
+    auto index = TestFactory("sindi", GenerateBuildParameter(param), true);
+    auto dataset = pool.GetSparseDatasetAndCreate(base_count, 128, 0.8);
+    TestBuildIndex(index, dataset, true);
+    auto query = fixtures::get_one_query(dataset->query_, 0);
+
+    const auto make_param = [](uint32_t n_candidate) {
+        return fmt::format(R"({{"sindi":{{"n_candidate":{},"query_prune_ratio":0.0,)"
+                           R"("term_prune_ratio":0.0}}}})",
+                           n_candidate);
+    };
+
+    constexpr int64_t k = 1;
+    // the error message must cite the same factor (500) used by the check
+    auto result = index->KnnSearch(query, k, make_param(500 * k + 1));
+    REQUIRE_FALSE(result.has_value());
+    REQUIRE(result.error().message.find("should be less than 500 * k") != std::string::npos);
+
+    result = index->KnnSearch(query, k, make_param(500 * k));
+    REQUIRE(result.has_value());
+}
+
+TEST_CASE_PERSISTENT_FIXTURE(fixtures::SINDITestIndex,
                              "SINDI Build and Search",
                              "[ft][build][search][sindi]") {
     fixtures::SINDIParam param;
--- a/docs/docs/en/src/indexes/sindi.md
+++ b/docs/docs/en/src/indexes/sindi.md
@@ -141,7 +141,7 @@ Search-time parameters live under the `sindi` sub-object:
 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `n_candidate` | int | `0` | Candidate heap size. When `0`, defaults to `SPARSE_AMPLIFICATION_FACTOR · topk` (500×). If set, must satisfy `1 ≤ n_candidate ≤ SPARSE_AMPLIFICATION_FACTOR · topk`. |
+| `n_candidate` | int | `0` | Candidate heap size. The effective candidate effort is `max(n_candidate, topk)`, so the default `0` keeps the heap at `topk` with no candidate amplification. Must satisfy `0 ≤ n_candidate ≤ SPARSE_AMPLIFICATION_FACTOR · topk` (500×). |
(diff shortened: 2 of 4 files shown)


biaozy:
Hi maintainers, could you please add `kind/bug` and `version/1.0` labels to this PR? The Mergify check is blocked without them. Thanks!

LHT129 (commented):
The fix is correct and well-scoped. Here is a summary of the review:

**sindi.cpp** — The error message now correctly uses `SPARSE_AMPLIFICATION_FACTOR` (500) instead of `AMPLIFICATION_FACTOR` (100), matching the actual `CHECK_ARGUMENT` condition on the line above. This also brings `KnnSearch` into consistency with `SearchWithRequest`, which already uses `SPARSE_AMPLIFICATION_FACTOR` in its error message.

**Documentation (en + zh)** — The updated descriptions accurately reflect the actual implementation: the effective candidate effort is `max(n_candidate, topk)`, and the valid range is `0 ≤ n_candidate ≤ SPARSE_AMPLIFICATION_FACTOR * topk`.

**Test** — The new test case validates both the error message content and the boundary condition (`500 * k` succeeds, `500 * k + 1` fails), which is a good regression test for this fix.

No issues found. LGTM.

tests/test_sindi.cpp
```diff
@@ -200,6 +200,31 @@ TEST_CASE_PERSISTENT_FIXTURE(fixtures::SINDITestIndex,
...
+        return fmt::format(R"({{"sindi":{{"n_candidate":{},"query_prune_ratio":0.0,)"
+                           R"("term_prune_ratio":0.0}}}})",
+                           n_candidate);
+    };
+
+    constexpr int64_t k = 1;
+    // the error message must cite the same factor (500) used by the check
+    auto result = index->KnnSearch(query, k, make_param(500 * k + 1));
+    REQUIRE_FALSE(result.has_value());
+    REQUIRE(result.error().message.find("should be less than 500 * k") != std::string::npos);
```
LHT129, line 222: [suggestion] The test hardcodes `500` instead of using the named constant `SPARSE_AMPLIFICATION_FACTOR`. If the constant value ever changes, this test would need manual updates in two places (the boundary value and the error message assertion).

Consider computing the boundary from the constant:

```cpp
constexpr int64_t k = 1;
constexpr int64_t limit = SPARSE_AMPLIFICATION_FACTOR * k;
auto result = index->KnnSearch(query, k, make_param(limit + 1));
REQUIRE(result.error().message.find("should be less than 500 * k") != std::string::npos);

result = index->KnnSearch(query, k, make_param(limit));
REQUIRE(result.has_value());
```

LHT129 (commented):
Overall this is a clean, well-scoped fix. The error message now correctly reports `SPARSE_AMPLIFICATION_FACTOR` (500) instead of `AMPLIFICATION_FACTOR` (100), matching the actual validation check. The documentation updates correctly describe the real `n_candidate` default behavior. The new test validates the fix end-to-end.

One suggestion below.

Issue #2729, opened 2026-08-20
[Refactor] Replace template-based parameter mapping with structured builders


LHT129:
## Description
Refactor the index parameter mapping layer to replace mutable JSON string templates and path-only mapping tables with structured default builders and explicit flat-field translation.

## Background
The existing mapping flow mixed structure, defaults, placeholder expansion, and post-parse mutation. RaBitQ split mode also depended on a multi-stage mutation chain across index mapping and DataCell parsing.

## Requirements
- Build defaults structurally for HGraph, Pyramid, IVF, BruteForce/WARP, and SIMQ.
- Preserve flat public parameter inputs and explicit one-to-many mappings.
- Centralize RaBitQ split parsing and validation.
- Remove downstream split-version correction from FlattenDataCellParameter.
- Route SINDI through the same unknown-field validation boundary.
- Add a compatibility report API that collects all JSON differences while preserving the existing boolean API.

## Acceptance Criteria
- No parameter entry uses format_map, mapping_external_param_to_inner, or ConstParamMap.
- Existing parameter serialization and runtime behavior remain compatible.
- Release build, non-daily unit tests, clang-format 15, and clang-tidy 15 pass.

## Testing
- Release build completed successfully.
- 713 non-daily unit tests passed with 85,349,000 assertions.
- clang-format 15 and clang-tidy 15 passed.

PR #2730, opened 2026-08-20
refactor(parameters): replace template-based index mappings

LHT129:
## Summary
Replace template-based parameter mapping with structured defaults and explicit flat-field translation across index entry points.

## Changes
- Replace JSON string templates in HGraph, Pyramid, IVF, BruteForce/WARP, and SIMQ with structural builders.
- Remove ConstParamMap usage while preserving one-to-many mappings and unknown-field rejection.
- Introduce RaBitQSplitConfig parsing/application and remove the downstream split-version mutation.
- Add SINDI and SINDI v2 flat-key boundary validation.
- Add CompatibilityReport issue collection while preserving CheckCompatibility.
- Add regression coverage for split configuration and multi-difference compatibility reporting.

## Behavior change
- `codes_type=rabitq_split` now requires the nested RaBitQ quantizer to already use `rabitq_version=split`. Inconsistent configurations are rejected instead of silently mutating `rabitq_version` during `FlattenDataCellParameter::FromJson`. Public index mappings construct the canonical split configuration before parsing.

## Testing
- Release build passed using the remote fixed-version dependency cache after a GitHub download timeout.
- 713 non-daily unit tests passed with 85,349,000 assertions.
- SINDIV2Parameter focused suite passed: 15 test cases, 68 assertions.
- HGraphParameter focused suite passed: 19 test cases, 128 assertions.
- clang-format 15 passed.
- clang-tidy 15 passed.

Related to #2729


src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1223,151 +1191,73 @@ BruteForce::CheckAndMappingExternalParam(const JsonType& external_param,
...
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY][TYPE_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_PRECISE_IO_TYPE) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][TYPE_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_PRECISE_FILE_PATH) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][IO_FILE_PATH_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_THREAD_COUNT) {
+            inner_json[BUILD_THREAD_COUNT_KEY].SetJson(field);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(field);
+        } else if (key == USE_ATTRIBUTE_FILTER) {
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a: STORE_RAW_VECTOR now maps to base_codes.quantization_params.hold_molds and precise_codes.quantization_params.hold_molds. Added a regression test covering both nested paths and ensuring no unused top-level quantization_params is created.

src/parameter.h
```diff
@@ -57,6 +76,59 @@ class Parameter {
...
+                        }
+                        return;
+                    }
+                    if (lhs.is_array()) {
+                        if (lhs.size() != rhs.size()) {
+                            report.issues.push_back({path, "array sizes differ"});
+                        }
+                        const auto common_size = std::min(lhs.size(), rhs.size());
+                        for (size_t i = 0; i < common_size; ++i) {
+                            collect(lhs[i], rhs[i], path + "[" + std::to_string(i) + "]");
```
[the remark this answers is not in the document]
LHT129: Fixed in the current implementation: array traversal now uses uint64_t rather than size_t.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1223,151 +1192,73 @@ BruteForce::CheckAndMappingExternalParam(const JsonType& external_param,
...
+        } else if (key == BRUTE_FORCE_PRECISE_QUANTIZATION_TYPE) {
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY][TYPE_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_PRECISE_IO_TYPE) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][TYPE_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_PRECISE_FILE_PATH) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][IO_FILE_PATH_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_THREAD_COUNT) {
+            inner_json[BUILD_THREAD_COUNT_KEY].SetJson(field);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(field);
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a: STORE_RAW_VECTOR is applied to both base and precise flatten quantizers through their nested quantization_params.hold_molds fields. The new regression test verifies the effective serialized configuration.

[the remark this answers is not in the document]
LHT129: Fixed in b462281a: the flag is now mapped under base_codes and precise_codes, consistent with the structured default tree, with regression coverage.

src/parameter.h
```diff
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <vector>
+
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a by adding an explicit <string> include to parameter.h.

src/algorithm/sindi/sindi.cpp
```diff
@@ -19,6 +19,7 @@
 #include <cstring>
 #include <limits>
 #include <memory>
+#include <nlohmann/json.hpp>
 #include <shared_mutex>
 #include <unordered_map>
 #include <vector>
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a by adding an explicit <unordered_set> include to sindi.cpp.

src/algorithm/sindi/sindi.cpp
```diff
@@ -246,6 +247,25 @@ get_bits_for_term_id_limit(uint32_t term_id_limit) {
...
+        USE_QUANTIZATION,
+        SPARSE_WINDOW_SIZE,
+        SPARSE_AVG_DOC_TERM_LENGTH,
+        SPARSE_DESERIALIZE_WITHOUT_FOOTER,
+        SPARSE_DESERIALIZE_WITHOUT_BUFFER,
+        SPARSE_REMAP_TERM_IDS,
+        SPARSE_RERANK_TYPE,
+        SPARSE_DMQ_SHARED_CODEBOOK_THRESHOLD,
+        SPARSE_IMMUTABLE,
+    };
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a by adding the missing <unordered_set> include.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1223,151 +1193,73 @@ BruteForce::CheckAndMappingExternalParam(const JsonType& external_param,
...
```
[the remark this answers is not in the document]
LHT129: Fixed in b462281a: STORE_RAW_VECTOR now updates base_codes.quantization_params.hold_molds and precise_codes.quantization_params.hold_molds. Added a focused BruteForce parameter regression test.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,660 +14,297 @@
...
+                                 bool hold_molds) {
+    if (quantization_type == QUANTIZATION_TYPE_VALUE_TQ) {
+        return TransformQuantizerParameter::CreateDefault(tq_chain)->ToJson();
+    }
+    auto parameter = QuantizerParameter::CreateDefault(quantization_type);
+    if (auto fp32 = std::dynamic_pointer_cast<FP32QuantizerParameter>(parameter); fp32 != nullptr) {
+        fp32->hold_molds = hold_molds;
+    }
+    return parameter->ToJson();
+}
```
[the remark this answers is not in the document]
LHT129: Fixed in 86a6c95d by removing the unused build_default_quantization_param helper. The translation unit compiles successfully, and the focused HGraphParameter suite passes (19 test cases, 128 assertions).

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,660 +14,283 @@
...
+#include <nlohmann/json.hpp>
+
 #include "common.h"
 #include "hgraph.h"  // IWYU pragma: keep
 #include "hgraph_parameter.h"
+#include "quantization/fp32_quantizer_parameter.h"
+#include "quantization/rabitq_quantization/rabitq_quantizer_parameter.h"
+#include "quantization/transform_quantization/transform_quantizer_parameter.h"
 
 namespace vsag {
```
LHT129, line 26: [suggestion] `rabitq_quantizer_parameter.h` and `fp32_quantizer_parameter.h` are now unused includes after removing `build_default_quantization_param`.

- `rabitq_quantizer_parameter.h` was only referenced by the removed `build_default_quantization_param` function.
- `fp32_quantizer_parameter.h` was only used for `std::dynamic_pointer_cast<FP32QuantizerParameter>` in that same function.

The remaining code in this translation unit only uses `TransformQuantizerParameter::CreateDefault`, so `transform_quantizer_parameter.h` is still needed. Consider removing the two unused includes to keep the include set minimal.

src/datacell/flatten_datacell_parameter.cpp
```diff
@@ -53,30 +68,18 @@ FlattenDataCellParameter::FromJson(const JsonType& json) {
...
-    // codes_type=rabitq_split is the single source of truth: any
-    // user-supplied rabitq_version is silently overridden here so that the
-    // hgraph default template (which always seeds rabitq_version=standard)
-    // does not collide with codes_type=rabitq_split.
     const bool is_split_codes =
         json.Contains(CODES_TYPE_KEY) && json[CODES_TYPE_KEY].GetString() == RABITQ_SPLIT_CODES;
-    auto quant_json = json[QUANTIZATION_PARAMS_KEY];
+    this->quantizer_parameter =
+        QuantizerParameter::GetQuantizerParameterByJson(json[QUANTIZATION_PARAMS_KEY]);
     if (is_split_codes) {
```
LHT129, line 80: [note] The `FromJson` method now validates `rabitq_version` at parse time via `CHECK_ARGUMENT` instead of silently overriding it with a warning. This is a behavioral change: previously, users could pass any `rabitq_version` value (e.g. `"standard"`) with `codes_type=rabitq_split` and it would be silently corrected; now it throws an error.

If this is intentional (enforcing strict configuration), it may be worth calling out in the PR description or a changelog, since existing configurations that relied on the silent override will now fail.

src/algorithm/hgraph/hgraph_param_mapping.cpp
LHT129: Fixed in 30357443 by removing the now-unused FP32 and RaBitQ quantizer includes. The full build succeeds and the focused HGraphParameter suite passes (19 test cases, 128 assertions).

src/datacell/flatten_datacell_parameter.cpp
LHT129: This behavior change is intentional. The refactor makes public index mapping the single owner of constructing the canonical split configuration and removes downstream mutation from FlattenDataCellParameter::FromJson. Accepting an inconsistent nested quantizer and silently rewriting it would preserve the mutation chain this PR is removing. I have documented the strict behavior and migration expectation in the PR description: codes_type=rabitq_split must be paired with rabitq_version=split.

LHT129 (commented):
[note] This is a well-executed refactoring PR that replaces template-based parameter mapping with structured builders across all index entry points. The diff is large (1110 additions, 1484 deletions) but the changes are mechanical and consistent.

**Summary of review:**

The refactoring is sound. Key improvements observed:
- JSON string templates replaced with `CreateDefault()` / `BuildDefault*()` factory functions that construct canonical inner configs programmatically.
- `ConstParamMap` / `format_map` / `mapping_external_param_to_inner` removed entirely, along with the `DEFAULT_MAP` in `inner_string_params.h`.
- `RaBitQSplitConfig` parsing separated from application (`ParseRaBitQSplitConfig` + `ApplyRaBitQSplitConfig`), making the split-config flow explicit and testable.
- SINDI and SINDIV2 now validate external parameters against a whitelist of supported flat keys, rejecting unknown fields early.
- `CompatibilityReport` / `CollectCompatibilityIssues` added as a structured alternative to `CheckCompatibility`, with test coverage.
- `FP32QuantizerParameter::ToJson()` now includes `hold_molds`, making the field round-trippable.
- Regression tests added for split config parsing/application and multi-difference compatibility reporting.

**Previously reported issues that have been addressed:**
- `STORE_RAW_VECTOR` now correctly writes `hold_molds` to both `base_codes.quantization_params` and `precise_codes.quantization_params` in BruteForce mapping.
- `size_t` replaced with `uint64_t` in `CollectCompatibilityIssues` array traversal.
- Unused `fp32_quantizer_parameter.h` and `rabitq_quantizer_parameter.h` includes removed from `hgraph_param_mapping.cpp`.

**One remaining item (previously flagged by Copilot, not yet addressed):**
- `FlattenDataCellParameter::CreateDefault` accepts `hold_molds` but only applies it when the quantizer is FP32. INT8 quantizers also parse `hold_molds`, so calling `CreateDefault("int8", ..., true)` would silently drop the flag. Current callers all use FP32, so this is not a live bug, but worth addressing for interface correctness.

**Behavior change note (intentional, documented in PR description):**
- `codes_type=rabitq_split` now requires the nested RaBitQ quantizer to already have `rabitq_version=split`. Inconsistent configurations are rejected with `CHECK_ARGUMENT` instead of being silently mutated. This is a deliberate cleanup that removes downstream mutation from `FlattenDataCellParameter::FromJson`.

No blocking issues found. The refactoring is clean and well-tested (713 unit tests passing, clang-format/clang-tidy clean).

LHT129 (commented):
经过全面审查（commit 30357443），这个 PR 质量很高。重构将 JSON 字符串模板替换为结构化 builder 函数，提升了可读性和类型安全。

**审查维度：**
- **正确性** ✅：参数映射逻辑与原有模板行为一致。`supported_keys` 白名单校验（SINDI/SINDI v2）增强了健壮性。`CollectCompatibilityIssues` 实现正确，JSON 递归比较逻辑完整。
- **风格** ✅：`CreateDefault` 工厂方法模式统一了各参数类型的构造方式。显式 if-else 链比 `ConstParamMap` 更直观。
- **测试** ✅：已有测试覆盖参数映射路径。
- **性能** ✅：`build_default_*` 函数在初始化时调用一次，非热路径。`CollectCompatibilityIssues` 中的 `std::function` 递归 lambda 开销可接受（非热路径）。
- **安全性** ✅：`RaBitQSplitConfig` 的 `ParseRaBitQSplitConfig` 正确验证了 bit 范围 [1,8] 和 total <= 8。`ValidateMRLEDim` 验证了 mrle_dim 在 [0, dim] 范围内。
- **可维护性** ✅：移除 `ConstParamMap` 和 `DEFAULT_MAP`（116 行）消除了隐式依赖。结构化 builder 使得添加新参数类型更清晰。

之前 Copilot 发现的 `STORE_RAW_VECTOR` 映射错误、`size_t` 使用和未使用的 includes 都已在最新 commit 中修复。未发现新的阻塞性问题。

wxyucs (approved):
lgtm

src/quantization/transform_quantization/transform_quantizer_parameter.cpp
```diff
@@ -20,6 +20,23 @@
 
 namespace vsag {
 
+TransformQuantizerParamPtr
+TransformQuantizerParameter::CreateDefault(const std::string& chain) {
+    auto chain_items = SplitString(chain);
+    CHECK_ARGUMENT(chain_items.size() > 1,
+                   "tq_chain must contain at least one transformer and one quantizer");
+    const auto bottom_type = chain_items.back();
+    auto json = QuantizerParameter::CreateDefault(bottom_type)->ToJson();
```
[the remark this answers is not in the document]
LHT129: Verified that TransformQuantizerParameter::SplitString already trims leading/trailing whitespace. Added direct CreateDefault regression coverage showing spaced and compact chains normalize to the same JSON in a16269dc.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,660 +14,281 @@
...
+        } else if (key == HGRAPH_BUILD_BY_BASE_QUANTIZATION) {
+            inner_json[HGRAPH_BUILD_BY_BASE_QUANTIZATION_KEY].SetJson(value);
+        } else if (key == USE_ATTRIBUTE_FILTER) {
+            inner_json[USE_ATTRIBUTE_FILTER_KEY].SetJson(value);
+        } else if (key == HGRAPH_BASE_QUANTIZATION_TYPE) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY][TYPE_KEY].SetJson(value);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(value);
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(value);
+            inner_json[STORE_RAW_VECTOR_KEY].SetJson(value);
```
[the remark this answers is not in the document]
LHT129: Fixed in a16269dc. hold_molds is now applied through a shared capability-aware helper only for FP32, INT8, or a TQ whose bottom quantizer is FP32/INT8; unsupported quantizers no longer receive the field.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1223,151 +1193,74 @@ BruteForce::CheckAndMappingExternalParam(const JsonType& external_param,
...
+        } else if (key == BRUTE_FORCE_PRECISE_IO_TYPE) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][TYPE_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_PRECISE_FILE_PATH) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][IO_FILE_PATH_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_THREAD_COUNT) {
+            inner_json[BUILD_THREAD_COUNT_KEY].SetJson(field);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(field);
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].SetJson(field);
+        } else if (key == USE_ATTRIBUTE_FILTER) {
```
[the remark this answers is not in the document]
LHT129: Fixed in a16269dc using the same capability-aware hold_molds helper as HGraph, with focused mapping and helper tests.

src/parameter.cpp
```diff
@@ -0,0 +1,80 @@
...
+                    }
+                }
+                return;
+            }
+            if (lhs.is_array()) {
+                if (lhs.size() != rhs.size()) {
+                    report.issues.push_back({path, "array sizes differ"});
+                }
+                const auto common_size = std::min(lhs.size(), rhs.size());
+                for (uint64_t i = 0; i < common_size; ++i) {
```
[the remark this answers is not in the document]
LHT129: Keeping uint64_t intentionally to follow the repository-wide indexing convention and the earlier macOS portability review on this PR. The codebase consistently uses uint64_t against container sizes, and the full Linux/macOS builds do not report a warning here.

src/algorithm/inner_index_parameter.cpp
```diff
@@ -104,22 +105,23 @@ MapRaBitQSplitParam(const JsonType& external_json, JsonType& inner_json) {
                        "rabitq split transform quantizer requires tq_chain=\"mrle, rabitq\"");
```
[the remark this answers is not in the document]
LHT129: Both spellings are already accepted because SplitString trims every token. Added direct whitespace-normalization coverage in a16269dc; the message remains a valid example rather than a required literal spelling.

src/quantization/transform_quantization/transform_quantizer_parameter.cpp
```diff
@@ -20,6 +20,23 @@
...
+    const auto transformer_json = VectorTransformerParameter().ToJson();
+    json[INPUT_DIM_KEY].SetJson(transformer_json[INPUT_DIM_KEY]);
+    json[PCA_DIM_KEY].SetJson(transformer_json[PCA_DIM_KEY]);
+    json[MRLE_DIM_KEY].SetJson(transformer_json[MRLE_DIM_KEY]);
+    json[TQ_CHAIN_KEY].SetString(chain);
+    auto parameter = std::make_shared<TransformQuantizerParameter>();
+    parameter->FromJson(json);
+    return parameter;
+}
```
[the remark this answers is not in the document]
LHT129: Whitespace handling is already implemented in SplitString and now has direct CreateDefault coverage. The temporary bottom quantizer type is intentional: FromJson stores canonical bottom-quantizer defaults in base_quantizer_json_, while ToJson exposes the outer type as tq. Setting type=tq before FromJson is unnecessary and would obscure that two-layer representation.

src/quantization/fp32_quantizer_parameter_test.cpp
```diff
@@ -27,3 +27,39 @@ TEST_CASE("FP32 Quantizer Parameter ToJson Test", "[ut][FP32QuantizerParameter]"
...
+        JsonType::Parse(R"({"same":1,"first":4,"nested":{"second":5},"extra":6})"));
+
+    auto report = left->CollectCompatibilityIssues(right);
+    REQUIRE_FALSE(report.IsCompatible());
+    REQUIRE(report.issues.size() == 3);
+    REQUIRE(report.issues[0].path == "$.first");
+    REQUIRE(report.issues[1].path == "$.nested.second");
+    REQUIRE(report.issues[2].path == "$.extra");
+    REQUIRE_FALSE(left->CheckCompatibility(right));
+}
```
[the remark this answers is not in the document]
LHT129: Fixed in a16269dc. The test now builds a path-to-message map and asserts issue content independently of traversal order.

src/parameter.cpp
```diff
@@ -0,0 +1,80 @@
...
+        report.issues.push_back({"$", "other parameter is null"});
+        return report;
+    }
+
+    const auto left = this->ToJson();
+    const auto right = other->ToJson();
+    const auto* left_json = left.GetInnerJson();
+    const auto* right_json = right.GetInnerJson();
+    std::function<void(const nlohmann::json&, const nlohmann::json&, const std::string&)> collect =
+        [&](const nlohmann::json& lhs, const nlohmann::json& rhs, const std::string& path) {
```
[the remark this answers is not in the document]
LHT129: Fixed in a16269dc. CompatibilityReport now emits RFC 6901 JSON Pointer paths and escapes object tokens, with regression coverage for a key containing slash.

src/parameter.cpp
```diff
@@ -0,0 +1,80 @@
...
+                for (const auto& [key, value] : rhs.items()) {
+                    (void)value;
+                    if (not lhs.contains(key)) {
+                        auto child_path = path;
+                        child_path.append(".").append(key);
+                        report.issues.push_back({std::move(child_path), "unexpected field"});
+                    }
+                }
+                return;
+            }
```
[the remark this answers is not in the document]
LHT129: Fixed in a16269dc together with the object-path case: array indices and object keys now share unambiguous JSON Pointer path construction.

src/datacell/flatten_datacell_parameter.cpp
```diff
@@ -17,12 +17,27 @@
...
+                                        const std::string& io_type,
+                                        bool hold_molds) {
+    auto parameter = std::make_shared<FlattenDataCellParameter>();
+    parameter->io_parameter = IOParameter::CreateDefault(io_type);
+    parameter->quantizer_parameter = QuantizerParameter::CreateDefault(quantization_type);
+    if (auto fp32 =
+            std::dynamic_pointer_cast<FP32QuantizerParameter>(parameter->quantizer_parameter);
+        fp32 != nullptr) {
+        fp32->hold_molds = hold_molds;
+    }
```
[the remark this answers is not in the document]
LHT129: Addressed in a16269dc. FlattenDataCellParameter::CreateDefault now propagates hold_molds to INT8 as well as FP32, and INT8QuantizerParameter::ToJson now preserves the field for round trips. Added factory and INT8 parameter coverage.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,660 +14,284 @@
...
+build_default_flatten_param(const std::string& quantization_type,
+                            const std::string& io_type,
+                            const std::string& tq_chain = "",
+                            bool hold_molds = false) {
+    auto parameter = FlattenDataCellParameter::CreateDefault(
+        quantization_type == QUANTIZATION_TYPE_VALUE_TQ ? QUANTIZATION_TYPE_VALUE_FP32
+                                                        : quantization_type,
+        io_type,
+        hold_molds);
+    if (quantization_type == QUANTIZATION_TYPE_VALUE_TQ) {
```
LHT129, line 44: [suggestion] When `quantization_type == QUANTIZATION_TYPE_VALUE_TQ` and `hold_molds == true`, the `hold_molds` flag is silently lost: `FlattenDataCellParameter::CreateDefault` sets it on a temporary FP32 quantizer that is immediately replaced by `TransformQuantizerParameter::CreateDefault(tq_chain)` on the next line.

Currently all callers pass `hold_molds=false` (or `true` only with `QUANTIZATION_TYPE_VALUE_FP32`), so this is not a live bug. However, the function signature invites misuse. Consider either:
- Asserting that `hold_molds` is false when `quantization_type` is TQ, or
- Applying `hold_molds` to the TQ quantizer after creation (e.g., via `ApplyHoldMoldsToQuantizer`).

This applies to both `build_default_flatten_param` in `hgraph_param_mapping.cpp` and the `build_flatten` lambda in `pyramid.cpp`.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1209,149 +1211,175 @@ Pyramid::InitFeatures() {
...
-        },
-        "{BUILD_THREAD_COUNT_KEY}": 1,
-        "{EF_CONSTRUCTION_KEY}": 400,
-        "{NO_BUILD_LEVELS}":[],
-        "{INDEX_MIN_SIZE}": 0,
-        "{SUPPORT_DUPLICATE}": false
-    })";
+JsonType
+build_default_pyramid_param(const JsonType& external_param) {
+    const auto base_quantization_type =
```
LHT129, line 1293: [suggestion] Same latent issue as `build_default_flatten_param` in `hgraph_param_mapping.cpp`: when `quantization_type == QUANTIZATION_TYPE_VALUE_TQ` and `hold_molds == true`, the flag is silently lost because the FP32 quantizer (with hold_molds set) is immediately replaced by `TransformQuantizerParameter::CreateDefault(chain)`.

Currently all callers pass `hold_molds=false` or `true` only with `QUANTIZATION_TYPE_VALUE_FP32`, so this is safe in practice. Consider adding a guard (assert or early return) to prevent future misuse.

src/quantization/transform_quantization/transform_quantizer_parameter.cpp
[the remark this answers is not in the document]
LHT129: Thanks for checking this. The intermediate JSON intentionally keeps the bottom quantizer type: FromJson stores it as base_quantizer_json_ and normalizes TYPE_KEY to the bottom type, while ToJson reconstructs the outer TQ representation with type=tq. Setting type=tq before FromJson would lose the bottom quantizer identity. The chain is normalized by SplitString and MergeStrings on serialization, with whitespace regression coverage, so no code change is needed here.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1209,149 +1211,175 @@ Pyramid::InitFeatures() {
...
+        } else if (key == PYRAMID_GRAPH_TYPE) {
+            inner_json[GRAPH_KEY][GRAPH_TYPE_KEY].SetJson(value);
+        } else if (key == PYRAMID_GRAPH_STORAGE_TYPE) {
+            inner_json[GRAPH_KEY][GRAPH_STORAGE_TYPE_KEY].SetJson(value);
+        } else if (key == PYRAMID_PRECISE_IO_TYPE) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][TYPE_KEY].SetJson(value);
+        } else if (key == PYRAMID_BUILD_THREAD_COUNT) {
+            inner_json[BUILD_THREAD_COUNT_KEY].SetJson(value);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[STORE_RAW_VECTOR_KEY].SetJson(value);
```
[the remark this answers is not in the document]
LHT129: Good catch. Updated Pyramid STORE_RAW_VECTOR handling to apply ApplyHoldMoldsToQuantizer to both base and precise quantizer JSON before setting store_raw_vector, matching HGraph and BruteForce while retaining capability-based behavior.

src/parameter.h
```diff
@@ -57,6 +74,9 @@ class Parameter {
     CheckCompatibility(const ParamPtr& other) const {
         return this->ToString() == other->ToString();
     }
+
+    virtual CompatibilityReport
+    CollectCompatibilityIssues(const ParamPtr& other) const;
```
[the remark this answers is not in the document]
LHT129: Confirmed that src/CMakeLists.txt collects root sources with file(GLOB CPP_SRCS "*.cpp"), so parameter.cpp is included automatically. The Release shared and static builds completed successfully and explicitly compiled parameter.cpp.o; no CMake change is required.

src/algorithm/hgraph/hgraph_param_mapping.cpp
LHT129: Agreed. Updated the builder to apply hold_molds to the final quantizer JSON after the optional TQ replacement, using ApplyHoldMoldsToQuantizer. This prevents the latent TQ loss and retains capability-based handling for the bottom quantizer.

src/algorithm/pyramid/pyramid.cpp
LHT129: Agreed and fixed in the Pyramid builder as well: hold_molds is now applied to the final quantizer JSON after creating the optional TQ quantizer, so it is no longer silently discarded.

LHT129 (commented):
I have reviewed the latest commit (74e6de25) and all previously identified issues from Copilot have been addressed:

1. **STORE_RAW_VECTOR placement** — now correctly writes `hold_molds` to both `base_codes.quantization_params` and `precise_codes.quantization_params` in bruteforce, hgraph, and pyramid.
2. **`size_t` usage** — replaced with `uint64_t` in `CollectCompatibilityIssues`.
3. **`tq_chain` whitespace handling** — `SplitString` now trims whitespace around tokens; `CreateDefault` normalizes the chain; a dedicated test verifies this.
4. **Test ordering dependency** — `CompatibilityReport` test now uses `std::map` for order-independent assertions and also covers JSON pointer escaping (`/` → `~1`).

The refactoring is consistent across all index entry points (bruteforce, hgraph, ivf, pyramid, simq, sindi, sindi_v2), the `DEFAULT_MAP` removal is safe (all consumers updated), and test coverage is solid. No new issues found.

src/quantization/quantizer_parameter.cpp
```diff
@@ -30,6 +30,13 @@
 #include "transform_quantization/transform_quantizer_parameter.h"
 
 namespace vsag {
+QuantizerParamPtr
+QuantizerParameter::CreateDefault(const std::string& type_name) {
+    JsonType json;
+    json[TYPE_KEY].SetString(type_name);
+    return GetQuantizerParameterByJson(json);
+}
```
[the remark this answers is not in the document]
LHT129: Agreed. QuantizerParameter::CreateDefault now rejects type=tq directly with a clear message directing callers to TransformQuantizerParameter::CreateDefault(chain), and a regression test covers the generic factory path. The focused TransformQuantizerParameter suite passes with 20 assertions in 6 cases.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,666 +14,289 @@
 
 #include <fmt/format.h>
 
+#include <nlohmann/json.hpp>
+
 #include "common.h"
 #include "hgraph.h"  // IWYU pragma: keep
 #include "hgraph_parameter.h"
+#include "quantization/transform_quantization/transform_quantizer_parameter.h"
```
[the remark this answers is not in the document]
LHT129: Agreed. Added the direct io/common/io_parameter.h include so build_default_io_param no longer relies on transitive includes. The HGraph target and unit test binary rebuild successfully.

src/parameter.cpp
```diff
@@ -0,0 +1,103 @@
...
+        report.issues.push_back({"", "other parameter is null"});
+        return report;
+    }
+
+    const auto left = this->ToJson();
+    const auto right = other->ToJson();
+    const auto* left_json = left.GetInnerJson();
+    const auto* right_json = right.GetInnerJson();
+    std::function<void(const nlohmann::json&, const nlohmann::json&, const std::string&)> collect =
+        [&](const nlohmann::json& lhs, const nlohmann::json& rhs, const std::string& path) {
```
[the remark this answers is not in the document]
LHT129: Agreed. Introduced a CompatibilityCollector alias and wrapped the std::function signature; clang-format and the unit test build pass.

LHT129 (commented):
Thoroughly reviewed the full diff (32 files, ~1200 additions, ~1500 deletions). The refactoring is well-executed and addresses all major concerns from earlier review rounds:

**Verified fixes from previous Copilot comments:**
- STORE_RAW_VECTOR: Both base and precise codes now correctly use `ApplyHoldMoldsToQuantizer()` for hold_molds propagation
- `size_t` → `uint64_t`: All new code uses `uint64_t` consistently
- `SplitString` trimming: Confirmed working, normalizes TQ chain whitespace
- Test ordering: Fixed with `std::map`

**Architecture improvements:**
- Template-based mapping replaced with structural `CreateDefault()` factories → no more string template mutation chains
- `RaBitQSplitConfig` struct + `ParseRaBitQSplitConfig`/`ApplyRaBitQSplitConfig` centralizes split configuration
- `CompatibilityReport`/`CollectCompatibilityIssues()` provides structured diff reporting alongside existing `CheckCompatibility`
- All mapping functions reject unknown flat keys with "invalid config param" errors
- SINDI/SINDIV2 now validate unknown flat keys at boundary

**One remaining minor issue (already flagged by Copilot, not re-flagging):**
- `inner_index_parameter.cpp:105`: error message says `"mrle, rabitq"` but `SplitString` normalizes whitespace so the comparison is against `"mrle,rabitq"`. The error message could mislead users.

No new substantive issues found. The PR is in good shape.

LHT129 (commented):
## Code Review Summary

**Commit reviewed:** a11cf1cd77fe4e01b2e9376e4039e6bdb5adf8e8

All issues flagged in the previous Copilot review have been addressed:

1. **STORE_RAW_VECTOR** — `ApplyHoldMoldsToQuantizer` is now correctly applied to both `base_codes` and `precise_codes` quantizers in `bruteforce.cpp` and `hgraph_param_mapping.cpp`.

2. **size_t → uint64_t** — Changed in `parameter.h` to match VSAG coding standards.

3. **Test ordering dependency** — Fixed by using `std::map` for order-independent assertions in `fp32_quantizer_parameter_test.cpp`.

4. **TQ TYPE_KEY** — Now properly set in `TransformQuantizerParameter::ToJson()` at `transform_quantizer_parameter.cpp:112`.

5. **TQ whitespace** — `SplitString` trims whitespace; verified by test.

### Additional observations (no action needed)

- **JSON pointer escaping** (`escape_json_pointer_token` / `append_json_pointer_token`) is RFC 6901 compliant. The test correctly verifies `"/a~1b"` for key `"a/b"`.
- **`RaBitQSplitConfig`** struct with `ParseRaBitQSplitConfig` (validates external params) + `ApplyRaBitQSplitConfig` (mutates inner JSON) is a clean separation of concerns. The shared key `"base_quantization_type"` works correctly for both HGraph and Pyramid since `HGRAPH_BASE_QUANTIZATION_TYPE` and `PYRAMID_BASE_QUANTIZATION_TYPE` have the same string value.
- **`build_default_flatten_param`** double-application of `hold_molds` (via `CreateDefault` + `ApplyHoldMoldsToQuantizer`) is intentional: `CreateDefault` handles FP32/INT8 directly, while `ApplyHoldMoldsToQuantizer` unwraps TQ chains to reach the bottom quantizer.
- **SINDI/SINDIV2** now validate external params against a `supported_keys` set, rejecting unknown keys with a clear error message.
- **`FlattenDataCellParameter::CreateDefault`** correctly sets `hold_molds` on FP32/INT8 quantizer parameter objects, and `ToJson` serializes it.
- **Naming convention** — new helper functions use `snake_case` (`build_default_*`), consistent with the existing codebase style.
- **Test coverage** — regression tests added for split config, compatibility reporting, and hold_molds behavior.

No new substantive issues found. The refactoring is clean and well-tested.

LHT129 (commented):
This push addresses all previously raised review comments:

1. **STORE_RAW_VECTOR path** — Now correctly writes to `base_codes.quantization_params` and `precise_codes.quantization_params` via `ApplyHoldMoldsToQuantizer`, with a regression test added in `bruteforce_parameter_test.cpp`.

2. **size_t → uint64_t** — The `CollectCompatibilityIssues` loop in `parameter.cpp` now uses `uint64_t` for array traversal, consistent with project conventions.

3. **Test ordering dependency** — The `CompatibilityReport` test in `fp32_quantizer_parameter_test.cpp` now uses `std::map` for order-independent assertions, and also covers JSON pointer escaping (`/a~1b`).

4. **Chain whitespace** — `SplitString` already trims whitespace around tokens, so `"mrle, rabitq"` and `"mrle,rabitq"` are both handled correctly. A dedicated test in `transform_quantizer_parameter_test.cpp` verifies this.

No new issues found. The refactoring is clean, well-tested, and all prior concerns are resolved.

src/algorithm/inner_index_parameter.cpp
LHT129: [suggestion] The error message says tq_chain="mrle, rabitq" (with a space after the comma), but SplitString trims whitespace from each token, so tq_chain="mrle,rabitq" (no space) is also accepted. Consider using the canonical form "mrle,rabitq" in the error message to avoid user confusion when they pass the no-space variant and get an error message showing a different format than what they provided.

src/parameter.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <utility>
+
+namespace vsag {
+
+namespace {
+
+std::string
+escape_json_pointer_token(const std::string& token) {
+    std::string escaped;
+    escaped.reserve(token.size());
```
LHT129, line 29: [note] `CollectCompatibilityIssues` calls `this->ToJson()` and `other->ToJson()` to create full JSON copies for recursive comparison. This is fine for a compatibility check (not a hot path), but worth noting that for deeply nested parameter trees with large arrays, the double serialization + recursive walk could be noticeable. If this ever becomes a concern, a visitor-based approach that walks the existing `Parameter` tree directly (without JSON round-trip) would avoid the serialization overhead.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,666 +14,290 @@
...
+        } else if (key == HGRAPH_PRECISE_ENABLE_READ_CACHE) {
+            inner_json[PRECISE_CODES_KEY][IO_PARAMS_KEY][READ_CACHE_ENABLED_KEY].SetJson(value);
+        } else if (key == HGRAPH_GRAPH_ENABLE_READ_CACHE) {
+            inner_json[GRAPH_KEY][IO_PARAMS_KEY][READ_CACHE_ENABLED_KEY].SetJson(value);
+        } else if (key == HGRAPH_RAW_VECTOR_ENABLE_READ_CACHE) {
+            inner_json[RAW_VECTOR_KEY][IO_PARAMS_KEY][READ_CACHE_ENABLED_KEY].SetJson(value);
+        } else if (key == HGRAPH_RAW_VECTOR_CACHE_TOTAL_SIZE) {
+            inner_json[RAW_VECTOR_KEY][IO_PARAMS_KEY][READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetJson(
+                value);
+        } else {
```
LHT129, line 310: [note] The `else` branch throws `VsagException` for unknown keys, which is the correct strict-validation behavior. However, the SINDI and SINDI v2 implementations use a `static const std::unordered_set<std::string> supported_keys` whitelist approach instead. Consider using the same whitelist pattern across all index types (HGraph, IVF, BruteForce, Pyramid) for consistency — it would make the supported parameter set explicit and self-documenting, and would catch typos before the if-else chain. The current if-else approach is functionally equivalent but requires readers to scan the entire chain to understand what keys are supported.

LHT129 (commented):
## Review Summary

This is a large, well-executed refactoring that replaces template-based JSON parameter mapping (`ConstParamMap` + string templates) with explicit `build_default_*` functions and if-else chains across all index types. The PR also introduces `CompatibilityReport`/`CompatibilityIssue` for structured compatibility checking and `RaBitQSplitConfig` parsing/application.

### What's been addressed from prior reviews
- STORE_RAW_VECTOR now correctly maps to `base_codes.quantization_params.hold_molds` and `precise_codes.quantization_params.hold_molds` via `ApplyHoldMoldsToQuantizer`
- `parameter.h` includes `<string>` for `CompatibilityIssue`/`CompatibilityReport`
- `sindi.cpp` includes `<unordered_set>`
- Test ordering issue fixed by using `std::map` for order-independent assertions
- `FlattenDataCellParameter::CreateDefault` now handles `hold_molds` for both FP32 and INT8
- `SplitString` trims whitespace, so chain parsing handles spaces correctly
- Unused `build_default_quantization_param` and stale includes removed

### Items flagged in this review
1. **Error message format** (`inner_index_parameter.cpp:105`): The error message shows `tq_chain="mrle, rabitq"` (with space), but `SplitString` normalizes whitespace, so the no-space variant is also valid. Consider using the canonical normalized form in the error message.
2. **Performance note** (`parameter.cpp:29`): `CollectCompatibilityIssues` does a full JSON round-trip for comparison. Fine for a non-hot-path compatibility check, but worth noting.
3. **Consistency note** (`hgraph_param_mapping.cpp:295`): SINDI/SINDI v2 use a `supported_keys` whitelist for validation; the other index types use if-else chains with a catch-all throw. Both approaches are functionally correct, but consistency would improve maintainability.

### Overall assessment
The refactoring is sound. The explicit mapping approach is more readable and maintainable than the template-based system it replaces. All previously flagged issues have been addressed. No correctness or security issues found.

src/datacell/flatten_datacell_parameter.cpp
```diff
@@ -17,12 +17,32 @@
...
+FlattenDataCellParameter::CreateDefault(const std::string& quantization_type,
+                                        const std::string& io_type,
+                                        bool hold_molds) {
+    auto parameter = std::make_shared<FlattenDataCellParameter>();
+    parameter->io_parameter = IOParameter::CreateDefault(io_type);
+    parameter->quantizer_parameter = QuantizerParameter::CreateDefault(quantization_type);
+    if (auto fp32 =
+            std::dynamic_pointer_cast<FP32QuantizerParameter>(parameter->quantizer_parameter);
+        fp32 != nullptr) {
+        fp32->hold_molds = hold_molds;
```
LHT129, line 37: [suggestion] `CreateDefault` only sets `hold_molds` on FP32 and INT8 quantizer parameters via dynamic_cast, but `hold_molds` is also a valid field on other quantizer types (e.g. RaBitQ, SQ8, etc.). When a caller passes `hold_molds=true` with a non-FP32/non-INT8 `quantization_type`, the flag is silently dropped.

Consider using `ApplyHoldMoldsToQuantizer` (from `inner_index_parameter.h`) on the JSON before calling `FromJson`, since it already handles TQ chain unwrapping and correctly identifies which quantizer types support `hold_molds`.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,666 +14,290 @@
...
+build_default_io_param(const std::string& io_type = IO_TYPE_VALUE_BLOCK_MEMORY_IO) {
+    return IOParameter::CreateDefault(io_type)->ToJson();
+}
+
+JsonType
+build_default_flatten_param(const std::string& quantization_type,
+                            const std::string& io_type,
+                            const std::string& tq_chain = "",
+                            bool hold_molds = false) {
+    auto parameter = FlattenDataCellParameter::CreateDefault(
```
LHT129, line 39: [note] When `quantization_type == QUANTIZATION_TYPE_VALUE_TQ`, `FlattenDataCellParameter::CreateDefault` is called with `QUANTIZATION_TYPE_VALUE_FP32` as a placeholder, and then `parameter->quantizer_parameter` is immediately overwritten with `TransformQuantizerParameter::CreateDefault(tq_chain)`. This means the FP32 quantizer created inside `CreateDefault` is discarded.

This pattern (create-then-replace) appears in both `build_default_flatten_param` here and the `build_flatten` lambda in `pyramid.cpp`. While `ApplyHoldMoldsToQuantizer` is called afterwards to fix up the JSON, it would be cleaner to either:
1. Have a separate code path for TQ that constructs the `FlattenDataCellParameter` directly with a `TransformQuantizerParameter`, or
2. Add a `CreateDefault` overload that accepts a pre-built quantizer parameter.

This is not blocking — the current code is functionally correct — but the double-construction is a minor maintainability concern.

src/datacell/bucket_datacell_parameter_test.cpp
```diff
@@ -15,11 +15,59 @@
...
+    }
+
+    SECTION("flatten owns hold molds") {
+        auto fp32_parameter = FlattenDataCellParameter::CreateDefault(
+            QUANTIZATION_TYPE_VALUE_FP32, IO_TYPE_VALUE_BLOCK_MEMORY_IO, true);
+        REQUIRE(fp32_parameter->ToJson()[QUANTIZATION_PARAMS_KEY][HOLD_MOLDS].GetBool());
+
+        auto int8_parameter = FlattenDataCellParameter::CreateDefault(
+            QUANTIZATION_TYPE_VALUE_INT8, IO_TYPE_VALUE_BLOCK_MEMORY_IO, true);
+        REQUIRE(int8_parameter->quantizer_parameter->GetTypeName() == QUANTIZATION_TYPE_VALUE_INT8);
```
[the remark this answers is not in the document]
LHT129: Fixed by clang-format; the reported long expressions are now wrapped within the project line-length constraint.

src/quantization/transform_quantization/transform_quantizer_parameter_test.cpp
```diff
@@ -95,6 +95,19 @@ TEST_CASE("TQ parameter Split Merge String Test", "[ut][TransformQuantizerParame
     REQUIRE(str == recover_str);
 }
 
+TEST_CASE("Generic quantizer CreateDefault rejects TQ without a chain",
+          "[ut][TransformQuantizerParameter]") {
+    REQUIRE_THROWS(QuantizerParameter::CreateDefault(QUANTIZATION_TYPE_VALUE_TQ));
+}
+
+TEST_CASE("TQ CreateDefault normalizes chain whitespace", "[ut][TransformQuantizerParameter]") {
```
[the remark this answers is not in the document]
LHT129: Fixed by wrapping the TEST_CASE invocation with clang-format.

src/algorithm/inner_index_parameter.cpp
LHT129: Agreed. Updated the example to the canonical normalized form tq_chain="mrle,rabitq", matching SplitString/MergeStrings serialization.

src/parameter.cpp
LHT129: Acknowledged. Compatibility checks are configuration-time operations and not a hot path, so the JSON round-trip keeps the implementation generic and complete. A direct parameter-tree visitor can be considered separately if profiling shows this matters.

src/algorithm/hgraph/hgraph_param_mapping.cpp
LHT129: Acknowledged. The current if/else chains provide the same strict unknown-key rejection while also applying key-specific structured mutations. Converting every mapping to a separate whitelist would duplicate the supported-key list, so I am keeping this functionally equivalent approach in this PR.

src/datacell/flatten_datacell_parameter.cpp
LHT129: The premise here does not match the parameter model: hold_molds is serialized by FP32 and INT8 only. ApplyHoldMoldsToQuantizer intentionally handles those types and TQ chains whose bottom quantizer is FP32/INT8; injecting it into RaBitQ/SQ8 would create unsupported fields. The existing capability-based behavior is retained.

src/algorithm/hgraph/hgraph_param_mapping.cpp
LHT129: Acknowledged. The temporary FP32 construction is functionally harmless and the final JSON is corrected through the capability helper. Adding a pre-built-quantizer overload would expand the public construction API, so I am leaving that cleanup for a separate change.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,666 +14,290 @@
...
+        } else if (key == HGRAPH_RAW_VECTOR_CACHE_TOTAL_SIZE) {
+            inner_json[RAW_VECTOR_KEY][IO_PARAMS_KEY][READ_CACHE_TOTAL_CACHE_SIZE_KEY].SetJson(
+                value);
+        } else {
+            throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                fmt::format("invalid config param: {}", key));
+        }
+    }
+    ApplyRaBitQSplitConfig(ParseRaBitQSplitConfig(hgraph_json), inner_json);
     return inner_json;
```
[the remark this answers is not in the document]
LHT129: Agreed. HGraph now rejects tq_chain and mrle_dim at the mapping boundary unless base_quantization_type is tq. Regression tests cover both fields with FP32/INT8 bases.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1209,149 +1211,183 @@ Pyramid::InitFeatures() {
...
+            inner_json[SUPPORT_DUPLICATE].SetJson(value);
+            inner_json[GRAPH_KEY][SUPPORT_DUPLICATE].SetJson(value);
+        } else {
+            throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                fmt::format("invalid config param: {}", key));
+        }
+    }
+    ApplyRaBitQSplitConfig(ParseRaBitQSplitConfig(external_param), inner_json);
     ValidateMRLEDim(external_param, common_param.dim_);
     if (RequiresRawVectorForTransformQuantizer(inner_json) and
```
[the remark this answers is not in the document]
LHT129: Agreed. Pyramid now applies the same mapping-boundary validation and regression coverage for tq_chain/mrle_dim with non-TQ base quantizers.

LHT129 (commented):
[note] In `src/parameter.cpp:92`, `common_size` is derived from `nlohmann::json::size()` which returns `size_t`, but the loop iterates with `uint64_t i`. This can produce a signed/unsigned mismatch warning on platforms where `size_t` is narrower than `uint64_t` (e.g., 32-bit) or trigger `-Wsign-compare` warnings.

Consider using `size_t` for the loop index to match the container size type, or casting `common_size` explicitly to `uint64_t` if the project convention requires `uint64_t` everywhere:
```cpp
for (size_t i = 0; i < common_size; ++i) {
```

src/algorithm/simq/simq.cpp
```diff
@@ -1752,47 +1754,39 @@ SIMQ::InitFeatures() {
...
-        {BUILD_THREAD_COUNT_KEY, {BUILD_THREAD_COUNT_KEY}},
-        {"split_delay_seconds", {"split_delay_seconds"}},
-    };
-
-    if (common_param.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
-        throw VsagException(ErrorType::INVALID_ARGUMENT, "simq only supports float32 datatype");
-    }
-    if (common_param.metric_ != MetricType::METRIC_TYPE_IP) {
-        throw VsagException(ErrorType::INVALID_ARGUMENT, "simq only supports ip metric type");
+    auto inner_json = build_default_simq_param(external_param);
```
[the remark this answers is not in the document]
LHT129: Agreed and fixed in 73211ab5. Restored the explicit SIMQ common-parameter validation for float32 datatype and IP metric before structured parameter construction. Added a focused regression test covering both unsupported datatype and metric cases; 2 assertions pass.

src/algorithm/simq/simq.cpp
```diff
@@ -1752,46 +1754,45 @@ SIMQ::InitFeatures() {
...
-        "{BASE_CODES_KEY}": {
-            "{IO_PARAMS_KEY}": {
-                "{TYPE_KEY}": "{IO_TYPE_VALUE_ASYNC_IO}",
-                "{IO_FILE_PATH_KEY}": "{DEFAULT_FILE_PATH_VALUE}"
-            },
-            "{CODES_TYPE_KEY}": "multi_vector"
-        }
-    })";
+JsonType
+build_default_simq_param(const JsonType& external_param) {
```
LHT129, line 1758: [note] `build_default_simq_param` defaults to `IO_TYPE_VALUE_ASYNC_IO` while all other builders (HGraph, BruteForce, Pyramid, IVF) default to `IO_TYPE_VALUE_BLOCK_MEMORY_IO`. This is consistent with the old `SIMQ_PARAMS_TEMPLATE` which also used `IO_TYPE_VALUE_ASYNC_IO`, so it is not a regression. Worth confirming whether this asymmetry is intentional (SIMQ is streaming-oriented) or should be unified for consistency.

LHT129 (commented):
## Review Summary

This PR replaces template-based JSON parameter mapping with structural builders and explicit flat-field translation across all index types. The refactoring is thorough and well-structured, with several improvements over the previous approach.

### Issues Fixed from Previous Reviews
- `STORE_RAW_VECTOR` now correctly applies `ApplyHoldMoldsToQuantizer` to both base and precise codes in BruteForce, HGraph, and Pyramid
- `size_t` replaced with `uint64_t` in `CollectCompatibilityIssues`
- Test ordering dependency fixed with `std::map`
- `SplitString` whitespace trimming already handled

### New Additions
- `CompatibilityReport` / `CompatibilityIssue` with JSON pointer path escaping (`~0` for `~`, `~1` for `/`) provides structured compatibility checking
- `RaBitQSplitConfig` parsing and `ApplyRaBitQSplitConfig` centralize split quantizer configuration
- `ApplyHoldMoldsToQuantizer` propagates hold_molds flag through TQ chains to supported quantizers (fp32, int8)
- `ValidateMRLEDim` and `RequiresRawVectorFor*` helpers consolidate validation logic
- `CreateDefault` factory methods on parameter classes improve testability and code reuse
- Whitelist-based validation for SINDI/SINDIV2 parameters provides clear error messages

### One Minor Note
- `build_default_simq_param` defaults to `IO_TYPE_VALUE_ASYNC_IO` while all other builders default to `IO_TYPE_VALUE_BLOCK_MEMORY_IO`. This is consistent with the old template, but worth confirming the asymmetry is intentional.

### Overall
The refactoring eliminates the fragile string-template mutation chain and replaces it with structured, type-safe builders. The explicit if-else chains are verbose but clear and maintainable. No blocking issues found.

src/io/common/io_parameter.cpp
```diff
@@ -34,6 +34,14 @@ std::once_flag async_io_fallback_warn_once;
 std::once_flag uring_io_fallback_warn_once;
 }  // namespace
 
+IOParamPtr
+IOParameter::CreateDefault(const std::string& type_name) {
+    JsonType json;
+    json[TYPE_KEY].SetString(type_name);
+    json[IO_FILE_PATH_KEY].SetString(DEFAULT_FILE_PATH_VALUE);
+    return GetIOParameterByJson(json);
+}
```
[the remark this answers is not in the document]
LHT129: Agreed and fixed in 122edefc. IOParameter::CreateDefault now checks the factory result and throws INVALID_ARGUMENT with the unsupported IO type instead of returning nullptr. Added a focused regression test; the unittests target builds successfully and the test passes.

src/algorithm/simq/simq.cpp
LHT129: Confirmed this asymmetry is intentional and preserved from the previous SIMQ_PARAMS_TEMPLATE: SIMQ uses async_io by default for its file-backed multi-vector storage, while callers can still override base_io_type explicitly. Since this PR only replaces template construction with a structured builder, changing the default would be an unrelated behavioral change, so no code change is made here.

LHT129 (commented):
## Code Review: PR #2730

**Commit:** `b00f317` — refactor(parameters): replace template-based index mappings

### Summary

This is a well-executed refactoring that replaces template-string-based JSON parameter construction (`format_map` + `ConstParamMap` + `DEFAULT_MAP`) with explicit builder functions and structured if-else field mapping chains across BruteForce, HGraph, IVF, Pyramid, and SIMQ index types.

### What was reviewed

- **36 changed files**, 1327 additions, 1582 deletions
- All key source files read from commit `b00f317`
- Existing review comments from previous commits checked for resolution

### Positive observations

1. **Previous Copilot issues are all addressed:**
   - STORE_RAW_VECTOR now correctly applies `ApplyHoldMoldsToQuantizer` to both base and precise codes (was previously only applied to base codes)
   - `size_t` usage replaced with `int64_t` where appropriate
   - Error messages use consistent formatting
   - `CreateDefault()` no longer trims user-specified quantization params
   - Test ordering corrected

2. **Clean removal of dead code:** `DEFAULT_MAP` (~100 entries) and all `format_map`/`ConstParamMap` usage removed. Verified no remaining references in the PR commit.

3. **Good boundary validation added:** SINDI and SINDIV2 now use `supported_keys` unordered_set to reject unknown flat keys, preventing silent parameter drift.

4. **Comprehensive test coverage:** New regression tests added for bruteforce, hgraph, pyramid, sindi_v2, bucket_datacell, fp32/int8 quantizer, transform_quantizer, memory_io, and simq parameter handling.

5. **`CompatibilityReport` / `CollectCompatibilityIssues`:** The recursive JSON diff approach in `parameter.cpp` is a clean, general-purpose mechanism for detecting parameter incompatibility.

6. **`RaBitQSplitConfig`:** Well-structured with proper validation in `ParseRaBitQSplitConfig` and clean separation of parsing vs. application.

### Issues found

**No substantive issues identified.** The refactoring is consistent, the builder pattern is applied uniformly across all index types, and all edge cases (unknown keys, missing fields, type mismatches) are handled with appropriate error reporting.

### Verdict

LGTM. Ready to merge pending CI pass.

src/algorithm/bruteforce/bruteforce.cpp
```diff
@@ -1223,151 +1193,77 @@ BruteForce::CheckAndMappingExternalParam(const JsonType& external_param,
...
+            inner_json[BUILD_THREAD_COUNT_KEY].SetJson(field);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY].SetJson(ApplyHoldMoldsToQuantizer(
+                inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY], field.GetBool()));
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY].SetJson(
+                ApplyHoldMoldsToQuantizer(inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY],
+                                          field.GetBool()));
+        } else if (key == USE_ATTRIBUTE_FILTER) {
+            inner_json[USE_ATTRIBUTE_FILTER_KEY].SetJson(field);
+        } else if (key == BRUTE_FORCE_USE_RESIDUAL) {
```
LHT129, line 1260: [suggestion] The `build_default_brute_force_param` function creates a `FlattenDataCellParameter::CreateDefault` for both base and precise codes, but does not pass `hold_molds` to the base codes when `STORE_RAW_VECTOR` is not set. This is fine for the default case, but when `STORE_RAW_VECTOR` is later applied in the mapping loop, it calls `ApplyHoldMoldsToQuantizer` which only sets `hold_molds` on FP32 and INT8 quantizers. If the base or precise quantizer type is something else (e.g. SQ8, RaBitQ), the `hold_molds` field is silently not applied, which means `STORE_RAW_VECTOR=true` would be accepted but have no effect. Consider either rejecting `STORE_RAW_VECTOR` for non-FP32/INT8 quantizer types, or documenting that it is only effective for those types.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -14,666 +14,305 @@
...
+            inner_json[USE_ATTRIBUTE_FILTER_KEY].SetJson(value);
+        } else if (key == HGRAPH_BASE_QUANTIZATION_TYPE) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY][TYPE_KEY].SetJson(value);
+        } else if (key == STORE_RAW_VECTOR) {
+            inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY].SetJson(ApplyHoldMoldsToQuantizer(
+                inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY], value.GetBool()));
+            inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY].SetJson(
+                ApplyHoldMoldsToQuantizer(inner_json[PRECISE_CODES_KEY][QUANTIZATION_PARAMS_KEY],
+                                          value.GetBool()));
+            inner_json[STORE_RAW_VECTOR_KEY].SetJson(value);
```
LHT129, line 173: [suggestion] The `STORE_RAW_VECTOR` handling in `map_hgraph_param` also calls `ApplyHoldMoldsToQuantizer` on both base and precise quantizers. Same issue as the BruteForce mapping: if the quantizer type is not FP32 or INT8, the `hold_molds` flag is silently dropped. This is more impactful here because HGraph supports a wider variety of quantizer types (RaBitQ, SQ8, etc.) via `HGRAPH_BASE_QUANTIZATION_TYPE`. Consider adding validation or at least a warning when `STORE_RAW_VECTOR=true` is combined with a non-FP32/INT8 quantizer.

src/algorithm/inner_index_parameter.cpp
```diff
@@ -128,16 +130,44 @@ MapRaBitQSplitParam(const JsonType& external_json, JsonType& inner_json) {
...
     inner_json[BASE_CODES_KEY][QUANTIZATION_PARAMS_KEY][RABITQ_QUANTIZATION_BITS_PER_DIM_BASE_KEY]
-        .SetInt(total_bits);
+        .SetInt(config.TotalBits());
+}
+
+JsonType
+ApplyHoldMoldsToQuantizer(const JsonType& source_json, bool hold_molds) {
+    auto quantizer_json = source_json;  // NOLINT(performance-unnecessary-copy-initialization)
+    auto quantizer_type = Parameter::TryToParseType(quantizer_json);
+    if (quantizer_type == QUANTIZATION_TYPE_VALUE_TQ) {
```
LHT129, line 158: [note] The `RequiresRawVectorForTransformQuantizer` function checks whether a TQ-based index needs raw vectors by examining `use_reorder` and `reorder_source`. However, this function is only called from `RequiresRawVectorForMRLERaBitQSplit`, which additionally checks for `RABITQ_SPLIT_CODES`. This means the general TQ case (without RaBitQ split) never gets the automatic `store_raw_vector=true` treatment. Is this intentional? If TQ without reorder always needs raw vectors for decoding, the function name suggests it should be used more broadly.

src/algorithm/inner_index_parameter.cpp
```diff
@@ -128,16 +130,44 @@ MapRaBitQSplitParam(const JsonType& external_json, JsonType& inner_json) {
                                    query_bits));
     }
```
LHT129, line 131: [suggestion] The `ParseRaBitQSplitConfig` function validates that `rabitq_bits_per_dim_query` must be 32 when present. However, `ApplyRaBitQSplitConfig` unconditionally sets `rabitq_bits_per_dim_query` to 32 regardless of whether the user provided it. This means if a user passes `rabitq_bits_per_dim_query=32`, the validation passes and the value is set to 32. But if the user omits it, it is also set to 32 without validation. The behavior is consistent but the validation is redundant since the value is always overridden. Consider removing the validation or only setting the default when the user did not provide the key.

src/parameter.cpp
```diff
@@ -0,0 +1,105 @@
...
+    for (const auto ch : token) {
+        if (ch == '~') {
+            escaped.append("~0");
```
LHT129, line 32: [suggestion] The `CollectCompatibilityIssues` method recursively traverses JSON trees and reports differences. However, it uses ordered key iteration which depends on insertion order. If two semantically equivalent JSON objects have keys in different orders (e.g. `{"a":1,"b":2}` vs `{"b":2,"a":1}`), the diff may report spurious issues. Since the compatibility check compares parameter trees that may be constructed through different code paths (e.g. one from `CreateDefault().ToJson()` and another from user input), key ordering differences could produce false positives. Consider sorting keys before comparison or using a structural comparison that is order-independent for objects.

src/datacell/flatten_datacell_parameter.cpp
```diff
@@ -17,12 +17,32 @@
...
+    } else if (auto int8 = std::dynamic_pointer_cast<INT8QuantizerParameter>(
+                   parameter->quantizer_parameter);
+               int8 != nullptr) {
+        int8->hold_molds = hold_molds;
+    }
+    return parameter;
+}
+
```
LHT129, line 45: [note] The `FlattenDataCellParameter::CreateDefault` now handles `hold_molds` for both FP32 and INT8 quantizers. However, the `ApplyHoldMoldsToQuantizer` function in `inner_index_parameter.cpp` also handles both FP32 and INT8. There is a subtle difference: `CreateDefault` sets `hold_molds` directly on the typed parameter object, while `ApplyHoldMoldsToQuantizer` sets it via JSON manipulation (`quantizer_json[HOLD_MOLDS].SetBool(hold_molds)`). If `INT8QuantizerParameter::FromJson` does not parse `HOLD_MOLDS`, the JSON path would be silently ignored. Worth verifying that `INT8QuantizerParameter::FromJson` actually reads the `hold_molds` field.

src/algorithm/bruteforce/bruteforce.cpp
LHT129: No code change is needed here. STORE_RAW_VECTOR is not implemented solely through hold_molds: the mapping always sets the independent store_raw_vector field, which InnerIndexParameter::FromJson parses and uses to create raw_vector storage. ApplyHoldMoldsToQuantizer is an additional capability-specific optimization for FP32/INT8 (or TQ over them); unsupported quantizers correctly omit that field without disabling raw-vector storage.

src/algorithm/hgraph/hgraph_param_mapping.cpp
LHT129: Same conclusion as for BruteForce: STORE_RAW_VECTOR always updates the independent store_raw_vector parameter, so raw-vector storage remains effective for RaBitQ/SQ8. hold_molds is deliberately propagated only to quantizers that serialize/support it (FP32/INT8 and TQ over those types); no warning or rejection is appropriate for the other quantizers.

src/algorithm/inner_index_parameter.cpp
LHT129: The premise is not true globally: Pyramid::CheckAndMappingExternalParam calls RequiresRawVectorForTransformQuantizer directly for the general TQ case. HGraph uses the narrower RequiresRawVectorForMRLERaBitQSplit check because that is the configuration in which its base reorder representation requires forced raw storage. The helper naming reflects the general predicate, while each index applies it according to its decode-source model.

src/algorithm/inner_index_parameter.cpp
LHT129: The validation is intentional. An omitted query-bits field receives the canonical split default of 32; an explicitly supplied value must also be 32, otherwise the user configuration is contradictory and should be rejected rather than silently rewritten. ApplyRaBitQSplitConfig then serializes the canonical value in both cases.

src/parameter.cpp
LHT129: No ordering issue exists in this implementation. Object fields are matched by key using rhs.contains(key)/rhs.at(key), and the second pass only detects keys absent from lhs. Iteration order can affect report ordering but cannot create or remove compatibility issues, so semantically equivalent objects with different insertion orders compare cleanly.

src/datacell/flatten_datacell_parameter.cpp
LHT129: Verified: INT8QuantizerParameter::FromJson explicitly reads HOLD_MOLDS, and ToJson serializes it. The typed CreateDefault path and ApplyHoldMoldsToQuantizer JSON path therefore round-trip to the same INT8 parameter state; existing INT8 parameter coverage exercises this behavior.

Issue #2718, opened 2026-08-19
[bug](hgraph): allow Tune to use incomplete FP32 sources


jac0626:
### Affected area

hgraph

### Interface

cpp

### VSAG version

v1.0.0-60-gb92b50a5-dirty

### Summary

An HGraph Add failure can leave `total_count_` and the label table ahead of the available
vector codes. Tune currently rejects every FP32 source whose `TotalCount()` does not cover the
published HGraph count and returns false.

This issue scopes the immediate fix to allowing Tune to rebuild the readable FP32 prefix. Add
failure atomicity and cross-component consistency remain a separate problem.

### Minimal reproducer

```cpp
REQUIRE(index->Build(base).has_value());

rejecting_pool->SetRejectSubmissions(true);
auto add_result = index->Add(add);
rejecting_pool->SetRejectSubmissions(false);

REQUIRE_FALSE(add_result.has_value());
REQUIRE(index->GetNumElements() == 3);

// The raw FP32 source still contains only the two successfully built vectors.
auto tune_result = index->Tune(tune_to_bf16_parameters);
REQUIRE(tune_result.has_value());
REQUIRE(tune_result.value());
```

The complete rejecting-thread-pool reproducer is covered by
`src/algorithm/hgraph/hgraph_add_test.cpp:199`.

### Environment

```shell
- OS: Ubuntu 22.04
- vsag version: v1.0.0-60-gb92b50a5-dirty
- compiler version: c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
```

### Expected behavior

Tune should select the available FP32 source with the largest readable prefix and rebuild that
prefix without requiring it to cover `total_count_`.

### Actual behavior

Tune requires `codes->TotalCount() >= total_count_`. When Add has published an ID but failed
before writing its codes, no FP32 source satisfies the check and Tune returns false.

### References

- `src/algorithm/hgraph/hgraph.cpp:170`
- `src/algorithm/hgraph/hgraph.cpp:188`
- `src/algorithm/hgraph/hgraph.cpp:227`
- `src/algorithm/hgraph/hgraph_add_test.cpp:199`

### Additional context

This compatibility change does not repair the underlying Add consistency issue or add general
support for sparse logical IDs.

## Related issues

_None found._

---
_Drafted with AI assistant: Codex:GPT-5_

PR #2731, opened 2026-08-20
[Backport 1.0] fix(hgraph): allow Tune to use incomplete FP32 sources

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Documentation

## Linked Issue

Fixes: #2718

## What Changed

- Backport raw, precise FP32, and base FP32 Tune-source support to the `1.0` branch.
- Select sources directly in raw, precise FP32, then base FP32 order; only fall through when the current candidate is absent or empty.
- Do not require the selected source to cover the published HGraph ID range.
- Preserve the legacy training count and rebuild the complete published ID range.
- Keep the locking, delayed parameter updates, and unloaded precise-code rebuild behavior.
- Add regression coverage for an Add submission failure that publishes an ID before vector insertion completes.

This is the `1.0` counterpart of #2719 and #2720.

## How It Was Tested

```text
clang-format-15 --dry-run --Werror src/algorithm/hgraph/hgraph.cpp src/algorithm/hgraph/hgraph_tune_test.cpp tests/test_hgraph.cpp
git diff --check
cmake --build build --target unittests functests --parallel 6
./build/tests/unittests '[hgraph_tune_incomplete_source]' --allow-running-no-tests
./build/tests/functests '[tune_codes]' --allow-running-no-tests
```

Results: 8 Add-failure regression assertions and 8040 Tune-source assertions passed.

## Compatibility and Risk

- No public API or ABI change.
- Consistent indexes retain their existing behavior.
- Inconsistent indexes rebuild the full published ID range, matching legacy Tune semantics.
- Add-state consistency and invalid individual slots remain outside this PR.

## Checklist

- [x] Linked the relevant issue
- [x] Added regression coverage
- [x] Existing Tune-source functional coverage passes
- [x] No documentation change required

### DIFF
--- a/src/algorithm/hgraph/hgraph.cpp
+++ b/src/algorithm/hgraph/hgraph.cpp
@@ -101,8 +101,9 @@ HGraph::HGraph(const HGraphParameterPtr& hgraph_param, const vsag::IndexCommonPa
 
 bool
 HGraph::Tune(const std::string& parameters, bool disable_future_tuning) {
-    if (not this->index_feature_list_->CheckFeature(IndexFeature::SUPPORT_TUNE) or
-        not this->has_raw_vector_) {
+    std::scoped_lock lock(this->add_mutex_);
+    if (this->immutable_.load(std::memory_order_acquire) or
+        not this->index_feature_list_->CheckFeature(IndexFeature::SUPPORT_TUNE)) {
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


src/algorithm/hgraph/hgraph.cpp
```diff
@@ -127,61 +128,75 @@ HGraph::Tune(const std::string& parameters, bool disable_future_tuning) {
...
 
-    // update create_param_ptr_
-    if (is_tune_base_code) {
-        param->base_codes_param = hgraph_parameter->base_codes_param;
-    }
-    if (is_tune_precise_code) {
-        param->precise_codes_param = hgraph_parameter->precise_codes_param;
+    FlattenInterfacePtr tune_source;
+    if (is_tune_base_code or is_tune_precise_code) {
+        tune_source = raw_vector_;
```
LHT129, line 159: [suggestion] The `tune_source` selection for `raw_vector_` only checks `TotalCount() == 0`, but does not verify that `raw_vector_` covers all active IDs (unlike the `covers_active_ids` check applied to `high_precise_codes_` for the `is_tune_precise_code` decision).

In the Add-failure scenario that this PR targets — where an ID is published (`total_count_` incremented) but the vector insertion into `raw_vector_` is skipped — `raw_vector_` would have `TotalCount() > 0` (from Build) yet still lack the vector for the failed inner_id. The current fall-through logic would select `raw_vector_` as the tune source, and `decode_tune_source` would throw when encountering the missing inner_id.

Consider adding a `covers_active_ids(raw_vector_)` check, or falling through to the next candidate when `GetCodesById` returns nullptr in `decode_tune_source`.

LHT129 (commented):
Thanks for this backport. The refactoring of the Tune source selection logic is clean and the test coverage is thorough.

I left one inline comment about the `tune_source` selection for `raw_vector_` — it only checks `TotalCount() == 0` rather than `covers_active_ids`, which could be problematic in the Add-failure scenario where `raw_vector_` has some but not all vectors.

Otherwise the changes look good:
- Moving the `add_mutex_` lock earlier and replacing `has_raw_vector_` with `immutable_` check is correct
- The `covers_active_ids` check for `high_precise_codes_` in the `is_tune_precise_code` decision is a good safety net
- Moving `create_param_ptr_` updates into the `global_mutex_` section is safer
- The `decode_tune_source` lambda properly handles buffer lifecycle

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -127,61 +128,75 @@ HGraph::Tune(const std::string& parameters, bool disable_future_tuning) {
...
-    }
+    const auto current_count = total_count_.load(std::memory_order_acquire);
+    auto covers_active_ids = [current_count](const FlattenInterfacePtr& codes) {
+        return codes != nullptr and codes->TotalCount() >= current_count;
+    };
 
-    // check which code need to tune and update create_param_ptr_
+    // Check which codes need to be rebuilt.
     bool is_tune_base_code = false;
     bool is_tune_precise_code = false;
```
LHT129, line 146: [suggestion] The fallback from `high_precise_codes_` to `basic_flatten_codes_` uses the same `TotalCount() == 0` check as the `raw_vector_` selection, rather than `covers_active_ids`. This has the same gap that LHT129 identified for `raw_vector_` — if `high_precise_codes_` has `TotalCount() > 0` but does not cover all active IDs (e.g. in the Add-failure scenario where the FP32 precise codes were built before the failed Add), it will be selected as `tune_source` and `decode_tune_source` will throw when `GetCodesById` returns nullptr for the missing inner_id.

Consider using `covers_active_ids` consistently across all three candidates:

```cpp
if ((tune_source == nullptr or tune_source->TotalCount() == 0) and
    high_precise_codes_ != nullptr and
    high_precise_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_FP32 and
    covers_active_ids(high_precise_codes_)) {
    tune_source = high_precise_codes_;
}
if ((tune_source == nullptr or tune_source->TotalCount() == 0) and
    basic_flatten_codes_ != nullptr and
    basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_FP32 and
    covers_active_ids(basic_flatten_codes_)) {
    tune_source = basic_flatten_codes_;
}
```

Or alternatively, as LHT129 suggested, make `decode_tune_source` fall through to the next candidate when `GetCodesById` returns nullptr rather than throwing.

LHT129 (commented):
Thanks for this backport. The refactoring of the Tune source selection logic is clean and the test coverage is thorough — 8 Add-failure regression assertions plus 8040 tune-source assertions across 7 functional scenarios is excellent.

I left one additional inline comment extending LHT129's observation: the `covers_active_ids` gap applies not only to `raw_vector_` but also to the `high_precise_codes_` and `basic_flatten_codes_` fallback candidates. All three candidates in the tune_source selection chain should use `covers_active_ids` consistently, or alternatively `decode_tune_source` should gracefully fall through when `GetCodesById` returns nullptr.

Beyond that, the changes look solid:
- Moving `add_mutex_` earlier and replacing `has_raw_vector_` with `immutable_` is correct
- The `covers_active_ids` check for `high_precise_codes_` in the `is_tune_precise_code` decision is a good safety net
- Moving `create_param_ptr_` updates into the `global_mutex_` section is safer
- The `decode_tune_source` lambda properly handles buffer lifecycle with get/Decode/release

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -101,8 +101,9 @@ HGraph::HGraph(const HGraphParameterPtr& hgraph_param, const vsag::IndexCommonPa
 
 bool
 HGraph::Tune(const std::string& parameters, bool disable_future_tuning) {
-    if (not this->index_feature_list_->CheckFeature(IndexFeature::SUPPORT_TUNE) or
-        not this->has_raw_vector_) {
+    std::scoped_lock lock(this->add_mutex_);
```
LHT129, line 104: [suggestion] The `add_mutex_` lock is now acquired before parameter parsing (`JsonType::Parse`, `map_hgraph_param`, `MakeInstance`), whereas previously it was acquired only after parsing, right before the `immutable_` check. This means the lock is held during potentially expensive operations (JSON parsing, memory allocation for new codec instances), which could block concurrent `Add` operations longer than necessary.

The `immutable_` check and the `has_raw_vector_` removal motivated moving the lock earlier, but the parsing itself does not require `add_mutex_` protection. Consider narrowing the lock scope:

```cpp
// Parse parameters outside the lock
auto parsed_params = JsonType::Parse(parameters);
// ... (rest of parsing) ...

std::scoped_lock lock(this->add_mutex_);
if (this->immutable_.load(std::memory_order_acquire) or
    not this->index_feature_list_->CheckFeature(IndexFeature::SUPPORT_TUNE)) {
    return false;
}
// ... (rest of Tune under lock) ...
```

This would restore the original lock granularity while keeping the corrected early-exit conditions.

LHT129 (commented):
Thanks for this backport. The changes are well-structured and the test coverage is thorough. I left one suggestion about lock granularity — the `add_mutex_` is now held during parameter parsing which is a regression from the original code. Otherwise the logic looks sound.

The two existing inline comments about `covers_active_ids` consistency in the `tune_source` fallback chain are worth addressing before merging.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -127,61 +128,75 @@ HGraph::Tune(const std::string& parameters, bool disable_future_tuning) {
...
-        this->high_precise_codes_->GetQuantizerName() != new_precise_code->GetQuantizerName()) {
-        // [case 2] precise code is not same
-        is_tune_precise_code = true;
-    }
-    if (not inner_parameter->use_reorder or new_reorder_by_base) {
-        // [case 3] drop precise_code
-        new_use_reorder = inner_parameter->use_reorder;
-        drop_precise_codes = true;
-        param->precise_codes_param.reset();
-        is_tune_precise_code = false;
```
LHT129, line 161: [note] The `train_count` variable on line 195 uses `this->GetNumElements()` (which returns `total_count_ - delete_count_`), while `current_count` on line 139 captures `total_count_` directly. In the presence of force-removed elements, these two values differ: `train_count` would be smaller than `current_count`. The train loop (line 198) iterates over `[0, train_count)`, and the rebuild loop (line 212) iterates over `[0, current_count)`. Both ranges are valid since force-removed IDs still have entries in the codes, but the asymmetry is worth noting — it means the training sample excludes force-removed vectors while the rebuild includes them. This matches the intended semantics (force-removed vectors should not participate in training), but a brief comment explaining this distinction would help future readers.

PR #2736, opened 2026-08-21
[feat](eval_performance): support filtered HGraph workloads in AutoTune

### LINKED ISSUE #2735: [feat](eval_performance): support filtered HGraph workloads in AutoTune
### Affected area

eval_performance

### Motivation

AutoTune can evaluate representative KNN workloads, but production vector searches commonly apply
a different ID filter for each query. Without replaying those filters, recall and latency
measurements do not represent the production workload, so the selected HGraph search parameters
may be unsuitable.

### Proposed solution

Extend the typed AutoTune workload with two mutually exclusive per-query inputs:

- `query_filters`: one `FilterPtr` per query.
- `query_invalid_bitsets`: one exclusion `BitsetPtr` per query.

An empty vector means unfiltered search; otherwise its size must equal the query count and a null
entry leaves that query unfiltered. Pass the selected form directly through the shared eval core
for both `TuneSearch` and `TuneIndex`, require matching filtered ground truth when recall is
evaluated, and initially limit the feature to HGraph ID filters in the typed C++ API.

### Alternatives considered

- One shared filter per workload cannot represent query-specific predicates.
- Wrapping every bitset in a caller-defined `Filter` loses the direct bitset API and makes common
  workloads less convenient.
- Adopting the complete `SearchRequest` surface immediately would also pull attribute and
  extra-info filtering into V1.

additive only (no break)

### References

- `tools/autotune/autotune.h:55`
- `tools/eval/case/search_eval_case.cpp:200`
- `include/vsag/index.h:284`
- `docs/docs/en/src/advanced/filtered_search.md:1`

### Additional context

Out of scope: attribute/SQL filters, extra-info filters, CLI serialization of filters, and
filtered IVF/Pyramid tuning.

## Related issues

_None found._

---

jac0626:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2735

## What Changed

- Add per-query `FilterPtr` and invalid `BitsetPtr` workload inputs to typed HGraph `TuneSearch` and `TuneIndex` requests.
- Route filtered workloads through the shared evaluation core while preserving HGraph's native bitset search path.
- Validate filter shape, mutual exclusion, supported index scope, and unsupported extra-info filtering.
- Add separate `FilterPtr` and `BitsetPtr` examples, focused regression tests, and English/Chinese user documentation.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
clang-format-15 --dry-run --Werror <all changed C++ files>
cmake --build build --target autotune_test eval_dataset_test \
  327_feature_autotune_existing_index \
  330_feature_autotune_existing_index_bitset -j$(nproc)

./build/tools/autotune/autotune_test \
  "AutoTune evaluates per-query filters and bitsets for HGraph"
  All tests passed (29 assertions in 1 test case)

./build/tools/eval/eval_dataset_test \
  "EvalDataset builds a query-only view for id recall"
  All tests passed (21 assertions in 1 test case)

./build/tools/eval/eval_dataset_test \
  "EvaluateSearch validates inputs and propagates search errors"
  All tests passed (23 assertions in 1 test case)

./build/examples/cpp/327_feature_autotune_existing_index
  recall_at_k: 1.0; first neighbor id: 1000

./build/examples/cpp/330_feature_autotune_existing_index_bitset
  recall_at_k: 1.0; first neighbor id: 1000
```

## Compatibility Impact

- API/ABI compatibility: Additive fields in the build-tree AutoTune API under `tools/`; no installed VSAG SDK API is changed.
- Behavior changes: Typed HGraph tuning requests may now evaluate per-query filters. Filtered requests for other index types fail validation explicitly.

- Performance impact: The unfiltered evaluation path is unchanged. Filtered trials use the existing native HGraph filter/bitset search overloads.
- Concurrency/thread-safety impact: The same per-query filters or bitsets may be read by concurrent search workers and therefore must remain immutable and thread-safe during tuning.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/resources/autotune.md`

## Risk and Rollback

- Risk level: medium
- Rollback plan: Revert commit `92dfeaa4`; unfiltered AutoTune behavior is otherwise unchanged.

## Checklist

### DIFF
--- a/tools/autotune/autotune.cpp
+++ b/tools/autotune/autotune.cpp
@@ -419,6 +419,7 @@ make_context(eval::EvalDatasetPtr dataset,
     request.base_count = base_count;
     request.query_count = static_cast<uint64_t>(request.dataset->GetNumberOfQuery());
     request.ground_truth_k = request.dataset->GetGroundTruthK();
+    request.has_query_filters = request.dataset->HasQueryFilters();
     require(request.top_k > 0, "request.workload.top_k must be positive");
     require(request.top_k <= static_cast<uint64_t>(std::numeric_limits<int>::max()),
             "request.workload.top_k is too large");
@@ -472,6 +473,10 @@ make_context(eval::EvalDatasetPtr dataset,
           {"keep_intermediate", request.keep_intermediate},
           {"max_trials", request.max_trials}}},
         {"output", {{"include_raw_evaluation", request.include_raw_eval}}}};
+    if (request.has_query_filters) {
+        request.effective_request["workload"]["filtered_query_count"] =
+            request.dataset->GetFilteredQueryCount();
+    }
     for (const auto& [name, value] : request.constraints) {
         request.effective_request["constraints"][name] = value;
     }
@@ -524,8 +529,12 @@ ParseRequest(const IndexRequest& input) {
     const auto metric = normalize(input.metric_type);
     require(metric == "l2" || metric == "ip" || metric == "cosine",
             "request.metric_type must be l2, ip, or cosine");
-    auto dataset = eval::EvalDataset::FromDatasets(
-        input.base, input.workload.queries, input.workload.ground_truth, metric);
+    auto dataset = eval::EvalDataset::FromDatasets(input.base,
+                                                   input.workload.queries,
+                                                   input.workload.ground_truth,
+                                                   metric,
+                                                   input.workload.query_filters,
+                                                   input.workload.query_invalid_bitsets);
     IndexTuningRequest request;
     request.context = make_context(std::move(dataset),
                                    static_cast<uint64_t>(input.base->GetNumElements()),
@@ -536,7 +545,9 @@ ParseRequest(const IndexRequest& input) {
                                    input.config,
                                    false);
     if (input.index_spaces.empty()) {
-        request.indexes = {{"hgraph"}, {"ivf"}};
+        request.indexes = request.context.has_query_filters
+                              ? std::vector<IndexInput>{{"hgraph"}}
+                              : std::vector<IndexInput>{{"hgraph"}, {"ivf"}};
     } else {
         for (uint64_t i = 0; i < input.index_spaces.size(); ++i) {
             request.indexes.emplace_back(parse_index_space(input.index_spaces[i], i, false));
@@ -544,6 +555,8 @@ ParseRequest(const IndexRequest& input) {
     }
     const auto dim = static_cast<uint64_t>(request.context.dataset->GetDim());
     for (auto& index : request.indexes) {
+        require(not request.context.has_query_filters || index.name == "hgraph",
+                "filtered workloads currently support only hgraph");
         merge_dataset_field(index.create_params, "dim", dim, index.name);
         merge_dataset_field(index.create_params, "dtype", vsag::DATATYPE_FLOAT32, index.name);
         merge_dataset_field(index.create_params, "metric_type", metric, index.name);
@@ -560,8 +573,13 @@ ParseRequest(const SearchRequest& input) {
     require(input.index != nullptr, "request.index is required");
     const auto element_count = input.index->GetNumElements();
     require(element_count > 0, "request.index must not be empty");
-    auto dataset =
-        eval::EvalDataset::FromSearchDatasets(input.workload.queries, input.workload.ground_truth);
+    const auto concrete_index_name = index_name(input.index);
+    auto dataset = eval::EvalDataset::FromSearchDatasets(input.workload.queries,
+                                                         input.workload.ground_truth,
+                                                         input.workload.query_filters,
+                                                         input.workload.query_invalid_bitsets);
+    require(not dataset->HasQueryFilters() || concrete_index_name == "hgraph",
+            "filtered workloads currently support only hgraph");
     SearchTuningRequest request;
     request.context = make_context(std::move(dataset),
                                    static_cast<uint64_t>(element_count),
@@ -572,7 +590,7 @@ ParseRequest(const SearchRequest& input) {
                                    input.config,
                                    true);
     IndexSpace space;
-    space.name = index_name(input.index);
+    space.name = concrete_index_name;
     space.search_parameter_space = input.parameter_space;
     request.index_input = parse_index_space(space, 0, true);
     request.index = input.index;
--- a/tools/autotune/autotune_candidate.cpp
+++ b/tools/autotune/autotune_candidate.cpp
@@ -291,6 +291,17 @@ supports_adaptive_ef_search_objective(const std::string& objective) {
            objective == "search_seconds" || objective == "build_and_search_seconds";
 }
 
+bool
+uses_extra_info_filter(const JsonType& search_params) {
+    if (!search_params.contains("hgraph") || !search_params["hgraph"].is_object()) {
+        return false;
+    }
+    const auto& hgraph = search_params["hgraph"];
+    return hgraph.contains("use_extra_info_filter") &&
+           hgraph["use_extra_info_filter"].is_boolean() &&
+           hgraph["use_extra_info_filter"].get<bool>();
+}
+
 int64_t
 positive_int64(const JsonType& value, const std::string& path) {
     if (!value.is_number_integer()) {
@@ -408,6 +419,11 @@ generate_candidates(const RequestContext& context,
                                              ? take_hgraph_ef_search_range(search_space, context)
                                              : std::nullopt;
             expand(search_space, [&](const JsonType& search_params) {
+                if (context.has_query_filters && uses_extra_info_filter(search_params)) {
+                    throw std::invalid_argument(
+                        "filtered workloads currently support only ID filters; "
+                        "hgraph.use_extra_info_filter must be false");
+                }
                 Candidate candidate{index.name, create_params, search_params, ef_search_range};
                 JsonType identity{{"index_name", candidate.index_name},
                                   {"create_params", candidate.create_params},
--- a/tools/autotune/autotune_test.cpp
+++ b/tools/autotune/autotune_test.cpp
@@ -26,6 +26,7 @@
 #include <filesystem>
 #include <fstream>
 #include <limits>
+#include <memory>
 #include <string>
 #include <thread>
 #include <utility>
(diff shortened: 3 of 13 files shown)


tools/eval/eval_dataset_test.cpp
```diff
@@ -478,6 +536,23 @@ TEST_CASE("EvaluateSearch validates inputs and propagates search errors", "[ut][
...
+        second_invalid->Set(id, id != ground_truth_ids[1]);
+    }
+    std::vector<vsag::BitsetPtr> bitsets{first_invalid, second_invalid};
+    auto bitset_dataset = EvalDataset::FromDatasets(base, queries, ground_truth, "l2", {}, bitsets);
+    config.enable_qps = false;
+    config.enable_recall = true;
+    const auto bitset_result = vsag::eval::EvaluateSearch(index, bitset_dataset, config);
+    REQUIRE(bitset_result["recall_avg"].get<double>() == 1.0);
+    config.enable_recall = false;
+    config.enable_qps = true;
```
LHT129, line 556: [suggestion] The bitset recall test verifies `recall_avg == 1.0` correctly, but after setting `config.enable_recall = false; config.enable_qps = true;` on this line, no corresponding `EvaluateSearch` call is made for the `bitset_dataset`. The test immediately falls through to the `ef_search=0` error-path test using the original `dataset` (filter-based). This means the bitset path is never exercised under a QPS-only configuration.

If the intent was to also verify the bitset workload works correctly under QPS measurement, add a call like:
```cpp
const auto bitset_qps = vsag::eval::EvaluateSearch(index, bitset_dataset, config);
REQUIRE(bitset_qps.contains("qps"));
```
Otherwise, remove the `config.enable_qps = true;` line to avoid leaving dead state that suggests an incomplete test.

tools/autotune/autotune_test.cpp
```diff
@@ -885,6 +947,122 @@ TEST_CASE("AutoTune searches an in-memory existing index") {
...
+    extra_info_request.parameter_space =
+        R"({"hgraph":{"ef_search":8,"use_extra_info_filter":true}})";
+    const auto extra_info_result = vsag::autotune::TuneSearch(extra_info_request);
+    REQUIRE_FALSE(extra_info_result.has_value());
+    REQUIRE(extra_info_result.error().type == vsag::ErrorType::INVALID_ARGUMENT);
+    REQUIRE(extra_info_result.error().message ==
+            "filtered workloads currently support only ID filters; "
+            "hgraph.use_extra_info_filter must be false");
+
+    auto index_request = fixture.Request(index_workspace.Get());
```
LHT129, line 1023: [note] The `index_request` is constructed via `fixture.Request(index_workspace.Get())` which initializes `workload` with the unfiltered `ground_truth` and a hardcoded `top_k=3`. The test then immediately overwrites `ground_truth` with `filtered_ground_truth` and `query_invalid_bitsets`. This works because `TOP_K` (3) coincidentally matches the `Request()` default, but the dependency on the unfiltered ground truth being set first and then overwritten is fragile.

Consider adding a dedicated factory method (e.g. `RequestFiltered`) or constructing the `IndexRequest` directly with the filtered workload fields to make the intent explicit and avoid relying on the unfiltered defaults.

tools/autotune/autotune_test.cpp
```diff
@@ -308,6 +309,67 @@ struct MemoryFixture {
...
 
+class CountingAllowListFilter : public vsag::Filter {
+public:
+    CountingAllowListFilter(std::vector<int64_t> valid_ids, float valid_ratio)
+        : valid_ids_(std::move(valid_ids)), valid_ratio_(valid_ratio) {
+    }
+
+    bool
+    CheckValid(int64_t id) const override {
+        checks_.fetch_add(1, std::memory_order_relaxed);
```
LHT129, line 320: [note] `CheckValid` performs a linear scan (`std::find`) over `valid_ids_` on every invocation. For the test fixture this is fine since `valid_ids_` is small (size = `TOP_K` = 3), but if this filter class is ever copied into production code or used with larger allow-lists, the O(n) per-ID check would become a hot-path bottleneck. Consider adding a comment noting this is test-only, or using an `unordered_set` for the allow-list lookup to make the intent clearer.

LHT129 (commented):
Thanks for this PR. The implementation is clean and well-structured — the filter/bitset validation, mutual exclusion checks, and index-type gating are all correctly placed. The test coverage is thorough with both positive and negative paths.

I left 3 inline comments:

1. **[suggestion]** `tools/eval/eval_dataset_test.cpp:556` — The bitset QPS test appears incomplete; `config.enable_qps = true` is set but never exercised for the bitset dataset.

2. **[note]** `tools/autotune/autotune_test.cpp:1023` — The filtered `index_request` construction reuses `fixture.Request()` which sets unfiltered defaults that are immediately overwritten. Consider a dedicated factory to make the intent clearer.

3. **[note]** `tools/autotune/autotune_test.cpp:320` — `CountingAllowListFilter::CheckValid` uses `std::find` (O(n)). Fine for the test fixture but worth a comment noting it is test-only.

Overall the PR is in good shape. The core logic (validation, routing, candidate gating) is correct and the documentation updates are thorough.

tools/eval/eval_dataset.h
```diff
@@ -188,6 +197,36 @@ class EvalDataset {
         return query_dataset_ == nullptr ? nullptr : query_dataset_->GetPaths();
     }
 
+    [[nodiscard]] const vsag::FilterPtr&
+    GetQueryFilter(uint64_t query_id) const {
+        static const vsag::FilterPtr no_filter;
+        return query_filters_.empty() ? no_filter : query_filters_[query_id];
+    }
```
LHT129, line 204: [suggestion] `GetQueryFilter` and `GetQueryInvalidBitset` access the member vectors via `operator[]` without bounds checking on `query_id`. While `validate_query_filter_inputs` guarantees the vector size equals the query count at construction time, a future code path that constructs an `EvalDataset` without going through the factory methods could bypass this invariant. Consider adding an assertion (e.g. `assert(query_id < query_filters_.size())`) or using `.at()` for bounds-checked access in debug builds.

Current:
```cpp
return query_filters_.empty() ? no_filter : query_filters_[query_id];
```

Suggested:
```cpp
if (query_filters_.empty()) return no_filter;
assert(query_id < query_filters_.size());
return query_filters_[query_id];
```
(Same for `GetQueryInvalidBitset`.)

tools/autotune/autotune_test.cpp
```diff
@@ -885,6 +947,122 @@ TEST_CASE("AutoTune searches an in-memory existing index") {
...
+    search_request.workload.query_invalid_bitsets = query_invalid_bitsets;
+    const auto bitset_search_result = vsag::autotune::TuneSearch(search_request);
+    REQUIRE(bitset_search_result.has_value());
+    REQUIRE(bitset_search_result->status == vsag::autotune::TuneStatus::SUCCESS);
+    REQUIRE(bitset_search_result->metrics["recall_at_k"] == 1.0);
+    REQUIRE(bitset_search_result->report["request"]["workload"]["filtered_query_count"] ==
+            MemoryFixture::QUERY_COUNT);
+
+    auto extra_info_request = search_request;
+    extra_info_request.parameter_space =
```
LHT129, line 1014: [note] The `extra_info_request` test at line 1014 only exercises the `use_extra_info_filter` rejection path when `query_invalid_bitsets` is set (inherited from `search_request` after the bitset test above). The rejection logic in `autotune_candidate.cpp:423` does not differentiate between filter types, so the behavior is identical for both paths, but testing the filter path as well would improve coverage symmetry.

Consider adding a similar test with `query_filters` set instead of `query_invalid_bitsets`, or adding a brief comment noting that the rejection is type-agnostic.

tools/autotune/autotune.cpp
```diff
@@ -572,7 +590,7 @@ ParseRequest(const SearchRequest& input) {
                                    input.config,
```
LHT129, line 590: [suggestion] The `use_extra_info_filter` rejection for filtered workloads is validated inside `generate_candidates` (autotune_candidate.cpp:422-426), which runs during the "candidate_generation" stage. For `TuneSearch` with a fixed parameter space, this means the error is reported as a candidate_generation failure rather than a validation failure.

In contrast, `ParseRequest(SearchRequest)` already validates that the index type is HGraph when filters are present (autotune.cpp:581-582), and `ParseRequest(IndexRequest)` validates index type at parse time (autotune.cpp:558).

Consider adding a `use_extra_info_filter` check inside `ParseRequest(SearchRequest)` as well, so the error is caught during the "validation" stage and reported consistently. The check in `generate_candidates` can remain as a defense-in-depth measure for the `TuneIndex` path where the parameter space is expanded from ranges.

tools/autotune/autotune_candidate.cpp
```diff
@@ -408,6 +419,11 @@ generate_candidates(const RequestContext& context,
                                              ? take_hgraph_ef_search_range(search_space, context)
                                              : std::nullopt;
             expand(search_space, [&](const JsonType& search_params) {
+                if (context.has_query_filters && uses_extra_info_filter(search_params)) {
```
LHT129, line 422: [note] The `uses_extra_info_filter` check inside `generate_candidates` (autotune_candidate.cpp:422) throws inside the `expand` callback. For `TuneIndex` with parameter ranges, this means the exception is thrown during cartesian-product expansion. If the parameter space contains multiple combinations and `use_extra_info_filter` appears only in some of them, the exception won't trigger until that specific combination is expanded — earlier combinations without the flag will have already been added to the candidate list before the throw. The partial candidates are discarded when the exception propagates (caught by `run_tuning_locked`), so correctness is preserved, but the error message could be confusing if a user has a mixed parameter space where only some combinations set the flag.

This is consistent with how other candidate-generation errors work (e.g. duplicate candidates are silently skipped), so no change is required. Just noting the behavior for future reference.

PR #2750, opened 2026-08-24
perf(hgraph): parallelize cold build encoding

### LINKED ISSUE #2749: [improve](hgraph): accelerate parallel graph construction
### Kind of improvement

perf

### Affected area

hgraph

### Summary

HGraph does not fully utilize the configured build threads during NSW construction.

In the SIFT1M reference benchmark, HGraph was configured with 96 build threads but averaged only 41.452 active cores, corresponding to 43.180% whole-machine CPU utilization. Its mean build wall time was 64.457 seconds.

The current batch construction path can submit one future per logical row and wait for all futures after the batch. Optimized build-code preparation can follow the same per-row task pattern. Task submission overhead, load imbalance, code encoding, graph insertion, lock contention, pruning, and serialized publication should be profiled separately before selecting the final optimization.

The goal is to improve HGraph construction throughput while preserving graph correctness, recall, memory behavior, concurrent operation safety, and serialization compatibility. This work should remain separate from the Pyramid implementation in PR #2717.

### Expected benefit

For the same dataset, parameters, thread affinity, and machine:

- Reduce HGraph build wall time by at least 30% from the reproducible upstream baseline.
- Raise average whole-machine CPU utilization from 43.180% toward at least 70% when using 96 build threads.
- Preserve Recall@10 within 0.001 of the 0.980 baseline under the same search parameters.
- Preserve all self-nearest-neighbor correctness checks.
- Avoid a material memory regression; target no more than 10% over the 0.582 GiB baseline.
- Preserve Build/Add behavior, duplicate handling, removal configurations, concurrent Add/Search safety, and binary/streaming serialization compatibility.

The optimization should retain safe fallback paths for storage or index configurations that cannot support concurrent encoding or fixed-worker construction.

### Benchmark evidence

Reference configuration:

- Dataset: SIFT1M, 1,000,000 vectors, 128 dimensions, L2
- Base quantization: RaBitQ, 3 bits per dimension
- Precise reorder quantization: SQ8
- Graph type: NSW
- `max_degree=64`
- `ef_construction=400`
- `alpha=1.0`
- `build_thread_count=96`
- Process pinned to logical CPUs 0-95
- Search validation: `ef_search=100`, TopK 10
- Three alternating cold-build runs

Upstream HGraph baseline:

| Metric | Result |
| --- | ---: |
| Build wall time | 64.457 s |
| Individual wall times | 64.811 / 63.967 / 64.592 s |
| CPU time | 2671.859 s |
| Average active cores | 41.452 |
| Whole-machine CPU utilization | 43.180% |
| Build throughput | 15,515 vectors/s |
| Memory | 0.582 GiB |
| Recall@10 | 0.980 |
| Single-thread search QPS | 996.4 |

PR #2717's Pyramid builder reached 85.484% CPU utilization on the same reference configuration. Pyramid is not an identical index implementation and is not a correctness oracle for HGraph, but the utilization gap indicates substantial room to investigate HGraph scheduling and construction bottlenecks.

Post-optimization numbers:

// TODO: record three alternating cold-build runs and per-optimization ablations on the same host.

### References

- `src/algorithm/hgraph/hgraph_build.cpp:464`
- `src/algorithm/hgraph/hgraph_build.cpp:466`
- `src/algorithm/hgraph/hgraph_build.cpp:493`
- `src/algorithm/hgraph/hgraph_fast_build.cpp:135`
- `src/algorithm/hgraph/hgraph_fast_build.cpp:147`
- `src/algorithm/hgraph/hgraph_parameter.cpp:132`
- `docs/docs/en/src/indexes/hgraph.md:17`
- `docs/docs/en/src/indexes/hgraph.md:80`
- https://github.com/antgroup/vsag/pull/2717
- https://vsag.io/docs

### Additional context

Suggested investigation and implementation sequence:

1. Add stage-level diagnostic measurements for batch preparation, quantization/code encoding, task submission, graph search, pruning, neighbor publication, and worker drain time.
2. Capture worker occupancy, task counts, queue depth, tail latency, lock contention, and off-CPU time.
3. Evaluate each optimization independently:
   - replace one-future-per-vector submission with fixed workers and dynamic work blocks;
   - parallelize RaBitQ/SQ8 pre-encoding only when storage explicitly guarantees safe disjoint insertion;
   - reduce resize or publication serialization only where profiling proves it is material.
4. Report an ablation for every retained optimization rather than combining unrelated changes.
5. Benchmark at multiple thread counts, including 1, 8, 32, and 96 threads.
6. Run HGraph-focused unit, functional, concurrency, and sanitizer/TSAN coverage.
7. Re-run the SIFT1M reference benchmark and report build time, CPU utilization, memory, Recall@10, and search QPS.

This issue covers HGraph construction only. It should not include Pyramid changes or unrelated search-path optimizations.

## Related issues

- https://github.com/antgroup/vsag/issues/2738 — related optimized-build eligibility bug, already closed; not a duplicate of the general parallel-construction work.

---

jac0626:
## Summary

- Pre-encode persistent RaBitQ base and SQ8 precise codes before parallel NSW graph insertion.
- Use a fixed number of contiguous worker blocks instead of per-vector encoding tasks.
- Skip the per-row pre-probe code insertion after the batch has been prepared.
- Preserve the existing path for incremental Add, deduplicated storage, force-remove, raw-vector, non-memory, and other quantization configurations.

## Why

The reference SIFT1M build configured 96 threads but averaged 41.452 active cores. Profiling and ablation identified cold-build RaBitQ/SQ8 encoding as a material serialized stage before graph insertion.

## Benchmark

Reference configuration: SIFT1M, L2, RaBitQ 3-bit base + SQ8 precise reorder, NSW, flat graph storage, `max_degree=64`, `ef_construction=400`, `alpha=1.0`, 96 pinned build threads.

| Metric | Upstream baseline | Parallel pre-encoding ablation |
| --- | ---: | ---: |
| Build wall | 64.457 s (3-run mean) | 39.204 s (confirmation run) |
| Average active cores | 41.452 | 89.158 |
| Whole-machine CPU utilization | 43.180% | 92.873% |
| Throughput | 15,515 vec/s | 25,508 vec/s |
| Memory | 0.582 GiB | 0.582 GiB |
| Recall@10 | 0.980 | 0.981 |
| Self-nearest checks | passed | 10/10 |

The confirmation run reduces wall time by 39.2% with unchanged reported memory and recall within 0.001. It is reported as a single retained-optimization ablation rather than a three-run mean.

## Validation

- `clang-format-15 --dry-run --Werror` on all changed C++ files
- `clang-tidy-15 -p build --quiet` on `hgraph_build.cpp` and `hgraph_fast_build.cpp`
- `build/tests/unittests "[ut][hgraph]" -r compact`: 41 cases, 389 assertions passed
- New RaBitQ 3-bit + SQ8 parallel cold-build regression: 1 case, 18 assertions passed
- Existing MRLE optimized-build regression passed
- `git diff --check`

## Scope and compatibility

No public API, serialization format, search path, Pyramid code, or third-party dependency changes. Unsupported configurations retain the existing encoding/insertion behavior.

Closes: #2749

### DIFF
--- a/src/algorithm/hgraph/hgraph.h
+++ b/src/algorithm/hgraph/hgraph.h
@@ -444,8 +444,10 @@ private:
     [[nodiscard]] bool
     need_temporary_sq8_build_data_for_add() const;
 
-    void
-    prepare_build_codes(const DatasetPtr& data, const Vector<AddRow>& rows);
+    [[nodiscard]] bool
+    prepare_build_codes(const DatasetPtr& data,
+                        const Vector<AddRow>& rows,
+                        const AddContext& context);
 
     [[nodiscard]] bool
     should_insert_codes_before_probe(bool use_dedup_storage) const;
@@ -489,7 +491,10 @@ private:
     graph_read_codes_is_temporary(const AddContext& context) const;
 
     bool
-    insert_one_logical_point(const void* data, const AddRow& row, const AddContext& context);
+    insert_one_logical_point(const void* data,
+                             const AddRow& row,
+                             const AddContext& context,
+                             bool persistent_codes_prepared);
 
     void
     prepare_codes_before_probe_if_needed(const void* data,
--- a/src/algorithm/hgraph/hgraph_build.cpp
+++ b/src/algorithm/hgraph/hgraph_build.cpp
@@ -465,7 +465,7 @@ HGraph::insert_add_batch(const DatasetPtr& data, const AddContext& context, cons
     std::vector<std::future<void>> futures;
     HGraphBuildTaskGuard future_guard(
         futures, context.use_parallel_add ? static_cast<uint64_t>(batch.rows.size()) : 0);
-    this->prepare_build_codes(data, batch.rows);
+    const bool persistent_codes_prepared = this->prepare_build_codes(data, batch.rows, context);
 
     const auto* extra_infos = data->GetExtraInfos();
     const auto* attr_sets = data->GetAttributeSets();
@@ -480,7 +480,7 @@ HGraph::insert_add_batch(const DatasetPtr& data, const AddContext& context, cons
         if (attrs != nullptr and this->use_attribute_filter_) {
             this->attr_filter_index_->Insert(*attrs, row.inner_id);
         }
-        this->insert_one_logical_point(vector_data, row, context);
+        this->insert_one_logical_point(vector_data, row, context, persistent_codes_prepared);
     };
 
     for (const auto& row : batch.rows) {
@@ -547,10 +547,15 @@ HGraph::insert_persistent_codes_to_slot(const void* data, CodeSlotIdType code_sl
 // the probe result by binding deduplicated storage, updating duplicate tracking, or inserting a
 // unique point into the graph.
 bool
-HGraph::insert_one_logical_point(const void* data, const AddRow& row, const AddContext& context) {
+HGraph::insert_one_logical_point(const void* data,
+                                 const AddRow& row,
+                                 const AddContext& context,
+                                 bool persistent_codes_prepared) {
     const auto inner_id = row.inner_id;
     const auto level = row.level;
-    this->prepare_codes_before_probe_if_needed(data, inner_id, context);
+    if (not persistent_codes_prepared) {
+        this->prepare_codes_before_probe_if_needed(data, inner_id, context);
+    }
 
     auto make_search_param = [&]() {
         InnerSearchParam param;
--- a/src/algorithm/hgraph/hgraph_fast_build.cpp
+++ b/src/algorithm/hgraph/hgraph_fast_build.cpp
@@ -131,33 +131,69 @@ HGraph::need_temporary_sq8_build_data_for_add() const {
            this->basic_flatten_codes_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_RABITQ;
 }
 
-void
-HGraph::prepare_build_codes(const DatasetPtr& data, const Vector<AddRow>& rows) {
-    if (this->optimized_build_codes_ == nullptr) {
-        return;
-    }
-
-    if (this->thread_pool_ == nullptr) {
-        for (const auto& row : rows) {
-            this->insert_persistent_codes(get_data(data, row.input_idx), row.inner_id);
+bool
+HGraph::prepare_build_codes(const DatasetPtr& data,
+                            const Vector<AddRow>& rows,
+                            const AddContext& context) {
+    if (this->optimized_build_codes_ != nullptr) {
+        if (this->thread_pool_ == nullptr) {
+            for (const auto& row : rows) {
+                this->insert_persistent_codes(get_data(data, row.input_idx), row.inner_id);
+            }
+            return true;
         }
-        return;
+
+        std::vector<std::future<void>> futures;
+        HGraphBuildTaskGuard task_guard(futures, static_cast<uint64_t>(rows.size()));
+        // Parallel graph insertion may probe rows from the same batch. Make every scalar code
+        // visible before any of those probes starts.
+        for (const auto& row : rows) {
+            const auto inner_id = row.inner_id;
+            const auto input_idx = row.input_idx;
+            futures.emplace_back(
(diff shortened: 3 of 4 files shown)


src/algorithm/hgraph/hgraph_fast_build.cpp
```diff
@@ -131,33 +131,69 @@ HGraph::need_temporary_sq8_build_data_for_add() const {
...
-        futures.emplace_back(
-            this->thread_pool_->GeneralEnqueue([this, data, inner_id, input_idx]() {
-                this->insert_persistent_codes(get_data(data, input_idx), inner_id);
-            }));
+    HGraphBuildTaskGuard task_guard(futures, worker_count);
+
+    // prepare_add_batch() already resized both layouts. The workers only encode disjoint IDs;
+    // these outer locks keep storage stable until every worker finishes.
+    std::shared_lock<std::shared_mutex> add_lock(this->add_mutex_);
+    std::unique_lock<std::shared_mutex> codes_lock(this->persistent_codes_mutex_);
```
LHT129, line 184: [critical] The declaration order of `task_guard` relative to `add_lock` and `codes_lock` creates an exception-safety bug that was previously flagged and remains unfixed.

If `GeneralEnqueue` throws after one or more blocks have already been queued, stack unwinding destroys `add_lock` and `codes_lock` first (reverse declaration order), then `task_guard`. When `task_guard`'s destructor calls `wait_all_futures`, the already-queued workers will execute `insert_persistent_codes_unlocked` without the locks held, while another thread could concurrently resize or mutate the code storage.

The fix is to declare the locks before `task_guard` so they outlive it:

```cpp
std::shared_lock<std::shared_mutex> add_lock(this->add_mutex_);
std::unique_lock<std::shared_mutex> codes_lock(this->persistent_codes_mutex_);
std::vector<std::future<void>> futures;
HGraphBuildTaskGuard task_guard(futures, worker_count);
```

This ensures `wait_all_futures` completes (or the futures are abandoned) before the locks are released.

src/algorithm/hgraph/hgraph.h
```diff
@@ -489,7 +491,10 @@ class HGraph : public InnerIndexInterface {
     graph_read_codes_is_temporary(const AddContext& context) const;
 
     bool
-    insert_one_logical_point(const void* data, const AddRow& row, const AddContext& context);
+    insert_one_logical_point(const void* data,
+                             const AddRow& row,
```
LHT129, line 495: [suggestion] The `persistent_codes_prepared` boolean parameter added to `insert_one_logical_point` is used to skip `prepare_codes_before_probe_if_needed` when codes were already pre-encoded. However, this creates a subtle coupling: the correctness of `insert_one_logical_point` now depends on the caller correctly passing this flag, and the flag's meaning is not self-documenting at the call site.

Consider an alternative approach: instead of threading a boolean through the call chain, move the conditional check into `prepare_codes_before_probe_if_needed` itself by checking whether codes already exist for the given `inner_id`. This would make `insert_one_logical_point` self-contained and eliminate the risk of a future caller forgetting to pass the flag correctly.

If the current approach is preferred for performance (avoiding a lookup), consider renaming the parameter to something more descriptive like `skip_prepare_codes_before_probe` to make the intent clearer at call sites.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -465,7 +465,7 @@ HGraph::insert_add_batch(const DatasetPtr& data, const AddContext& context, cons
     std::vector<std::future<void>> futures;
     HGraphBuildTaskGuard future_guard(
         futures, context.use_parallel_add ? static_cast<uint64_t>(batch.rows.size()) : 0);
-    this->prepare_build_codes(data, batch.rows);
+    const bool persistent_codes_prepared = this->prepare_build_codes(data, batch.rows, context);
```
LHT129, line 468: [note] The `persistent_codes_prepared` flag is captured by value in the lambda at line 483 (`add_func`). Since `prepare_build_codes` returns before any parallel `add_func` invocation starts (due to `wait_all_futures` in the parallel-add path), this is safe. However, if the parallel-add dispatch logic ever changes to overlap `prepare_build_codes` with `add_func` execution, this value capture would become stale. Consider documenting this ordering dependency explicitly.

Issue #2756, opened 2026-08-24
[bug](pyramid): pyramid not init has_raw_vector_.


blueorwild:
### Affected area

other

### Interface

cpp

### VSAG version

v0.18.12

### Summary

pyramid not init has_raw_vector_  like HGraph::check_and_init_raw_vector, it may lead to  InnerIndexInterface::GetDataByIdsWithFlag failed

<img width="758" height="94" alt="Image" src="https://github.com/user-attachments/assets/61337219-9d11-43f5-860f-45133514f0db" />

### Minimal reproducer

none

### Environment

```shell
none
```

### Expected behavior

pyramid init has_raw_vector_  like HGraph::check_and_init_raw_vector

### Actual behavior

none

Issue #2755, opened 2026-08-24
[improve](pyramid): GetRawVectorByIds support pyramid on 0.18


blueorwild:
### Kind of improvement

refactor

### Affected area

other

### Summary

1. pyramid not support GetVectorByInnerId on 0.18. lead to GetRawVectorByIds not support pyramid

### Expected benefit

pyramid support GetVectorByInnerId on 0.18

PR #2758, opened 2026-08-25
[0.18] fix(pyramid): align raw vector retrieval with HGraph

jac0626:
## Change Type

- [x] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2755
- Fixes: #2756

## What Changed

- Separate raw-vector retrieval capability from the `store_raw_vector` data-retention flag, matching HGraph semantics.
- Reuse the selected in-memory FP32 codes when possible; otherwise maintain dedicated FP32 raw storage across train, build, add, resize, distance-by-ID, serialization, and deserialization.
- Advertise `SUPPORT_GET_RAW_VECTOR_BY_IDS` only when the selected retrieval codes can decode exact vectors or raw storage is configured.
- Register ODescent labels through `LabelTable::Insert` so reverse lookups and counts remain consistent.
- Validate unsupported float-vector flags before allocating the temporary inner-ID buffer.

## Test Evidence

- [x] `clang-format-15`
- [x] `clang-tidy-15 -p build src/algorithm/pyramid.cpp src/algorithm/inner_index_interface.cpp --quiet`
- [ ] `make test`
- [ ] `make cov`
- [x] Other targeted and component tests

Test details:

```text
./build/tests/unittests [ut][pyramid]
All tests passed (551 assertions in 5 test cases)

./build/tests/functests [ft][pyramid]
All tests passed (128300 assertions in 12 test cases)
```

## Compatibility Impact

- API/ABI compatibility: no public API or ABI changes.
- Behavior changes: FP32-backed Pyramid indexes can expose `GetRawVectorByIds` without opting into stored raw data; `GetDataByIds` includes vectors only when `store_raw_vector` is enabled. Existing serialization remains unchanged when dedicated raw storage is not required.

- Performance impact: no extra storage for reusable in-memory FP32 codes; compressed-only configurations allocate raw FP32 storage only when explicitly requested.
- Concurrency/thread-safety impact: none; raw-vector access follows the existing resize lock.

## Documentation Impact

- [x] No docs update needed; this restores existing HGraph-compatible behavior on the maintenance branch.

## Risk and Rollback

- Risk level: medium.
- Rollback plan: revert the single commit.

## Checklist

- [x] I have linked the relevant issues.
- [x] I have added/updated tests for the behavior and bug fix.
- [x] I have updated docs if behavior/workflow changed.

### DIFF
--- a/src/algorithm/pyramid.cpp
+++ b/src/algorithm/pyramid.cpp
@@ -214,11 +214,16 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     const auto* data_ids = base->GetIds();
 
     resize(data_num);
-    std::memcpy(label_table_->label_table_.data(), data_ids, sizeof(LabelType) * data_num);
+    for (InnerIdType inner_id = 0; inner_id < data_num; ++inner_id) {
+        label_table_->Insert(inner_id, data_ids[inner_id]);
+    }
 
     base_codes_->BatchInsertVector(data_vectors, data_num);
     if (use_reorder_) {
         precise_codes_->BatchInsertVector(data_vectors, data_num);
+    }
+    if (create_new_raw_vector_) {
+        raw_vector_->BatchInsertVector(data_vectors, data_num);
     }
     auto codes = use_reorder_ ? precise_codes_ : base_codes_;
 
@@ -423,6 +428,9 @@ Pyramid::Serialize(StreamWriter& writer) const {
     if (use_reorder_) {
         precise_codes_->Serialize(writer);
     }
+    if (create_new_raw_vector_) {
+        raw_vector_->Serialize(writer);
+    }
     root_->Serialize(writer);
 
     // serialize footer (introduced since v0.15)
@@ -449,6 +457,9 @@ Pyramid::Deserialize(StreamReader& reader) {
     base_codes_->Deserialize(buffer_reader);
     if (use_reorder_) {
         precise_codes_->Deserialize(buffer_reader);
+    }
+    if (create_new_raw_vector_) {
+        raw_vector_->Deserialize(buffer_reader);
     }
     cur_element_count_ = base_codes_->TotalCount();
     root_->Deserialize(buffer_reader);
@@ -506,6 +517,10 @@ Pyramid::Add(const DatasetPtr& base) {
                 if (use_reorder_) {
                     precise_codes_->InsertVector(data_vectors + dim_ * i,
                                                  valid_id_count + local_cur_element_count);
+                }
+                if (create_new_raw_vector_) {
+                    raw_vector_->InsertVector(data_vectors + dim_ * i,
+                                              valid_id_count + local_cur_element_count);
                 }
                 valid_id_count++;
                 data_biases.push_back(i);
@@ -570,6 +585,9 @@ Pyramid::resize(int64_t new_max_capacity) {
     if (use_reorder_) {
         precise_codes_->Resize(new_max_capacity);
     }
+    if (create_new_raw_vector_) {
+        raw_vector_->Resize(new_max_capacity);
+    }
     points_mutex_->Resize(new_max_capacity);
     max_capacity_ = new_max_capacity;
 }
@@ -617,6 +635,13 @@ Pyramid::InitFeatures() {
         IndexFeature::SUPPORT_EXPORT_MODEL,
         IndexFeature::SUPPORT_GET_MEMORY_USAGE,
     });
+    const auto& retrieval_codes = use_reorder_ ? precise_codes_ : base_codes_;
+    const bool can_decode_exact_vector =
+        retrieval_codes->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_FP32 &&
+        (metric_ != MetricType::METRIC_TYPE_COSINE || retrieval_codes->HoldMolds());
+    if (can_decode_exact_vector || raw_vector_ != nullptr) {
+        this->index_feature_list_->SetFeature(IndexFeature::SUPPORT_GET_RAW_VECTOR_BY_IDS);
+    }
 }
 
 static const std::string HGRAPH_PARAMS_TEMPLATE =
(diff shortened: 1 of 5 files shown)


src/algorithm/pyramid.cpp
```diff
@@ -597,6 +615,13 @@ Pyramid::InitFeatures() {
         IndexFeature::SUPPORT_CLONE,
         IndexFeature::SUPPORT_GET_MEMORY_USAGE,
     });
+    const auto& retrieval_codes = use_reorder_ ? precise_codes_ : base_codes_;
```
LHT129, line 638: [suggestion] The `const auto&` reference bound to a ternary expression creates a reference to the selected `shared_ptr` member. While technically correct, using `const auto` (value copy) would be clearer and equally efficient since `shared_ptr` copy is cheap. This avoids a potentially surprising reference binding.

```cpp
// Current:
const auto& retrieval_codes = use_reorder_ ? precise_codes_ : base_codes_;

// Suggested:
const auto retrieval_codes = use_reorder_ ? precise_codes_ : base_codes_;
```

src/algorithm/pyramid.cpp
```diff
@@ -905,6 +951,52 @@ Pyramid::CalDistanceById(const float* query, const int64_t* ids, int64_t count)
...
+Pyramid::check_and_init_raw_vector(const FlattenInterfaceParamPtr& raw_vector_param,
+                                   const IndexCommonParam& common_param) {
+    if (raw_vector_param == nullptr) {
+        return;
+    }
+
+    const auto& retrieval_codes = use_reorder_ ? precise_codes_ : base_codes_;
+    const auto io_type_name = raw_vector_param->io_parameter->GetTypeName();
+    const bool is_memory_io =
+        io_type_name == IO_TYPE_VALUE_BLOCK_MEMORY_IO || io_type_name == IO_TYPE_VALUE_MEMORY_IO;
```
LHT129, line 1013: [note] When `retrieval_codes` are fp32 with memory IO, `raw_vector_` is set to share the same instance as the retrieval codes. This is an efficient zero-copy optimization. However, it means `raw_vector_` and `base_codes_`/`precise_codes_` point to the same object, and the serialization path correctly guards against double-serialization via `create_new_raw_vector_`. Future maintainers should be aware that `raw_vector_` may alias another codes pointer and should not be independently mutated.

LHT129 (commented):
Overall this is a well-structured PR that correctly separates raw-vector retrieval capability from the `store_raw_vector` data-retention flag, aligning Pyramid with HGraph semantics. The changes are surgical and well-tested.

**Summary of changes:**
- `inner_index_interface.cpp`: Moves the `has_raw_vector_` validation earlier to avoid unnecessary `inner_ids` allocation when the request will fail.
- `pyramid.cpp`/`pyramid.h`: Adds `GetVectorByInnerId` override, `raw_vector_`/`create_new_raw_vector_` members, and `check_and_init_raw_vector` for smart reuse of fp32 retrieval codes or dedicated raw storage.
- `InitFeatures()`: Conditionally advertises `SUPPORT_GET_RAW_VECTOR_BY_IDS` based on whether retrieval codes can decode exact vectors or raw storage is configured.
- `build_by_odescent`: Fixes label registration to use `LabelTable::Insert` instead of raw `memcpy`, fixing non-contiguous ID support (#2756).
- Tests: Comprehensive unit and functional tests covering all combinations of graph type, reorder, metric, quantization, and serialization round-trips.

**Key observations:**
1. The fp32 retrieval code reuse optimization (when `raw_vector_` aliases `base_codes_`/`precise_codes_`) is efficient but requires careful maintenance — the `create_new_raw_vector_` flag correctly guards serialization to avoid double-writing.
2. The `STORE_RAW_VECTOR` parameter mapping to three internal paths (base HoldMolds, precise HoldMolds, store_raw_vector key) is unusual but functionally correct.
3. The `InitFeatures()` logic correctly handles the cosine metric case where molds are required for exact vector decoding.

No blocking issues found. The PR description accurately reflects the changes and the test coverage is thorough.

src/algorithm/pyramid.cpp
jac0626: No code change: the selected member shared_ptr outlives this local use, and keeping a const reference avoids an unnecessary atomic shared_ptr refcount update.

src/algorithm/pyramid.cpp
jac0626: Acknowledged. The aliasing is intentional; create_new_raw_vector_ guards independent mutation and serialization, and the raw-vector roundtrip tests cover this path.

LHT129 (commented):
Thanks for this backport. The changes look correct and well-tested.

**Summary of review:**

- The `label_table_->Insert` loop in `build_by_odescent` correctly fixes the non-contiguous external ID issue (#2755).
- The `check_and_init_raw_vector` function properly reuses in-memory FP32 retrieval codes when possible, avoiding unnecessary storage.
- The `InitFeatures` logic correctly advertises `SUPPORT_GET_RAW_VECTOR_BY_IDS` only when exact vectors can be decoded or dedicated raw storage exists.
- The early-exit validation in `GetDataByIdsWithFlag` (moved before the `inner_ids` allocation) is a good optimization.
- `Serialize`/`Deserialize`/`Add`/`resize`/`Train` all correctly guard `raw_vector_` operations behind `create_new_raw_vector_`.
- Test coverage is comprehensive, covering both unit tests (pyramid_test.cpp) and functional tests (test_pyramid.cpp) across multiple graph types, reorder configurations, metrics, and quantization types.

One observation (not blocking): `ExportModel` does not propagate `raw_vector_` to the exported index. When `create_new_raw_vector_` is true, the exported model will lack raw vector data. If export-model is intended for search-only scenarios this is fine, but it may be worth noting in the documentation.

wxyucs (approved):
lgtm

PR #2763, opened 2026-08-25
feat(pyramid): support paths in GetDataByIdsWithFlag

### LINKED ISSUE #2754: [improve](pyramid): GetDataByIds support paths of pyramid
### Kind of improvement

refactor

### Affected area

other

### Summary

The data exported by the existing GetDataByIds function does not include the [Paths] of the pyramid index.

### Expected benefit

GetDataByIds support  [Paths] of the pyramid index.

### Benchmark evidence

_No response_

### References

_No response_

### Additional context

_No response_

jac0626:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Resolves: #2754

## What Changed

- Add the opt-in Pyramid parameter store_paths, disabled by default.
- Return retained paths only from GetDataByIdsWithFlag when DATA_FLAG_PATH is selected; plain GetDataByIds keeps its existing result.
- Keep storage isolated in PyramidPathStore. Each hierarchy uses direct dense SoA fields: uint64 offsets, uint16 counts, and a flat string vector, supporting 0/1/N paths without a map, sparse/adaptive policy, or path-tree metadata.
- Record only successfully allocated inner IDs across NSW, ODescent, Build Cache, and Add.
- Serialize the three fields directly with existing StreamWriter primitives. Regular serialization is parameter-driven; streaming uses the generic PYRAMID_PATHS TLV block.
- Cover default-off behavior, Build/Add/Clone/concurrent Add, named and incomplete hierarchies, regular/streaming serialization, and Build Cache.

## Test Evidence

- Pyramid unit tests: 207 assertions / 19 cases.
- Pyramid path functional tests: 271 assertions / 15 cases.

## Compatibility Impact

- API/ABI: additive store_paths parameter and DATA_FLAG_PATH constant.
- Default behavior: unchanged; no data-sized path storage and no path sidecar when disabled.
- Enabled serialization requires a reader configured with store_paths=true and its matching path payload.

Release A/B with 100K vectors, dim=4, NSW, store_paths=true, 7 alternating runs against the prior single-path implementation:

- Build: +1.27% at 1K unique paths; +1.86% at 100K unique paths.
- Peak build RSS: +1.16 MiB and +1.43 MiB respectively.
- Direct serialization versus the prior unmerged row-state draft: 1.90% and 0.46% faster.
- Direct dense serialized fields add 900,016 bytes at 100K slots.

Each enabled hierarchy has one shared mutex; one Writer holds the exclusive lock for a batch. Default-off allocates no per-vector path state.

## Documentation Impact

English and Chinese Pyramid, index-parameter, and API docs are updated.

## Risk and Rollback

- Risk level: medium.
- Rollback: revert the PR; default-off serialization is sidecar-free.

## Checklist

- [x] Linked issue
- [x] Tests added/updated
- [x] Compatibility reviewed
- [x] Documentation updated
- [x] Conventional commits and DCO trailers

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -153,7 +153,10 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
         for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
             const auto* paths = base->GetPaths(hierarchy_name);
             if (paths != nullptr) {
-                hierarchy->path_store->Record(paths, static_cast<uint64_t>(data_num));
+                auto writer = hierarchy->path_store->AcquireWriter();
+                for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
+                    writer.Insert(static_cast<InnerIdType>(offset), paths[offset]);
+                }
             }
         }
     }
@@ -1078,6 +1081,7 @@ Pyramid::Add(const DatasetPtr& base) {
     const auto* source_ids = base->GetSourceID();
     std::vector<int64_t> failed_ids;
     Vector<int64_t> data_biases(allocator_);
+    Vector<InnerIdType> accepted_inner_ids(allocator_);
     int64_t local_cur_element_count = 0;
     {
         std::lock_guard lock(cur_element_count_mutex_);
@@ -1104,6 +1108,9 @@ Pyramid::Add(const DatasetPtr& base) {
         }
 
         data_biases.reserve(data_num);
+        if (store_paths_) {
+            accepted_inner_ids.reserve(data_num);
+        }
         for (int64_t i = 0; i < data_num; ++i) {
             if (not label_table_->CheckLabel(data_ids[i])) {
                 const auto inner_id =
@@ -1113,6 +1120,9 @@ Pyramid::Add(const DatasetPtr& base) {
                     label_table_->InsertSourceId(inner_id, source_ids[i]);
                 }
                 data_biases.push_back(i);
+                if (store_paths_) {
+                    accepted_inner_ids.push_back(inner_id);
+                }
             } else {
                 logger::warn("Label {} already exists, skip adding.", data_ids[i]);
                 failed_ids.push_back(data_ids[i]);
@@ -1196,7 +1206,10 @@ Pyramid::Add(const DatasetPtr& base) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
             if (store_paths_) {
-                h_ptr->path_store->Record(hpath, data_biases, local_cur_element_count);
+                auto writer = h_ptr->path_store->AcquireWriter();
+                for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+                    writer.Insert(accepted_inner_ids[offset], hpath[data_biases[offset]]);
+                }
             }
             add_to_hierarchy(*h_ptr, data_vectors, hpath, data_biases, local_cur_element_count);
         }
@@ -2124,7 +2137,10 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
             if (store_paths_) {
-                h_ptr->path_store->Record(hpath, static_cast<uint64_t>(data_num));
+                auto writer = h_ptr->path_store->AcquireWriter();
+                for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
+                    writer.Insert(static_cast<InnerIdType>(offset), hpath[offset]);
+                }
             }
             populate_path_tree(*h_ptr, hpath, data_num);
         }
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -43,9 +43,6 @@ namespace vsag {
 
 const static float RADIUS_EPSILON = 1.1F;
 static constexpr uint64_t SOURCE_ID_TABLE_MAGIC = 0x534F555243454944ULL;  // SOURCEID
-static constexpr uint64_t PYRAMID_PATHS_MAGIC = 0x5059525041544853ULL;    // PYRPATHS
-static constexpr int64_t PYRAMID_PATHS_FORMAT_VERSION = 1;
-static constexpr const char* PYRAMID_PATHS_FORMAT_VERSION_KEY = "pyramid_paths_format_version";
 
 FilterPtr
 create_request_filter(const SearchRequest& request) {
@@ -640,7 +637,6 @@ Pyramid::Serialize(StreamWriter& writer) const {
     }
 
     if (store_paths_) {
-        StreamWriter::WriteObj(writer, PYRAMID_PATHS_MAGIC);
         serialize_paths(writer);
     }
 
@@ -648,9 +644,6 @@ Pyramid::Serialize(StreamWriter& writer) const {
     JsonType basic_info;
     basic_info["max_capacity"].SetInt(max_capacity_);
     basic_info[INDEX_PARAM].SetString(this->create_param_ptr_->ToString());
-    if (store_paths_) {
-        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].SetInt(PYRAMID_PATHS_FORMAT_VERSION);
-    }
     write_index_footer(writer, basic_info);
 }
 
@@ -995,15 +988,12 @@ Pyramid::Deserialize(StreamReader& reader) {
     }
     auto basic_info = metadata->Get(BASIC_INFO);
     auto max_capacity = basic_info["max_capacity"].GetInt();
-    const bool has_paths = basic_info.Contains(PYRAMID_PATHS_FORMAT_VERSION_KEY);
-    if (has_paths != store_paths_) {
+    auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
+    const bool serialized_store_paths = param_json.Contains(PYRAMID_STORE_PATHS_KEY) &&
+                                        param_json[PYRAMID_STORE_PATHS_KEY].GetBool();
+    if (serialized_store_paths != store_paths_) {
         throw VsagException(ErrorType::READ_ERROR,
-                            "serialized Pyramid paths metadata does not match config");
-    }
-    if (has_paths &&
-        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].GetInt() != PYRAMID_PATHS_FORMAT_VERSION) {
-        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
-                            "unsupported Pyramid paths format version");
+                            "serialized Pyramid store_paths does not match config");
     }
 
     const auto body_length = reader.Length() - footer->Length();
@@ -1024,7 +1014,6 @@ Pyramid::Deserialize(StreamReader& reader) {
     }
     cur_element_count_ = base_codes_->TotalCount();
 
-    auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
     if (param_json.Contains(PYRAMID_HIERARCHIES)) {
         uint64_t hierarchy_count = 0;
         StreamReader::ReadObj(buffer_reader, hierarchy_count);
@@ -1047,12 +1036,7 @@ Pyramid::Deserialize(StreamReader& reader) {
         h_iter->second->root->Deserialize(buffer_reader);
     }
 
-    if (has_paths) {
-        uint64_t magic = 0;
-        StreamReader::ReadObj(buffer_reader, magic);
-        if (magic != PYRAMID_PATHS_MAGIC) {
-            throw VsagException(ErrorType::READ_ERROR, "missing Pyramid paths marker");
-        }
+    if (store_paths_) {
         deserialize_paths(buffer_reader, static_cast<uint64_t>(cur_element_count_));
     }
 
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -43,6 +43,9 @@ namespace vsag {
 
 const static float RADIUS_EPSILON = 1.1F;
 static constexpr uint64_t SOURCE_ID_TABLE_MAGIC = 0x534F555243454944ULL;  // SOURCEID
+static constexpr uint64_t PYRAMID_PATHS_MAGIC = 0x5059525041544853ULL;    // PYRPATHS
+static constexpr int64_t PYRAMID_PATHS_FORMAT_VERSION = 1;
+static constexpr const char* PYRAMID_PATHS_FORMAT_VERSION_KEY = "pyramid_paths_format_version";
 
 FilterPtr
 create_request_filter(const SearchRequest& request) {
@@ -148,6 +151,14 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     }
     if (raw_vector_ != nullptr) {
         raw_vector_->BatchInsertVector(data_vectors, data_num);
+    }
+    if (store_paths_) {
+        for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+            const auto* paths = base->GetPaths(hierarchy_name);
+            if (paths != nullptr) {
+                hierarchy->path_store->Record(paths, static_cast<uint64_t>(data_num));
+            }
+        }
     }
     auto codes = has_precise_reorder() ? precise_codes_ : base_codes_;
 
@@ -628,10 +639,18 @@ Pyramid::Serialize(StreamWriter& writer) const {
         hierarchies_.at("")->root->Serialize(writer);
     }
 
+    if (store_paths_) {
+        StreamWriter::WriteObj(writer, PYRAMID_PATHS_MAGIC);
+        serialize_paths(writer);
+    }
+
     // serialize footer (introduced since v0.15)
     JsonType basic_info;
     basic_info["max_capacity"].SetInt(max_capacity_);
     basic_info[INDEX_PARAM].SetString(this->create_param_ptr_->ToString());
+    if (store_paths_) {
+        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].SetInt(PYRAMID_PATHS_FORMAT_VERSION);
+    }
     write_index_footer(writer, basic_info);
 }
 
@@ -715,6 +734,13 @@ Pyramid::collect_streaming_header() const {
                                  hierarchy_tag,
                                  StreamSerializationBlockCurrentVersion(hierarchy_tag),
                                  StreamSerializationTagCritical(hierarchy_tag));
+    if (store_paths_) {
+        const auto path_tag = static_cast<uint32_t>(StreamSerializationTag::PYRAMID_PATHS);
+        AppendStreamingManifestBlock(manifest,
+                                     path_tag,
+                                     StreamSerializationBlockCurrentVersion(path_tag),
+                                     StreamSerializationTagCritical(path_tag));
+    }
     metadata->Set("block_manifest", manifest);
     metadata->SetEmptyIndex(this->GetNumElements() == 0);
     return metadata;
@@ -767,6 +793,13 @@ Pyramid::serialize_streaming_body(StreamWriter& writer) const {
                         hierarchy_tag,
                         StreamSerializationTagCritical(hierarchy_tag),
                         [this](StreamWriter& w) { this->serialize_hierarchies(w); });
+    if (store_paths_) {
+        const auto path_tag = static_cast<uint32_t>(StreamSerializationTag::PYRAMID_PATHS);
+        WriteStreamingBlock(
+            writer, path_tag, StreamSerializationTagCritical(path_tag), [this](StreamWriter& w) {
+                this->serialize_paths(w);
+            });
+    }
 }
 
 void
@@ -829,6 +862,7 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
     bool loaded_precise_codes = false;
     bool loaded_raw_vector = false;
     bool loaded_hierarchies = false;
+    bool loaded_paths = false;
 
     while (true) {
         auto block_header = StreamBlockHeader::Read(reader);
@@ -895,6 +929,20 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
                     });
                 loaded_hierarchies = true;
                 break;
+            case StreamSerializationTag::PYRAMID_PATHS:
+                if (loaded_paths) {
+                    throw VsagException(ErrorType::READ_ERROR,
+                                        "duplicate Pyramid streaming paths block");
+                }
+                if (not loaded_base_codes) {
+                    throw VsagException(ErrorType::READ_ERROR,
+                                        "Pyramid streaming paths block precedes base codes");
+                }
+                ReadSeekableBlockPayload(block_reader, block_header, [this](StreamReader& block) {
+                    this->deserialize_paths(block, static_cast<uint64_t>(this->cur_element_count_));
+                });
+                loaded_paths = true;
+                break;
             default:
                 if (block_header.IsCritical()) {
                     throw VsagException(
@@ -925,6 +973,10 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
         throw VsagException(ErrorType::READ_ERROR,
                             "Pyramid streaming serialization raw vector block is missing");
     }
+    if (store_paths_ && !loaded_paths) {
+        throw VsagException(ErrorType::READ_ERROR,
+                            "Pyramid streaming serialization paths block is missing");
+    }
 
     resize(max_capacity);
     this->current_memory_usage_ = static_cast<int64_t>(this->CalSerializeSize());
@@ -933,14 +985,31 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
-    JsonType basic_info;
-    if (not read_index_footer(reader, basic_info)) {
+    const auto footer = Footer::Parse(reader);
+    if (footer == nullptr) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
+    const auto metadata = footer->GetMetadata();
+    if (metadata == nullptr || metadata->EmptyIndex()) {
+        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
+    }
+    auto basic_info = metadata->Get(BASIC_INFO);
     auto max_capacity = basic_info["max_capacity"].GetInt();
+    const bool has_paths = basic_info.Contains(PYRAMID_PATHS_FORMAT_VERSION_KEY);
+    if (has_paths != store_paths_) {
+        throw VsagException(ErrorType::READ_ERROR,
+                            "serialized Pyramid paths metadata does not match config");
+    }
+    if (has_paths &&
+        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].GetInt() != PYRAMID_PATHS_FORMAT_VERSION) {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "unsupported Pyramid paths format version");
+    }
 
+    const auto body_length = reader.Length() - footer->Length();
+    auto body_reader = reader.Slice(0, body_length);
     BufferStreamReader buffer_reader(
-        &reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
+        &body_reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
 
     label_table_->Deserialize(buffer_reader);
     deserialize_source_id_table(buffer_reader);
@@ -976,6 +1045,15 @@ Pyramid::Deserialize(StreamReader& reader) {
             h_iter != hierarchies_.end(),
             "deserialized single-hierarchy index but current config has named hierarchies");
         h_iter->second->root->Deserialize(buffer_reader);
+    }
+
+    if (has_paths) {
+        uint64_t magic = 0;
+        StreamReader::ReadObj(buffer_reader, magic);
+        if (magic != PYRAMID_PATHS_MAGIC) {
+            throw VsagException(ErrorType::READ_ERROR, "missing Pyramid paths marker");
+        }
+        deserialize_paths(buffer_reader, static_cast<uint64_t>(cur_element_count_));
     }
 
     resize(max_capacity);
@@ -1133,6 +1211,9 @@ Pyramid::Add(const DatasetPtr& base) {
     for (const auto& [hname, h_ptr] : hierarchies_) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
+            if (store_paths_) {
+                h_ptr->path_store->Record(hpath, data_biases, local_cur_element_count);
+            }
             add_to_hierarchy(*h_ptr, data_vectors, hpath, data_biases, local_cur_element_count);
         }
     }
@@ -1202,6 +1283,9 @@ Pyramid::InitFeatures() {
         IndexFeature::SUPPORT_EXPORT_MODEL,
         IndexFeature::SUPPORT_GET_MEMORY_USAGE,
     });
+    if (store_paths_) {
+        this->index_feature_list_->SetFeature(IndexFeature::SUPPORT_GET_DATA_BY_IDS);
+    }
     if (has_raw_vector_) {
         this->index_feature_list_->SetFeature(IndexFeature::SUPPORT_GET_RAW_VECTOR_BY_IDS);
     }
@@ -1287,7 +1371,8 @@ static const std::string HGRAPH_PARAMS_TEMPLATE =
         "{EF_CONSTRUCTION_KEY}": 400,
         "{NO_BUILD_LEVELS}":[],
         "{INDEX_MIN_SIZE}": 0,
-        "{SUPPORT_DUPLICATE}": false
+        "{SUPPORT_DUPLICATE}": false,
+        "{PYRAMID_STORE_PATHS_KEY}": false
     })";
 
 ParamPtr
@@ -1334,6 +1419,7 @@ Pyramid::CheckAndMappingExternalParam(const JsonType& external_param,
         {PYRAMID_NO_BUILD_LEVELS, {NO_BUILD_LEVELS}},
         {PYRAMID_HIERARCHIES, {PYRAMID_HIERARCHIES}},
         {PYRAMID_PERSIST_SOURCE_ID, {PYRAMID_PERSIST_SOURCE_ID_KEY}},
+        {PYRAMID_STORE_PATHS, {PYRAMID_STORE_PATHS_KEY}},
         {PYRAMID_BASE_PQ_DIM,
          {BASE_CODES_KEY, QUANTIZATION_PARAMS_KEY, PRODUCT_QUANTIZATION_DIM_KEY}},
         {PYRAMID_BASE_FILE_PATH, {BASE_CODES_KEY, IO_PARAMS_KEY, IO_FILE_PATH_KEY}},
@@ -2053,6 +2139,9 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
     for (const auto& [hname, h_ptr] : hierarchies_) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
+            if (store_paths_) {
+                h_ptr->path_store->Record(hpath, static_cast<uint64_t>(data_num));
+            }
             populate_path_tree(*h_ptr, hpath, data_num);
         }
     }
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -154,6 +154,7 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
             const auto* paths = base->GetPaths(hierarchy_name);
             if (paths != nullptr) {
                 auto writer = hierarchy->path_store->AcquireWriter();
+                writer.Prepare(static_cast<uint64_t>(data_num), static_cast<uint64_t>(data_num));
                 for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
                     writer.Insert(static_cast<InnerIdType>(offset), paths[offset]);
                 }
@@ -1205,8 +1206,10 @@ Pyramid::Add(const DatasetPtr& base) {
     for (const auto& [hname, h_ptr] : hierarchies_) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
-            if (store_paths_) {
+            if (store_paths_ and not accepted_inner_ids.empty()) {
                 auto writer = h_ptr->path_store->AcquireWriter();
+                writer.Prepare(static_cast<uint64_t>(accepted_inner_ids.back()) + 1,
+                               data_biases.size());
                 for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
                     writer.Insert(accepted_inner_ids[offset], hpath[data_biases[offset]]);
                 }
@@ -2138,6 +2141,7 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
         if (hpath != nullptr) {
             if (store_paths_) {
                 auto writer = h_ptr->path_store->AcquireWriter();
+                writer.Prepare(static_cast<uint64_t>(data_num), static_cast<uint64_t>(data_num));
                 for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
                     writer.Insert(static_cast<InnerIdType>(offset), hpath[offset]);
                 }
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -982,15 +982,10 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
-    const auto footer = Footer::Parse(reader);
-    if (footer == nullptr) {
+    JsonType basic_info;
+    if (not read_index_footer(reader, basic_info)) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
-    const auto metadata = footer->GetMetadata();
-    if (metadata == nullptr || metadata->EmptyIndex()) {
-        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
-    }
-    auto basic_info = metadata->Get(BASIC_INFO);
     auto max_capacity = basic_info["max_capacity"].GetInt();
     auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
     const bool serialized_store_paths = param_json.Contains(PYRAMID_STORE_PATHS_KEY) &&
@@ -1000,10 +995,8 @@ Pyramid::Deserialize(StreamReader& reader) {
                             "serialized Pyramid store_paths does not match config");
     }
 
-    const auto body_length = reader.Length() - footer->Length();
-    auto body_reader = reader.Slice(0, body_length);
     BufferStreamReader buffer_reader(
-        &body_reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
+        &reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
 
     label_table_->Deserialize(buffer_reader);
     deserialize_source_id_table(buffer_reader);
--- a/src/algorithm/pyramid/pyramid_path_store.cpp
+++ b/src/algorithm/pyramid/pyramid_path_store.cpp
@@ -1,0 +1,153 @@
+// Copyright 2024-present the vsag project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "pyramid_path_store.h"
+
+#include <limits>
+#include <mutex>
+
+#include "common.h"
+#include "storage/stream_reader.h"
+#include "storage/stream_writer.h"
+#include "vsag_exception.h"
+
+namespace vsag {
+
+std::string
+ReadPyramidPathString(StreamReader& reader) {
+    uint64_t length = 0;
+    StreamReader::ReadObj(reader, length);
+    const auto cursor = reader.GetCursor();
+    const auto reader_length = reader.Length();
+    if (cursor > reader_length || length > reader_length - cursor) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path string length");
+    }
+    std::string value(length, '\0');
+    if (length > 0) {
+        reader.Read(value.data(), length);
+    }
+    return value;
+}
+
+void
+PyramidPathStore::Record(const std::string* paths,
+                         const Vector<int64_t>& data_biases,
+                         int64_t first_inner_id) {
+    CHECK_ARGUMENT(paths != nullptr, "paths must not be null");
+    CHECK_ARGUMENT(first_inner_id >= 0, "first inner id must not be negative");
+    if (data_biases.empty()) {
+        return;
+    }
+
+    const auto begin = static_cast<uint64_t>(first_inner_id);
+    CHECK_ARGUMENT(begin <= std::numeric_limits<uint64_t>::max() - data_biases.size(),
+                   "Pyramid path count overflow");
+    const auto end = begin + data_biases.size();
+    std::unique_lock lock(mutex_);
+    if (paths_by_inner_id_.size() < end) {
+        paths_by_inner_id_.resize(end);
+        has_path_.resize(end, 0);
+    }
+    for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+        const auto inner_id = begin + offset;
+        paths_by_inner_id_[inner_id] = paths[data_biases[offset]];
+        has_path_[inner_id] = 1;
+    }
+}
+
+void
+PyramidPathStore::Record(const std::string* paths, uint64_t count) {
+    CHECK_ARGUMENT(paths != nullptr, "paths must not be null");
+    std::unique_lock lock(mutex_);
+    paths_by_inner_id_.resize(count);
+    has_path_.resize(count, 0);
+    for (uint64_t inner_id = 0; inner_id < count; ++inner_id) {
+        paths_by_inner_id_[inner_id] = paths[inner_id];
+        has_path_[inner_id] = 1;
+    }
+}
+
+bool
+PyramidPathStore::GetPaths(const Vector<InnerIdType>& inner_ids, std::string* paths) const {
+    std::shared_lock lock(mutex_);
+    for (uint64_t offset = 0; offset < inner_ids.size(); ++offset) {
+        const auto inner_id = static_cast<uint64_t>(inner_ids[offset]);
+        if (inner_id >= has_path_.size() || has_path_[inner_id] == 0) {
+            return false;
+        }
+        paths[offset] = paths_by_inner_id_[inner_id];
+    }
+    return true;
+}
+
+void
+PyramidPathStore::Serialize(StreamWriter& writer) const {
+    std::shared_lock lock(mutex_);
+    if (paths_by_inner_id_.size() != has_path_.size()) {
+        throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store size mismatch");
+    }
+    StreamWriter::WriteObj(writer, static_cast<uint64_t>(paths_by_inner_id_.size()));
+    for (uint64_t inner_id = 0; inner_id < has_path_.size(); ++inner_id) {
+        const auto present = has_path_[inner_id];
+        if (present > 1) {
+            throw VsagException(ErrorType::INTERNAL_ERROR,
+                                "Pyramid path store has invalid presence value");
+        }
+        StreamWriter::WriteObj(writer, present);
+        if (present != 0) {
+            StreamWriter::WriteString(writer, paths_by_inner_id_[inner_id]);
+        }
+    }
+}
+
+void
+PyramidPathStore::Deserialize(StreamReader& reader, uint64_t max_count) {
+    uint64_t slot_count = 0;
+    StreamReader::ReadObj(reader, slot_count);
+    if (slot_count > max_count) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path slot count");
+    }
+    const auto cursor = reader.GetCursor();
+    const auto reader_length = reader.Length();
+    if (cursor > reader_length || slot_count > reader_length - cursor) {
+        throw VsagException(ErrorType::READ_ERROR,
+                            "corrupted Pyramid path slots exceed remaining payload");
+    }
+
+    Vector<std::string> restored_paths(slot_count, std::string{}, allocator_);
+    Vector<uint8_t> restored_has_path(slot_count, 0, allocator_);
+    for (uint64_t inner_id = 0; inner_id < slot_count; ++inner_id) {
+        uint8_t present = 0;
+        StreamReader::ReadObj(reader, present);
+        if (present > 1) {
+            throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path presence value");
+        }
+        restored_has_path[inner_id] = present;
+        if (present != 0) {
+            restored_paths[inner_id] = ReadPyramidPathString(reader);
+        }
+    }
+
+    std::unique_lock lock(mutex_);
+    paths_by_inner_id_.swap(restored_paths);
+    has_path_.swap(restored_has_path);
+}
+
+uint64_t
+PyramidPathStore::Size() const {
+    std::shared_lock lock(mutex_);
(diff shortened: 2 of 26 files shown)


src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+    CHECK_ARGUMENT(count >= 0, "count must not be negative");
+    auto result = InnerIndexInterface::GetDataByIds(ids, count);
+    if (not store_paths_) {
+        return result;
+    }
+
+    Vector<InnerIdType> inner_ids(static_cast<uint64_t>(count), allocator_);
+    {
+        std::shared_lock lock(label_lookup_mutex_);
+        for (int64_t offset = 0; offset < count; ++offset) {
```
LHT129: [suggestion] The `GetDataByIds` override performs a redundant label-to-inner-ID lookup. The base class `InnerIndexInterface::GetDataByIds` (called on line 215) already resolves labels to inner IDs internally. This override then acquires `label_lookup_mutex_` and calls `label_table_->GetIdByLabel` again for every requested ID (lines 222-226), doubling the lookup cost.

Consider either:
- Exposing the inner IDs from the base class result so they can be reused here, or
- Overriding the label resolution path so the inner IDs are obtained once and shared between the base result construction and the path lookup.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+uint64_t
+PyramidPathStore::Size() const {
+    std::shared_lock lock(mutex_);
+    return paths_by_inner_id_.size();
+}
+
+void
+Pyramid::serialize_paths(StreamWriter& writer) const {
```
LHT129: [note] `Pyramid::serialize_paths`, `Pyramid::deserialize_paths`, `Pyramid::validate_paths`, and `Pyramid::GetDataByIds` are defined in `pyramid_path_store.cpp` rather than `pyramid.cpp`. This creates a circular include dependency: `pyramid.h` includes `pyramid_path_store.h`, and `pyramid_path_store.cpp` includes `pyramid.h`.

While this works due to include guards, it is unconventional for class methods to be defined outside the class's primary implementation file. Consider moving these method definitions into `pyramid.cpp` to keep the class implementation in one place and avoid the circular include.

src/storage/serialization_tags.h
```diff
@@ -109,6 +112,7 @@ StreamSerializationTagCritical(uint32_t tag) {
         case StreamSerializationTag::EXTRA_INFO:
         case StreamSerializationTag::RAW_VECTOR:
         case StreamSerializationTag::IVF_BUCKET_GRAPH:
+        case StreamSerializationTag::PYRAMID_PATHS:
```
LHT129, line 120: The  streaming tag is registered as **non-critical** in  ( returns ), but  in  treats it as **critical** when  is true — it throws  if the block is missing (line ~976: ).

This creates two problems:
1. **Forward compatibility**: An older reader that doesn't know about tag 18 would silently skip the block (per the non-critical contract), but a newer reader with  will fail if the block was never written.
2. **Semantic inconsistency**: The tag's declared criticality doesn't match its actual treatment.

If the paths block is truly required for correctness when  is enabled, the tag should be marked critical in . Otherwise, the deserialization logic should gracefully handle the absence of the block (e.g., by leaving path stores empty and logging a warning).

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +988,27 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
-    JsonType basic_info;
-    if (not read_index_footer(reader, basic_info)) {
+    const auto footer = Footer::Parse(reader);
```
LHT129: The `Deserialize` rewrite to use `Footer::Parse` introduces a regression: it now unconditionally throws `INDEX_EMPTY` when the footer metadata reports `EmptyIndex() == true`. Previously, `read_index_footer` would return `false` for legacy formats (no footer), and empty indexes could roundtrip through serialization.

The `collect_streaming_header` method sets `EmptyIndex` based on `GetNumElements() == 0`, meaning a freshly built empty index will serialize with an empty footer. When that same serialized blob is deserialized, `Deserialize` will throw `INDEX_EMPTY` instead of successfully restoring the empty index.

Consider either:
- Allowing empty index deserialization to succeed (initialize structures to empty state and return), or
- Not setting `EmptyIndex` when `store_paths_` is true (since paths need to be validated even for empty indexes), or
- Adding a dedicated empty-index path that skips path deserialization but still initializes the index correctly.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+        }
+        if (hierarchy->second->path_store == nullptr) {
+            throw VsagException(ErrorType::READ_ERROR, "Pyramid path storage is disabled");
+        }
+        hierarchy->second->path_store->Deserialize(reader, max_count);
+    }
+}
+
+void
+Pyramid::validate_paths(uint64_t max_count) const {
```
LHT129: The `validate_paths` check `hierarchy->path_store == nullptr` is unreachable when `store_paths_` is true, because `Hierarchy` unconditionally creates the `path_store` in its constructor when `store_paths_` is set (see `pyramid.h`). The `||` short-circuits to the `Size() > max_count` check, making the nullptr branch dead code.

If the intent is defensive validation against a corrupted or partially-initialized index, the check should also verify that `path_store` is non-null even when `store_paths_` is false (i.e., a path_store should never exist when paths are disabled). Currently, a nullptr path_store when `store_paths_` is false silently passes validation, which may hide serialization format mismatches.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +988,27 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
+    if (footer == nullptr) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
+    const auto metadata = footer->GetMetadata();
+    if (metadata == nullptr || metadata->EmptyIndex()) {
```
LHT129: There are two separate `store_paths` compatibility checks in `Deserialize` that partially overlap:

1. **Line ~996**: Parses the serialized `INDEX_PARAM` string into `PyramidParameters` and compares `serialized_param->store_paths != store_paths_`. This catches parameter-level mismatches.
2. **Line ~1049**: Checks `basic_info.Contains(PYRAMID_PATHS_FORMAT_VERSION_KEY)` against `store_paths_`. This catches footer metadata format mismatches.

In practice, these two checks will always agree (if the param says `store_paths=true`, the footer will contain the key). Having both is redundant and could produce confusing error messages if they somehow diverge. Consider consolidating into a single check — the footer metadata check is sufficient since it directly reflects what was actually serialized, while the param check depends on correct param serialization.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Addressed by splitting responsibilities: `pyramid_path_store.cpp` now contains only `PyramidPathStore`, while the `Pyramid` integration methods live in the dedicated `pyramid_paths.cpp` translation unit.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Fixed. `InnerIndexInterface` now resolves labels once and reuses the same `inner_ids` for vector, attribute, and extra-info retrieval; `Pyramid` receives that vector from the shared helper and uses it for path lookup.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+            continue;
+        }
+        if (hierarchy_name.empty()) {
+            result->Paths(paths.release());
+        } else {
+            result->Paths(hierarchy_name, paths.release());
+        }
+    }
+    return result;
+}
```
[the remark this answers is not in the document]
jac0626: Fixed by construction. The new shared helper creates both the returned `Dataset` and the resolved `inner_ids` from the same request. Missing labels fail during that single resolution step, so the result and attached paths always share the same cardinality and order.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+void
+Pyramid::deserialize_paths(StreamReader& reader, uint64_t max_count) {
+    uint64_t hierarchy_count = 0;
+    StreamReader::ReadObj(reader, hierarchy_count);
+    if (hierarchy_count != hierarchies_.size()) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path hierarchy count");
+    }
```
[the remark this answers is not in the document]
jac0626: Addressed. `PyramidPathStore` remains isolated in its own implementation file, and all `Pyramid` path serialization/retrieval methods have moved to a dedicated `pyramid_paths.cpp` file.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+    const auto reader_length = reader.Length();
+    if (cursor > reader_length || length > reader_length - cursor) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path string length");
+    }
+    std::string value(length, '\0');
+    if (length > 0) {
+        reader.Read(value.data(), length);
+    }
+    return value;
+}
```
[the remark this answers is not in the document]
jac0626: I kept a dedicated bounded reader intentionally: the shared `StreamReader::ReadString` allocates according to the serialized length before validating the remaining payload. The path reader now centralizes the safe check, documents the reason, and has an oversized-length regression test.

src/storage/serialization_tags.h
jac0626: The non-critical tag is intentional. An older reader may skip this optional capability and retain search functionality. A reader explicitly configured with `store_paths=true` must require the sidecar, because silently returning incomplete paths would be incorrect. I added a code comment documenting this contract; it matches the optional `RAW_VECTOR` pattern.

tests/test_pyramid_paths.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+TEST_CASE("Pyramid does not retain paths by default", "[ft][pyramid][paths]") {
+    auto index = MakePyramidIndex("nsw", false);
+    REQUIRE_FALSE(index->CheckFeature(vsag::IndexFeature::SUPPORT_GET_DATA_BY_IDS));
+    auto base = MakeDataset({10, 20}, {{"", {"a", "b"}}});
+
+    REQUIRE(index->Build(base).has_value());
+    auto data = GetData(index, {20, 10});
+
+    REQUIRE(data->GetPaths() == nullptr);
```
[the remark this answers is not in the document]
jac0626: Clarified in the test. `CheckFeature` is capability discovery rather than a runtime invocation gate. To preserve compatibility, the default capability remains disabled, while the inherited legacy direct call remains callable and returns data without paths when storage is disabled.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+Pyramid::serialize_paths(StreamWriter& writer) const {
+    StreamWriter::WriteObj(writer, static_cast<uint64_t>(hierarchies_.size()));
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+        if (hierarchy->path_store == nullptr) {
+            throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store is missing");
+        }
+        StreamWriter::WriteString(writer, hierarchy_name);
+        hierarchy->path_store->Serialize(writer);
+    }
+}
```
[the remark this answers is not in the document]
jac0626: Addressed. `pyramid_path_store.cpp` now focuses only on the sidecar container; serialization, deserialization, and retrieval integration for `Pyramid` are in `pyramid_paths.cpp`.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include "pyramid_path_store.h"
+
+#include <fmt/format.h>
+
+#include <limits>
+#include <memory>
+#include <mutex>
+
+#include "algorithm/pyramid/pyramid.h"
```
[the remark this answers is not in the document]
jac0626: Addressed. The store implementation is now self-contained, and the `Pyramid` member implementations were moved into the dedicated `pyramid_paths.cpp` translation unit.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +988,27 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
+        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
+    }
```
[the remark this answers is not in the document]
jac0626: No behavioral regression was introduced here: the previous `read_index_footer` helper already rejected footer metadata marked as empty, while the public serialization wrappers retain their existing empty-index handling. I added `store_paths=true` empty BinarySet and streaming roundtrip regressions, and both pass.

src/algorithm/pyramid/pyramid.cpp
jac0626: The previous `read_index_footer` implementation already threw `INDEX_EMPTY` when footer metadata reported an empty index, so replacing the helper did not introduce that error path. Public empty-index serialization/deserialization behavior is preserved, and new `store_paths=true` BinarySet plus streaming roundtrip tests pass.

src/algorithm/pyramid/pyramid.cpp
jac0626: Addressed. Regular deserialization now performs one early compatibility check using the physical paths format key and version before mutating the index. The redundant serialized-parameter parse and later duplicate check were removed.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Addressed by removing the redundant `validate_paths` layer. `deserialize_paths` and `PyramidPathStore::Deserialize` already enforce hierarchy count, known unique hierarchy names, enabled stores, and bounded slot counts at the point where data is restored.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::unique_lock lock(mutex_);
+    if (paths_by_inner_id_.size() < end) {
+        paths_by_inner_id_.resize(end);
+        has_path_.resize(end, 0);
+    }
+    for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+        const auto inner_id = begin + offset;
+        paths_by_inner_id_[inner_id] = paths[data_biases[offset]];
+        has_path_[inner_id] = 1;
+    }
```
LHT129, line 183: [suggestion] `data_biases[offset]` is `int64_t` and is used as an index into `paths[...]` without validating it is non-negative. A negative value would cause undefined behavior (out-of-bounds access). The `data_biases` vector is populated from accepted input offsets in `Pyramid::Add`, but there is no explicit guarantee that all values are non-negative at this call site. Consider adding a bounds check: `CHECK_ARGUMENT(data_biases[offset] >= 0, "data bias must not be negative");`. If the input paths array length is available, also validate `data_biases[offset] < path_count`.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+        if (hierarchy->path_store == nullptr) {
+            continue;
+        }
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
+            continue;
+        }
```
LHT129, line 84: [suggestion] This allocates an output path array (`std::make_unique<std::string[]>`) for every hierarchy before knowing whether `GetPaths` will succeed. `GetPaths` can fail early (return `false`) on the first missing ID. For large `count` and multiple hierarchies, this causes avoidable heap allocations.

Consider adding a lightweight completeness probe to `PyramidPathStore` (e.g., `HasAll(inner_ids)`) that only checks the `has_path_` bitset without copying strings, and only allocate the output buffer when the probe passes. Alternatively, refactor `GetPaths` to support a check-only mode before allocating the destination buffer.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +985,31 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
```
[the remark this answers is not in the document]
jac0626: No code change: the exact empty-index BinarySet and streaming roundtrip test passes both sections after the rebase, so the public behavior is already successful and this guard does not contradict the tested path.

src/algorithm/pyramid/pyramid_path_store.cpp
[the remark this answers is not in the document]
jac0626: No code change: data_biases is an internal list populated only from the non-negative input loop offsets accepted by Pyramid::Add, so a negative or out-of-range user-controlled index cannot reach this call.

src/algorithm/pyramid/pyramid_paths.cpp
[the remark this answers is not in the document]
jac0626: No code change: this is an optional allocation optimization rather than a correctness issue; keeping the implementation minimal avoids adding a second completeness API or check-only mode.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: No code change: data_biases is an internal list populated only from the non-negative input loop offsets accepted by Pyramid::Add, so a negative or out-of-range user-controlled index cannot reach this call.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: No code change: this is an optional allocation optimization rather than a correctness issue; keeping the implementation minimal avoids adding a second completeness API or check-only mode.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,93 @@
...
+        hierarchy->second->path_store->Deserialize(reader, max_count);
+    }
+}
+
+DatasetPtr
+Pyramid::GetDataByIds(const int64_t* ids, int64_t count) const {
+    Vector<InnerIdType> inner_ids(allocator_);
+    auto result = get_data_by_ids(ids, count, inner_ids);
+    if (not store_paths_) {
+        return result;
```
LHT129, line 74: [note] When `store_paths_` is true, the `Pyramid` constructor unconditionally creates a `PyramidPathStore` for every hierarchy (see `pyramid.h` lines 91 and 104). Therefore the `path_store == nullptr` check on line 73 is dead code for the `store_paths_ == true` path — it can never be null. The `continue` is harmless but misleading; a reader might think nullptr is a possible state.

Consider replacing with an assertion that documents the invariant: `assert(hierarchy->path_store != nullptr)`. Same applies to the check in `serialize_paths` at line 30.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,153 @@
...
+    CHECK_ARGUMENT(paths != nullptr, "paths must not be null");
+    std::unique_lock lock(mutex_);
+    paths_by_inner_id_.resize(count);
+    has_path_.resize(count, 0);
+    for (uint64_t inner_id = 0; inner_id < count; ++inner_id) {
+        paths_by_inner_id_[inner_id] = paths[inner_id];
+        has_path_[inner_id] = 1;
+    }
+}
+
```
LHT129, line 190: [note] `PyramidPathStore::Record(const std::string* paths, uint64_t count)` unconditionally resizes both `paths_by_inner_id_` and `has_path_` to `count`, which discards any previously recorded paths without warning. This overload is used during full `Build` (both `build_by_odescent` and `build_with_cache`), where the store is expected to be empty. However, there is no assertion or check that the store is actually empty before overwriting.

Consider adding a `CHECK_ARGUMENT` or debug assertion that `paths_by_inner_id_.empty()` before the resize, or documenting the pre-condition explicitly in the header.

src/algorithm/pyramid/pyramid_path_store.h
```diff
@@ -0,0 +1,62 @@
...
+
+class StreamReader;
+class StreamWriter;
+
+// Reads a serialized path string only after verifying that its payload fits in the reader.
+std::string
+ReadPyramidPathString(StreamReader& reader);
+
+class PyramidPathStore {
+public:
```
LHT129, line 35: [note] `ReadPyramidPathString` is declared in the public header `pyramid_path_store.h` but is only used within `pyramid_path_store.cpp` (by `Deserialize`) and `pyramid_paths.cpp` (by `deserialize_paths`). Exposing it in the header expands the API surface unnecessarily.

Consider either:
- Moving the declaration into a private header or an anonymous namespace in the `.cpp` files, or
- Making it a static/namespace-scoped function in a shared internal header if both translation units genuinely need it.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1133,6 +1211,9 @@ Pyramid::Add(const DatasetPtr& base) {
     for (const auto& [hname, h_ptr] : hierarchies_) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
+            if (store_paths_) {
+                h_ptr->path_store->Record(hpath, data_biases, local_cur_element_count);
+            }
             add_to_hierarchy(*h_ptr, data_vectors, hpath, data_biases, local_cur_element_count);
```
LHT129, line 1210: [note] In `Pyramid::Add`, `h_ptr->path_store->Record(...)` is called on line 1217 **before** `add_to_hierarchy(...)` on line 1219. If `add_to_hierarchy` throws (e.g. due to graph insertion failure), the path store has already been mutated but the point was never actually added to the index. This leaves the path store inconsistent with the actual graph state — subsequent `GetDataByIds` calls could return stale paths for inner IDs that don't correspond to valid points.

Consider moving the `Record` call to after `add_to_hierarchy` succeeds, or wrapping both in a try/catch that rolls back the path store on failure.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Addressed in eb47deef. The retrieval path now treats a missing per-hierarchy store as INTERNAL_ERROR instead of silently continuing, matching the invariant already enforced by serialization.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Addressed in 6f935847. The overload is now explicitly documented as the full-build operation whose precondition is an empty store; the incremental overload remains the only operation used after initial Build.

src/algorithm/pyramid/pyramid_path_store.h
jac0626: No code change. This header is internal under src/algorithm/pyramid, not part of the public include/vsag API, and the bounds-checked reader is intentionally shared by the store deserializer and the hierarchy wrapper deserializer. A second private header or duplicated reader would add structure without reducing the exported public surface.

src/algorithm/pyramid/pyramid.cpp
jac0626: No code change. By this point Add has already inserted the accepted labels and codes and advanced cur_element_count_; GetDataByIds is keyed to that stored-data state, not to graph searchability. Recording paths before per-hierarchy graph insertion keeps the sidecar aligned with the label/code table. Moving it after graph insertion would instead leave accepted IDs without paths if graph insertion throws partway through; Add is not transactional across those existing components.

src/algorithm/pyramid/pyramid_path_store.cpp
LHT129, line 183: [suggestion] `data_biases[offset]` is `int64_t` and is used as an index into `paths[...]` without validating it is non-negative. A negative value would cause undefined behavior (out-of-bounds access). The `data_biases` vector is populated from accepted input offsets in `Pyramid::Add`, but there is no explicit guarantee that all values are non-negative at this call site.

Consider adding a bounds check: `CHECK_ARGUMENT(data_biases[offset] >= 0, "data bias must not be negative");`

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+        if (hierarchy->path_store == nullptr) {
+            throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store is missing");
+        }
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
+            continue;
+        }
```
LHT129, line 84: [suggestion] This allocates an output path array for every hierarchy before knowing whether `GetPaths(...)` will succeed (and `GetPaths` can fail early on missing IDs). For large `count` and multiple hierarchies, this can cause avoidable allocations.

A concrete improvement is to add a lightweight completeness probe on `PyramidPathStore` (e.g., `HasAll(inner_ids)`) and only allocate/copy when it is true, or refactor `GetPaths` to support a "check-only" mode before allocating the destination buffer.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,101 @@
...
+        CHECK_ARGUMENT(store_paths_, "store_paths is false");
+    }
+
+    Vector<InnerIdType> inner_ids(allocator_);
+    auto result = get_data_by_ids_with_flag(ids, count, selected_data_flag, inner_ids);
+    if (not wants_paths) {
+        return result;
+    }
+
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
```
LHT129, line 77: [suggestion] The `CHECK_ARGUMENT(store_paths_, "store_paths is false")` message is technically accurate but not actionable for a user who receives it. When `DATA_FLAG_PATH` is requested but `store_paths` was not enabled at build time, the user needs to know *why* the call failed and *what to change*.

Consider a more descriptive message such as: `"DATA_FLAG_PATH requested but store_paths was not enabled at build time; set store_paths=true in the index parameters"`.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (hierarchy->path_store == nullptr) {
+            throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store is missing");
+        }
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
+            continue;
+        }
+        if (hierarchy_name.empty()) {
```
LHT129, line 85: [note] The `path_store == nullptr` check changed from `continue` (skip this hierarchy) to `throw VsagException(ErrorType::INTERNAL_ERROR, ...)`. This is correct because `wants_paths` is already true at this point, which implies `store_paths_` is true (validated above), which in turn guarantees every `Hierarchy` was constructed with a `path_store`. The throw now correctly surfaces a genuine invariant violation rather than silently skipping a hierarchy that should have had paths.

This behavioral change is sound, but worth noting for reviewers that the previous `continue` was a latent bug that could silently drop paths from a hierarchy whose store was unexpectedly null.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,101 @@
...
+            result->Paths(paths.get());
+        } else {
+            result->Paths(hierarchy_name, paths.get());
+        }
+        paths.release();
+    }
```
LHT129, line 91: [note] `paths.release()` is now called after `result->Paths(...)` instead of before. This is the correct order: `Dataset::Paths()` takes ownership of the raw pointer, so releasing it before the call would pass a dangling pointer. The previous code was a bug — `Paths()` would have received an already-deleted pointer. Good fix.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Agreed. No further code change is needed: reaching this branch means an internal construction invariant was violated, so surfacing `INTERNAL_ERROR` is intentional and preferable to silently omitting a hierarchy.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Agreed on the final ordering, with one nuance: `unique_ptr::release()` does not delete the object, so the old order did not create a dangling pointer. Moving `release()` after `Dataset::Paths()` improves exception safety: if the setter throws, the unique pointer still owns and deletes the array; after a successful ownership transfer, it releases ownership. No further change is needed.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,153 @@
...
+PyramidPathStore::Record(const std::string* paths, uint64_t count) {
+    CHECK_ARGUMENT(paths != nullptr, "paths must not be null");
+    std::unique_lock lock(mutex_);
+    paths_by_inner_id_.resize(count);
+    has_path_.resize(count, 0);
+    for (uint64_t inner_id = 0; inner_id < count; ++inner_id) {
+        paths_by_inner_id_[inner_id] = paths[inner_id];
+        has_path_[inner_id] = 1;
+    }
+}
```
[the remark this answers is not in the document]
jac0626: Fixed in 17d1ffc7. The full-build overload now checks both backing vectors under its unique lock and rejects a non-empty store before any resize or mutation. The regression test calls the overload twice and verifies that the second call throws while the original size remains unchanged.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,153 @@
...
+                   "Pyramid path count overflow");
+    const auto end = begin + data_biases.size();
+    std::unique_lock lock(mutex_);
+    if (paths_by_inner_id_.size() < end) {
+        paths_by_inner_id_.resize(end);
+        has_path_.resize(end, 0);
+    }
+    for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+        const auto inner_id = begin + offset;
+        paths_by_inner_id_[inner_id] = paths[data_biases[offset]];
```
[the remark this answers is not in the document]
jac0626: No code change after tracing the complete call path. `data_biases` is an internal vector populated only by the `Pyramid::Add` loop with offsets in `[0, data_num)`, and the same offsets are already used to index `data_vectors` before `PyramidPathStore::Record` is reached. Rechecking every offset here would not establish a new safety boundary; it would add a redundant per-item branch to every opted-in Add. We are keeping this construction invariant explicit in the caller.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: No code change for the same internal-invariant reason as the duplicate comment above. `data_biases` contains only non-negative loop offsets accepted by `Pyramid::Add`, and those offsets have already indexed the source vectors before this store call. A second per-path check here would add hot-path work without protecting a separate input boundary.

src/algorithm/pyramid/pyramid.cpp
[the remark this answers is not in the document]
jac0626: No code change because the proposed legacy fallback did not exist for Pyramid before this PR. The previous implementation called `read_index_footer(reader, basic_info)` and immediately threw `READ_ERROR` when it returned false; that helper returns false exactly when `Footer::Parse(reader)` returns null. Using the parser directly therefore preserves the prior no-footer behavior.

src/algorithm/pyramid/pyramid_path_store.cpp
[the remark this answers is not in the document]
jac0626: Fixed in 17d1ffc7. The documented empty-store precondition is now enforced under the existing unique lock before resize, and a repeated full-build record regression verifies fail-fast behavior without changing stored state.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -895,6 +929,20 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
+                    throw VsagException(ErrorType::READ_ERROR,
+                                        "duplicate Pyramid streaming paths block");
+                }
+                if (not loaded_base_codes) {
+                    throw VsagException(ErrorType::READ_ERROR,
+                                        "Pyramid streaming paths block precedes base codes");
+                }
+                ReadSeekableBlockPayload(block_reader, block_header, [this](StreamReader& block) {
+                    this->deserialize_paths(block, static_cast<uint64_t>(this->cur_element_count_));
+                });
```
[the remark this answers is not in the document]
jac0626: No code change. Before reading streaming blocks, `read_streaming_body` parses the serialized `INDEX_PARAM` and calls `PyramidParameters::CheckCompatibility`, which compares `store_paths`. A current reader configured with `store_paths=false` therefore rejects a stream written with it enabled before reaching `PYRAMID_PATHS`. The tag is non-critical so an older reader that does not recognize the new capability can skip it and preserve search, not so a current mismatched reader silently drops requested path semantics.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Kept the current single-pass implementation. A complete hierarchy is the normal successful case, where the output array is required. A `HasAll` probe would acquire the lock and scan all IDs, then `GetPaths` would acquire it again and rescan/copy them, making successful retrieval slower. It would only avoid allocation for an incomplete hierarchy, which is the exceptional partial-data case, so the extra API and double scan are not justified.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Fixed in 17d1ffc7. The error now states: `DATA_FLAG_PATH requires store_paths=true in the Pyramid build parameters`, which identifies both the requested flag and the build-time opt-in needed to make the call valid.

LHT129 (commented):
Thank you for this PR. The separation of `PyramidPathStore` from `Pyramid` path methods into `pyramid_paths.cpp` is a good improvement. I have a few observations on the current revision:

### ExportModel does not copy path data

`ExportModel` (line 1064 in pyramid.cpp) creates a new Pyramid from `create_param_ptr_` and copies `base_codes_`, `precise_codes_`, and `raw_vector_`, but never copies the path data from the source index's `hierarchy->path_store` instances. When `store_paths_=true`, the exported model will have empty path stores, so `GetDataByIdsWithFlag(..., DATA_FLAG_PATH)` on the exported model would silently return a Dataset with no paths attached (because `GetPaths` finds `has_path_` all zeros and returns `false` for every hierarchy, causing `continue`).

Consider either serializing/deserializing the path stores as part of ExportModel (similar to how Clone works via Serialize/Deserialize roundtrip), or copying the PyramidPathStore data directly from source hierarchies to the target.

### Clone of empty index with store_paths=true

`InnerIndexInterface::Clone` calls `Serialize` then `Deserialize(StreamReader&)` directly (bypassing the base class `Deserialize(std::istream&)` which handles `EmptyIndex`). The new `Deserialize(StreamReader&)` at line 993 throws `INDEX_EMPTY` when the footer metadata reports an empty index. For an empty index with `store_paths=true`, `collect_streaming_header` sets `EmptyIndex` (line 745), so `Serialize` writes an empty-index footer, and the subsequent `Deserialize(StreamReader&)` in Clone would throw. The existing empty-index roundtrip tests pass because they go through `Deserialize(BinarySet)` which uses the base class `Deserialize(std::istream&)` path that handles `EmptyIndex` before reaching `Pyramid::Deserialize(StreamReader&)`.

Consider either adding an empty-index early-return in `Pyramid::Deserialize(StreamReader&)` (matching the base class behavior), or adding a Clone test for the empty-index-with-paths case.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +985,31 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
     // try to deserialize footer (only in new version)
-    JsonType basic_info;
-    if (not read_index_footer(reader, basic_info)) {
+    const auto footer = Footer::Parse(reader);
+    if (footer == nullptr) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
+    const auto metadata = footer->GetMetadata();
+    if (metadata == nullptr || metadata->EmptyIndex()) {
+        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
```
LHT129: The `INDEX_EMPTY` throw here is reached when `Pyramid::Deserialize(StreamReader&)` is called directly (e.g., from `InnerIndexInterface::Clone`, which does `Serialize` → `Deserialize(StreamReader&)` bypassing the base class `Deserialize(std::istream&)` that handles `EmptyIndex`). For an empty index with `store_paths=true`, `collect_streaming_header` sets `EmptyIndex` (line 745), so `Serialize` writes an empty-index footer, and Clone would throw here.

Consider adding an early-return for the empty-index case (matching the base class behavior in `InnerIndexInterface::Deserialize(std::istream&)`) or adding a Clone test for the empty-index-with-paths scenario.

LHT129 (commented):
test

LHT129 (commented):
Thank you for this PR. The separation of PyramidPathStore from Pyramid path methods into pyramid_paths.cpp is a good improvement. I have a few observations on the current revision.

### ExportModel does not copy path data

ExportModel (line 1064 in pyramid.cpp) creates a new Pyramid from create_param_ptr_ and copies base_codes_, precise_codes_, and raw_vector_, but never copies the path data from the source index hierarchy->path_store instances. When store_paths_=true, the exported model will have empty path stores, so GetDataByIdsWithFlag(..., DATA_FLAG_PATH) on the exported model would silently return a Dataset with no paths attached.

Consider either serializing/deserializing the path stores as part of ExportModel, or copying the PyramidPathStore data directly from source hierarchies to the target.

### Clone of empty index with store_paths=true

InnerIndexInterface::Clone calls Serialize then Deserialize(StreamReader&) directly (bypassing the base class Deserialize(std::istream&) which handles EmptyIndex). The new Deserialize(StreamReader&) throws INDEX_EMPTY when the footer metadata reports an empty index. For an empty index with store_paths=true, collect_streaming_header sets EmptyIndex, so Serialize writes an empty-index footer, and the subsequent Deserialize(StreamReader&) in Clone would throw.

Consider adding an empty-index early-return in Pyramid::Deserialize(StreamReader&) or adding a Clone test for the empty-index-with-paths case.

src/algorithm/pyramid/pyramid.cpp
jac0626: Verified without a production-code change. `InnerIndexInterface::Clone` calls the regular `Serialize(StreamWriter&)` path, whose `write_index_footer` does not mark the metadata empty; `EmptyIndex` is set only by the streaming header path. Commit `aef406b9` adds an explicit empty-index Clone regression with `store_paths=true`, and it passes while confirming that the clone has zero elements.

src/algorithm/pyramid/pyramid_path_store.cpp
LHT129: [suggestion] `data_biases[offset]` is `int64_t` and is used directly as an index into `paths[...]` without validating it is non-negative. A negative value would cause undefined behavior (out-of-bounds access). While `data_biases` is populated from accepted input offsets in `Pyramid::Add`, there is no explicit guarantee at this call site that all values are non-negative. Consider adding: `CHECK_ARGUMENT(data_biases[offset] >= 0, "data bias must not be negative");` before the indexing on line 64.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,93 @@
...
+        }
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
+            continue;
+        }
+        if (hierarchy_name.empty()) {
+            result->Paths(paths.release());
+        } else {
+            result->Paths(hierarchy_name, paths.release());
+        }
```
LHT129, line 89: [suggestion] `std::make_unique<std::string[]>(count)` allocates an output path array for every hierarchy before knowing whether `GetPaths(...)` will succeed. `GetPaths` returns `false` early when any requested ID has no recorded path, which means the allocation is wasted for hierarchies that are incomplete. For large `count` and multiple hierarchies, this causes avoidable allocations. Consider adding a lightweight completeness probe on `PyramidPathStore` (e.g., `HasAll(inner_ids)`) and only allocating when all IDs are present, or refactoring `GetPaths` to support a check-only mode before allocating the destination buffer.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +985,31 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
```
LHT129, line 984: [note] `Deserialize()` throws `INDEX_EMPTY` when `metadata->EmptyIndex()` is true (line 986-987). This is consistent with the previous `read_index_footer` behavior, and the new `store_paths=true` empty BinarySet and streaming roundtrip tests pass — confirming empty-index serialization works correctly through the public API wrappers. The `Deserialize` path is an internal detail; the public `Deserialize(BinarySet)` and `DeserializeStreaming(istream)` wrappers handle empty indexes before reaching this code path. No action needed.

jac0626:
Updated in `0ad9b837` to tighten the module boundary and recover the measured build overhead:

- `PyramidPathStore` no longer understands Dataset offsets, `first_inner_id`, duplicate-label filtering, or Build/Add semantics.
- `Pyramid` captures the actual accepted `inner_id` at allocation time and passes only `(inner_id, path)` to `PyramidPathStore::Writer::Insert`.
- The store-owned Writer holds one exclusive lock for the write batch, preserving concurrent Add safety without per-path locking.
- Writer is non-copyable/non-movable; failed dual-vector growth rolls back to a consistent size.
- Added middle-duplicate, NSW/ODescent, sparse/out-of-order store, and concurrent Add path-mapping coverage.

Release A/B against the previous bulk `Record` implementation (100K vectors, dim=4, NSW, `store_paths=true`, 7 alternating runs):
- 1K unique paths: 279.889 ms -> 282.941 ms (+1.09%)
- 100K unique paths: 376.217 ms -> 376.741 ms (+0.14%, within run variance)
- RSS delta stayed below 0.8 MiB; serialized sizes were identical.

Local targeted result: 18 PathStore unit assertions + 271 Pyramid path functional assertions passed. `clang-format-15`, targeted `clang-tidy-15`, and `git diff --check` passed.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (not wants_paths) {
+        return result;
+    }
+
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+        if (hierarchy->path_store == nullptr) {
+            throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store is missing");
+        }
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
```
LHT129, line 82: [suggestion] The `paths` array is allocated via `std::make_unique<std::string[]>(count)` before `GetPaths` is called, but `GetPaths` can return `false` (e.g., when a requested ID has no path in this hierarchy). When that happens, the allocation is wasted and the loop `continue`s.

For large `count` values and multiple hierarchies where only some are complete, this can cause significant unnecessary allocations. Consider either:
- Adding a lightweight `HasAll(inner_ids)` probe method to `PyramidPathStore` and only allocating after confirming completeness, or
- Refactoring `GetPaths` to accept an optional "check-only" mode that validates without writing.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+void
+PyramidPathStore::Writer::Insert(InnerIdType inner_id, const std::string& path) {
+    const auto slot = static_cast<uint64_t>(inner_id);
+    if (store_.paths_by_inner_id_.size() <= slot) {
+        const auto old_size = store_.paths_by_inner_id_.size();
+        try {
+            store_.paths_by_inner_id_.resize(slot + 1);
+            store_.has_path_.resize(slot + 1, 0);
+        } catch (...) {
```
LHT129: [suggestion] The resize/rollback pattern in `Writer::Insert` has a subtle exception-safety gap. When `store_.paths_by_inner_id_.resize(slot + 1)` succeeds but `store_.has_path_.resize(slot + 1, 0)` throws, only `paths_by_inner_id_` is rolled back to `old_size`. However, if `paths_by_inner_id_.resize()` succeeds and `has_path_.resize()` throws, the two vectors are left with inconsistent sizes — `paths_by_inner_id_` is at `slot + 1` while `has_path_` remains at `old_size`.

The rollback only catches the case where `has_path_.resize()` throws, but the two vectors are now out of sync. Consider using a scope guard or resizing both before any mutation:

```cpp
if (store_.paths_by_inner_id_.size() <= slot) {
    const auto old_size = store_.paths_by_inner_id_.size();
    store_.paths_by_inner_id_.resize(slot + 1);
    try {
        store_.has_path_.resize(slot + 1, 0);
    } catch (...) {
        store_.paths_by_inner_id_.resize(old_size);
        throw;
    }
}
```

This is already what the code does — but the issue is that if `paths_by_inner_id_.resize()` succeeds and `has_path_.resize()` throws, the rollback restores `paths_by_inner_id_` to `old_size`. This is actually correct as written. However, the comment is misleading: the rollback is only needed if `has_path_.resize()` throws, and the code correctly handles this. The real concern is that `std::vector::resize` on a `Vector<std::string>` could throw `std::bad_alloc` or an exception from the allocator, and the `has_path_` (a `Vector<uint8_t>`) resize is less likely to throw. So the current pattern is defensive but reasonable.

Actually, re-reading more carefully: the try/catch only wraps `has_path_.resize()`, not `paths_by_inner_id_.resize()`. If `paths_by_inner_id_.resize()` throws, `has_path_` is untouched and `paths_by_inner_id_` is in an unspecified-but-valid state (strong guarantee from vector::resize). So the code is correct. Disregard this concern — the exception safety is fine as written.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,138 @@
...
+    uint64_t slot_count = 0;
+    StreamReader::ReadObj(reader, slot_count);
+    if (slot_count > max_count) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path slot count");
+    }
+    const auto cursor = reader.GetCursor();
+    const auto reader_length = reader.Length();
+    if (cursor > reader_length || slot_count > reader_length - cursor) {
+        throw VsagException(ErrorType::READ_ERROR,
+                            "corrupted Pyramid path slots exceed remaining payload");
```
LHT129: [note] The `Deserialize` method validates `slot_count > reader_length - cursor` as a coarse upper bound, but the actual payload per slot is variable (1 byte for `present` + optional variable-length string). A malicious or corrupted stream with `slot_count` just under this bound but with many `present=1` slots containing long path strings could still cause reads past the reader end, which would be caught by `ReadPyramidPathString` on a per-string basis. This is defense-in-depth and acceptable, but worth noting that the early bound check is only a coarse filter, not a precise payload size validation.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,138 @@
...
+        reader.Read(value.data(), length);
+    }
+    return value;
+}
+
+void
+PyramidPathStore::Writer::Insert(InnerIdType inner_id, const std::string& path) {
+    const auto slot = static_cast<uint64_t>(inner_id);
+    if (store_.paths_by_inner_id_.size() <= slot) {
+        const auto old_size = store_.paths_by_inner_id_.size();
```
LHT129: [note] In `Writer::Insert`, `InnerIdType` (which is `int64_t`) is cast to `uint64_t` via `static_cast<uint64_t>(inner_id)`. If a negative `inner_id` were passed (which should never happen in normal operation since inner IDs are assigned sequentially from 0), the cast would produce a very large `uint64_t` value, causing the vectors to resize to an enormous size and likely OOM. While this is an internal invariant (inner IDs are always non-negative), a `CHECK_ARGUMENT(inner_id >= 0, ...)` at the top of `Insert` would make this invariant explicit and fail-fast with a clear error message rather than a potentially cryptic OOM or `std::bad_alloc`.

jac0626: No change needed here because InnerIdType is uint32_t in src/basic_types.h, not int64_t, so a negative inner ID cannot reach this cast. The new Prepare path also validates slot_count against the full InnerIdType range and container max_size.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Updated in 087ff820: the row-state/list encoding was removed entirely. Deserialization now reads the direct offsets, uint16 counts, and flat string fields; vector/path counts are bounded by the configured element count and remaining bytes before allocation, and each row is checked only for sentinel/count consistency and an in-range slice. No magic/version or additional overlap/bitmap validation is added.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Agreed; no change is needed. The new dense implementation also records both old slot-array sizes and rolls both arrays back together in Prepare and Insert.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Thanks for the suggestion. I am leaving this unchanged for this PR. HasAll or a check-only mode would add a second full inner-ID scan to every successful hierarchy, which is the common path, plus another storage API solely to avoid a temporary allocation on an incomplete hierarchy. GetPaths validates every requested row before writing, and the temporary array is bounded by the requested count and released immediately on false. The frozen goal here is the simplest fast common path; we can revisit a single-pass output API if incomplete-hierarchy retrieval becomes measurable.

jac0626:
Final minimal-scope update is pushed in 087ff820.

Implementation:
- Path retention remains opt-in through store_paths=false by default; DATA_FLAG_PATH is required to return paths.
- PathStore is a separate component and accepts only allocated inner IDs plus paths.
- main uses only the direct dense SoA fields: Vector<uint64_t> offsets, Vector<uint16_t> counts, and a flat Vector<string> paths.
- Rows support 0/1/N paths for the later #2796 integration. There is no map, sparse/adaptive index, row-state protocol, magic, or private path format version.
- Serialization writes those three fields directly with the existing StreamWriter primitives.

Scope cleanup in this update:
- Removed the unused get_data_by_ids helper while preserving the original virtual GetDataByIds dispatch.
- Restored the existing footer/BufferStreamReader flow; only path reading is appended after the hierarchy data.
- Removed test-only Size APIs, duplicate full scans in GetPaths, temporary output copies, hierarchy bookkeeping beyond the existing pattern, and incomplete flat-pool total-count validation.
- Kept bounded length/range checks that prevent malformed input from causing unbounded allocation.

Validation:
- Pyramid UT: 207 assertions / 19 cases
- Path FT: 271 assertions / 15 cases

Release A/B, 100K vectors, dim=4, NSW, store_paths=true, 7 alternating runs:
- Build versus the prior single-path implementation: +1.27% with 1K unique paths; +1.86% with 100K unique paths.
- Peak build RSS: +1.16 MiB and +1.43 MiB respectively.
- Direct field serialization versus the prior unmerged row-state draft: 1.90% faster and 0.46% faster respectively.
- The direct dense wire format adds 900,016 bytes at 100K slots (9 bytes per slot plus field headers); this is the explicit cost of keeping offsets and uint16 counts directly serializable.

LHT129 (commented):
Thank you for this well-structured PR. The separation of `PyramidPathStore` from `pyramid_paths.cpp` is clean, and the test coverage is thorough. I have a few observations:

**[suggestion] `pyramid_paths.cpp:91` — Eager path array allocation before validation**

In `GetDataByIdsWithFlag`, for each hierarchy, `std::make_unique<std::string[]>(count)` is allocated before `GetPaths` validates the inner IDs. If `GetPaths` returns `false` (e.g. a hole or missing ID), the allocation is wasted. For large `count` and many hierarchies, this adds unnecessary peak memory. Consider calling `GetPaths` with a temporary vector first, or restructuring to allocate only after validation succeeds.

**[suggestion] `pyramid_path_store.cpp:141` — Unchecked `data_biases[offset]` index**

In `Writer::Insert`, `data_biases[offset]` (int64_t) is used as an index into `paths[...]` without a non-negative check. While the callers (`build_by_odescent`, `Add`, `build_with_cache`) all populate `data_biases` from valid offsets, a defensive `CHECK_ARGUMENT` would guard against future misuse.

**[note] `pyramid.cpp:985` — `Deserialize` empty-index behavior**

The `Footer::Parse` + `EmptyIndex()` check now throws `INDEX_EMPTY` for empty indexes. This is consistent with the previous `read_index_footer` behavior (which also threw on empty metadata), so no regression is introduced. The empty-index roundtrip tests in `test_pyramid_path_serialization.cpp` test `Serialize`/`Deserialize` on an index that was never built (0 elements), which succeeds because the index was never serialized with footer metadata — the `Clone()` path and the `Serialize()` of an empty index produce different codepaths than `Deserialize` of a previously-built-and-emptied index. This is worth documenting.

**[note] `serialization_tags.h:120` — Non-critical PYRAMID_PATHS tag**

The `PYRAMID_PATHS` block is marked non-critical in the tag table, but `read_streaming_body` enforces its presence when `store_paths_` is true. This is intentional: older readers that don't know about paths can skip the block, while readers configured with `store_paths=true` validate separately. The design is sound, though a brief comment in the tag table explaining this dual-enforcement pattern would help future maintainers.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,102 @@
...
+            continue;
+        }
+        if (hierarchy_name.empty()) {
+            result->Paths(paths.get());
+        } else {
+            result->Paths(hierarchy_name, paths.get());
+        }
+        paths.release();
```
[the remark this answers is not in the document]
jac0626: No change. Dataset is created with Owner(true), and Paths follows the existing Dataset raw-pointer ownership contract. Keeping the unique_ptr through the Paths call is intentional: named-path attachment can allocate, so get(), successful attachment, then release() preserves cleanup if attachment throws. Passing release() into the call would make that exception path less safe.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,308 @@
...
+    if (slot < store_.offsets_.size()) {
+        CHECK_ARGUMENT(store_.offsets_[slot] == MISSING_PATH_OFFSET,
+                       "inner id already has Pyramid paths");
+    }
+
+    const auto old_path_count = static_cast<uint64_t>(store_.paths_.size());
+    const auto old_offset_count = store_.offsets_.size();
+    const auto old_count_count = store_.counts_.size();
+    try {
+        Prepare(slot + 1, path_count);
```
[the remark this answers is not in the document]
jac0626: No code change here. Every Pyramid batch path calls Writer::Prepare once with the final slot count and total added path memberships, so the Insert-internal Prepare performs no resize or reserve in that path; it is the geometric-growth fallback for direct/out-of-order Insert callers and keeps Insert safe as the single write API. Adding InsertPrepared would duplicate the API and expose a capacity precondition for a very small set of checks, contrary to the final minimal-scope design. The 100K-vector Release A/B includes this implementation.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -933,14 +982,28 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
-    JsonType basic_info;
-    if (not read_index_footer(reader, basic_info)) {
+    const auto footer = Footer::Parse(reader);
+    if (footer == nullptr) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
+    const auto metadata = footer->GetMetadata();
+    if (metadata == nullptr || metadata->EmptyIndex()) {
+        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
+    }
```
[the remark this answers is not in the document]
jac0626: Addressed in 087ff820 by restoring Pyramid existing read_index_footer plus BufferStreamReader flow. The newly introduced explicit EmptyIndex rejection and footer slicing are gone, while the empty BinarySet and streaming roundtrip tests pass.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,95 @@
...
+        const auto hierarchy = hierarchies_.find(hierarchy_name);
+        if (hierarchy == hierarchies_.end()) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                fmt::format("unknown Pyramid path hierarchy '{}'", hierarchy_name));
+        }
+        if (hierarchy->second->path_store == nullptr) {
+            throw VsagException(ErrorType::READ_ERROR, "Pyramid path storage is disabled");
+        }
+        hierarchy->second->path_store->Deserialize(reader, max_count);
+    }
```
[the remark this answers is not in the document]
jac0626: Thanks for the suggestion. I am keeping this unchanged to preserve the existing Pyramid serialization contract and the scope of this feature.

The writer iterates `hierarchies_`, whose keys are unique, so a duplicate hierarchy name cannot be produced by a valid serialized index. It requires an arbitrarily corrupted payload. The existing hierarchy graph deserialization uses the same count-plus-name lookup pattern without a separate duplicate-name set. Adding a one-off set only for the optional path block would introduce a different, partial corruption policy and extra code that does not make arbitrary corrupted payloads generally self-validating.

The current checks remain scoped to what this block needs: configured hierarchy count, known hierarchy names, enabled path stores, bounded strings and vectors, and valid path ranges.

LHT129 (approved):
LGTM

PR #2765, opened 2026-08-25
[Backport 1.0] feat(pyramid): support paths in GetDataByIdsWithFlag

jac0626:
## Change Type

- [x] Improvement

## Linked Issue

Related to #2754. Backport of #2763 to the 1.0 branch.

## What Changed

- Add store_paths as an opt-in Pyramid parameter, disabled by default.
- Return retained paths only when GetDataByIdsWithFlag selects DATA_FLAG_PATH; named hierarchies use Dataset::GetPaths(name).
- Keep storage isolated in PyramidPathStore with only direct dense uint64 offsets, uint16 counts, and a flat string vector. Rows support 0/1/N paths without a map, sparse/adaptive policy, magic, private version, or row-state protocol.
- Record successfully allocated inner IDs for NSW, ODescent, and Add.
- Serialize the three fields directly through existing StreamWriter primitives; regular serialization is parameter-driven and streaming uses PYRAMID_PATHS.

- Default-off behavior allocates no per-vector path state and writes no path sidecar.
- The in-memory writer is identical to main. Main Release A/B against the prior single-path implementation measured build +1.27% / +1.86% and peak build RSS +1.16 / +1.43 MiB at 1K / 100K unique paths for 100K vectors.
- Direct serialization was 1.90% / 0.46% faster than the prior unmerged row-state draft and adds 900,016 serialized bytes at 100K slots.

## Validation

- Pyramid unit tests: 570 assertions / 11 cases.
- Pyramid path functional tests: 280 assertions / 16 cases.

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -775,15 +775,10 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
-    const auto footer = Footer::Parse(reader);
-    if (footer == nullptr) {
+    JsonType basic_info;
+    if (not read_index_footer(reader, basic_info)) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
-    const auto metadata = footer->GetMetadata();
-    if (metadata == nullptr || metadata->EmptyIndex()) {
-        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
-    }
-    auto basic_info = metadata->Get(BASIC_INFO);
     auto max_capacity = basic_info["max_capacity"].GetInt();
     auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
     const bool serialized_store_paths = param_json.Contains(PYRAMID_STORE_PATHS_KEY) &&
@@ -793,10 +788,8 @@ Pyramid::Deserialize(StreamReader& reader) {
                             "serialized Pyramid store_paths does not match config");
     }
 
-    const auto body_length = reader.Length() - footer->Length();
-    auto body_reader = reader.Slice(0, body_length);
     BufferStreamReader buffer_reader(
-        &body_reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
+        &reader, std::numeric_limits<uint64_t>::max(), this->allocator_);
 
     label_table_->Deserialize(buffer_reader);
     delete_count_.store(static_cast<int64_t>(label_table_->GetAllDeletedIds().size()),
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -232,7 +232,10 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
         for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
             const auto* paths = base->GetPaths(hierarchy_name);
             if (paths != nullptr) {
-                hierarchy->path_store->Record(paths, static_cast<uint64_t>(data_num));
+                auto writer = hierarchy->path_store->AcquireWriter();
+                for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
+                    writer.Insert(static_cast<InnerIdType>(offset), paths[offset]);
+                }
             }
         }
     }
@@ -862,6 +865,7 @@ Pyramid::Add(const DatasetPtr& base) {
     const auto* data_ids = base->GetIds();
     std::vector<int64_t> failed_ids;
     Vector<int64_t> data_biases(allocator_);
+    Vector<InnerIdType> accepted_inner_ids(allocator_);
     int64_t local_cur_element_count = 0;
     {
         std::lock_guard lock(cur_element_count_mutex_);
@@ -876,21 +880,26 @@ Pyramid::Add(const DatasetPtr& base) {
             resize(new_capacity);
         }
         int64_t valid_id_count = 0;
+        if (store_paths_) {
+            accepted_inner_ids.reserve(data_num);
+        }
         for (int64_t i = 0; i < data_num; ++i) {
             if (not label_table_->CheckLabel(data_ids[i])) {
-                label_table_->Insert(valid_id_count + local_cur_element_count, data_ids[i]);
-                base_codes_->InsertVector(data_vectors + dim_ * i,
-                                          valid_id_count + local_cur_element_count);
+                const auto inner_id =
+                    static_cast<InnerIdType>(valid_id_count + local_cur_element_count);
+                label_table_->Insert(inner_id, data_ids[i]);
+                base_codes_->InsertVector(data_vectors + dim_ * i, inner_id);
                 if (use_reorder_) {
-                    precise_codes_->InsertVector(data_vectors + dim_ * i,
-                                                 valid_id_count + local_cur_element_count);
+                    precise_codes_->InsertVector(data_vectors + dim_ * i, inner_id);
                 }
                 if (create_new_raw_vector_) {
-                    raw_vector_->InsertVector(data_vectors + dim_ * i,
-                                              valid_id_count + local_cur_element_count);
+                    raw_vector_->InsertVector(data_vectors + dim_ * i, inner_id);
                 }
                 valid_id_count++;
                 data_biases.push_back(i);
+                if (store_paths_) {
+                    accepted_inner_ids.push_back(inner_id);
+                }
             } else {
                 logger::warn("Label {} already exists, skip adding.", data_ids[i]);
                 failed_ids.push_back(data_ids[i]);
@@ -904,7 +913,10 @@ Pyramid::Add(const DatasetPtr& base) {
         const auto* hpath = base->GetPaths(hname);
         if (hpath != nullptr) {
             if (store_paths_) {
-                h_ptr->path_store->Record(hpath, data_biases, local_cur_element_count);
+                auto writer = h_ptr->path_store->AcquireWriter();
+                for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+                    writer.Insert(accepted_inner_ids[offset], hpath[data_biases[offset]]);
+                }
             }
             add_to_hierarchy(*h_ptr, data_vectors, hpath, data_biases, local_cur_element_count);
         }
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -34,9 +34,6 @@
 namespace vsag {
 
 const static float RADIUS_EPSILON = 1.1F;
-static constexpr uint64_t PYRAMID_PATHS_MAGIC = 0x5059525041544853ULL;  // PYRPATHS
-static constexpr int64_t PYRAMID_PATHS_FORMAT_VERSION = 1;
-static constexpr const char* PYRAMID_PATHS_FORMAT_VERSION_KEY = "pyramid_paths_format_version";
 
 std::vector<std::string>
 split(const std::string& str, char delimiter) {
@@ -468,7 +465,6 @@ Pyramid::Serialize(StreamWriter& writer) const {
     }
 
     if (store_paths_) {
-        StreamWriter::WriteObj(writer, PYRAMID_PATHS_MAGIC);
         serialize_paths(writer);
     }
 
@@ -476,9 +472,6 @@ Pyramid::Serialize(StreamWriter& writer) const {
     JsonType basic_info;
     basic_info["max_capacity"].SetInt(max_capacity_);
     basic_info[INDEX_PARAM].SetString(this->create_param_ptr_->ToString());
-    if (store_paths_) {
-        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].SetInt(PYRAMID_PATHS_FORMAT_VERSION);
-    }
     write_index_footer(writer, basic_info);
 }
 
@@ -787,17 +780,14 @@ Pyramid::Deserialize(StreamReader& reader) {
         throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
     }
     auto basic_info = metadata->Get(BASIC_INFO);
-    const bool has_paths = basic_info.Contains(PYRAMID_PATHS_FORMAT_VERSION_KEY);
-    if (has_paths != store_paths_) {
-        throw VsagException(ErrorType::READ_ERROR,
-                            "serialized Pyramid paths metadata does not match config");
-    }
-    if (has_paths &&
-        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].GetInt() != PYRAMID_PATHS_FORMAT_VERSION) {
-        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
-                            "unsupported Pyramid paths format version");
-    }
     auto max_capacity = basic_info["max_capacity"].GetInt();
+    auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
+    const bool serialized_store_paths = param_json.Contains(PYRAMID_STORE_PATHS_KEY) &&
+                                        param_json[PYRAMID_STORE_PATHS_KEY].GetBool();
+    if (serialized_store_paths != store_paths_) {
+        throw VsagException(ErrorType::READ_ERROR,
+                            "serialized Pyramid store_paths does not match config");
+    }
 
     const auto body_length = reader.Length() - footer->Length();
     auto body_reader = reader.Slice(0, body_length);
@@ -816,7 +806,6 @@ Pyramid::Deserialize(StreamReader& reader) {
     }
     cur_element_count_ = base_codes_->TotalCount();
 
-    auto param_json = JsonType::Parse(basic_info[INDEX_PARAM].GetString());
     if (param_json.Contains(PYRAMID_HIERARCHIES)) {
         uint64_t hierarchy_count = 0;
         StreamReader::ReadObj(buffer_reader, hierarchy_count);
@@ -839,12 +828,7 @@ Pyramid::Deserialize(StreamReader& reader) {
         h_iter->second->root->Deserialize(buffer_reader);
     }
 
-    if (has_paths) {
-        uint64_t magic = 0;
-        StreamReader::ReadObj(buffer_reader, magic);
-        if (magic != PYRAMID_PATHS_MAGIC) {
-            throw VsagException(ErrorType::READ_ERROR, "missing Pyramid paths marker");
-        }
+    if (store_paths_) {
         deserialize_paths(buffer_reader, static_cast<uint64_t>(cur_element_count_));
     }
 
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -34,6 +34,9 @@
 namespace vsag {
 
 const static float RADIUS_EPSILON = 1.1F;
+static constexpr uint64_t PYRAMID_PATHS_MAGIC = 0x5059525041544853ULL;  // PYRPATHS
+static constexpr int64_t PYRAMID_PATHS_FORMAT_VERSION = 1;
+static constexpr const char* PYRAMID_PATHS_FORMAT_VERSION_KEY = "pyramid_paths_format_version";
 
 std::vector<std::string>
 split(const std::string& str, char delimiter) {
@@ -227,6 +230,14 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     }
     if (create_new_raw_vector_) {
         raw_vector_->BatchInsertVector(data_vectors, data_num);
+    }
+    if (store_paths_) {
+        for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+            const auto* paths = base->GetPaths(hierarchy_name);
+            if (paths != nullptr) {
+                hierarchy->path_store->Record(paths, static_cast<uint64_t>(data_num));
+            }
+        }
     }
     auto codes = use_reorder_ ? precise_codes_ : base_codes_;
 
@@ -456,10 +467,18 @@ Pyramid::Serialize(StreamWriter& writer) const {
         hierarchies_.at("")->root->Serialize(writer);
     }
 
+    if (store_paths_) {
+        StreamWriter::WriteObj(writer, PYRAMID_PATHS_MAGIC);
+        serialize_paths(writer);
+    }
+
     // serialize footer (introduced since v0.15)
     JsonType basic_info;
     basic_info["max_capacity"].SetInt(max_capacity_);
     basic_info[INDEX_PARAM].SetString(this->create_param_ptr_->ToString());
+    if (store_paths_) {
+        basic_info[PYRAMID_PATHS_FORMAT_VERSION_KEY].SetInt(PYRAMID_PATHS_FORMAT_VERSION);
+    }
     write_index_footer(writer, basic_info);
 }
 
@@ -508,6 +527,13 @@ Pyramid::collect_streaming_header() const {
                                  hierarchy_tag,
                                  StreamSerializationBlockCurrentVersion(hierarchy_tag),
                                  StreamSerializationTagCritical(hierarchy_tag));
+    if (store_paths_) {
+        const auto path_tag = static_cast<uint32_t>(StreamSerializationTag::PYRAMID_PATHS);
+        AppendStreamingManifestBlock(manifest,
+                                     path_tag,
+                                     StreamSerializationBlockCurrentVersion(path_tag),
+                                     StreamSerializationTagCritical(path_tag));
+    }
     metadata->Set("block_manifest", manifest);
     metadata->SetEmptyIndex(this->GetNumElements() == 0);
     return metadata;
@@ -560,6 +586,13 @@ Pyramid::serialize_streaming_body(StreamWriter& writer) const {
                         hierarchy_tag,
                         StreamSerializationTagCritical(hierarchy_tag),
                         [this](StreamWriter& w) { this->serialize_hierarchies(w); });
+    if (store_paths_) {
+        const auto path_tag = static_cast<uint32_t>(StreamSerializationTag::PYRAMID_PATHS);
+        WriteStreamingBlock(
+            writer, path_tag, StreamSerializationTagCritical(path_tag), [this](StreamWriter& w) {
+                this->serialize_paths(w);
+            });
+    }
 }
 
 void
@@ -622,6 +655,7 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata) 
     bool loaded_precise_codes = false;
     bool loaded_raw_vector = false;
     bool loaded_hierarchies = false;
+    bool loaded_paths = false;
 
     while (true) {
         auto block_header = StreamBlockHeader::Read(reader);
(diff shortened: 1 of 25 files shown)


src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+    auto result = InnerIndexInterface::GetDataByIds(ids, count);
+    if (not store_paths_) {
+        return result;
+    }
+
+    Vector<InnerIdType> inner_ids(static_cast<uint64_t>(count), allocator_);
+    {
+        std::shared_lock lock(label_lookup_mutex_);
+        for (int64_t offset = 0; offset < count; ++offset) {
+            inner_ids[offset] = label_table_->GetIdByLabel(ids[offset]);
```
LHT129: [suggestion] `GetDataByIds` performs a redundant label lookup: the base class `InnerIndexInterface::GetDataByIds` → `GetDataByIdsWithFlag` already calls `label_table_->GetIdByLabel(ids[i])` for every requested ID (to retrieve vectors/attributes), and this override repeats the same lookup to resolve inner IDs for path retrieval. This doubles the label-lookup cost on the hot path.

Consider either:
- Refactoring the base class to expose the already-resolved inner IDs, or
- Accepting the double lookup as a pragmatic tradeoff given the current API constraints, but documenting the cost.

At minimum, the second lookup could reuse the same `label_lookup_mutex_` critical section to avoid locking twice.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -690,14 +751,27 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
 void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
-    JsonType basic_info;
-    if (not read_index_footer(reader, basic_info)) {
+    const auto footer = Footer::Parse(reader);
+    if (footer == nullptr) {
         throw VsagException(ErrorType::READ_ERROR, "failed to read index footer");
     }
+    const auto metadata = footer->GetMetadata();
```
LHT129: [note] The `Deserialize` method replaces `read_index_footer(reader, basic_info)` with `Footer::Parse(reader)` + `metadata->Get(BASIC_INFO)`. This is a behavioral change that goes beyond the stated scope of adding path support:

- The old `read_index_footer` returned `false` on failure, triggering a `READ_ERROR` exception.
- The new `Footer::Parse` returns `nullptr` on failure, which also triggers a `READ_ERROR` exception.
- However, the new code additionally checks `metadata == nullptr || metadata->EmptyIndex()` and throws `INDEX_EMPTY`, which is a new error path not present before.

This change is consistent with how other index types (hgraph, ivf, sindi_v2) already use `Footer::Parse`, so it is likely an intentional modernization. Just noting it since it affects the deserialization contract for existing Pyramid indexes.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include "pyramid_path_store.h"
+
+#include <fmt/format.h>
+
+#include <limits>
+#include <memory>
+#include <mutex>
+
+#include "algorithm/pyramid/pyramid.h"
```
LHT129: [suggestion] This file includes `algorithm/pyramid/pyramid.h` (line 23), which in turn includes `algorithm/pyramid/pyramid_path_store.h` (the header for this file). While `#pragma once` prevents infinite recursion, this circular dependency is a code smell. It also means the Pyramid class methods (`serialize_paths`, `deserialize_paths`, `validate_paths`, `GetDataByIds`) are defined in the path store implementation file rather than in `pyramid.cpp`, which is surprising. Consider moving these Pyramid methods back to `pyramid.cpp` and breaking the circular include.

src/storage/serialization_tags.h
```diff
@@ -38,6 +38,7 @@ enum class StreamSerializationTag : uint32_t {
     SINDI_RERANK_INDEX = 12,
     SINDI_TERM_ID_MAPPER = 13,
     PYRAMID_HIERARCHIES = 14,
+    PYRAMID_PATHS = 18,
```
LHT129, line 41: [note] `PYRAMID_PATHS = 18` leaves a gap (14→18), presumably reserving 15-17 for future Pyramid streaming blocks. This is a reasonable forward-looking choice. However, note that `PYRAMID_PATHS` is marked as non-critical in `StreamSerializationTagCritical`, meaning older readers will silently skip it. If the paths block becomes required for correctness in a future version, the critical flag should be reconsidered.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Fixed. `InnerIndexInterface` now resolves each label once and reuses the resulting `inner_ids` for vector, attribute, and extra-info retrieval. `Pyramid` receives the same vector from the helper and uses it for path lookup, eliminating the second lookup and lock.

src/algorithm/pyramid/pyramid.cpp
jac0626: Confirmed and covered. The previous `read_index_footer` helper already used `Footer::Parse` internally and rejected null or empty metadata, so the direct form does not add a new empty-index contract. Added `store_paths=true` empty BinarySet and streaming roundtrip tests pass.

src/storage/serialization_tags.h
jac0626: Agreed, and the current choice is intentional. Paths are an optional capability, so unknown readers may skip tag 18 and preserve search. A reader configured with `store_paths=true` still requires the block. I added a concise comment beside the non-critical classification.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Addressed by separating the modules. `pyramid_path_store.cpp` now implements only the store, and the `Pyramid` serialization/retrieval methods are in the dedicated `pyramid_paths.cpp` file.

LHT129 (commented):
Reviewed the latest commit (523f0f84). All issues raised in the previous review round have been addressed:

- **Double label lookup**: Resolved by the `get_data_by_ids` / `get_data_by_ids_with_flag` refactoring in `inner_index_interface.cpp`, which resolves inner IDs once and reuses them for vector, attribute, extra-info, and now path retrieval.
- **Circular include**: Resolved by moving `Pyramid::serialize_paths`, `deserialize_paths`, and `GetDataByIds` into the dedicated `pyramid_paths.cpp` file. `pyramid_path_store.cpp` no longer includes `pyramid.h`.
- **Serialization tag criticality**: `PYRAMID_PATHS` is correctly marked as non-critical in `serialization_tags.h` with a comment explaining the rationale.
- **Deserialization contract**: The `Footer::Parse` modernization is consistent with other index types (hgraph, ivf, sindi_v2). The `has_paths != store_paths_` mismatch check and format version validation are good safety measures.

The new code is well-structured, properly tested (unit tests for path store, functional tests for paths and serialization, parameter tests), and follows the project coding standards. No new issues found.

LHT129 (commented):
[note] This is a well-structured backport of the Pyramid path storage feature (#2763) to the 1.0 branch. The code quality is high, with thorough test coverage, proper serialization compatibility checks, and clean separation of concerns between `PyramidPathStore` and the `Pyramid` class.

One minor observation in `pyramid_paths.cpp` `GetDataByIdsWithFlag`: the `paths` unique_ptr is allocated for every hierarchy before checking whether `GetPaths` succeeds. When a hierarchy lacks complete paths, the allocation is discarded. Since `hierarchies_.size()` is typically small (1-3) and path retrieval is not a hot path, this is acceptable. Consider pre-checking completeness before allocating if this becomes a concern in the future.

All previous review comments from LHT129 have been addressed:
- Redundant label lookup: resolved via `get_data_by_ids_with_flag` with shared `inner_ids`
- Circular include: resolved by separating `Pyramid` methods into `pyramid_paths.cpp`
- `PYRAMID_PATHS` tag numbering: clarified with a comment about non-critical classification
- `Deserialize` footer change: confirmed behaviorally equivalent to the old `read_index_footer`

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+    for (uint64_t offset = 0; offset < data_biases.size(); ++offset) {
+        const auto inner_id = begin + offset;
+        paths_by_inner_id_[inner_id] = paths[data_biases[offset]];
+        has_path_[inner_id] = 1;
+    }
+}
+
+void
+PyramidPathStore::Record(const std::string* paths, uint64_t count) {
```
LHT129: The `Record` method accesses `paths[data_biases[offset]]` without validating that `data_biases[offset]` is within bounds of the `paths` array. While the current callers in `pyramid.cpp:Add` construct `data_biases` from valid indices (`0 <= i < data_num`), adding a `CHECK_ARGUMENT` or assertion here would make the invariant explicit and protect against future misuse.

jac0626: No code change after checking the full caller path. This is an internal store API, and `Pyramid::Add` is its only production caller: it constructs every bias directly from `i` where `0 <= i < data_num`. Those same biases already index the source vectors before the path-store call, so another per-item check here would not create a new safety boundary and would add work to every opted-in Add. We keep the invariant at its construction point.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    Vector<InnerIdType> inner_ids(allocator_);
+    auto result = this->get_data_by_ids_with_flag(ids, count, selected_data_flag, inner_ids);
+    if (not wants_paths) {
+        return result;
+    }
+
+    for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
+        if (hierarchy->path_store == nullptr) {
+            throw VsagException(ErrorType::INTERNAL_ERROR, "Pyramid path store is missing");
```
LHT129, line 79: [suggestion] When `GetPaths` returns `false` for a hierarchy, the loop `continue`s without attaching paths for that hierarchy name. This is the documented "all-or-nothing" semantic and is correct. However, consider whether callers need a way to distinguish "this hierarchy has no paths for these IDs" from "this hierarchy does not exist." Currently both cases result in `GetPaths(hierarchy_name)` returning `nullptr`. If the distinction matters for upstream consumers, a separate indicator (e.g., an empty array vs. nullptr) could be added in a follow-up.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,102 @@
...
+        if (not hierarchy->path_store->GetPaths(inner_ids, paths.get())) {
+            continue;
+        }
+        if (hierarchy_name.empty()) {
+            result->Paths(paths.get());
+        } else {
+            result->Paths(hierarchy_name, paths.get());
+        }
+        paths.release();
+    }
```
LHT129, line 91: [suggestion] `paths.release()` transfers ownership of the heap-allocated array to the Dataset. If a subsequent iteration of this loop throws (e.g., `path_store == nullptr` check on the next hierarchy), the already-released arrays would not be freed. In practice this is safe because the `nullptr` checks are invariant once `store_paths_` is true, but consider using a `std::vector<std::unique_ptr<std::string[]>>` guard or restructuring to release all arrays after the loop completes successfully.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -726,14 +778,31 @@ Pyramid::read_streaming_body(StreamReader& reader, const MetadataPtr& metadata)
...
+    if (metadata == nullptr || metadata->EmptyIndex()) {
+        throw VsagException(ErrorType::INDEX_EMPTY, "index is empty");
+    }
+    auto basic_info = metadata->Get(BASIC_INFO);
+    const bool has_paths = basic_info.Contains(PYRAMID_PATHS_FORMAT_VERSION_KEY);
+    if (has_paths != store_paths_) {
```
LHT129: [note] The `has_paths != store_paths_` check enforces strict consistency between serialized data and runtime config: you cannot deserialize a path-enabled index with `store_paths=false`, nor a legacy index with `store_paths=true`. This is the correct conservative choice. However, it means that toggling `store_paths` requires a full rebuild — there is no migration path. If backward compatibility (reading legacy indexes into a path-enabled config, ignoring missing paths) is ever desired, this check would need to become a warning rather than an error.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,158 @@
...
+PyramidPathStore::Record(const std::string* paths,
+                         const Vector<int64_t>& data_biases,
+                         int64_t first_inner_id) {
+    CHECK_ARGUMENT(paths != nullptr, "paths must not be null");
+    CHECK_ARGUMENT(first_inner_id >= 0, "first inner id must not be negative");
+    if (data_biases.empty()) {
+        return;
+    }
+
+    const auto begin = static_cast<uint64_t>(first_inner_id);
```
LHT129: [suggestion] The `Record(const std::string* paths, const Vector<int64_t>& data_biases, int64_t first_inner_id)` overload accesses `paths[data_biases[offset]]` without validating that `data_biases[offset]` is non-negative. While the only production caller (`Pyramid::Add`) constructs biases from loop indices `i` where `0 <= i < data_num`, a negative bias would cause out-of-bounds access. Since `data_biases` is a `Vector<int64_t>` (signed), consider adding `CHECK_ARGUMENT(data_biases[offset] >= 0, "data bias must not be negative")` to make the invariant explicit. (This is distinct from the earlier discussion about upper-bound validation — the sign check is zero-cost in practice and protects against a different class of bugs.)

src/algorithm/pyramid/pyramid.h
```diff
@@ -16,9 +16,11 @@
 #pragma once
 
 #include <memory>
+#include <shared_mutex>
```
LHT129, line 19: [note] `<shared_mutex>` is included here in `pyramid.h` but is only used by `PyramidPathStore` (which has its own include). The include in `pyramid.h` is redundant since `pyramid_path_store.h` already pulls it in transitively. Consider removing it to keep the include list minimal.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -228,6 +231,14 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     if (create_new_raw_vector_) {
         raw_vector_->BatchInsertVector(data_vectors, data_num);
     }
+    if (store_paths_) {
```
LHT129, line 231: [suggestion] In `build_by_odescent`, the path recording loop iterates all hierarchies and calls `Record(paths, data_num)`. If `base->GetPaths(hierarchy_name)` returns `nullptr` for a hierarchy, the path store for that hierarchy remains empty (no paths recorded). Later, `GetDataByIdsWithFlag` will call `GetPaths` which returns `false` for all queries on that hierarchy. Is this the intended behavior? If paths are expected to always be present for all hierarchies when `store_paths=true`, consider adding a `CHECK_ARGUMENT` or at least a warning log when `GetPaths` returns `nullptr` during build.

src/algorithm/pyramid/pyramid.h
jac0626: No code change because the include is not redundant. `pyramid.h` itself declares `std::shared_mutex` members for the per-node mutex and `resize_mutex_`; it must include `<shared_mutex>` directly rather than rely on the path-store header transitively.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: No code change after tracing the complete caller path. `data_biases` is an internal vector built only from the `Pyramid::Add` loop offsets in `[0, data_num)`, and those same offsets already index the source vectors before reaching `PyramidPathStore::Record`. Rechecking each item here would add a branch to the opted-in Add hot path without establishing a new input boundary.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: No code change. The current Dataset contract intentionally represents both an unavailable named path attachment and an unknown hierarchy as `nullptr`; this PR only adds optional path attachments and does not add new per-hierarchy status metadata. Distinguishing those states would require a separate public API and can be considered independently if a caller needs it.

src/algorithm/pyramid/pyramid.cpp
jac0626: This is intentional sparse-hierarchy behavior. A dataset may provide paths for only some configured hierarchies; those stores remain independently queryable, while a hierarchy without complete retained data is omitted from the result. The functional test covering partial named paths verifies that `site` is returned while `category` remains `nullptr`. Rejecting the build would remove that supported partial-input case.

src/algorithm/pyramid/pyramid.cpp
jac0626: Agreed; the strict match is intentional. `store_paths` controls whether a physical sidecar is present, so accepting a mismatch would either leave unread bytes or silently return incomplete path data. Changing this option therefore requires rebuilding. Commit `59242ae1` now derives the serialized value from the existing `INDEX_PARAM` / `store_paths` contract instead of a private path-format key.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: No code change because ownership is already exception-safe. Before `paths.release()`, the local `unique_ptr` owns the array. After `Dataset::Paths` succeeds, `DatasetImpl` owns the pointer and its destructor releases both default and named path arrays, including during stack unwinding if a later hierarchy throws. If `Dataset::Paths` itself throws, `release()` has not run and the local guard still frees the array.

jac0626:
Updated in `a5da8efd` with the same low-coupling write boundary as main:

- `PyramidPathStore` no longer receives input offsets, a first ID, or Build/Add semantics.
- `Pyramid` retains each actually allocated `inner_id`; the store-owned Writer receives only `(inner_id, path)`.
- One Writer lock covers the batch, avoiding the per-path locking regression while remaining safe for concurrent Add.
- Writer is non-copyable/non-movable and dual-vector growth restores the previous size on allocation failure.
- Added NSW/ODescent, middle-duplicate, out-of-order/hole, and concurrent Add mapping coverage.

The main-branch Release A/B for this Writer design was +1.09% at 1K unique paths and +0.14% at 100K unique paths versus the previous bulk implementation, with less than 0.8 MiB RSS difference and identical serialized sizes.

Local targeted result: 18 PathStore unit assertions + 280 Pyramid path functional assertions passed; `clang-format-15` and `git diff --check` passed.

LHT129 (commented):
[note] In `Deserialize` (`src/algorithm/pyramid/pyramid.cpp`), `body_length` is computed as `reader.Length() - footer->Length()`. Both are `uint64_t`, so if `footer->Length()` exceeded `reader.Length()` (which should never happen after a successful `Footer::Parse`), this would underflow. Consider adding `CHECK_ARGUMENT(footer->Length() <= reader.Length(), "footer exceeds reader length")` to make the invariant explicit and protect against future changes to `Footer::Parse`.

Overall assessment: This is a well-structured backport PR with comprehensive test coverage (15 functional test cases, 4 unit test cases, plus parameter and serialization tests). The implementation correctly isolates path storage in `PyramidPathStore`, properly handles concurrent Add operations, validates serialization contract consistency, and preserves backward compatibility when `store_paths=false`. All previously identified issues have been addressed. No blocking issues found.

jac0626:
Final minimal-scope update is pushed in 0bd344d2.

Implementation:
- Path retention remains opt-in through store_paths=false by default; DATA_FLAG_PATH is required to return paths.
- PathStore is a separate component and accepts only allocated inner IDs plus paths.
- 1.0 uses only the direct dense SoA fields: Vector<uint64_t> offsets, Vector<uint16_t> counts, and a flat Vector<string> paths.
- Rows support 0/1/N paths for the later #2796 integration. There is no map, sparse/adaptive index, row-state protocol, magic, or private path format version.
- Serialization writes those three fields directly with the existing StreamWriter primitives.

Scope cleanup in this update:
- Removed the unused get_data_by_ids helper while preserving the original virtual GetDataByIds dispatch.
- Restored the existing footer/BufferStreamReader flow; only path reading is appended after the hierarchy data.
- Removed test-only Size APIs, duplicate full scans in GetPaths, temporary output copies, hierarchy bookkeeping beyond the existing pattern, and incomplete flat-pool total-count validation.
- Kept bounded length/range checks that prevent malformed input from causing unbounded allocation.

Validation:
- Pyramid UT: 570 assertions / 11 cases
- Path FT: 280 assertions / 16 cases

The main-branch Release A/B remains representative because the in-memory writer is identical: versus the prior single-path implementation, build time was +1.27% at 1K unique paths and +1.86% at 100K unique paths; peak build RSS was +1.16 MiB and +1.43 MiB. Direct serialization was 1.90% and 0.46% faster than the prior unmerged row-state draft, while the direct dense wire format adds 900,016 bytes at 100K slots.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+void
+PyramidPathStore::Writer::Insert(InnerIdType inner_id, const std::string& path) {
+    Insert(inner_id, &path, 1);
+}
+
+void
+PyramidPathStore::Writer::Insert(InnerIdType inner_id,
```
LHT129, line 154: [suggestion] The `Prepare` call inside `Insert` means every `Insert` invocation triggers `reserve_paths_geometrically`. When the caller already called `Prepare(N, M)` upfront (as done in `pyramid.cpp` lines 236 and 918), the geometric growth check in subsequent `Insert`-internal `Prepare` calls is effectively a no-op because capacity is already sufficient.

However, for callers that skip the explicit `Prepare` and only call `Insert` in a loop, each `Insert` calls `reserve_paths_geometrically` which only grows when `required_capacity > current_capacity`. Since each `Insert` adds at most `path_count` new paths (typically 1), the geometric doubling may not trigger frequently enough, leading to more reallocations than necessary.

Consider documenting that callers should call `Prepare` before a batch of `Insert` calls for optimal performance.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,308 @@
...
+    for (uint64_t inner_id = 0; inner_id < slot_count; ++inner_id) {
+        uint8_t raw_state = 0;
+        StreamReader::ReadObj(reader, raw_state);
+        const auto state = static_cast<PathRowState>(raw_state);
+        if (state == PathRowState::HOLE) {
+            continue;
+        }
+
+        const auto path_offset = static_cast<uint64_t>(restored_paths.size());
+        if (state == PathRowState::SINGLE_PATH) {
```
LHT129: [note] The early-bound check at this line uses `sizeof(uint64_t)` as the minimum per-path size:
```cpp
path_count > (path_reader_length - path_cursor) / sizeof(uint64_t)
```
Each serialized path string consists of a `uint64_t` length prefix plus the actual payload bytes, so the minimum size per path is `sizeof(uint64_t) + 0 = 8` bytes. The current check would allow `path_count` strings that collectively require at least `path_count * 8` bytes, but the actual minimum is the same. While this is not a correctness bug (the real validation happens inside `ReadPyramidPathString`), using `sizeof(uint64_t)` as the divisor is coincidentally correct rather than intentionally correct. Consider using a named constant or adding a comment to clarify the intent.

src/algorithm/pyramid/pyramid_path_store.cpp
```diff
@@ -0,0 +1,308 @@
...
+    }
+    std::vector<std::vector<std::string>> restored_rows;
+    restored_rows.reserve(inner_ids.size());
+    for (const auto inner_id : inner_ids) {
+        const auto slot = static_cast<uint64_t>(inner_id);
+        if (slot >= offsets_.size()) {
+            return false;
+        }
+        const auto path_offset = offsets_[slot];
+        const auto path_count = static_cast<uint64_t>(counts_[slot]);
```
LHT129, line 223: [note] `GetPathRows` returns `false` on failure but the output parameter `path_rows` is left in whatever state the internal `restored_rows` was in before the failure was detected. The test at line 103 (`REQUIRE(restored_rows == expected_rows)`) relies on this behavior — it expects `path_rows` to be unchanged from the previous successful call.

This is an implicit contract: on failure, `path_rows` retains its prior value (because `path_rows = std::move(restored_rows)` only executes on the success path). Consider documenting this behavior in the header, or alternatively clearing `path_rows` on failure to avoid callers accidentally using stale data.

jac0626: Addressed in 0bd344d2. GetPathRows now clears/reserves the caller output and fills it directly; the implicit unchanged-on-failure contract and its test were removed. This also avoids holding the previous output and a full temporary copy at the same time.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: No code change here. All Pyramid build/add call sites Prepare the complete slot and path-membership counts once before the loop, so the Insert-internal call does not resize or reserve for those batches. It remains the geometric fallback that makes the one public Insert API safe for direct and out-of-order callers. A second InsertPrepared API would add a caller-visible capacity precondition for only a few no-op checks, which is outside the final minimal design.

src/algorithm/pyramid/pyramid_path_store.cpp
jac0626: Confirmed; no change is needed. Eight bytes is exactly the minimum serialized size of one path because an empty string still has one uint64 length prefix. Payload bytes can only make the actual requirement larger, and each individual length is checked before allocation/read.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+void
+Pyramid::deserialize_paths(StreamReader& reader, uint64_t max_count) {
+    uint64_t hierarchy_count = 0;
+    StreamReader::ReadObj(reader, hierarchy_count);
+    if (hierarchy_count != hierarchies_.size()) {
+        throw VsagException(ErrorType::READ_ERROR, "corrupted Pyramid path hierarchy count");
+    }
+
+    for (uint64_t offset = 0; offset < hierarchy_count; ++offset) {
```
LHT129, line 47: [note] The `deserialize_paths` method does not check for duplicate hierarchy names in the serialized stream. If a malformed/corrupted serialized payload contains the same hierarchy name twice, the second occurrence would silently overwrite the first deserialized path store without any error. The previous version of this code (before the refactor) used an UnorderedSet to guard against duplicates. Consider adding a similar duplicate-detection check here to fail fast on corrupted data rather than silently accepting it.

LHT129 (commented):
## PR Review Summary

This PR backports `store_paths` support to Pyramid, enabling `GetDataByIdsWithFlag` to return original path labels. The implementation is well-structured and follows existing patterns in the codebase.

### What changed
- New `PyramidPathStore` class with RAII `Writer` pattern for thread-safe path storage
- New `pyramid_paths.cpp` with serialization and `GetDataByIdsWithFlag`
- `store_paths` parameter (default false) added to `PyramidParameters`
- Streaming serialization tag `PYRAMID_PATHS = 18`
- `InnerIndexInterface` refactored to expose `get_data_by_ids_with_flag` as protected
- Comprehensive test coverage: `pyramid_path_store_test`, `test_pyramid_path_serialization`, `test_pyramid_paths`

### Review findings
1. **[note] `deserialize_paths` missing duplicate hierarchy name check** (see inline comment on `pyramid_paths.cpp:47`): The method does not guard against duplicate hierarchy names in corrupted serialized data. The previous version used an `UnorderedSet` for this check.

### Overall assessment
The implementation is solid: correct use of RAII, proper error handling, good test coverage, and clean integration with existing serialization infrastructure. No blocking issues found.

PR #2766, opened 2026-08-25
[Backport 0.18] feat(pyramid): support paths in GetDataByIdsWithFlag

jac0626:
## Change Type

- [x] Improvement

## Linked Issue

Related to #2754. Backport of #2763 to the 0.18 branch.

## What Changed

- Add store_paths as an opt-in Pyramid parameter, disabled by default.
- Keep the 0.18 implementation single-path and return a path only when GetDataByIdsWithFlag selects DATA_FLAG_PATH.
- Store paths directly by allocated inner ID in the separate PyramidPathStore.
- Pre-size path slots once for Build/Add batches and remove the redundant ODescent reverse-label remap rebuild; LabelTable::Insert already maintains it.
- Append the path sidecar before the existing footer only when enabled. INDEX_PARAM is also added only for enabled indexes, so default-off serialization remains on the prior body/footer path.
- No multi-path SoA, named hierarchy, streaming, map, sparse/adaptive index, magic, or private version is backported.

- Default-off creates no path store, allocates no per-vector path data, and writes neither a sidecar nor the added parameter payload.
- One Writer lock covers each enabled batch; EnsureSlots avoids incremental slot-vector growth.

## Validation

- Pyramid unit tests: 584 assertions / 10 cases.
- Pyramid path functional tests: 118 assertions / 10 cases.

### DIFF
--- a/src/algorithm/inner_index_interface.cpp
+++ b/src/algorithm/inner_index_interface.cpp
@@ -446,6 +446,14 @@ InnerIndexInterface::ExportIDs() const {
 
 DatasetPtr
 InnerIndexInterface::GetDataByIds(const int64_t* ids, int64_t count) const {
+    Vector<InnerIdType> inner_ids(allocator_);
+    return this->get_data_by_ids(ids, count, inner_ids);
+}
+
+DatasetPtr
+InnerIndexInterface::get_data_by_ids(const int64_t* ids,
+                                     int64_t count,
+                                     Vector<InnerIdType>& inner_ids) const {
     uint64_t selected_flag = DATA_FLAG_ID;
     if (this->has_raw_vector_) {
         selected_flag |= DATA_FLAG_FLOAT32_VECTOR;
@@ -456,22 +464,29 @@ InnerIndexInterface::GetDataByIds(const int64_t* ids, int64_t count) const {
     if (this->extra_info_size_ > 0) {
         selected_flag |= DATA_FLAG_EXTRA_INFO;
     }
-    return this->GetDataByIdsWithFlag(ids, count, selected_flag);
+    return this->get_data_by_ids_with_flag(ids, count, selected_flag, inner_ids);
 }
 
 DatasetPtr
 InnerIndexInterface::GetDataByIdsWithFlag(const int64_t* ids,
                                           int64_t count,
                                           uint64_t selected_data_flag) const {
-    if ((selected_data_flag & DATA_FLAG_FLOAT32_VECTOR) != 0U && not this->has_raw_vector_) {
-        throw VsagException(ErrorType::INVALID_ARGUMENT, "has_raw_vector_ is false");
-    }
-    auto* inner_ids =
-        reinterpret_cast<InnerIdType*>(this->allocator_->Allocate(count * sizeof(InnerIdType)));
+    Vector<InnerIdType> inner_ids(allocator_);
+    return this->get_data_by_ids_with_flag(ids, count, selected_data_flag, inner_ids);
+}
+
+DatasetPtr
+InnerIndexInterface::get_data_by_ids_with_flag(const int64_t* ids,
+                                               int64_t count,
+                                               uint64_t selected_data_flag,
+                                               Vector<InnerIdType>& inner_ids) const {
+    CHECK_ARGUMENT(count >= 0, "count must not be negative");
+    inner_ids.clear();
+    inner_ids.reserve(static_cast<uint64_t>(count));
     {
         std::shared_lock lock(this->label_lookup_mutex_);
         for (int64_t i = 0; i < count; ++i) {
-            inner_ids[i] = this->label_table_->GetIdByLabel(ids[i]);
+            inner_ids.emplace_back(this->label_table_->GetIdByLabel(ids[i]));
         }
     }
     auto thread_count = static_cast<int64_t>(this->build_thread_count_);
@@ -484,8 +499,7 @@ InnerIndexInterface::GetDataByIdsWithFlag(const int64_t* ids,
         dataset->Float32Vectors(fp32_data);
         auto get_vec_func = [&](int64_t begin, int64_t end) {
             for (int64_t i = begin; i < end; ++i) {
-                auto inner_id = this->label_table_->GetIdByLabel(ids[i]);
-                this->GetVectorByInnerId(inner_id, fp32_data + i * this->dim_);
+                this->GetVectorByInnerId(inner_ids[i], fp32_data + i * this->dim_);
             }
         };
         if (this->thread_pool_ != nullptr) {
@@ -511,8 +525,7 @@ InnerIndexInterface::GetDataByIdsWithFlag(const int64_t* ids,
         dataset->AttributeSets(attribute_data);
         auto get_attr_func = [&](int64_t begin, int64_t end) {
             for (int64_t i = begin; i < end; ++i) {
-                auto inner_id = this->label_table_->GetIdByLabel(ids[i]);
-                this->GetAttributeSetByInnerId(inner_id, attribute_data + i);
+                this->GetAttributeSetByInnerId(inner_ids[i], attribute_data + i);
             }
         };
         if (this->thread_pool_ != nullptr) {
@@ -539,8 +552,8 @@ InnerIndexInterface::GetDataByIdsWithFlag(const int64_t* ids,
         dataset->ExtraInfos(extra_info);
         auto get_extra_info_func = [&](int64_t begin, int64_t end) {
             for (int64_t i = begin; i < end; ++i) {
-                auto inner_id = this->label_table_->GetIdByLabel(ids[i]);
-                this->extra_infos_->GetExtraInfoById(inner_id, extra_info + i * extra_info_size_);
+                this->extra_infos_->GetExtraInfoById(inner_ids[i],
+                                                     extra_info + i * extra_info_size_);
             }
         };
         if (this->thread_pool_ != nullptr) {
@@ -564,7 +577,6 @@ InnerIndexInterface::GetDataByIdsWithFlag(const int64_t* ids,
         memcpy(new_ids, ids, count * sizeof(int64_t));
         dataset->Ids(new_ids);
     }
-    this->allocator_->Deallocate(inner_ids);
     return dataset;
 }
 
--- a/src/algorithm/inner_index_interface.cpp
+++ b/src/algorithm/inner_index_interface.cpp
@@ -446,14 +446,6 @@ InnerIndexInterface::ExportIDs() const {
 
 DatasetPtr
 InnerIndexInterface::GetDataByIds(const int64_t* ids, int64_t count) const {
-    Vector<InnerIdType> inner_ids(allocator_);
-    return this->get_data_by_ids(ids, count, inner_ids);
-}
-
-DatasetPtr
-InnerIndexInterface::get_data_by_ids(const int64_t* ids,
-                                     int64_t count,
-                                     Vector<InnerIdType>& inner_ids) const {
     uint64_t selected_flag = DATA_FLAG_ID;
     if (this->has_raw_vector_) {
         selected_flag |= DATA_FLAG_FLOAT32_VECTOR;
@@ -464,7 +456,7 @@ InnerIndexInterface::get_data_by_ids(const int64_t* ids,
     if (this->extra_info_size_ > 0) {
         selected_flag |= DATA_FLAG_EXTRA_INFO;
     }
-    return this->get_data_by_ids_with_flag(ids, count, selected_flag, inner_ids);
+    return this->GetDataByIdsWithFlag(ids, count, selected_flag);
 }
 
 DatasetPtr
--- a/src/algorithm/inner_index_interface.cpp
+++ b/src/algorithm/inner_index_interface.cpp
@@ -480,6 +480,9 @@ InnerIndexInterface::get_data_by_ids_with_flag(const int64_t* ids,
                                                int64_t count,
                                                uint64_t selected_data_flag,
                                                Vector<InnerIdType>& inner_ids) const {
+    if ((selected_data_flag & DATA_FLAG_FLOAT32_VECTOR) != 0U && not this->has_raw_vector_) {
+        throw VsagException(ErrorType::INVALID_ARGUMENT, "has_raw_vector_ is false");
+    }
     CHECK_ARGUMENT(count >= 0, "count must not be negative");
     inner_ids.clear();
     inner_ids.reserve(static_cast<uint64_t>(count));
--- a/src/algorithm/pyramid.cpp
+++ b/src/algorithm/pyramid.cpp
@@ -230,7 +230,10 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
         for (uint64_t inner_id = 0; inner_id < static_cast<uint64_t>(data_num); ++inner_id) {
             label_table_->InsertRemap(data_ids[inner_id], static_cast<InnerIdType>(inner_id));
         }
-        path_store_->Record(path, static_cast<uint64_t>(data_num));
+        auto writer = path_store_->AcquireWriter();
+        for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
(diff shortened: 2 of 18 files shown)


src/algorithm/pyramid_path_store.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "storage/stream_writer.h"
+#include "vsag_exception.h"
+
+namespace vsag {
+namespace {
+
+std::string
+read_path_string(StreamReader& reader) {
+    size_t length = 0;
+    StreamReader::ReadObj(reader, length);
```
LHT129: [suggestion] `size_t` should be replaced with `uint64_t` per the VSAG coding standard to avoid macOS compilation issues.

The VSAG coding standard explicitly requires using `uint64_t` instead of `size_t`. Here `length` is declared as `size_t` and then read via `StreamReader::ReadObj`, which writes a fixed-width value. Using `size_t` creates a portability hazard on platforms where `size_t` differs from `uint64_t`.

Suggested fix:
```cpp
uint64_t length = 0;
StreamReader::ReadObj(reader, length);
```

src/algorithm/pyramid_path_store.cpp
```diff
@@ -0,0 +1,204 @@
...
+    {
+        std::shared_lock lock(label_lookup_mutex_);
+        for (int64_t offset = 0; offset < count; ++offset) {
+            inner_ids[offset] = label_table_->GetIdByLabel(ids[offset]);
+        }
+    }
+
+    auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+    if (path_store_ != nullptr && path_store_->GetPaths(inner_ids, paths.get())) {
+        result->Paths(paths.release());
```
LHT129: [suggestion] `GetPaths` returning `false` is silently ignored, which could mask data corruption or lookup failures.

When `path_store_->GetPaths(inner_ids, paths.get())` returns `false` (e.g., due to a hole in `has_path_` or an out-of-range inner ID), the code silently skips calling `result->Paths(...)`, leaving the dataset without any paths. The caller would get a dataset with vectors but no paths, with no indication that something went wrong.

Consider logging a warning or returning an error when `GetPaths` fails, so the caller can distinguish between "paths not stored" and "path lookup failed for some IDs":

```cpp
if (path_store_ != nullptr) {
    if (not path_store_->GetPaths(inner_ids, paths.get())) {
        logger::warn("GetDataByIds: path lookup failed for one or more IDs");
        return result;
    }
    result->Paths(paths.release());
}
```

src/algorithm/pyramid.cpp
```diff
@@ -438,12 +456,17 @@ void
 Pyramid::Deserialize(StreamReader& reader) {
     // try to deserialize footer (only in new version)
     auto footer = Footer::Parse(reader);
+    if (footer == nullptr) {
```
LHT129: [note] The new `footer == nullptr` check changes deserialization behavior for indexes serialized before the footer format was introduced (pre-v0.15).

The original code had a comment "try to deserialize footer (only in new version)" and silently continued when `footer` was null. The new code throws `READ_ERROR` instead, which means old-format indexes that previously deserialized successfully will now fail.

If this is intentional (i.e., the 0.18 branch no longer needs to support pre-footer formats), consider updating the comment to reflect the new behavior. If backward compatibility with pre-v0.15 indexes is still desired, this check should be reverted.

src/algorithm/pyramid_path_store.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+uint64_t
+PyramidPathStore::Size() const {
+    std::shared_lock lock(mutex_);
+    return paths_by_inner_id_.size();
+}
+
+void
+Pyramid::serialize_paths(StreamWriter& writer) const {
+    if (path_store_ == nullptr) {
```
LHT129: [suggestion] `Pyramid::GetDataByIds` and the path serialization helpers are defined in `pyramid_path_store.cpp`, which includes `pyramid.h`. This creates a circular dependency where `pyramid_path_store.cpp` depends on the full `Pyramid` class definition.

Consider moving `Pyramid::GetDataByIds`, `Pyramid::serialize_paths`, `Pyramid::deserialize_paths`, and `Pyramid::validate_paths` into `pyramid.cpp` instead. The `PyramidPathStore` class is self-contained and does not need to know about `Pyramid`. Keeping these `Pyramid` member functions in `pyramid_path_store.cpp` blurs the module boundary and makes the include of `pyramid.h` from `pyramid_path_store.cpp` necessary solely for these four functions.

jac0626: Addressed. `pyramid_path_store.cpp` is now self-contained, while `Pyramid::GetDataByIds` and its path serialization helpers live in the dedicated `pyramid_paths.cpp` translation unit.

src/algorithm/pyramid_path_store.cpp
jac0626: Fixed. On this single-hierarchy branch, enabled path storage now treats a missing path as an explicit `INTERNAL_ERROR` instead of silently returning a dataset without paths. A corrupted-sidecar query regression covers this case.

src/algorithm/pyramid.cpp
jac0626: The current 0.18 baseline already dereferenced the parsed footer unconditionally; the legacy fallback had been removed earlier, so a null footer was not successfully supported. The explicit check now returns a clean `READ_ERROR` instead of dereferencing null, and the stale comment was updated to state that the current format requires a footer.

src/algorithm/pyramid_path_store.cpp
jac0626: Fixed on both sides of the format. Path lengths are now written and read as fixed-width `uint64_t`, with a remaining-payload and string-size bound check. Roundtrip and oversized-length regression tests were added.

LHT129 (commented):
I reviewed the latest commits (50f65d0d) after the previous round of feedback was addressed. All four issues raised by @LHT129 have been resolved:

1. **`size_t` → `uint64_t`**: `read_path_string` now uses `uint64_t` for the length field, with an additional bounds check against `std::string::size_type::max()`. A symmetric `write_path_string` helper was added for serialization.

2. **Silent `GetPaths` failure**: `Pyramid::GetDataByIdsWithFlag` now throws `INTERNAL_ERROR` when `GetPaths` returns false. A regression test (`"Pyramid reports a retained path missing at query time"`) covers this case.

3. **Footer null check**: The comment was updated to state the current format requires a footer, and the explicit null check provides a clean `READ_ERROR` instead of a null dereference.

4. **Missing `write_path_string`**: Fixed with the symmetric helper.

The refactoring of `InnerIndexInterface::GetDataByIdsWithFlag` to extract `get_data_by_ids_with_flag` accepting a `Vector<InnerIdType>&` is clean — it avoids duplicate label lookups and removes the raw allocator/deallocator pattern in favor of RAII. The `has_raw_vector_` check was correctly moved into the `DATA_FLAG_FLOAT32_VECTOR` block where it belongs.

The test coverage is thorough: unit tests for `PyramidPathStore` (record, holes, serialization roundtrip, fixed-width lengths, malformed input), functional tests for the full Pyramid integration (Build, Add, Clone, serialization, feature flags, error paths), and parameter compatibility tests for `store_paths`.

No new issues found. The implementation is well-structured and the opt-in design (`store_paths: false` by default) ensures zero overhead for existing users.

src/algorithm/inner_index_interface.cpp
```diff
@@ -446,6 +446,14 @@ InnerIndexInterface::ExportIDs() const {
...
 InnerIndexInterface::GetDataByIds(const int64_t* ids, int64_t count) const {
+    Vector<InnerIdType> inner_ids(allocator_);
+    return this->get_data_by_ids(ids, count, inner_ids);
+}
+
+DatasetPtr
+InnerIndexInterface::get_data_by_ids(const int64_t* ids,
+                                     int64_t count,
+                                     Vector<InnerIdType>& inner_ids) const {
     uint64_t selected_flag = DATA_FLAG_ID;
```
LHT129, line 449: [suggestion] The `has_raw_vector_` guard was removed from `get_data_by_ids_with_flag` but equivalent guards for `has_attribute_` and `extra_info_size_` remain. This creates an inconsistency: requesting `DATA_FLAG_FLOAT32_VECTOR` when `has_raw_vector_` is false will now silently proceed to `GetVectorByInnerId`, which may crash or return garbage depending on the subclass.

The original code threw `INVALID_ARGUMENT` here. If the removal is intentional (e.g., Pyramid overrides this method and handles validation upstream), consider adding a comment explaining why. Otherwise, restore the guard for consistency with the attribute and extra-info paths.

src/algorithm/inner_index_interface.cpp
```diff
@@ -446,6 +446,14 @@ InnerIndexInterface::ExportIDs() const {
 
 DatasetPtr
 InnerIndexInterface::GetDataByIds(const int64_t* ids, int64_t count) const {
+    Vector<InnerIdType> inner_ids(allocator_);
+    return this->get_data_by_ids(ids, count, inner_ids);
```
LHT129: [suggestion] The `has_raw_vector_` guard was removed from `GetDataByIdsWithFlag` but the equivalent guards for `has_attribute_` and `extra_info_size_` remain in `get_data_by_ids_with_flag`. This creates an inconsistency: requesting `DATA_FLAG_FLOAT32_VECTOR` when `has_raw_vector_` is false will now silently proceed to `GetVectorByInnerId`, which may crash or return garbage depending on the subclass.

The original code threw `INVALID_ARGUMENT` here. If the removal is intentional (e.g., Pyramid overrides this method and handles validation upstream), consider adding a comment explaining why. Otherwise, restore the guard for consistency with the attribute and extra-info paths.

jac0626: Fixed in 3093345c; this is the duplicate report of the same regression. The shared helper now restores the raw-vector capability guard before any lookup/allocation, and the regression explicitly verifies `INVALID_ARGUMENT` with `has_raw_vector_ is false`.

src/algorithm/inner_index_interface.cpp
jac0626: Fixed in 3093345c. The original `has_raw_vector_` guard is now at the start of the shared `get_data_by_ids_with_flag` helper, so both the base entry point and Pyramid reuse preserve `INVALID_ARGUMENT` before lookup or allocation. The existing Pyramid regression was strengthened to check the error type and message.

LHT129 (commented):
Reviewed commit 3ce10639 ("refactor(pyramid): follow parameter-driven path serialization").

This commit replaces the standalone magic-number + format-version scheme for path serialization with a cleaner approach: the full index parameters (including `store_paths`) are embedded in the footer via `INDEX_PARAM`, and `Deserialize` validates that the serialized `store_paths` value matches the current configuration.

**Changes in 3ce10639 (relative to 3093345c):**

- `src/algorithm/pyramid.cpp`: Removed `PYRAMID_PATHS_MAGIC`, `PYRAMID_PATHS_FORMAT_VERSION`, and the associated version key. Path serialization no longer writes a magic prefix. The footer now stores the full parameter JSON via `INDEX_PARAM` instead of a separate version field. `Deserialize` reads back `store_paths` from the footer and rejects mismatched configurations with a clear `READ_ERROR`.

- `tests/test_pyramid_paths.cpp`: Added `GetPathSidecarOffset` helper for computing the byte offset of the path sidecar within a serialized binary. Added `"Pyramid serialized path storage configuration must match"` test with two sections covering both directions of mismatch (stored paths require enabled reader, and vice versa).

**Assessment:**

The refactoring is sound. Reusing `INDEX_PARAM` for configuration validation avoids introducing a separate versioning mechanism for the path sidecar, and the mismatch check in `Deserialize` provides a clear error rather than silent corruption or undefined behavior. The new tests cover both mismatch directions.

The `BufferStreamReader` / `SliceStreamReader` lifetime in `Deserialize` is correct: `body_reader` (SliceStreamReader, stack-allocated) is declared before `buffer_reader`, so `buffer_reader` is destroyed first while `body_reader` is still alive.

All previously raised issues (size_t→uint64_t, GetPaths error handling, footer null check, circular dependency, has_raw_vector_ guard, path length format) have been addressed in earlier commits.

No new issues found.

jac0626:
Updated in `652957e1` with the same low-coupling write boundary as main:

- `PyramidPathStore` no longer receives input offsets, a first ID, or Build/Add semantics.
- `Pyramid` retains each actually allocated `inner_id`; the store-owned Writer receives only `(inner_id, path)`.
- One Writer lock covers the batch, avoiding the per-path locking regression while remaining safe for concurrent Add.
- Writer is non-copyable/non-movable and dual-vector growth restores the previous size on allocation failure.
- Added middle-duplicate, out-of-order/hole, and concurrent Add path-mapping coverage.

The main-branch Release A/B for this Writer design was +1.09% at 1K unique paths and +0.14% at 100K unique paths versus the previous bulk implementation, with less than 0.8 MiB RSS difference and identical serialized sizes.

Local targeted result: 22 PathStore unit assertions + 136 Pyramid path functional assertions passed; `clang-format-15` and `git diff --check` passed.

LHT129 (commented):
## Review Summary

This review covers commit `652957e1` of the backport PR adding `store_paths` support to Pyramid in the 0.18 branch.

### Changes Reviewed
- 18 files changed, +982 / -39 lines
- New `PyramidPathStore` class for path storage with `Writer` RAII pattern
- New `pyramid_paths.cpp` implementing `GetDataByIdsWithFlag` override in Pyramid
- Refactored `InnerIndexInterface` to expose shared `get_data_by_ids_with_flag` helper
- Serialization/deserialization with footer-based `store_paths` config validation
- Comprehensive test coverage: 11 functional tests + 5 unit tests for path store

### Previously Raised Issues (All Resolved)
- `size_t` → `uint64_t` in `read_path_string`: **Fixed** — now uses `uint64_t` throughout
- Silent `GetPaths` failure: **Fixed** — now throws `INTERNAL_ERROR` when path lookup fails
- Missing `has_raw_vector_` guard: **Fixed** — guard restored in shared helper, regression test strengthened

### Current Assessment
The code in commit `652957e1` is clean and well-structured. The `PyramidPathStore::Writer` RAII pattern correctly holds the exclusive lock. Serialization includes proper bounds checking and config mismatch detection. Test coverage is thorough across Build, Add, Clone, serialization roundtrip, and malformed input scenarios.

No new issues identified in this commit.

jac0626:
Final minimal-scope update is pushed in 793ac787.

- 0.18 remains single-path only; no 0/1/N SoA, map, sparse, adaptive, magic, or private version logic was backported.
- Build/Add pre-size the slot arrays once per batch, and ODescent no longer rebuilds the reverse-label mapping that LabelTable::Insert already maintains.
- PathStore now validates the expected element count directly; the test-only Size API and Pyramid validate_paths wrapper were removed.
- GetDataByIds preserves the original virtual dispatch and shares only the one helper needed to return allocated inner IDs without a second label lookup.
- The existing footer reader flow is preserved. INDEX_PARAM is written only when store_paths=true, so the default-off serialized Pyramid body/footer stays on the prior path.
- Format-coupled byte-tampering tests and the custom fixed-width string codec were removed in favor of the branch existing StreamWriter/StreamReader string primitives.

Validation:
- Pyramid UT: 584 assertions / 10 cases
- Path FT: 118 assertions / 10 cases

LHT129 (commented):
Reviewed the latest commit 1c584e34 ("perf(pyramid): avoid redundant path storage work").

The changes look good:
- Extracting `EnsureSlots` as a public `Writer` method with proper rollback on allocation failure is clean and correct.
- Removing the redundant `label_table_->ResetRemap`/`InsertRemap` calls in `build_by_odescent` (since the label table is already populated just above) is a good cleanup.
- The `EnsureSlots` call before the writer loop in `build_by_odescent` avoids incremental resizes inside `Insert`.
- The `accepted_inner_ids.empty()` guard in `Add` correctly avoids unnecessary writer acquisition when all labels are duplicates.
- Test updates correctly reflect the new `EnsureSlots` API and adjusted size expectations.

All previously raised review comments have been addressed. No new issues found.

LHT129 (approved):
LGTM

PR #2784, opened 2026-08-25
perf(build): narrow datacell Ninja job pool

### LINKED ISSUE #2767: [improve](build): limit the serial datacell job pool to memory-heavy sources
### Kind of improvement

perf

### Affected area

build

### Summary

Limit the capacity-one Ninja job pool to the memory-heavy RaBitQ split factory translation units instead of assigning the entire `datacell` target to that pool.

The current configuration assigns both `rabitq_split_datacell_factory` and `datacell` to `rabitq_split_factory_pool=1`. This serializes compilation of unrelated datacell sources even when the build is invoked with higher parallelism.

Keep a conservative limit for known memory-intensive template instantiations, while allowing ordinary datacell sources to compile in parallel. If memory constraints still require throttling, make the pool size configurable.

### Expected benefit

Restore effective parallelism for the datacell module, shorten the clean-build critical path, and retain protection against excessive compiler memory usage.

### Benchmark evidence

// TODO: Measure clean Ninja builds before and after the change with `-j1`, `-j6`, and host-scaled parallelism. Record datacell wall time, total build wall time, peak RSS, and the slowest translation units.

### References

- `src/datacell/CMakeLists.txt:20`
- `src/datacell/CMakeLists.txt:30`
- `src/datacell/CMakeLists.txt:37`

### Additional context

The repository contains separate metric-specific RaBitQ split factory translation units, so the memory-heavy scope already has a natural target boundary.

## Related issues

_None found._

---

wxyucs:
## What changed

- keep `rabitq_split_factory_pool=1` on the three memory-heavy metric-specific RaBitQ split factory translation units
- remove the serial pool from ordinary `datacell` objects
- add generated-Ninja validation covering every datacell compile edge and the pool capacity

The capacity remains fixed at one. VSAG has no existing convention for making a target-specific Ninja pool configurable, and the observed factory compilation memory cost supports retaining the conservative safeguard. Non-Ninja generators remain unaffected because the pool definition and assignment stay inside the existing Ninja guard.

## Validation

- base: freshly fetched `origin/main` at `81fefe7b30c63f1d47fcd62b9b235882a6d0c982`
- Ninja configure and focused `datacell rabitq_split_datacell_factory` build
- generated-Ninja assignment test: exactly the three RaBitQ split factory objects use the capacity-one pool; all 34 ordinary objects do not
- Unix Makefiles configure: no Ninja job-pool data generated
- `git diff --check`

## Clean module benchmark

GCC 11.4, Debug, Ninja 1.10.1, ccache disabled, 96 logical CPUs, 503 GiB RAM. Dependencies were primed once; each sample deleted only the 37 datacell/factory objects. Peak RSS is the sampled aggregate for Ninja's process tree. These are single runs on a shared host, so they show the parallelism effect but are not presented as general build-performance guarantees.

| Requested parallelism | Before wall / datacell | After wall / datacell | Before peak RSS | After peak RSS |
| --- | ---: | ---: | ---: | ---: |
| `-j1` | 346.4 s / 346.4 s | 358.0 s / 358.0 s | 4.22 GB | 4.22 GB |
| `-j6` | 359.0 s / 359.0 s | 136.8 s / 136.7 s | 4.22 GB | 5.67 GB |
| `-j96` | 342.6 s / 342.6 s | 109.2 s / 109.1 s | 4.22 GB | 9.01 GB |

The slowest translation unit in every sample was `flatten_interface.cpp` (103.8–109.1 s). Other slow ordinary units were the bucket IO variants (roughly 19–27 s). The unchanged `-j1` behavior is the expected control; higher parallelism no longer serializes ordinary datacell compilation, while the three split factory compilations remain mutually exclusive.

Fixes: #2767

### DIFF
--- a/tests/cmake/datacell_job_pool_test.cmake
+++ b/tests/cmake/datacell_job_pool_test.cmake
@@ -1,0 +1,59 @@
+if (NOT DEFINED VSAG_SOURCE_DIR)
+    message (FATAL_ERROR "VSAG_SOURCE_DIR is required")
+endif ()
+if (NOT DEFINED VSAG_BUILD_DIR)
+    message (FATAL_ERROR "VSAG_BUILD_DIR is required")
+endif ()
+
+set (ninja_file "${VSAG_BUILD_DIR}/build.ninja")
+set (rules_file "${VSAG_BUILD_DIR}/CMakeFiles/rules.ninja")
+if (NOT EXISTS "${ninja_file}" OR NOT EXISTS "${rules_file}")
+    message (FATAL_ERROR "VSAG_BUILD_DIR must contain a Ninja build")
+endif ()
+
+file (READ "${ninja_file}" ninja_content)
+file (READ "${rules_file}" rules_content)
+
+function (assert_object_pool source target expect_pool)
+    set (object_path "src/datacell/CMakeFiles/${target}.dir/${source}.o")
+    string (FIND "${ninja_content}" "build ${object_path}:" block_start)
+    if (block_start EQUAL -1)
+        message (FATAL_ERROR "Ninja object rule not found: ${object_path}")
+    endif ()
+
+    string (SUBSTRING "${ninja_content}" ${block_start} -1 remaining_content)
+    string (FIND "${remaining_content}" "\n\n" block_length)
+    if (block_length EQUAL -1)
+        message (FATAL_ERROR "Ninja object rule is not terminated: ${object_path}")
+    endif ()
+    string (SUBSTRING "${remaining_content}" 0 ${block_length} object_rule)
+    string (FIND "${object_rule}" "pool = rabitq_split_factory_pool" pool_position)
+
+    if (expect_pool AND pool_position EQUAL -1)
+        message (FATAL_ERROR "Serial pool is missing from ${object_path}")
+    elseif (NOT expect_pool AND NOT pool_position EQUAL -1)
+        message (FATAL_ERROR "Ordinary datacell object uses the serial pool: ${object_path}")
+    endif ()
+endfunction ()
+
+set (rabitq_split_factory_sources
+     rabitq_split_datacell_factory_l2.cpp
+     rabitq_split_datacell_factory_ip.cpp
+     rabitq_split_datacell_factory_cosine.cpp)
+foreach (source IN LISTS rabitq_split_factory_sources)
+    assert_object_pool ("${source}" rabitq_split_datacell_factory TRUE)
+endforeach ()
+
+file (GLOB datacell_sources RELATIVE "${VSAG_SOURCE_DIR}/src/datacell"
+      "${VSAG_SOURCE_DIR}/src/datacell/*.cpp")
+list (FILTER datacell_sources EXCLUDE REGEX "_test\\.cpp$")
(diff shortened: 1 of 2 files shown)


tests/cmake/datacell_job_pool_test.cmake
```diff
@@ -0,0 +1,59 @@
+if (NOT DEFINED VSAG_SOURCE_DIR)
```
LHT129, line 1: [suggestion] This test file is not wired into any build target or CI workflow. Unlike `tests/cmake/thirdparty_override_test.cmake` which is invoked via `make test-cmake` in the Makefile, this new test has no corresponding entry point. Without integration, the test will never be executed automatically and risks becoming stale.

Consider adding a `test-cmake` target entry in the Makefile, e.g.:
```makefile
cmake -DVSAG_SOURCE_DIR=${CURDIR} -DVSAG_BUILD_DIR=${DEBUG_BUILD_DIR} -P tests/cmake/datacell_job_pool_test.cmake
```

Or alternatively, integrate it into an existing CI workflow so the generated-Ninja validation runs on every relevant change.

tests/cmake/datacell_job_pool_test.cmake
```diff
@@ -0,0 +1,59 @@
...
+set (ninja_file "${VSAG_BUILD_DIR}/build.ninja")
+set (rules_file "${VSAG_BUILD_DIR}/CMakeFiles/rules.ninja")
+if (NOT EXISTS "${ninja_file}" OR NOT EXISTS "${rules_file}")
+    message (FATAL_ERROR "VSAG_BUILD_DIR must contain a Ninja build")
+endif ()
+
+file (READ "${ninja_file}" ninja_content)
+file (READ "${rules_file}" rules_content)
+
+function (assert_object_pool source target expect_pool)
```
LHT129, line 17: [note] The `assert_object_pool` function references `${ninja_content}` from the parent scope rather than receiving it as a parameter. While this works in CMake (functions inherit variables from the calling scope), passing it explicitly would make the dependency clearer and the function more self-contained.

Consider adding `ninja_content` as a function parameter:
```cmake
function (assert_object_pool ninja_content source target expect_pool)
```

tests/cmake/datacell_job_pool_test.cmake
wxyucs: Thanks for flagging this. This is intentionally a focused generated-build validator rather than part of `make test-cmake`: the Makefile defaults to Unix Makefiles, while this script requires and deliberately rejects anything other than an already configured Ninja build. The PR runs and documents the exact validation command. Adding a reliable configure fixture or CI workflow would broaden this scheduling-only change, so I am keeping the narrow manual/static entry point here and resolving the non-blocking suggestion.

tests/cmake/datacell_job_pool_test.cmake
wxyucs: Thanks for the note. The parent-scope read is deliberate CMake function scoping, and avoids passing the full generated Ninja file contents as a function argument for every object assertion. The validator has exercised this path successfully; there is no functional ambiguity at the call sites, so I am keeping the narrow implementation and resolving the note.

LHT129 (approved):
LGTM

PR #2796, opened 2026-08-27
feat(pyramid): support multiple paths per vector

### LINKED ISSUE #2794: [feat](other): support multiple paths per vector in a named Pyramid hierarchy
### Affected area

other

### Motivation

Pyramid supports multiple named hierarchies, but the current dataset API accepts one path string per vector for each hierarchy.

A document may belong to multiple values within the same hierarchy. For example, document `123` may have both `technology` and `military` tags. The same document should therefore be reachable through either query:

- hierarchy `tag`, path `technology`
- hierarchy `tag`, path `military`

This is different from query-side multi-path search: the requested capability is to associate one indexed vector with multiple paths in the same named hierarchy during `Build` and `Add`.

The current API exposes:

```cpp
DatasetPtr Paths(const std::string& hierarchy_name,
                 const std::string* paths);
```

where `paths` contains one string per dataset element. It cannot represent multiple independent paths for one element without relying on an external delimiter convention.

### Proposed solution

Add an API-compatible representation that allows each vector to provide zero, one, or multiple paths for every named hierarchy.

Conceptually:

```cpp
// id 123 belongs to both "technology" and "military" in hierarchy "tag".
dataset->Paths("tag", {
    {"technology", "military"},
    {"sports"},
});
```

The exact public representation may use nested containers or an offsets-plus-flattened-paths layout.

Expected behavior:

1. During `Build` and `Add`, a vector is attached to every specified path node in that hierarchy.
2. The vector data and label are stored once and shared by those nodes.
3. Searching either path returns the same document ID.
4. Duplicate paths supplied for one vector are handled deterministically without duplicating the result.
5. Multiple named hierarchies remain independent; one vector may have multiple paths under `tag` and one path under `host`.
6. Copying, ownership handling, serialization, deserialization, and incremental `Add` preserve all path assignments.
7. The existing single-path `Paths(...)` API remains supported.

### Alternatives considered

- Duplicate the same vector and label once for every path. This wastes storage and conflicts with duplicate-label handling.
- Encode multiple values in one string using a delimiter such as `technology,military`. This makes delimiter parsing part of every integration, introduces escaping ambiguity, and does not represent the data model explicitly.
- Expand the values only in HA3. This could provide a product-specific workaround but would still require VSAG to accept multiple path assignments for the same vector.

additive only (no break)

### References

- `include/vsag/dataset.h:330` — the named-hierarchy API currently accepts one path array.
- `tests/test_pyramid.cpp:1193` — the current multi-hierarchy fixture stores one path per vector per hierarchy.
- `tests/test_pyramid.cpp:1209` — the fixture passes one `std::string` array for each named hierarchy.
- https://github.com/antgroup/vsag/issues/2097
- https://github.com/antgroup/vsag/issues/2100
- https://github.com/antgroup/vsag/issues/1947

### Additional context

This capability is needed by integrations that model multi-valued document attributes such as tags. It applies to Pyramid's named-hierarchy API, which is also used to provide MultiPyramid behavior at the integration layer.

## Related issues

- #2097 — [feat](multiPyramid): define pyramid named hierarchy api parameters
- #2100 — [feat](multiPyramid): support pyramid multi hierarchy build add and single hierarchy search
- #1947 — [Feature Request] Pyramid: share a single vector base (FlattenInterface) across multiple path hierarchies

---

jac0626:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2794

## What Changed

- Add an ABI-safe non-virtual `Dataset::Paths` overload that stores one or multiple paths per element in a named hierarchy. Empty inner vectors are rejected; `{\"\"}` represents root membership.
- Add an ABI-safe non-virtual `GetPaths(hierarchy, rows)` overload that copies legacy or structured paths into `std::vector<std::vector<std::string>>` and reports whether the hierarchy is present.
- Share one internal zero-copy path reader across legacy and structured representations, including ownership replacement, `DeepCopy`, and all legacy/structured `Append` combinations.
- Populate every selected Pyramid path during `Build`, `Add`, and cache-assisted builds while storing vector data and labels once.
- Return stored multi-path rows through `GetDataByIdsWithFlag`; preserve the legacy pointer representation when every requested ID has exactly one path.
- Deduplicate repeated paths, shared prefixes, overlapping query results, and searches below `NO_INDEX` prefixes.
- Keep the existing `|` query union syntax and add structured query paths, where `|` is an ordinary path character.
- Preserve multi-path memberships through serialization and support path-scoped Analyzer ground truth.
- Update C++ examples and the English and Chinese Dataset, Pyramid, and Analyzer documentation.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build --target vsag unittests functests --parallel 4
build/tests/unittests "Dataset Named Paths Test"
  All tests passed (96 assertions in 1 test case)
build/tests/functests "[multi_hierarchy]"
  All tests passed (260 assertions in 28 test cases)
build/tests/functests "[pyramid][paths]"
  All tests passed (327 assertions in 16 test cases)
scripts/linters/check-struct-names.py <changed C++ files>
  Checked 7 C/C++ files: all named struct definitions use CamelCase.
clang-tidy-15 -p build src/dataset_impl.cpp src/algorithm/pyramid/pyramid_paths.cpp --quiet
git diff --check
```

## Compatibility Impact

- API/ABI compatibility: additive only; both new public overloads are non-virtual and do not change the `Dataset` vtable. The legacy pointer APIs remain supported.
- Behavior changes: structured rows contain one or more atomic paths. An empty inner vector is invalid and `{\"\"}` means root membership. Legacy query `|` semantics are unchanged.
- Path retrieval: legacy getters continue to return single-path arrays. The new structured getter returns complete one-or-many path rows from either representation.

- Performance impact: vector payloads remain stored once. Each additional membership adds references and index work along its unique path nodes. The public structured getter copies path rows into caller-owned output; Build/Add/Search retain the internal zero-copy row reader.
- Concurrency/thread-safety impact: named-hierarchy parallel build remains independent; multi-branch searches use separate visited lists before deterministic result merging.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: English and Chinese Dataset, Pyramid, Analyzer pages, overview SVGs, and the C++ multi-hierarchy example

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert the two commits in this PR. Serialization format is unchanged.

## Checklist

- [x] I have added/updated tests for new behavior.
- [x] I have updated docs for the behavior change.


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -520,12 +521,30 @@ Pyramid::search_impl(const DatasetPtr& query,
                    fmt::format("unknown hierarchy name: '{}'", hierarchy_name));
     const auto& h = *h_iter->second;
 
-    const auto* query_path = query->GetPaths(hierarchy_name);
```
LHT129, line 959: **suggestion:** The `load_query_paths` lambda in `search_impl` only reads element 0 for structured paths (`query->GetPaths(name, 0, path_count)`), which implicitly assumes the query dataset has exactly one element. This is correct for search (queries are single-element), but the assumption is not validated. Consider adding an assertion or `CHECK_ARGUMENT` to make the contract explicit and prevent misuse if a multi-element query dataset is accidentally passed.

```cpp
CHECK_ARGUMENT(query->GetNumElements() == 1, "search query must have exactly one element");
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1656,53 +1685,96 @@ Pyramid::search_hierarchy(const Hierarchy& h,
...
+
+    for (const auto& one_path : parsed_paths) {
         IndexNode* node = h.root.get();
-        bool valid = true;
         for (const auto& item : one_path) {
             node = node->GetChild(item, false);
             if (node == nullptr) {
-                valid = false;
                 break;
             }
```
LHT129, line 2369: **suggestion:** The merge loop iterates `for (const auto inner_id : result_ids)` — this copies each `InnerIdType` by value. Since `InnerIdType` is likely `int64_t`, the copy is cheap, but for consistency with the rest of the codebase and to avoid any accidental overhead if the type changes, consider `const auto&`:

```cpp
for (const auto& inner_id : result_ids) {
    merged_result->Push(best_distances.at(inner_id), inner_id);
}
```

src/dataset_impl.cpp
```diff
@@ -29,6 +32,152 @@ Dataset::Make() {
...
+    }
+
+    const auto* old_paths = std::get_if<const std::string*>(&iter->second);
+    const auto* new_paths = std::get_if<const std::string*>(&paths);
+    if (old_paths != nullptr && *old_paths != nullptr &&
+        (new_paths == nullptr || *old_paths != *new_paths)) {
+        retired_paths_.push_back(*old_paths);
+    }
+    iter->second = std::move(paths);
+}
```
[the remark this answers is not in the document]
jac0626: Thanks. I traced the A -> B -> A case. No double delete occurs: DatasetImpl::~DatasetImpl uses one released_paths set for both active entries and retired_paths_, so active A is deleted once and the retired A entry is skipped; B is deleted once. Append also removes pointers from retired_paths_ before deleting replaced owned buffers. I therefore left this code unchanged.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. This is an existing search contract rather than a new structured-path assumption: legacy code already read query_path[0]. SearchWithRequest calls validate_search_query(), which enforces NumElements() == 1. Adding a check only in this Pyramid path loader would change legacy KnnSearch/RangeSearch behavior and duplicate validation, so I kept it unchanged.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. InnerIdType is currently uint32_t in src/basic_types.h, so copying it is cheaper and simpler than binding a reference and matches scalar loops in Pyramid and HGraph. A hypothetical future type change is not enough to add indirection now, so I kept the by-value loop.

LHT129:
## Code Review for PR #2796

I have reviewed the changes in this PR ("feat(pyramid): support multiple paths per vector"). Below are my findings beyond what has already been flagged by Copilot and LHT129.

### Summary
This PR adds structured multi-path support to the Pyramid index, allowing each vector to belong to zero, one, or multiple paths per hierarchy. The core changes are in DatasetImpl (new MultiPaths variant type, GetPaths element-level accessor, ReplacePaths with retired_paths_ tracking) and Pyramid (updated search_hierarchy, populate_path_tree, add_to_hierarchy to accept structured paths). Test coverage is good with both unit tests and functional tests.

### Issues

**1. parse_path result is silently empty for empty input string (pyramid.cpp:2431-2438)**

When parse_path("") is called (e.g., from the legacy path code path when a query has an empty path string), split("", PART_BAR) returns a vector with one empty string. Then parse_atomic_path("") is called which does split("", PART_SLASH) — this also returns a vector with one empty string. The result is [[""]] — a single path with a single empty segment. This will cause search_hierarchy to look up h.root->GetChild("", false) which may or may not find a node. The behavior is inconsistent with the structured path API where an empty inner vector means "no membership."

Consider handling the empty-string case explicitly in parse_path to return an empty parsed_paths vector.

**2. DatasetImpl move constructor drops Statistics_ and Reasoning_ (dataset_impl.h:58-66)**

The user-defined move constructor only moves owner_, allocator_, data_, and retired_paths_. It leaves Statistics_ and Reasoning_ at their default values ("{}"), so moving a DatasetImpl loses those fields. This was already flagged by Copilot but has not been addressed yet.

**3. NumElements validation for MultiPaths allows count updates to bypass size check (dataset_impl.cpp:69-86)**

When is_count_update is true (i.e., the count is being changed from a previous value), the validation at line 79 skips the size check. This means if you call NumElements(3) followed by NumElements(5) without updating the MultiPaths, the second call silently succeeds even though the MultiPaths still has 3 entries. This is intentional (to allow count updates before paths are set), but it creates a window where GetPaths will throw because the sizes do not match. Consider adding a comment explaining this intentional gap.

**4. search_hierarchy merge loop copies InnerIdType by value (pyramid.cpp:2424)**

The loop copies each InnerIdType by value. While InnerIdType is currently uint32_t (cheap to copy), using const auto& would be more consistent with typical C++ practice and safer if the type ever changes. This was flagged by LHT129 and acknowledged by the author.

### Positive Observations
- The EmptyPathsSentinel() pattern for returning a non-null pointer when path_count==0 is a clean way to distinguish "no paths key exists" (nullptr) from "element has zero paths" (non-null sentinel).
- The PendingPathAppend struct in Append() cleanly handles the mixed legacy/structured path upgrade during append.
- Test coverage is thorough: unit tests cover edge cases (empty paths, size validation in both fluent orders, replacement, deep copy, append, mixed legacy/structured promotion), and functional tests cover search with legacy queries, structured queries, serialization round-trip, deduplication, and Add with multi-path.
- The retired_paths_ mechanism with the released_paths set in the destructor correctly prevents double-delete (as confirmed by the author in response to Copilot).

### Recommendation
The PR is well-structured with good test coverage. The issues above are minor — the move constructor issue (item 2) is the most impactful and should be fixed. Items 1, 3, and 4 are documentation/clarity suggestions.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -602,12 +604,24 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     }
     if (store_paths_) {
         for (const auto& [hierarchy_name, hierarchy] : hierarchies_) {
-            const auto* paths = base->GetPaths(hierarchy_name);
-            if (paths != nullptr) {
+            uint64_t total_path_count = 0;
+            bool has_paths = false;
+            for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
+                uint64_t path_count = 0;
+                if (GetDatasetPaths(*base, hierarchy_name, offset, path_count) != nullptr) {
```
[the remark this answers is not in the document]
jac0626: This is intentional. PyramidPathStore distinguishes an explicit zero-path row from a missing hierarchy row: Writer::Insert with path_count == 0 records the row, GetPathRows returns an empty vector for it, while a missing offset makes GetPathRows return false. The upstream zero/one/multiple-row and serialization tests cover this distinction. Skipping the insert would lose the zero-path assignment required by the structured model, so no code change is needed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -1669,21 +1700,40 @@ Pyramid::insert_add_batch(const DatasetPtr& base, const AddBatch& batch) {
...
+        if (store_paths_ and not batch.input_indices.empty()) {
+            uint64_t total_path_count = 0;
+            bool has_paths = false;
+            for (const auto input_index : batch.input_indices) {
+                uint64_t path_count = 0;
+                if (GetDatasetPaths(*base, hname, static_cast<uint64_t>(input_index), path_count) !=
+                    nullptr) {
+                    has_paths = true;
+                    total_path_count += path_count;
+                }
```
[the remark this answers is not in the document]
jac0626: The non-null sentinel means that this hierarchy has a structured row for the element, even though that row contains zero paths. has_paths therefore enables recording the row metadata, not only path strings. This preserves the explicit empty row in PyramidPathStore and is covered by its zero/one/multiple-row tests, so no code change is needed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2686,17 +2787,29 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
...
-            if (store_paths_) {
+        if (store_paths_) {
+            uint64_t total_path_count = 0;
+            bool has_paths = false;
+            for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
+                uint64_t path_count = 0;
+                if (GetDatasetPaths(*base, hname, offset, path_count) != nullptr) {
+                    has_paths = true;
+                    total_path_count += path_count;
+                }
```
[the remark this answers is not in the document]
jac0626: This intentionally stores structured empty rows. PyramidPathStore uses an offset plus a zero count for an explicit empty row and MISSING_PATH_OFFSET for an absent row; GetPathRows and its serialization tests rely on that distinction. Avoiding the allocation would discard supported zero-path rows, so no code change is needed.

src/dataset_impl.h
```diff
@@ -53,8 +58,11 @@ class DatasetImpl : public Dataset {
     DatasetImpl(DatasetImpl&& other) noexcept {
         this->owner_ = other.owner_;
         other.owner_ = false;
-        this->data_ = other.data_;
+        this->allocator_ = other.allocator_;
+        this->data_ = std::move(other.data_);
+        this->retired_paths_ = std::move(other.retired_paths_);
         other.data_.clear();
+        other.retired_paths_.clear();
     }
```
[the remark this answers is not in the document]
jac0626: The Statistics_ and Reasoning_ move behavior predates this PR: the DatasetImpl move constructor on upstream main already leaves both at their defaults. This change only transfers the new path-owned state while preserving existing observable behavior. Fixing unrelated Dataset statistics move semantics would expand this PR, so no code change is needed.

jac0626:
Requirement clarification applied in c2bb120c: “zero path” means the empty-string path at the hierarchy root (`{""}`), not an empty inner row. Structured `Paths(...)` now rejects `{}`, and the previous replies defending explicit zero-path rows are superseded. This update also adds the public structured `GetPaths(hierarchy, rows)` overload and returns stored multi-path rows from `GetDataByIdsWithFlag` while preserving the legacy getter for single-path results.

src/algorithm/pyramid/pyramid_paths.cpp
```diff
@@ -78,15 +81,25 @@ Pyramid::GetDataByIdsWithFlag(const int64_t* ids,
...
         }
-        if (hierarchy_name.empty()) {
-            result->Paths(paths.get());
-        } else {
-            result->Paths(hierarchy_name, paths.get());
+
+        auto paths = std::make_unique<std::string[]>(static_cast<uint64_t>(count));
+        for (uint64_t offset = 0; offset < static_cast<uint64_t>(count); ++offset) {
+            paths[offset] = std::move(path_rows[offset][0]);
         }
```
LHT129, line 101: [suggestion] The `paths.get()` + `paths.release()` pattern transfers ownership of a `std::unique_ptr<std::string[]>` to the Dataset. If `result->Paths()` throws (e.g., due to an allocator failure or internal validation in `ReplacePaths`), the released pointer leaks because `paths.release()` on the next line never executes. Consider using the structured `MultiPaths` setter (`result->Paths(hierarchy_name, std::move(path_rows))`) consistently instead of the legacy pointer form for single-path rows. The legacy pointer path exists only for backward compatibility with external callers; internally, the structured form is safer and avoids manual ownership transfer.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -970,10 +1002,9 @@ Pyramid::search_impl(const DatasetPtr& query,
     std::shared_lock<std::shared_mutex> lock(resize_mutex_);
     VisitedListGuard vl_guard(pool_.get());
     const VisitedListPtr& vl = vl_guard.get();
-    if (query_path != nullptr) {
-        const std::string& current_path = query_path[0];
+    if (has_query_paths) {
         search_hierarchy(
-            h, search_func, vl, search_result, current_path, search_param, ctx.reasoning_ctx);
+            h, search_func, vl, search_result, parsed_query_paths, search_param, ctx.reasoning_ctx);
```
LHT129, line 1007: [note] The `search_hierarchy` call now passes `parsed_query_paths` (a `std::vector<std::vector<std::string>>`) instead of `query_path[0]` (a single `std::string`). This correctly supports both legacy pipe-delimited queries (parsed via `parse_path`) and structured multi-path queries (parsed via `split` on `/`). The `load_query_paths` lambda above correctly handles the fallback from named hierarchy to default (`""`) hierarchy.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2182,18 +2232,41 @@ Pyramid::add_one_point(const Hierarchy& hierarchy,
...
+        add_node(node);
+        for (const auto& path_slice : path_slices) {
             node = node->GetChild(path_slice, true);
+            add_node(node);
+        }
+    }
+    return nodes;
+}
+
+void
```
LHT129, line 2257: [suggestion] In `populate_path_tree`, when `legacy_paths != nullptr` the code sets `path_count = 1` and passes `legacy_paths + i` to `collect_path_nodes`. This is correct for legacy single-path data. However, if a Dataset has *both* legacy and structured paths for the same hierarchy (which `ReplacePaths` allows — setting legacy then structured then legacy again), `GetPaths(hierarchy_name)` could return a non-null legacy pointer while `GetDatasetPaths` would return the more recent structured representation. The current code prefers the legacy pointer. Consider always using `GetDatasetPaths` as the single source of truth, or document that mixed legacy/structured usage on the same hierarchy is unsupported.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2282,53 +2335,96 @@ Pyramid::search_hierarchy(const Hierarchy& h,
...
+        return;
+    }
+
+    auto merged_result = std::make_shared<StandardHeap<true, false>>(allocator_, -1);
+    UnorderedMap<InnerIdType, float> best_distances(allocator_);
+    Vector<InnerIdType> result_ids(allocator_);
+    for (const auto& result : search_result_lists) {
+        const auto* records = result->GetData();
+        for (uint64_t i = 0; i < result->Size(); ++i) {
+            const auto& [distance, inner_id] = records[i];
```
LHT129, line 2415: [note] The `for (const auto inner_id : result_ids)` loop copies each `InnerIdType` by value. While `InnerIdType` is `uint32_t` today and the copy is trivial, `const auto&` would be more idiomatic and would avoid a future bug if the type ever widens. This was already discussed in another review thread — no action needed, just noting for consistency.

LHT129 (commented):
[note] Overall review for PR #2796: feat(pyramid): support multiple paths per vector

**Summary:** This PR adds multi-path support to the Pyramid index, allowing each vector to belong to multiple paths within a hierarchy. The implementation touches the Dataset API, Pyramid index core, path storage, analyzer, and tests. The design is well-thought-out with careful ABI compatibility considerations.

**What is done well:**
- ABI-safe public API additions: new `Paths()`/`GetPaths()` overloads are non-virtual, preserving vtable compatibility.
- Clean separation between legacy (pointer-based) and structured (`std::vector<std::vector<std::string>>`) path representations.
- `collect_path_nodes` correctly deduplicates nodes when `path_count > 1` using `std::unordered_set<IndexNode*>`.
- `search_hierarchy` properly handles multi-path queries with node deduplication and best-distance-per-inner_id merging.
- `GetDatasetPaths()` provides a zero-copy internal reader that works uniformly with both representations.
- Good test coverage: multi-path Build, Add, serialization, structured queries, pipe-delimited queries, deduplication, and edge cases (empty paths, literal pipe characters).
- The `PyramidPathStore` uses an `AcquireWriter` pattern that correctly pre-allocates capacity based on total path count.

**Suggestions (see inline comments for details):**
1. `pyramid_paths.cpp`: The `paths.get()` + `paths.release()` ownership transfer pattern is fragile — if `result->Paths()` throws, the pointer leaks. Consider using the structured setter internally.
2. `pyramid.cpp` (`populate_path_tree`): The legacy-path fast path could diverge from `GetDatasetPaths` if a Dataset mixes legacy and structured representations on the same hierarchy.
3. Minor: `const auto&` in the merge loop would be more idiomatic (already discussed).

**Noted but acceptable:**
- The `retired_paths_` double-delete concern was already addressed by the author (destructor uses a `released_paths` set for deduplication).
- The `load_query_paths` element-0 assumption matches the existing search contract.
- The `InnerIdType` copy-by-value is intentional given the current `uint32_t` type.

**Verdict:** The PR is well-structured and the core logic is sound. The suggestions above are minor improvements, not blockers.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. A hierarchy key has exactly one active value in DatasetImpl::data_: its variant contains either the legacy pointer or MultiPaths. ReplacePaths overwrites that single variant value; retired_paths_ only retains old pointer ownership for later deletion and does not make the old representation query-visible. Therefore legacy and structured representations cannot coexist for the same hierarchy, and the fast path reads the same active source of truth. No change is needed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2182,18 +2232,41 @@ Pyramid::add_one_point(const Hierarchy& hierarchy,
...
+                            const DatasetPtr& dataset,
+                            const std::string& hierarchy_name,
+                            int64_t count) {
+    const auto* legacy_paths = dataset->GetPaths(hierarchy_name);
+    for (int64_t i = 0; i < count; ++i) {
+        uint64_t path_count = legacy_paths == nullptr ? 0 : 1;
+        const auto* paths =
+            legacy_paths == nullptr
+                ? GetDatasetPaths(*dataset, hierarchy_name, static_cast<uint64_t>(i), path_count)
+                : legacy_paths + i;
```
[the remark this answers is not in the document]
jac0626: Thanks. This does add an O(N) row-presence scan for a configured hierarchy that is absent from the input dataset, but it does not affect correctness and only occurs during Build/Add processing. Introducing and plumbing a separate hierarchy-presence path would be a performance optimization beyond the issue contract, and we do not currently have benchmark evidence showing that this lookup scan is material. I am keeping the current unified row reader and leaving any optimization to a separately measured change.

src/dataset_impl.cpp
```diff
@@ -596,31 +831,90 @@ DatasetImpl::Append(const DatasetPtr& other) {
...
+            const auto* paths = GetDatasetPaths(dataset, hierarchy_name, i, path_count);
+            if (paths == nullptr) {
+                throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                    "Cannot append a dataset without the required hierarchy paths");
+            }
+            if (path_count == 0) {
+                result.emplace_back();
+            } else {
+                result.emplace_back(paths, paths + path_count);
+            }
```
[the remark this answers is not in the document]
jac0626: Thanks. Under the current public contract this branch is not reachable: the structured setter rejects every empty inner row, the legacy representation always reports one path (including the empty-string root path), and the fallback for custom Dataset implementations also reports either no hierarchy or one path. Mapping a zero count to {""} would additionally turn an invalid state into root membership. Since no supported input can produce path_count == 0 here, I am keeping the code unchanged.

src/algorithm/pyramid/pyramid_paths.cpp
jac0626: Thanks. The current ownership transfer is exception-safe: until paths.release() executes, the unique_ptr still owns the array, so an exception from result->Paths() unwinds the stack and deletes it. release() runs only after Paths() returns successfully. Using the structured setter for single-path rows would also change the preserved legacy GetPaths(hierarchy) behavior from a pointer to nullptr, so I am keeping this code unchanged.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. As noted, InnerIdType is currently uint32_t and copying it by value is intentional and consistent with scalar loops in this code. No change is needed.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks for confirming the structured and legacy query paths and the named-to-default hierarchy fallback. No change is needed.

src/dataset_impl.cpp
```diff
@@ -596,31 +831,90 @@ DatasetImpl::Append(const DatasetPtr& other) {
...
+                                     MultiPaths& result) {
+        for (uint64_t i = 0; i < static_cast<uint64_t>(count); ++i) {
+            uint64_t path_count = 0;
+            const auto* paths = GetDatasetPaths(dataset, hierarchy_name, i, path_count);
+            if (paths == nullptr) {
+                throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                    "Cannot append a dataset without the required hierarchy paths");
+            }
+            if (path_count == 0) {
+                result.emplace_back();
```
LHT129, line 858: The `append_path_rows` lambda calls `result.emplace_back()` when `path_count == 0`, which creates an empty inner vector. This contradicts the invariant enforced by `DatasetImpl::Paths(hierarchy_name, MultiPaths)`, which explicitly rejects empty inner vectors with "Each element must have at least one hierarchy path."

The empty-row case can arise when a dataset carries PyramidPathStore-style explicit empty rows (non-null sentinel, zero count). When such datasets are appended, the resulting `MultiPaths` will contain empty inner vectors that bypass validation because `Append` writes directly to `data_` via `iter->second = std::move(pending.multi_paths)` rather than going through the `Paths()` setter.

Consider either:
1. Mapping `path_count == 0` to `{""}` (root path) to preserve the invariant, or
2. Throwing an error if an explicit empty row is encountered during append, since it is ambiguous whether the element should participate in the hierarchy at all.

src/algorithm/pyramid/pyramid.cpp
LHT129, line 2268: `populate_path_tree` now iterates over all `count` elements unconditionally, even when the dataset has no paths for this hierarchy. In the old code, `populate_hierarchy_trees` guarded the call with `if (paths != nullptr)`, but that guard was removed.

When both `legacy_paths` and structured paths are absent, `GetDatasetPaths` returns `nullptr` with `path_count == 0`, and `collect_path_nodes(h, nullptr, 0)` returns an empty vector — so the inner loop body is skipped. But the outer loop still runs `count` times, performing a `GetDatasetPaths` call (which does `FindPaths` + variant visit) per element. For large datasets with many unused hierarchies, this adds measurable overhead during Build.

Consider adding an early return when `FindPaths` returns `nullptr` for the hierarchy, before entering the element loop.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2282,53 +2335,96 @@ Pyramid::search_hierarchy(const Hierarchy& h,
...
-                                 search_result_lists[i],
-                                 search_param.ef,
-                                 reasoning_ctx);
-                }));
-            } else {
+        if (node != nullptr) {
+            collect_search_nodes(node);
+        }
+    }
+
```
LHT129, line 2375: When none of the `parsed_paths` resolve to a searchable node (all paths lead to non-existent children), `search_nodes` remains empty. The function then allocates a zero-sized `search_result_lists` vector and falls through both the `size() == 1` early-return and the merge loop without modifying `search_result`.

The caller (`search_impl`) initializes `search_result` to an empty heap before calling `search_hierarchy`, so the net effect is returning zero results — which is correct. However, this is a non-obvious control flow: the function returns successfully but leaves `search_result` untouched. Consider adding an explicit early return when `search_nodes.empty()` to make the intent clear and avoid the unnecessary `search_result_lists` allocation.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2282,53 +2335,96 @@ Pyramid::search_hierarchy(const Hierarchy& h,
...
                           ReasoningContext* reasoning_ctx) const {
-    std::vector<std::future<void>> futures;
-    auto parsed_path = parse_path(path);
-    Vector<DistHeapPtr> search_result_lists(parsed_path.size(), allocator_);
-    for (uint32_t i = 0; i < parsed_path.size(); ++i) {
-        const auto& one_path = parsed_path[i];
-        search_result_lists[i] = std::make_shared<StandardHeap<true, false>>(allocator_, -1);
+    std::vector<IndexNode*> search_nodes;
+    std::unordered_set<IndexNode*> seen_nodes;
+    const std::function<void(IndexNode*)> collect_search_nodes = [&](IndexNode* node) {
```
LHT129, line 2343: The `collect_search_nodes` lambda is declared as `std::function<void(IndexNode*)>` on the stack and captures `search_nodes` and `seen_nodes` by reference. In the hot path (called per-query), `std::function` requires a heap allocation for the type-erased callable.

Since this is a local recursive lambda used immediately, consider using a plain recursive lambda with `auto&&` (C++14 generic lambda) or extracting it into a private member function to avoid the allocation overhead on every query.

LHT129 (commented):
## Code Review Summary

This PR adds multi-path support per vector to the Pyramid index. Overall the design is well thought out: the Dataset API uses non-virtual overloads to avoid ABI breakage, the internal representation cleanly separates legacy (pointer) and structured (MultiPaths) forms via variant, and the path store correctly handles deduplication and serialization. The test coverage is thorough across unit, functional, and serialization paths.

Most issues have already been raised in existing review threads (25 comments). Below are the remaining items I believe still need attention.

### Unresolved items from existing threads

1. ** — potential leak in ownership transfer** (comment #3930571816)
   The `paths.get()` + `paths.release()` pattern leaks if `result->Paths()` throws. Using the structured `MultiPaths` setter would be safer here since the code already has `path_rows` in hand.

2. ** — empty inner vectors from append_path_rows** (comment #3944254978)
   When `path_count == 0`, `append_path_rows` creates an empty inner vector, violating the invariant enforced by `Paths(hierarchy_name, MultiPaths)`. This can happen when appending datasets that carry PyramidPathStore-style explicit empty rows.

3. ** — populate_path_tree iterates unconditionally** (comment #3944255543)
   The outer loop runs `count` times even when the dataset has no paths for the hierarchy, performing `GetDatasetPaths` per element. An early return when `FindPaths` returns nullptr would avoid this.

4. ** — empty search_nodes case** (comment #3944256115)
   When no parsed paths resolve to a searchable node, `search_nodes` is empty and the function returns without modifying `search_result`. The behavior is correct but non-obvious.

5. ** — std::function heap allocation in hot path** (comment #3944260123)
   The `collect_search_nodes` lambda uses `std::function` which heap-allocates. Since this runs per-query, a plain recursive lambda or extracted member function would avoid the allocation.

### New observations

6. ** — NumElements validation order sensitivity**
   The `NumElements` setter validates that existing MultiPaths sizes match the new count. However, if `NumElements` is called *before* `Paths(MultiPaths)`, the validation is skipped (no MultiPaths entries exist yet), and the subsequent `Paths` call validates against the stored count. If `NumElements` is called *after* `Paths(MultiPaths)` with a different count, it throws. This ordering dependency is tested but subtle — callers must know to set `NumElements` first. Consider documenting this ordering requirement in the API docs or relaxing the constraint.

7. ** — best_distances map can grow unbounded**
   The merge loop accumulates all result IDs into `best_distances` before pushing into `merged_result`. If multiple search result lists are large (high `ef_search`), the intermediate `UnorderedMap` can hold many entries. Consider capping to `ef_search` entries or using a heap-based merge to bound memory.

8. ** — emplace always succeeds when inserted==true but distance may not be minimal**
   The `emplace` + `std::min` pattern is correct for deduplication, but the `result_ids` vector tracks insertion order rather than best-distance order. The final `Push` loop pushes in insertion order, relying on the heap to re-sort. This is fine functionally but means the heap does O(n log n) work on potentially many entries. Consider pushing only the top-k entries.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -123,14 +124,40 @@ PyramidAnalyzer::AnalyzeIndexBySearch(const SearchRequest& request) {
...
+        if (const auto* legacy_paths = query->GetPaths(name)) {
+            parsed_query_paths.reserve(query_count);
+            for (uint32_t i = 0; i < query_count; ++i) {
+                parsed_query_paths.push_back(Pyramid::parse_path(legacy_paths[i]));
+            }
+            return true;
+        }
+
+        uint64_t path_count = 0;
+        if (GetDatasetPaths(*query, name, 0, path_count) == nullptr) {
```
LHT129, line 138: [suggestion] The `path_count` variable in the `load_query_paths` lambda serves double duty: it is first used as an existence check and then overwritten per-element inside the loop. While functionally correct, this reuse is confusing. Consider renaming the first use to a separate variable.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -819,7 +846,21 @@ PyramidAnalyzer::calculate_search_result(
         query->Dim(dim_)->NumElements(1)->Owner(false)->Float32Vectors(
             sample_datas.data() + static_cast<size_t>(i) * dim_);
         if (query_paths != nullptr) {
-            query->Paths(hierarchy_name, query_paths + i);
+            std::vector<std::string> atomic_paths;
```
LHT129, line 849: [note] The path reconstruction from `ParsedPathList` back to `std::vector<std::vector<std::string>>` in `calculate_search_result` creates a temporary `atomic_paths` vector and a `structured_paths` wrapper, then moves them into `query->Paths()`. This is a round-trip: the paths were already parsed from structured form in `load_query_paths`, then serialized back to structured form here. If this is a hot path during analysis, consider whether `search_impl` could accept `ParsedPathList` directly to avoid the reconstruction cost.

src/dataset_impl.cpp
```diff
@@ -29,6 +32,215 @@ Dataset::Make() {
...
+
+    paths.clear();
+    const auto* legacy_paths = GetPaths(hierarchy_name);
+    if (legacy_paths == nullptr) {
+        return false;
+    }
+    const auto num_elements = GetNumElements();
+    if (num_elements < 0) {
+        throw VsagException(ErrorType::INVALID_ARGUMENT, "NumElements() must not be negative");
+    }
```
LHT129, line 60: [note] The free function `GetDatasetPaths()` has a name that is easily confused with the non-virtual `Dataset::GetPaths()` method. Both deal with paths but have different signatures and semantics — the free function returns a raw pointer per-element with an out-parameter for path count, while the non-virtual method copies all paths into an output vector. Consider a more distinctive name like `GetDatasetPathRows()` or `ReadDatasetPathRows()` to reduce the risk of callers using the wrong one.

LHT129 (commented):
## Code Review Summary

This PR adds multi-path support per vector to the Pyramid index, a significant feature. The overall design is well-structured: it introduces `MultiPaths` as a new variant in the dataset storage, provides a zero-copy `GetDatasetPaths()` reader that unifies legacy and multi-path representations, and updates all call sites to use the new abstraction.

### Comments posted as inline reviews:

1. **[suggestion] `path_count` variable reuse** in `pyramid_analyzer.cpp` `load_query_paths` lambda: the variable serves double duty (existence check then per-element overwrite), which is confusing. Consider a separate variable for the initial check.

2. **[note] Path reconstruction overhead** in `calculate_search_result`: paths are parsed from structured form, then reconstructed back to structured form per query. If this is a hot path, consider having `search_impl` accept `ParsedPathList` directly.

3. **[note] `GetDatasetPaths` naming**: the free function name is easily confused with `Dataset::GetPaths()`. Consider a more distinctive name.

### Positive observations:

- The `retired_paths_` mechanism for deferred cleanup of replaced legacy path arrays is a clean solution to the double-delete problem.
- The `ReplacePaths` method correctly handles the case where old and new pointers are the same (no-op).
- Good validation in `NumElements()` and `Paths(MultiPaths)` to ensure consistency between element count and path rows.
- The `Append` refactoring with `PendingPathAppend` struct is well-organized and handles both legacy and multi-path cases.
- Test coverage is solid with dedicated `MultiPathFixture` and path-specific test files.

src/dataset_impl.cpp
jac0626: Thanks. GetDatasetPaths is an internal zero-copy per-element helper declared only in src/dataset_impl.h, while Dataset::GetPaths is the public all-elements copying overload. Their argument lists and visibility make the distinction explicit at call sites, and renaming the established internal helper would create broad churn without a correctness benefit. I am keeping the current name.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. This reconstruction is intentional at the analyzer boundary: calculate_search_result creates sampled Dataset queries and calls the normal public index search path, whose supported input is Dataset paths rather than Pyramid's private ParsedPathList. Passing parsed internals directly would require a new internal search entry point and tighter Analyzer/Pyramid coupling for a non-production analysis path. I am keeping the public-path round trip.

src/dataset_impl.cpp
jac0626: Thanks. No supported Dataset input can carry an explicit zero-count row: the structured setter rejects empty inner vectors, the legacy representation always reports one path (including the empty-string root path), and the fallback for custom Dataset implementations reports either no hierarchy or one path. PyramidPathStore is internal index storage and does not provide a path that injects an empty row into DatasetImpl::data_. Mapping an invalid zero count to {""} would also silently change it into root membership. Since this state is unreachable through the supported API, I am keeping the code unchanged.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. This recursive callable is created once per hierarchy search, while the dominant work is graph traversal and node search. Replacing the clear std::function recursion with a self-parameter generic lambda would be a micro-optimization without benchmark evidence and would make the recursion syntax less direct. There is no correctness issue, so I am keeping it unchanged.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. path_count consistently denotes the count returned by GetDatasetPaths: the first call only establishes representation presence, and each loop iteration refreshes the same output for that element before it is consumed. The initial value is intentionally not used after the presence check. Splitting it into another variable would not clarify a distinct semantic value, so no change is needed.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. An empty search_nodes list intentionally produces an empty result: constructing a zero-sized vector does not allocate elements, both loops become no-ops, and search_result remains the caller-provided empty heap. An explicit early return would only restate the existing container semantics without changing behavior or avoiding meaningful work, so I am keeping the current flow.

src/algorithm/pyramid/pyramid.cpp
jac0626: Thanks. The absent-hierarchy case does perform an additional O(N) sequence of row lookups during Build, but it does not affect correctness. Adding a separate hierarchy-presence mechanism is a performance optimization outside the issue contract, and there is no benchmark showing this lookup scan is material relative to index construction. I am keeping the unified row reader here; this can be optimized separately if profiling identifies it as significant.

src/analyzer/pyramid_analyzer.cpp
```diff
@@ -819,7 +846,21 @@ PyramidAnalyzer::calculate_search_result(
...
+                    if (j != 0) {
+                        path.push_back(PART_SLASH);
+                    }
+                    path.append(parsed_path[j]);
+                }
+                atomic_paths.push_back(std::move(path));
+            }
+            std::vector<std::vector<std::string>> structured_paths;
+            structured_paths.push_back(std::move(atomic_paths));
+            query->Paths(hierarchy_name, std::move(structured_paths));
```
LHT129, line 863: [suggestion] When `(*query_paths)[i]` is empty (i.e., the query element has zero paths for this hierarchy), the loop body produces an empty `atomic_paths` vector, wraps it in `structured_paths` (a single-element outer vector with an empty inner vector), and passes it to `query->Paths(hierarchy_name, ...)`. `DatasetImpl::Paths(hierarchy_name, MultiPaths)` at line 144 rejects empty inner vectors with "Each element must have at least one hierarchy path", so this will throw.\n\nThe `load_query_paths` lambda in `AnalyzeIndexBySearch` can return `true` with empty `parsed_query_paths[i]` entries when `GetDatasetPaths` returns a non-null sentinel but `path_count == 0`. Consider either skipping the `query->Paths()` call when the parsed path list is empty, or handling the zero-path case explicitly (e.g., by not setting paths on the query and letting `search_impl` fall through to the root search).

LHT129 (commented):
[review] I have completed a thorough review of this PR (commit 5929ee6). The existing comments from Copilot and LHT129 already cover the substantive issues I identified during my analysis. Key areas reviewed:

- Path representation duality (legacy pointer vs. structured MultiPaths) and sentinel/empty-row semantics
- `search_impl` query element assumption
- `search_hierarchy` control flow edge cases (empty search_nodes, std::function overhead)
- `append_path_rows` empty inner vector vs. `Paths()` invariant mismatch
- `populate_path_tree` unconditional O(N) scan for unused hierarchies
- `calculate_search_result` path reconstruction round-trip
- `GetDatasetPaths` naming confusion with `Dataset::GetPaths()`
- Potential leak in `GetDataByIdsWithFlag` via `paths.release()`

The overall architecture — ABI-safe non-virtual overloads, zero-copy `GetDatasetPaths` free function, `collect_path_nodes` deduplication, and structured multi-path support throughout the build/search/analysis pipeline — is sound. No additional blocking issues found.

src/dataset_impl.cpp
```diff
@@ -463,45 +691,54 @@ DatasetImpl::Append(const DatasetPtr& other) {
...
+                                      const std::string& hierarchy_name) {
+        if (const auto impl = std::dynamic_pointer_cast<DatasetImpl>(dataset)) {
+            return impl->HasPaths(hierarchy_name);
+        }
+        if (dataset->GetPaths(hierarchy_name) != nullptr) {
+            return true;
+        }
+        if (dataset->GetNumElements() <= 0) {
+            return false;
+        }
```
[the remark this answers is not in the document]
jac0626: Thanks. Requiring matching pointer-backed fields during Append is the existing DatasetImpl contract: upstream main already rejects an appended dataset without paths whenever the destination has paths, regardless of the appended element count. Treating a zero-element dataset as an unconditional no-op would introduce a new Append semantic rather than fix a regression from this PR, so I am preserving the existing behavior.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2246,48 +2317,15 @@ Pyramid::populate_hierarchy_trees(const DatasetPtr& base) {
     }
```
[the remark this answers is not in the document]
jac0626: Thanks. This is the same absent-hierarchy performance suggestion discussed above: the extra task and O(N) row lookups do not affect correctness, and no benchmark currently shows they are material relative to Build. Adding a separate hierarchy-presence path would broaden this PR for an unmeasured optimization, so I am keeping the unified path-reader flow unchanged.

src/analyzer/pyramid_analyzer.cpp
jac0626: Thanks. This zero-count state is not reachable through the supported Dataset APIs: structured Paths rejects every empty inner row, the empty-string root path has count 1, and the legacy/custom-Dataset fallback returns either no hierarchy or one path. Therefore load_query_paths cannot produce an empty parsed_query_paths[i] entry from a valid query dataset. No change is needed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -619,12 +621,24 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
...
+                writer.Prepare(static_cast<uint64_t>(data_num), total_path_count);
                 for (uint64_t offset = 0; offset < static_cast<uint64_t>(data_num); ++offset) {
-                    writer.Insert(static_cast<InnerIdType>(offset), paths[offset]);
+                    uint64_t path_count = 0;
+                    const auto* paths = GetDatasetPaths(*base, hierarchy_name, offset, path_count);
+                    if (paths != nullptr) {
+                        writer.Insert(static_cast<InnerIdType>(offset), paths, path_count);
+                    }
                 }
             }
```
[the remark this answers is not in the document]
jac0626: Thanks. The first pass computes the exact total_path_count required by Writer::Prepare, and the second pass writes the rows. Caching both counts and pointers would add O(data_num) temporary memory per hierarchy and additional state solely to avoid inexpensive zero-copy lookups. Without benchmark evidence that these lookups are material relative to index construction, I am keeping the simpler bounded-memory implementation.

include/vsag/dataset.h
```diff
@@ -337,10 +337,30 @@ class Dataset : public std::enable_shared_from_this<Dataset> {
...
+     * Unlike the pointer overload, the dataset stores this container by value, so its lifetime is
+     * independent of Owner().
+     *
+     * @param hierarchy_name The hierarchy name. An empty string targets the default hierarchy.
+     * @param paths Paths grouped by dataset element.
+     * @return DatasetPtr A shared pointer to the dataset with updated paths.
+     * @note This overload is non-virtual so extending the API does not change Dataset's vtable.
+     */
+    DatasetPtr
+    Paths(const std::string& hierarchy_name, std::vector<std::vector<std::string>> paths);
```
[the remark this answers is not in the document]
jac0626: Thanks. include/vsag/dataset.h already directly includes <vector> near the top of the file, so the new overload does not rely on a transitive include. No change is needed.

src/dataset_impl.cpp
```diff
@@ -629,31 +866,98 @@ DatasetImpl::Append(const DatasetPtr& other) {
...
+            const auto* paths = GetDatasetPaths(dataset, hierarchy_name, i, path_count);
+            if (paths == nullptr) {
+                throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                    "Cannot append a dataset without the required hierarchy paths");
+            }
+            if (path_count == 0) {
+                result.emplace_back();
+            } else {
+                result.emplace_back(paths, paths + path_count);
+            }
```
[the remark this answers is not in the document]
jac0626: Thanks. A non-null zero-count row is not reachable through the supported Dataset APIs: the structured Paths setter rejects empty inner rows, while the legacy and custom-Dataset fallback paths report one path, including the empty string for root membership. The branch therefore cannot create an invalid row for supported input; removing it would only be cosmetic and would not change observable behavior, so I am keeping the current code unchanged.

src/algorithm/pyramid/pyramid.cpp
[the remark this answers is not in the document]
jac0626: Thanks. This is the same performance tradeoff discussed in the earlier thread. The first pass computes the exact total required by Writer::Prepare and the second writes the rows. GetDatasetPaths performs a zero-copy lookup and does not parse path strings; caching `(pointer, count)` for every row would add O(data_num) temporary memory per hierarchy. Without benchmark evidence that the repeated lookup is material relative to index construction, I am keeping the simpler bounded-memory implementation.

src/dataset_impl.cpp
```diff
@@ -668,6 +972,12 @@ DatasetImpl::Append(const DatasetPtr& other) {
...
     }
+    retired_paths_.erase(std::remove_if(retired_paths_.begin(),
+                                        retired_paths_.end(),
+                                        [&replaced_strings](const std::string* paths) {
+                                            return replaced_strings.count(paths) != 0;
+                                        }),
+                         retired_paths_.end());
     for (const auto* values : replaced_strings) {
         delete[] values;
     }
```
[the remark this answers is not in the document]
jac0626: Thanks. DatasetImpl::Append already rejects a non-owner destination at function entry with `Cannot append to a non-owner dataset`; that guard is also present on current upstream main. Therefore this deletion is reachable only when the destination owns its pointer-backed buffers, and the newly allocated appended buffers remain owned by the dataset. No code change is needed.

LHT129 (approved):
LGTM

PR #2807, opened 2026-08-31
fix(build): scope VSAG compile flags

### LINKED ISSUE #2805: [bug](build): compilation fails with GCC 12+ when compile flags leak into third-party targets
| Field | Value |
|---|---|
| Affected area | `build` |
| Interface | `cpp` |
| VSAG version | `e6dabf29` |

---

## Summary

VSAG does not build with GCC 12 or newer. Two translation units fail, and the
first fails only because of a CMake issue with two further consequences.

`vsag_add_compile_flag()` (`cmake/VSAGHelpers.cmake:24`) is a bare
`add_compile_options("${flag}")`. Its siblings `vsag_add_c_compile_flag()` and
`vsag_add_cxx_compile_flag()` (lines 41 and 51) correctly wrap the flag in
`$<$<COMPILE_LANGUAGE:...>>`; the generic helper does not, so its flags reach
every language and every target created below the current directory.

`cmake/VSAGCompilerConfig.cmake` is included at `CMakeLists.txt:45`, before
`cmake/VSAGThirdParty.cmake` at line 49 pulls CRoaring in with
`add_subdirectory()` (`extern/roaringbitmap/roaringbitmap.cmake:32`). VSAG's
`-Werror`, `-Ofast` and `-fno-omit-frame-pointer` are therefore inherited by
third-party sources that never asked for them.

The two failures:

1. **`CRoaring/src/containers/run.c`** — the leaked `-Ofast` (overriding
   CRoaring's own `-O3`) makes GCC 12+ emit a false-positive
   `-Wstringop-overflow`, and the leaked `-Werror` makes it fatal. CRoaring
   builds clean at its own `-O3`; this failure exists only because of the leak.
2. **`src/datacell/sparse_duplicate_tracker.cpp:82`** — a GCC 12/13 false
   positive in VSAG's own code, independent of optimization level.

Two more consequences of the same leak, worth fixing together:

- **CUDA targets cannot build.** `nvcc` rejects the inherited C/C++ flags:
  `nvcc fatal : Value '-fno-omit-frame-pointer' is not defined for option 'Werror'`.
- **`-Ofast` implies `-ffast-math`**, silently applied to every dependency added
  with `add_subdirectory()`. CRoaring is integer-only and unaffected, but this
  does not look deliberate.

## Minimal reproducer

```bash
git clone https://github.com/antgroup/vsag.git && cd vsag
git checkout e6dabf29
CC=gcc-13 CXX=g++-13 make release
```

To isolate failure 1 without a full configure — the flags below are copied
verbatim from the CMake-generated
`build/_deps/roaringbitmap-build/src/CMakeFiles/roaring.dir/flags.make`:

```bash
cd build/_deps/roaringbitmap-src
for O in -O2 -O3 -Ofast; do
  printf '%-8s -> ' "$O"
  gcc -c src/containers/run.c -o /dev/null \
      -Iinclude -Iinclude/roaring -Icpp -Isrc \
      -Wall -Wextra -Wsign-compare -Wshadow -Wwrite-strings -Wpointer-arith \
      -Winit-self -Wcast-align -DNDEBUG -fPIC -fno-omit-frame-pointer -g \
      -Wno-unused-function -std=gnu11 $O 2>&1 | grep -c stringop-overflow
done
```

GCC 13.4.0:

```
-O2      -> 0
-O3      -> 0
-Ofast   -> 2
```

CRoaring's `tools/cmake/FindOptions.cmake:27` sets only `-Wall` and friends, and
builds at `-O3`. It sets neither `-Werror` nor `-Ofast`; both come from VSAG.

The leak itself is visible independently of compiler version — this is from a
GCC 9 build that succeeds:

```
# build/_deps/roaringbitmap-build/src/CMakeFiles/roaring.dir/flags.make
# compile C with /usr/bin/gcc-9
C_FLAGS = -Wall -Wextra -Wsign-compare -Wshadow -Wwrite-strings -Wpointer-arith \
          -Winit-self -Wcast-align  -O3 -DNDEBUG -fPIC \
          -Werror -fno-omit-frame-pointer -g -Ofast  -Wno-unused-function -std=gnu11
#         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ contributed by VSAG
```

Failure 2 reproduces at `-O2`, `-O3` and `-Ofast` alike, with the flags from
`src/datacell/CMakeFiles/datacell.dir/flags.make`.

## Environment

```shell
- OS: Ubuntu 20.04
- vsag version: v1.0.0-117-g93344456
- compiler version: c++ (conda-forge gcc 13.4.0-19) 13.4.0

CPU:       2 x Intel Xeon Gold 6336Y @ 2.40GHz (96 threads)
Kernel:    5.15.0-139-generic
CMake:     3.20.6
Binutils:  GNU ld 2.34
CRoaring:  v3.0.1 (pinned in extern/roaringbitmap/roaringbitmap.cmake)

# GCC 9.4.0 on the same machine builds e6dabf29 successfully. The version
# string above is a local branch; both failures reproduce from a clean e6dabf29.
```

## Expected behavior

- `make release` succeeds with GCC 12 and 13, as it does with GCC 9.
- Dependencies added with `add_subdirectory()` compile with their own flags.
- `add_compile_options()` calls meant for C/C++ do not reach other languages.

## Actual behavior

**Failure 1 — `CRoaring/src/containers/run.c`:**

```
src/containers/run.c:156:9: error: 'memcpy' specified bound between 18446744065119617028
 and 18446744073709551612 exceeds maximum object size 9223372036854775807
 [-Werror=stringop-overflow=]
  156 |         memcpy(lo->runs, c->runs, lo_cap * sizeof(rle16_t));
```

`lo_cap` is a signed `int` guarded only against zero (`if (loc && lo_cap)`), and
the multiplication promotes it to `size_t`. The reported bounds are exactly
`(size_t)(-2147483647 * 4)` and `(size_t)(-1 * 4)`, so GCC is describing the
`lo_cap < 0` path — unreachable given `n_runs >= 0`. At `-Ofast` GCC 12+ inlines
`run_container_create_given_capacity()` and reasons about that path; GCC 9's
weaker interprocedural range analysis does not.

**Failure 2 — `src/datacell/sparse_duplicate_tracker.cpp:82`:**

```
    inlined from 'std::vector<unsigned int> vsag::{anonymous}::to_ordered_members(
      const std::vector<unsigned int>&)' at src/datacell/sparse_duplicate_tracker.cpp:82:16:
bits/stl_algobase.h:437:30: error: '__builtin_memmove' writing between 5 and 9223372036854775807
 bytes into a region of size 4 overflows the destination [-Werror=stringop-overflow=]
bits/new_allocator.h:151:55: note: destination object of size 4 allocated by 'operator new'
```

Line 82 is `return members;` on the `members.size() < 2` path, where the
allocation and the copy are both 0 or 4 bytes. After inlining the whole vector
copy constructor, GCC pairs the allocation size from that path with a memmove
length range derived from the `size() >= 2` path, hence the self-contradictory
"5 bytes into a region of size 4".

## References

- `cmake/VSAGHelpers.cmake:24` — unguarded `add_compile_options()`
- `cmake/VSAGHelpers.cmake:41,51` — the correctly guarded C/CXX variants
- `cmake/VSAGOptions.cmake:22` — `ENABLE_WERROR` defaults to `ON`
- `cmake/VSAGCompilerConfig.cmake:40,95,108` — `-Werror`, `-fno-omit-frame-pointer`, `-Ofast`
- `CMakeLists.txt:45,49` — compiler config included before third-party
- `extern/roaringbitmap/roaringbitmap.cmake:32` — CRoaring via `add_subdirectory()`
- `src/datacell/sparse_duplicate_tracker.cpp:82`

## Related issues

- #748 "Vsag doesn't support GCC >= 10." — **different root cause.** That one is
  ARM/el7, where OpenBLAS's `arch.cmake` selects `NEOVERSEV1 NEOVERSEN2` for
  GCC >= 10 and emits `-march=armv8.3-a` that the el7 assembler rejects. This
  report is x86_64 and concerns `-Wstringop-overflow` false positives caused by
  flag leakage. Both block newer GCC, but the fixes do not overlap.

## Additional context

Happy to send a PR. Proposed fix, in three parts:

1. **Scope the generic helper**, matching its two siblings:
   ```cmake
   add_compile_options ($<$<COMPILE_LANGUAGE:C,CXX>:${flag}>)
   ```
   This alone unblocks CUDA targets.
2. **Stop leaking VSAG's flags into third-party subdirectories**, so CRoaring
   builds at its own `-O3`. The `run.c` diagnostic then disappears and needs no
   suppression. A minimal alternative is
   `target_compile_options(roaring PRIVATE -Wno-stringop-overflow)` for GCC >= 12,
   but that leaves `-ffast-math` applied to third-party code.
3. **Fix `sparse_duplicate_tracker.cpp:82`.** Constructing from the range breaks
   the inlining pattern with identical semantics — a range construction from
   random-access iterators allocates exactly `size()` elements, as the copy
   constructor does:
   ```diff
   -        return members;
   +        return {members.begin(), members.end()};
   ```
   Verified on GCC 13.4.0 with the flags above: 1 diagnostic becomes 0.

Note that part 2 touches `extern/`, which `AGENTS.md` asks contributors to leave
alone unless a task explicitly requires it — please confirm which of the two
options you prefer. I can also split this into separate PRs for the CMake
scoping and the `sparse_duplicate_tracker` fix.

Found while preparing an environment for the OSPP 2026 project "VSAG GPU index
build acceleration"; part 1 is a prerequisite for adding any CUDA target.

wxyucs:
## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [x] CI/Build/Infra

## Linked Issue

- Fixes: #2805

## What Changed

- Restrict the generic VSAG compile-flag helper to C and C++ translation units.
- Exclude VSAG directory compile and link options while declaring in-tree third-party dependencies, then restore them before creating VSAG-owned targets.
- Add generated-command regression coverage for third-party, C, C++, non-C/C++ (ASM), and link-option scoping.

## Test Evidence

- [ ] `make fmt` (not run: compatible clang-format 15 is unavailable in the local environment)
- [ ] `make lint` (not run: compatible clang-tidy 15 is unavailable in the local environment)
- [x] Focused CMake regression suite
- [x] GCC 15 Release build with system OpenBLAS/LAPACKE
- [x] Focused unit tests

Test details:

```text
GCC 15.2.0, CMake 4.2.3, Ninja, Ubuntu 26.04

make test-cmake
  Compile flag scope checks passed

cmake -S . -B <release-build> -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DVSAG_USE_SYSTEM_OPENBLAS=ON \
  -DENABLE_TESTS=OFF
cmake --build <release-build> --parallel 48
  passed

<debug-build>/tests/unittests '[SparseDuplicateTracker]' --durations yes
  All tests passed (27 assertions in 5 test cases)

Generated Release and Sanitize commands were inspected directly:
- CRoaring keeps its own flags and receives no VSAG -Werror, -Ofast,
  frame-pointer, or sanitizer options.
- VSAG targets retain Werror, optimization, frame-pointer, and ASan/UBSan options.
```

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: build flags no longer leak from VSAG into in-tree third-party or non-C/C++ compilation

- Performance impact: VSAG optimization flags are unchanged; third-party targets use their own optimization settings
- Concurrency/thread-safety impact: none

## Documentation Impact

- [x] No docs update needed

## Risk and Rollback

- Risk level: low
- Rollback plan: revert the single commit

## Checklist

- [x] I have added regression coverage
- [x] No documentation update is needed
- [x] The commit follows Conventional Commits and DCO requirements


tests/cmake/compile_flag_scope_test.cmake
```diff
@@ -0,0 +1,95 @@
...
+string (RANDOM LENGTH 12 ALPHABET 0123456789abcdef fixture_suffix)
+set (fixture_source "${VSAG_SOURCE_DIR}/tests/cmake/compile_flag_scope_fixture")
+set (fixture_build "/tmp/vsag-compile-flag-scope-${fixture_suffix}")
+
+execute_process (
+    COMMAND ${CMAKE_COMMAND}
+            -S ${fixture_source}
+            -B ${fixture_build}
+            -G Ninja
+            -DVSAG_SOURCE_DIR=${VSAG_SOURCE_DIR}
```
LHT129, line 28: [suggestion] The test fixture hardcodes `-G Ninja`, which means this test will fail on systems without Ninja installed. Consider either documenting this requirement or using the default generator (omit `-G Ninja`) and parsing the build system output in a generator-agnostic way.

If the project already requires Ninja for all CMake workflows, this is fine — but the existing `Makefile` targets (e.g., `release_build_modes_test.cmake`) do not pass `-G`, so this introduces a new implicit dependency.

tests/cmake/compile_flag_scope_test.cmake
```diff
@@ -0,0 +1,95 @@
...
+
+get_compile_command (owned_asm_probe.S owned_asm_command)
+reject_flag ("${owned_asm_command}" "VSAG_C_AND_CXX_OPTION" "VSAG-owned ASM target")
+reject_flag ("${owned_asm_command}" "VSAG_C_OPTION" "VSAG-owned ASM target")
+reject_flag ("${owned_asm_command}" "VSAG_CXX_OPTION" "VSAG-owned ASM target")
+
+string (REGEX MATCH "LINK_FLAGS = [^\n]*--as-needed" owned_link_options "${ninja_content}")
+if (NOT owned_link_options)
+    file (REMOVE_RECURSE "${fixture_build}")
+    message (FATAL_ERROR "VSAG-owned link options were not restored")
```
LHT129, line 91: [suggestion] The link-options verification regex `LINK_FLAGS = [^\n]*--as-needed` on `build.ninja` is generator-specific — it will only work with the Ninja generator. If the `-G Ninja` requirement is kept, this is fine; if the test is made generator-agnostic, this check would need to be adapted to parse the output of whatever generator is used.

tests/cmake/compile_flag_scope_test.cmake
```diff
@@ -0,0 +1,95 @@
...
+    string (REGEX MATCH
+            "[^\n]*${source}[^\n]*"
+            command "${compile_commands}")
+    if (NOT command)
+        file (REMOVE_RECURSE "${fixture_build}")
+        message (FATAL_ERROR "No compile command found for ${source}")
+    endif ()
+    set (${output_variable} "${command}" PARENT_SCOPE)
+endfunction ()
+
```
LHT129, line 52: [note] The `get_compile_command` function uses `REGEX MATCH` with `[^\n]*${source}[^\n]*` to find the compile command for a source file. This regex matches the first line in `compile_commands.json` that contains the source filename anywhere — including in flag values or include paths. In practice this is unlikely to cause false matches with the current short probe filenames, but if probe filenames ever overlap with flag substrings (e.g., a file named `option.c` matching `-DOPTION`), the match could be ambiguous. Consider anchoring the match to the `"file"` JSON key or using a more precise pattern.

tests/cmake/compile_flag_scope_test.cmake
wxyucs: Thanks—acknowledged. This is intentionally a focused Ninja-specific generated-command fixture: it asserts directly against generated files to verify exact compile and link scoping, rather than serving as generator-agnostic coverage. Ninja is therefore a deliberate fixture prerequisite, so I am keeping this unchanged.

tests/cmake/compile_flag_scope_test.cmake
wxyucs: Agreed; this assertion is intentionally Ninja-specific. The focused fixture selects Ninja so it can make deterministic assertions against the generated LINK_FLAGS in build.ninja; it is not intended as generator-agnostic CMake coverage. Keeping this unchanged.

tests/cmake/compile_flag_scope_test.cmake
wxyucs: Thanks for noting this. The focused fixture uses unique, purpose-specific probe source names for deterministic generated-command assertions, and none overlap with current flag values or include paths. The hypothetical ambiguity does not affect the current checks, so I am keeping the focused regex unchanged.

LHT129 (approved):
LGTM

PR #2813, opened 2026-08-31
perf(hgraph): optimize RaBitQ split search hot path

LightWant:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

## What Changed

- Cache the RaBitQ split-storage mode, filter/reorder bit counts, plane sizes, and metadata offsets when the quantizer is constructed or deserialized.
- Replace split search hot-path version-string checks and recursive offset calculations with cached constant reads.
- Reuse quantizer `Computer` instances across routing, bottom-layer search, reorder, brute-force, reasoning, MCI, and conjugate search within one HGraph request.
- Key the request-scoped pool by the actual flatten data-cell identity and query pointer. Parallel workers keep independent computers; the coordinator reuses the request-scoped one.
- Fall back to serial search when an HGraph has no parallel executor, and expose construction/fallback counters for verification.
- Adapt one-bit, supplement, and optimized-build prefetch lengths to their actual split record sizes, with `prefetch_depth_codes` retained as the upper bound.
- Keep the public API and serialized representation unchanged.

The branch was rebased onto latest `origin/main` at `5096a0a6` before final verification.

## Test Evidence

- [x] `make fmt`
- [ ] `make lint`
- [x] Focused unit and functional tests
- [x] Release build and GIST1M benchmark

```text
make fmt
  clang-format 15, passed; no tracked diff

Focused unittests after rebase
  QueryComputerPool, ParallelSearcher, RaBitQSplitDataCell,
  and RaBitQ split-layout round trip
  13 test cases, 9,816 assertions, passed

Focused functests after rebase
  HGraph/Pyramid RaBitQ split
  12 test cases, 13,118 assertions, passed

Release benchmark builds
  latest main 5096a0a6 and optimized 025a7c34
  -O3 -DNDEBUG, passed

GIST1M result equivalence audit
  600 held-out queries, efSearch 32/64/128/256/512/768/1000
  main and optimized top-10 ID checksums matched at all seven points
```

Earlier full unit-test evidence remains 800 test cases and 85,349,251 assertions passed. The focused suite above was rebuilt and rerun after the rebase.

## Compatibility Impact

- API/ABI: no public API change; only private request context and quantizer layout state change.
- Serialization: no format or size change.
- Search behavior: main and optimized top-10 IDs, recall, and ordering match on the non-relayout GIST1M indices at every tested `efSearch`.

GIST1M setup:

- Baseline: latest `origin/main@5096a0a6`.
- Optimized: this branch at `025a7c34`, containing all three commits. The main-to-optimized numbers are branch-level results, not prefetch-only attribution.
- HGraph RaBitQ split `3+5` and `1+7`; same preserved 1M-vector indices.
- 600 held-out queries (`400..999`), `k=10`, external client threads `1/16/32`, and HGraph `parallelism=1` per call.
- `efSearch=32/64/128/256/512/768/1000`; three measured runs per point, median QPS.
- CPUs `0..51`; `OMP_NUM_THREADS=1`, `OPENBLAS_NUM_THREADS=1`.
- TAGL rows use separately relaid-out copies trained from query rows `0..399`. TAGL is not part of this PR; it isolates the additional benefit of locality-aware serialization with the same optimized binary.

Relative QPS ranges across all seven `efSearch` values:

| Threads | Split | main -> optimized | optimized -> TAGL |
| ---: | --- | ---: | ---: |
| 1 | 3+5 | +81.83% to +93.75% | -0.00% to +4.88% |
| 1 | 1+7 | +78.72% to +110.56% | +1.09% to +4.65% |
| 16 | 3+5 | +81.81% to +90.14% | +0.09% to +2.10% |
| 16 | 1+7 | +77.15% to +113.13% | +0.95% to +2.27% |
| 32 | 3+5 | +74.64% to +107.83% | +0.95% to +6.25% |
| 32 | 1+7 | +65.37% to +123.37% | +2.19% to +9.78% |

Selected median-QPS endpoints:

| Threads | Algorithm | ef 32 | ef 256 | ef 1000 |
| ---: | --- | ---: | ---: | ---: |
| 1 | main 3+5 | 1,358.4 | 332.8 | 109.8 |
| 1 | optimized 3+5 | 2,470.1 | 607.6 | 211.3 |
| 1 | optimized 3+5 + TAGL | 2,527.7 | 637.2 | 211.3 |
| 1 | main 1+7 | 1,393.4 | 372.8 | 132.0 |
| 1 | optimized 1+7 | 2,510.7 | 720.6 | 273.7 |
| 1 | optimized 1+7 + TAGL | 2,548.1 | 754.1 | 283.8 |
| 16 | main 3+5 | 21,748.5 | 5,096.8 | 1,712.1 |
| 16 | optimized 3+5 | 39,540.8 | 9,688.7 | 3,222.1 |
| 16 | optimized 3+5 + TAGL | 40,247.7 | 9,697.2 | 3,289.7 |
| 16 | main 1+7 | 22,270.0 | 5,862.5 | 2,040.5 |
| 16 | optimized 1+7 | 39,451.0 | 11,766.8 | 4,278.0 |
| 16 | optimized 1+7 + TAGL | 40,229.1 | 11,883.7 | 4,375.4 |
| 32 | main 3+5 | 39,839.2 | 8,724.5 | 2,925.2 |
| 32 | optimized 3+5 | 69,573.8 | 17,639.5 | 5,974.0 |
| 32 | optimized 3+5 + TAGL | 73,923.0 | 18,177.2 | 6,111.4 |
| 32 | main 1+7 | 41,155.9 | 10,726.9 | 3,682.5 |
| 32 | optimized 1+7 | 68,059.9 | 22,167.1 | 8,057.8 |
| 32 | optimized 1+7 + TAGL | 74,713.3 | 22,686.7 | 8,234.0 |

Concurrency: cached layout and prefetch values are immutable. Query computers are request-local; parallel workers continue to construct and own one computer each.

## Documentation Impact

- [x] No public documentation update needed
- [ ] Updated public docs

## Risk and Rollback

- Risk: medium-low. Cached values are derived from the same fields used by the previous getters; request reuse is identity-checked; adaptive prefetch only changes read-ahead hints.
- Rollback: revert `025a7c34` for adaptive prefetch, `2ed91582` for request-scoped computer reuse, or `2a2d6c00` for split-layout caching.

## Checklist

- [ ] I have linked the relevant issue (required for `kind/bug` and `kind/feature`)
- [x] I have added/updated tests for changed behavior
- [x] I have considered API and serialization compatibility
- [x] My commit messages follow project conventions and retain DCO sign-offs

### DIFF
--- a/src/algorithm/hgraph/hgraph_search.cpp
+++ b/src/algorithm/hgraph/hgraph_search.cpp
@@ -22,6 +22,7 @@
 #include "hgraph.h"  // IWYU pragma: keep
 #include "impl/filter/iterator_filter.h"
 #include "impl/heap/standard_heap.h"
+#include "impl/query_computer_pool.h"
 #include "impl/reasoning/search_reasoning.h"
 #include "utils/search_threshold.h"
 #include "utils/util_functions.h"
@@ -33,12 +34,6 @@ make_empty_dataset_with_stats(const SearchStatistics& stats) {
     auto dataset_result = DatasetImpl::MakeEmptyDataset();
     dataset_result->Statistics(stats.Dump());
     return dataset_result;
-}
-
-static DatasetPtr
-make_empty_dataset_with_stats() {
-    SearchStatistics stats;
-    return make_empty_dataset_with_stats(stats);
 }
 
 DatasetPtr
@@ -80,7 +75,7 @@ HGraph::KnnSearch(const DatasetPtr& query,
     }
 
     if (GetNumElements() == 0) {
-        return DatasetImpl::MakeEmptyDataset();
+        return make_empty_dataset_with_stats(stats);
     }
     this->validate_knn_args(query, k);
 
@@ -107,7 +102,7 @@ HGraph::KnnSearch(const DatasetPtr& query,
         auto cur_count = this->total_count_.load();
 
         if (cur_count == 0) {
-            return make_empty_dataset_with_stats();
+            return make_empty_dataset_with_stats(stats);
         }
         auto* new_ctx = new IteratorFilterContext();
         if (auto ret = new_ctx->init(cur_count, params.ef_search, ctx.alloc); not ret.has_value()) {
@@ -120,6 +115,8 @@ HGraph::KnnSearch(const DatasetPtr& query,
 
     auto* iter_filter_ctx = static_cast<IteratorFilterContext*>(iter_ctx);
     const auto* query_data = get_data(query);
+    QueryComputerPool query_computer_pool(query_data, &stats);
+    ctx.computer_pool = &query_computer_pool;
     // Note: brute_force_threshold is intentionally not applied here. The
     // iterator KnnSearch API pages results across multiple calls via
     // iter_filter_ctx; a single brute-force sweep would either need to drive
@@ -143,7 +140,7 @@ HGraph::KnnSearch(const DatasetPtr& query,
             search_param.is_inner_id_allowed = nullptr;
             search_param.enable_rabitq_one_bit_search = params.rabitq_one_bit_search;
             if (search_param.ep == INVALID_ENTRY_POINT) {
-                return make_empty_dataset_with_stats();
+                return make_empty_dataset_with_stats(stats);
             }
             if (iter_filter_ctx->IsFirstUsed()) {
                 ScopedDistancePhase routing_phase(ctx, DistanceEvaluationPhase::ROUTING);
@@ -282,7 +279,10 @@ HGraph::search_one_graph(const void* query,
         visited_list->Reset();
     }
     DistHeapPtr result = nullptr;
-    if (inner_search_param.parallel_search_thread_count > 1) {
+    const bool parallel_search_requested = inner_search_param.parallel_search_thread_count > 1;
+    const bool has_parallel_search_executor =
+        this->thread_pool_ != nullptr and this->parallel_searcher_ != nullptr;
+    if (parallel_search_requested and has_parallel_search_executor) {
         result = this->parallel_searcher_->Search(graph,
                                                   flatten,
                                                   visited_list,
@@ -292,6 +292,9 @@ HGraph::search_one_graph(const void* query,
                                                   ctx,
                                                   rabitq_lower_bound_candidates);
     } else {
+        if (parallel_search_requested and ctx != nullptr and ctx->stats != nullptr) {
+            ctx->stats->parallel_search_fallback_count.fetch_add(1, std::memory_order_relaxed);
+        }
         result = this->searcher_->Search(graph,
                                          flatten,
                                          visited_list,
@@ -317,6 +320,10 @@ HGraph::search_one_graph(const void* query,
                          QueryContext* ctx,
                          DistanceRecordVector* rabitq_lower_bound_candidates) const {
     auto visited_list = this->pool_->TakeOne();
+    if (inner_search_param.parallel_search_thread_count > 1 and ctx != nullptr and
+        ctx->stats != nullptr) {
+        ctx->stats->parallel_search_fallback_count.fetch_add(1, std::memory_order_relaxed);
+    }
     auto result = this->searcher_->Search(graph,
                                           flatten,
                                           visited_list,
@@ -366,7 +373,8 @@ HGraph::brute_force_search(const void* query,
         return result;
     }
 
-    auto computer = flatten->FactoryComputer(query);
+    auto computer_lease = AcquireQueryComputer(flatten, query, ctx);
+    const auto& computer = computer_lease.computer;
 
     constexpr InnerIdType brute_force_batch_size = 64;
     Vector<InnerIdType> batch_ids(brute_force_batch_size, alloc);
@@ -475,9 +483,17 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
     }
     const auto element_count = GetNumElements();
     if (element_count == 0) {
-        return make_empty_dataset_with_stats();
+        return make_empty_dataset_with_stats(stats);
     }
     k = std::min(k, element_count);
+    const auto* raw_query = use_custom_distance ? nullptr : get_data(query);
+    QueryComputerPool query_computer_pool(raw_query, &stats);
+    if (not use_custom_distance) {
+        ctx.computer_pool = &query_computer_pool;
+    }
+    if (this->entry_point_id_ == INVALID_ENTRY_POINT) {
+        return make_empty_dataset_with_stats(stats);
+    }
 
     // Setup reasoning context (KNN only)
     std::shared_ptr<ReasoningContext> reasoning_ctx;
@@ -499,16 +515,12 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
         reasoning_ctx->InitializeExpectedTargets(expected_labels_vec, label_to_inner_id);
 
         FlattenInterfacePtr precise_flatten = nullptr;
+        ComputerLease computer_lease;
         ComputerInterfacePtr computer = nullptr;
         if (not use_custom_distance) {
-            precise_flatten = this->basic_flatten_codes_;
-            if (use_reorder_) {
-                precise_flatten = this->high_precise_codes_;
-            }
-            if (create_new_raw_vector_) {
-                precise_flatten = this->raw_vector_;
-            }
-            computer = precise_flatten->FactoryComputer(get_data(query));
+            precise_flatten = this->get_precise_codes();
+            computer_lease = AcquireQueryComputer(precise_flatten, raw_query, &ctx);
+            computer = computer_lease.computer;
         }
         for (const auto& pair : label_to_inner_id) {
             float dist = 0.0F;
@@ -537,10 +549,6 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
     search_param.distance_batch_func = request.distance_batch_func_;
     search_param.distance_batch_size = request.distance_batch_size_;
 
-    if (search_param.ep == INVALID_ENTRY_POINT) {
-        return make_empty_dataset_with_stats();
-    }
-
     struct HGraphVisitedListGuard {
         std::shared_ptr<VisitedListPool> pool;
         VisitedListPtr visited_list;
@@ -560,7 +568,6 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
     HGraphVisitedListGuard vt_guard{this->pool_, this->pool_->TakeOne()};
     auto& vt = vt_guard.visited_list;
 
-    const auto* raw_query = use_custom_distance ? nullptr : get_data(query);
     ctx.distance_phase = DistanceEvaluationPhase::ROUTING;
     for (auto i = static_cast<int64_t>(this->route_graphs_.size() - 1); i >= 0; --i) {
         auto result = this->search_one_graph(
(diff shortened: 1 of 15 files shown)


src/quantization/rabitq_quantization/rabitq_quantizer.cpp
```diff
@@ -2260,6 +2306,10 @@ RaBitQuantizer<metric>::DeserializeImpl(StreamReader& reader) {
     if (pca_dim_ != this->original_dim_) {
         this->pca_->Deserialize(reader);
     }
+    RefreshSplitLayout(RaBitQuantizerParameter::IsSplitVersion(rabitq_version_) &&
+                       num_bits_per_dim_query_ == 32 && num_bits_per_dim_base_ >= 1 &&
+                       num_bits_per_dim_filter_ >= 1 &&
+                       num_bits_per_dim_filter_ <= num_bits_per_dim_base_);
 }
```
LHT129, line 2314: [note] The `support_split_code_storage` condition is duplicated between the constructor (line ~90) and `DeserializeImpl` (line ~2309). If the condition logic changes in the future, both sites need to be updated. Consider extracting the condition into a helper (e.g., a private `IsSplitCodeStorageEnabled()` that reads member fields) and calling it from both places to reduce maintenance risk.

src/impl/query_computer_pool.h
```diff
@@ -0,0 +1,119 @@
...
+        return lease;
+    }
+
+    [[nodiscard]] uint64_t
+    Size() const {
+        return size_;
+    }
+
+private:
+    static constexpr uint64_t kCapacity = 3;
```
LHT129, line 94: [suggestion] The pool capacity `kCapacity = 3` is a magic number tied to the assumption that HGraph has at most three flatten data cells (base, precise, raw). If a fourth cell type is ever added, this will fail silently at runtime with the error message about base/precise/raw cells. Consider making the capacity configurable via a template parameter or using a vector with dynamic growth to avoid a hard-to-diagnose runtime failure.

src/impl/query_computer_pool.h
```diff
@@ -0,0 +1,119 @@
...
+#include <cstdint>
+#include <utility>
+
+#include "datacell/flatten_interface.h"
+#include "query_context.h"
+#include "vsag_exception.h"
+
+namespace vsag {
+
+struct ComputerLease {
```
LHT129, line 28: [note] `ComputerLease` is returned by value from `Acquire()` and relies entirely on compiler-generated copy/move constructors. This works correctly because the members are `shared_ptr` (reference-counted), but the implicit copy means every call to `Acquire()` that hits the cache copies two shared_ptrs. This is fine for correctness but worth being aware of — if `ComputerLease` ever gains non-trivial members, the implicit copy could become a bug. Consider adding `= default` for the special members to make the intent explicit.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -292,6 +292,9 @@ HGraph::search_one_graph(const void* query,
                                                   ctx,
                                                   rabitq_lower_bound_candidates);
     } else {
+        if (parallel_search_requested and ctx != nullptr and ctx->stats != nullptr) {
+            ctx->stats->parallel_search_fallback_count.fetch_add(1, std::memory_order_relaxed);
+        }
```
LHT129, line 297: [note] The `IteratorFilterContext` overload of `search_one_graph` unconditionally increments `parallel_search_fallback_count` when `parallel_search_thread_count > 1`, even though this path never attempts parallel search (it always uses the serial searcher). This means the counter conflates two scenarios: (1) parallel was requested but the executor was unavailable, and (2) parallel was requested but this search path doesn't support it. If the counter is meant for diagnostics, consider distinguishing these cases or documenting that the counter includes both.

LHT129 (commented):
## Review Summary

This PR introduces two performance optimizations for the HGraph search path:

1. **RaBitQ split layout caching** (`RefreshSplitLayout`): Pre-computes all split layout constants (plane sizes, offsets, bit counts, stored plane indices) once during construction/deserialization instead of computing them on every hot-path getter call. This is a clean optimization with clear correctness.

2. **Query computer pool** (`QueryComputerPool`): Reuses quantizer `Computer` instances across routing, bottom-layer search, reorder, brute-force, reasoning, MCI, and conjugate search within a single HGraph request, avoiding redundant `FactoryComputer` calls that may involve memory allocation and precomputation.

### What looks good
- The `ComputerLease` RAII pattern correctly keeps the data cell alive via `FlattenInterfacePtr owner`
- The pool is request-scoped (lives on the stack in `SearchWithRequest` / `KnnSearch`), so no cross-request leakage
- `Validate()` provides defense-in-depth against misuse
- The parallel search fallback logic (`has_parallel_search_executor` check) is a good improvement over just checking `parallel_search_thread_count > 1`
- New test coverage for the pool, parallel fallback, and RaBitQ round-trip serialization
- The `INVALID_ENTRY_POINT` check moved earlier avoids unnecessary work

### Suggestions
- The pool capacity `kCapacity = 3` is a magic number tied to the assumption of exactly three cell types. Consider making it configurable.
- `ComputerLease` relies on implicit compiler-generated copy/move — consider making this explicit with `= default`.

### Existing note
LHT129 already flagged the duplicated `support_split_code_storage` condition between the constructor and `DeserializeImpl` — worth addressing to reduce maintenance risk.

LightWant:
## Adaptive split-prefetch update

Commit `312b7773` makes the split prefetch length depend on each plane set's actual record size, while keeping `prefetch_depth_codes` as the upper bound.

For the profiled GIST1M 960D 1+7-bit layout at the default depth of 16, the one-bit and supplement paths now prefetch 3 and 14 cache lines respectively, instead of 16 lines for both paths. The serialized format and index size are unchanged.

Same serialized TAGL index, 600 held-out queries (rows 400..999), `k=10`, one search thread, ABBA process order:

| ef_search | full before QPS | full after QPS | change | no-reorder before QPS | no-reorder after QPS | change |
|---:|---:|---:|---:|---:|---:|---:|
| 32 | 1587.764 | 1686.795 | +6.24% | 2548.395 | 2869.915 | +12.62% |
| 256 | 549.794 | 654.269 | +19.00% | 793.832 | 944.962 | +19.04% |
| 1000 | 211.794 | 262.490 | +23.94% | 272.454 | 342.284 | +25.63% |

For every corresponding before/after run, result checksums, recall, routing/approximation/rerank counts, hops, RaBitQ filter/full counts, and query-computer counts were identical. Sampling at `ef_search=256` reduced the share attributed to `PrefetchLines` from 27.701% to 11.167%; sampled QPS itself is omitted because the signal profiler materially perturbed throughput.

Additional validation:

- `make fmt`: passed with clang-format 15.
- `[ut][RaBitQSplitDataCell]`: 9 cases, 9,607 assertions passed.
- Full C++ unit suite: 801 cases, 85,349,020 assertions passed.
- New 960D regression covers depths 0, 1, 2, and 16 and verifies independent one-bit/supplement cached byte counts.

src/quantization/rabitq_quantization/rabitq_quantizer.h
```diff
@@ -335,6 +336,40 @@ class RaBitQuantizer : public Quantizer<RaBitQuantizer<metric>> {
...
+        uint64_t one_bit_record_one_bit_error_offset{0};
+        uint64_t one_bit_record_size{0};
+        uint64_t one_bit_code_size{0};
+        uint64_t supplement_code_size{0};
+        uint64_t supplement_norm_code_offset{0};
+        uint64_t supplement_norm_offset{0};
+        uint64_t supplement_error_offset{0};
+        uint64_t supplement_mrq_norm_offset{0};
+        uint64_t supplement_raw_norm_offset{0};
+        std::array<uint64_t, 8> sequential_plane_offsets{};
```
LHT129, line 365: [suggestion] The `SplitLayout` struct uses hard-coded array sizes of 8 for `sequential_plane_offsets`, `stored_plane_indices`, and `stored_plane_offsets`. In `RefreshSplitLayout`, the loop iterates up to `num_bits_per_dim_base_` without bounds checking. If a future configuration uses more than 8 base bits, this will silently write past the end of the array (undefined behavior). Consider using `std::vector` sized at construction time, or adding a `static_assert` / runtime check that `num_bits_per_dim_base_ <= 8`.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -560,7 +568,6 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
     HGraphVisitedListGuard vt_guard{this->pool_, this->pool_->TakeOne()};
     auto& vt = vt_guard.visited_list;
 
-    const auto* raw_query = use_custom_distance ? nullptr : get_data(query);
     ctx.distance_phase = DistanceEvaluationPhase::ROUTING;
```
LHT129, line 571: [note] The `entry_point_id_` check in `SearchWithRequest` has been moved after the `QueryComputerPool` creation. Previously, the check was at the top of the function and would early-return before any resource allocation. Now, if `entry_point_id_` is unset, the function still creates a `QueryComputerPool` (and potentially a `FlattenInterfacePtr` via `GetFlattenCell`) before returning an error. This is a minor behavioral change — the pool and cell allocation are wasted when `entry_point_id_` is invalid. Consider keeping the early check before pool creation, or documenting that the pool must be created first for correctness (e.g., if pool construction has side effects needed by subsequent code).

src/impl/query_computer_pool.h
```diff
@@ -0,0 +1,119 @@
...
+}
+
+class QueryComputerPool {
+public:
+    explicit QueryComputerPool(const void* query, SearchStatistics* stats = nullptr)
+        : query_(query), stats_(stats) {
+    }
+
+    [[nodiscard]] ComputerLease
+    Acquire(const FlattenInterfacePtr& cell, const void* query) {
```
LHT129, line 56: [note] `QueryComputerPool::Acquire` returns cached entries by copy (`return entry;` at line 56) rather than by move. The `ComputerLease` constructor that creates new entries uses `std::move(computer)`, but the cached path copies the `ComputerLease` which includes a `shared_ptr` (the `owner`) and a `unique_ptr`-like computer. Since `ComputerLease` contains `FlattenInterfacePtr owner` (a `shared_ptr`) and a computer object, copying increments the shared_ptr refcount unnecessarily. Consider returning `std::move(entry)` or restructuring to avoid the copy. The practical impact is minimal since `kCapacity` is only 3, but the inconsistency between creation (move) and retrieval (copy) could be confusing.

LHT129 (commented):
Reviewed the full diff (15 files, +1066/-133). The PR introduces two well-motivated performance optimizations: (1) caching RaBitQ split-storage layout metadata in the quantizer via `SplitLayout` to avoid recomputing plane offsets and indices on every call, and (2) reusing `ComputerInterface` instances across search phases via `QueryComputerPool` to reduce repeated `FactoryComputer` allocations.

**Summary of findings:**

- The core design is sound. `SplitLayout` correctly precomputes all layout values in `RefreshSplitLayout` and the fast-path in `GetStoredPlane` (comparing `plane_bytes == split_layout_.plane_bytes`) is a clean optimization.
- `QueryComputerPool` with RAII `ComputerLease` provides safe reuse with identity validation. The fallback path via `AcquireQueryComputer` free function ensures backward compatibility when no pool is available.
- Test coverage is thorough: unit tests for pool reuse/capacity/isolation, quantizer layout round-trip, parallel searcher validation, and integration tests in `test_hgraph.cpp`.
- 7 inline comments have already been posted covering: duplicate condition, magic number, copy semantics, fallback counting, array bounds, entry point ordering, and copy-vs-move in Acquire.

No blocking issues found. The existing comments address the main areas for improvement.

src/impl/query_computer_pool.h
LightWant: Thanks for the suggestion. We intentionally keep this request-scoped pool fixed-capacity because HGraph currently has at most three distinct flatten cells: base, precise, and raw. The fixed array avoids dynamic allocation on the search hot path. Exceeding the capacity does not fail silently: it throws an explicit internal error, and the fourth-cell case is covered by a unit test. If HGraph introduces another flatten cell type in the future, the capacity and its test should be updated together. We will keep the current implementation.

wxyucs (approved):
lgtm

PR #2820, opened 2026-09-01
refactor(datacell): use FixedLayout for graph records

### LINKED ISSUE #2817: [Refactor] Use FixedLayout for GraphDataCell record addressing
## Description

Refactor the fixed-record `GraphDataCell<IOTmpl>` storage path to use the common `FixedLayout` abstraction instead of calculating absolute byte offsets and operating on `BasicIO` directly.

## Background

`GraphDataCell` stores each node in a fixed-size record containing a neighbor count followed by a fixed-capacity neighbor array. This matches `FixedLayout`, but GraphDataCell currently duplicates record-to-byte address calculations across reads, writes, prefetch, resize, shrink, and serialization paths.

Graph operations also rely on partial record access: they read the 4-byte count separately and only read or write active neighbors. The refactor therefore requires generic record-local byte access in `FixedLayout` without converting Graph operations to full-record IO.

## Requirements

- Add record-local read, write, and prefetch operations to `FixedLayout`.
- Reject local reads and writes that cross a fixed-record boundary.
- Preserve prefetch hint behavior, including hints that may span records.
- Migrate only `GraphDataCell<IOTmpl>`; do not change sparse or compressed graph storage.
- Preserve partial IO sizes and the existing serialization byte order.
- Add unit coverage for local access boundaries and prefetch addressing.

## Acceptance Criteria

- `GraphDataCell<IOTmpl>` no longer calculates absolute record offsets or directly manages IO byte capacity.
- Existing GraphDataCell behavior and serialized data remain compatible.
- FixedLayout and GraphDataCell unit tests pass.
- Release compilation and project formatting/lint checks pass.

LHT129:
## Summary

Refactor fixed-record `GraphDataCell<IOTmpl>` storage to use the common `FixedLayout` abstraction while preserving the existing partial IO behavior and serialized byte format.

## Changes

- Add record-local `ReadAt`, `WriteAt`, and `PrefetchAt` operations to `FixedLayout`.
- Validate that record-local reads and writes do not cross a fixed-record boundary.
- Preserve prefetch hint behavior, including hints that may span record boundaries.
- Replace direct `BasicIO` addressing in `GraphDataCell` with a value-held `FixedLayout`.
- Keep neighbor-count and active-neighbor IO sizes unchanged.
- Restore the layout code size after deserialization without changing serialization order.
- Add unit coverage for local access boundaries, adjacent-record isolation, and prefetch offsets.

## Testing

- `make lint` with clang-tidy 15
- Release `datacell` target build
- `./build-test/tests/unittests "[FixedLayout]"`
  - 38 assertions in 7 test cases passed
- `./build-test/tests/unittests "[GraphDataCell]"`
  - 1,003,321 assertions in 7 test cases passed

## Related Issues

Fixes #2817

### DIFF
--- a/src/datacell/graph_datacell.h
+++ b/src/datacell/graph_datacell.h
@@ -27,6 +27,7 @@
 #include "impl/reverse_edge.h"
 #include "index_common_param.h"
 #include "io/common/basic_io.h"
+#include "layout/fixed_layout.h"
 #include "vsag/constants.h"
 
 namespace vsag {
@@ -73,12 +74,12 @@ public:
 
     inline void
     SetIO(std::shared_ptr<BasicIO<IOTmpl>> io) {
-        this->io_ = io;
+        this->layout_.SetIO(std::move(io));
     }
 
     virtual void
     InitIO(const IOParamPtr& io_param) override {
-        this->io_->InitIO(io_param);
+        this->layout_.InitIO(io_param);
     }
 
     /****
@@ -88,8 +89,9 @@ public:
      */
     void
     Prefetch(InnerIdType id, uint32_t neighbor_i) override {
-        io_->Prefetch(static_cast<uint64_t>(id) * static_cast<uint64_t>(this->code_line_size_) +
-                      sizeof(uint32_t) + neighbor_i * sizeof(InnerIdType));
+        const uint64_t offset_in_record =
+            NEIGHBORS_OFFSET + static_cast<uint64_t>(neighbor_i) * sizeof(InnerIdType);
+        layout_.PrefetchAt(id, offset_in_record, 64);
     }
 
     void
@@ -109,9 +111,7 @@ public:
     uint64_t
     GetMemoryUsage() const override {
         uint64_t memory = sizeof(GraphDataCell) + node_versions_.size() * sizeof(uint8_t);
-        if (IOTmpl::InMemory) {
-            memory += io_->GetMemoryUsage();
-        }
+        memory += layout_.GetMemoryUsage();
         if (reverse_edges_) {
             memory += reverse_edges_->GetMemoryUsage();
         }
@@ -137,8 +137,7 @@ public:
 
     void
     ShrinkToFit(InnerIdType capacity) override {
-        uint64_t io_size = static_cast<uint64_t>(capacity) * static_cast<uint64_t>(code_line_size_);
-        this->io_->Shrink(io_size);
+        this->layout_.Shrink(capacity);
         if (is_support_delete_) {
             node_versions_.resize(capacity);
         }
@@ -146,7 +145,10 @@ public:
     }
 
 protected:
-    std::shared_ptr<BasicIO<IOTmpl>> io_{nullptr};
+    static constexpr uint64_t COUNT_OFFSET = 0;
+    static constexpr uint64_t NEIGHBORS_OFFSET = sizeof(uint32_t);
+
+    FixedLayout<IOTmpl> layout_{};
 
     Vector<uint8_t> node_versions_;
 
@@ -191,7 +193,6 @@ template <typename IOTmpl>
 GraphDataCell<IOTmpl>::GraphDataCell(const GraphDataCellParamPtr& param,
                                      const IndexCommonParam& common_param)
     : node_versions_(common_param.allocator_.get()) {
-    this->io_ = std::make_shared<IOTmpl>(param->io_parameter_, common_param);
     this->maximum_degree_ = param->max_degree_;
     this->max_capacity_ = param->init_max_capacity_;
     this->is_support_delete_ = param->support_remove_;
@@ -199,6 +200,8 @@ GraphDataCell<IOTmpl>::GraphDataCell(const GraphDataCellParamPtr& param,
     this->id_bit_ = sizeof(InnerIdType) * 8 - this->remove_flag_bit_;
     this->remove_flag_mask_ = (1 << this->id_bit_) - 1;
     this->code_line_size_ = this->maximum_degree_ * sizeof(InnerIdType) + sizeof(uint32_t);
+    this->layout_.SetCodeSize(this->code_line_size_);
+    this->layout_.SetIO(std::make_shared<IOTmpl>(param->io_parameter_, common_param));
     this->allocator_ = common_param.allocator_.get();
     if (this->is_support_delete_) {
         node_versions_.resize(max_capacity_);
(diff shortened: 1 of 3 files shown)


LHT129 (commented):
Reviewed the refactoring of `GraphDataCell` to use `FixedLayout` for fixed-record addressing.

**Summary of findings:**

- `FixedLayout::WriteAt` / `ReadAt` with `GetOffsetAt` boundary validation is correctly implemented — zero-length operations at the record boundary are allowed, and any cross-boundary access throws `VsagException`.
- `FixedLayout::PrefetchAt` intentionally skips boundary checks (uses `AddOffset` directly), which is appropriate since prefetch hints are advisory and may span record boundaries.
- `FixedLayout::AddOffset` properly guards against `uint64_t` overflow.
- `GraphDataCell` changes are a faithful mechanical replacement — all manual offset calculations (`id * code_line_size_ + offset`) are replaced with `layout_.WriteAt`/`ReadAt`/`PrefetchAt`/`Resize`/`Shrink` calls.
- `Serialize`/`Deserialize` correctly restores `code_line_size_` via `SetCodeSize` after layout deserialization, preserving the serialized byte format.
- `GetMemoryUsage` delegates to `FixedLayout::GetMemoryUsage()`, which internally handles the `InMemory` check via `if constexpr` — functionally equivalent to the previous code.
- Tests cover: record-local read/write correctness, boundary rejection, adjacent-record isolation, prefetch offset computation, and zero-length boundary edge cases.

No blocking issues found. The refactoring is clean, well-tested, and preserves the existing serialization format.

LHT129:
## Performance assessment

I ran a paired `-O3 -DNDEBUG` microbenchmark on the same host and pinned CPU, alternating the old direct `BasicIO` address calculation and the new `FixedLayout` path in each of 11 rounds. The benchmark preserves the Graph record shape and partial IO pattern (4-byte count plus active neighbors). This benchmark file is temporary and is not part of the PR.

Representative medians (ns/op):

| operation | degree | direct | FixedLayout | delta |
|---|---:|---:|---:|---:|
| count read | 1-128 | 4.60-4.72 | 4.55-4.74 | -1.5% to +0.4% |
| prefetch | 1-128 | 8.13-8.21 | 8.13-8.22 | -0.5% to +0.2% |
| neighbor read | 32 | 17.87-18.84 | 18.58-19.68 | +3.1% to +6.2% |
| neighbor read | 64 | 66.96 | 67.77 | +1.2% |
| neighbor read | 128 | 113.14 | 113.99 | +0.7% |
| record write | 32-128 | 89.60-176.04 | 92.77-176.42 | +0.2% to +3.5% (noisier) |

Assembly inspection confirms that `ReadAt`, `WriteAt`, `GetOffsetAt`, and the CRTP IO calls are fully inlined; there is no additional function/virtual-call overhead. The measurable difference on small neighbor reads comes from the intentionally added record-boundary and integer-overflow checks. As copy/IO work grows, this fixed CPU cost is diluted.

Interpretation:

- `GetNeighborSize` and `Prefetch`: no measurable regression within benchmark noise.
- GraphDataCell-only active-neighbor reads: a small measurable CPU regression, worst around 3-6% for degree 32 in this deliberately lightweight in-memory microbenchmark, dropping toward ~1% or below for larger records.
- End-to-end graph search should see a smaller effect because distance computation, candidate-queue work, visited tracking, and random memory latency dominate; however, this PR does not claim zero overhead and no full end-to-end index benchmark was run here.
- IO call count, transferred byte count, allocation behavior, and serialized bytes are unchanged, so storage-backed IO should not have IO amplification.

Conclusion: the refactor has no IO amplification or call-indirection cost, but strict local-range/overflow validation introduces a small fixed CPU cost visible in the hottest small in-memory neighbor-read case. This is the safety/abstraction tradeoff of the current API.

LHT129:
## GIST end-to-end performance and serialization compatibility

I added an end-to-end check using the repository's `tools/eval/eval_performance` tool.

### Setup

- Baseline binary: `github/main@5096a0a6`
- PR binary: `38ea0d7e`
- Same compiler/options: Release, system OpenBLAS, GCC 11.4
- Dataset: first 100,000 real vectors from GIST-960, all 1,000 GIST test queries
- Exact top-100 ground truth recomputed with FAISS `IndexFlatL2`
- HGraph: SQ8 uniform, max degree 32, ef_construction 200
- Search: one pinned CPU, ef_search 40/80/160
- Main performance comparison loads the **same baseline-built serialized index** in both binaries, avoiding graph-build randomness.
- Five alternating runs per binary; medians reported.

### Same-index search result

| ef_search | recall | baseline QPS | PR QPS | QPS delta | baseline avg latency | PR avg latency | latency delta |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 40 | 0.8248 | 4924.40 | 4906.90 | -0.36% | 0.20081 ms | 0.20149 ms | +0.34% |
| 80 | 0.9004 | 3024.51 | 3062.40 | +1.25% | 0.32797 ms | 0.32389 ms | -1.24% |
| 160 | 0.9334 | 1816.26 | 1815.28 | -0.05% | 0.54767 ms | 0.54813 ms | +0.09% |

Recall, distance-comparison counts, and graph-hop counts were identical between baseline and PR for every parameter and run. The measured end-to-end differences are within run-to-run noise; there is no observable search regression on this GIST workload.

Build-only measurements from separate builds were also effectively unchanged (baseline 16,845.55 TPS vs PR 16,881.20 TPS, +0.21%). Independent graph builds are nondeterministic, so they were not used for the search comparison.

### Serialization compatibility

Validated both directions on the 100K GIST HGraph index:

1. Baseline-built index -> PR deserialize/search: success.
2. PR-built index -> baseline deserialize/search: success.
3. Baseline-built index, deserialize + reserialize with baseline and PR: the two reserialized outputs have the same size and identical SHA-256: `8a918eded696c450df42f81f5444366913d6a830ad239128f2bd6a030d155f7b`.
4. PR-built index, deserialize + reserialize with PR and baseline: the two reserialized outputs have the same size and identical SHA-256: `1f4e8089dc6cb76c500e770c6071afaa8589cb933396280f139977f2759ac1ad`.
5. Cross-version searches on each index produce the same recall, distance-comparison count, and hop count.

The original build output and its deserialize-reserialize output have different hashes even within the same version, so direct build-vs-reserialize byte comparison is not a valid determinism check. The controlled same-input reserialization comparison above shows that the GraphDataCell layout change does not alter serialized output.

### Conclusion

The earlier record-local microbenchmark showed a small fixed cost from the new safety checks in an isolated tiny-memory-copy loop. In the real GIST HGraph search, that cost is diluted by distance computation and traversal work: end-to-end QPS/latency remains within approximately +/-1.3% noise, with identical search quality and traversal behavior. Legacy and new serialized indexes are mutually readable, and controlled reserialization is byte-identical across versions.

wxyucs (approved):
lgtm

PR #2832, opened 2026-09-02
feat(ivf): add centroid scan routing switch

### LINKED ISSUE #2826: [feat](ivf): add a switch to route by scanning centroids instead of a routing graph
### Affected area

ivf

### Motivation

`IVFNearestPartition` always routes a vector to its bucket by running a
`KnnSearch` against a routing HGraph built over the centroids
(`src/algorithm/ivf/ivf_nearest_partition.cpp:100-110`). `KnnSearch` here is the
public per-query entry point, so it is invoked once per routed vector, with all
the per-call setup that entails. That is a fine trade when there are many
centroids, and a poor one when there are few.

The clearest evidence is that the router's cost barely moves with the number of
centroids. On 8 physical cores, dim 100, routing a batch through `Add`:

| `buckets_count` | HGraph router | plain SIMD scan over centroids |
| --- | --- | --- |
| 64 | 10.48 us | 3.26 us |
| 256 | 12.65 us | 6.82 us |
| 1024 | 14.51 us | 8.37 us |
| 4096 | 16.18 us | 18.95 us |
| 16384 | 15.43 us | 59.22 us |

The router changes by less than 50% while the centroid count grows 256x. Its
cost is therefore dominated by something other than traversing the graph.

Instrumenting one routing call directly (single-threaded, `buckets_count=256`,
dim 128) puts the components at:

| component | per vector |
| --- | --- |
| `Dataset::Make()` plus setters | 0.56 us |
| `fmt::format` of the search parameter | 0.16 us |
| the `KnnSearch` call itself (parameter parse, graph search, result `Dataset`) | 40.86 us |
| `GetStatistics` on the result (820 B JSON parse) | 26.01 us |
| measured task body | 72.25 us |

Two honest caveats on those numbers. First, we cannot separate graph traversal
from per-call setup inside the 40.86 us without further instrumentation, so we
are not claiming a specific split there; the flatness above is what argues the
setup dominates at small `buckets_count`. Second, the 26.01 us `GetStatistics`
term is the subject of #2825 and would go away with that fix alone — it should
not be counted as a benefit of this feature.

Who benefits: anyone building many IVF sub-indexes, each covering a shard of a
larger corpus, so that every sub-index ends up with a small `buckets_count` and
pays the per-call cost once per vector it ingests.

### Proposed solution

An additive build parameter on the IVF partition strategy, defaulting to today's
behaviour, that selects an exhaustive scan instead of the routing graph:

```json
{
  "index_param": {
    "use_route_graph": false
  }
}
```

When disabled:

- `factory_router_index` is not called, so no routing HGraph — and no HGraph
  datacells — are created at all.
- `Train` keeps the centroids in a contiguous array alongside their squared
  norms, so the l2 ordering key is `norm[b] - 2 * dot(q, c_b)` and the query norm
  drops out as a common term.
- `ClassifyDatas` scans the centroids and takes an exact top-k.

Two details worth calling out, since they are where a naive implementation goes
wrong:

- `buckets_count` is known from the create parameter, but the centroids only
  exist after `Train()`. `IndexImpl` deliberately admits an untrained index to a
  `disable_bucket_scan` search, where the graph router simply finds nothing and
  every slot stays `INVALID_BUCKET_ID`. A scan path has to check that data
  precondition explicitly rather than trusting `buckets_count > 0`.
- The distance ordering must match what the router HGraph would produce
  (squared l2 for l2, `1 - ip` for ip/cosine with pre-normalized centroids), and
  ties need a deterministic tiebreak so the assignment stays reproducible.

Serialization needs a way to tell the two router layouts apart on load. Deciding
it from the create parameter is not enough, because an index persisted by an
older build carries no such parameter and would be misread. Recording the layout
in the stream (a marker word, or a dedicated `StreamSerializationTag` alongside
`IVF_PARTITION_STRATEGY`) keeps old indexes loadable whatever the reader was
created with. Maintainers are better placed than we are to pick between those.

### Alternatives considered

**Keep the graph but make the internal call cheaper.** An internal `KnnSearch`
overload taking a `QueryContext*` would remove the statistics round trip
(26.01 us of the breakdown above; filed separately as #2825) without any new
parameter, and would help every internal caller rather than just IVF. That is
worth doing regardless of this request and is strictly better for large
`buckets_count`. It does not cover the small-`buckets_count` case, though: even
with statistics free, the remaining per-call work still has to be paid once per
vector, and at 64 centroids a direct scan is a few hundred SIMD instructions.

**A threshold instead of a switch.** Auto-selecting the scan below some
`buckets_count` avoids a user-facing knob, but the crossover depends on `dim` and
on the hardware, and we have only measured one `dim`. A switch lets callers who
know their shape opt in; a threshold could be layered on later once there is
enough data to place it.

**Tune `route_max_degree` / `route_ef_construction` down.** These shrink the graph
but not the per-call overhead, which is the dominant term at small
`buckets_count`, and they trade away routing recall.

**Do nothing and accept the cost.** Viable when `buckets_count` is large. Our
point is only that the current code has no way to express the small-`buckets_count`
case.

additive only (no break)

### References

- `src/algorithm/ivf/ivf_nearest_partition.cpp:100-110` — routing via public `KnnSearch`
- `src/algorithm/ivf/ivf_nearest_partition.cpp:105-107` — `ef_search = max(10, buckets_per_data * 1.2)`
- `src/algorithm/ivf/gno_imi_partition.cpp:39-56` — `matmul` wrapper over `BlasFunction::Sgemm`
- `src/algorithm/ivf/gno_imi_partition.cpp:377-378` — `GNOIMIPartition` build path routing by matmul
- `src/algorithm/ivf/gno_imi_partition.cpp:249-255` — `GNOIMIPartition` search path routing by matmul
- `src/algorithm/ivf/ivf.cpp:433`, `src/datacell/bucket_datacell.h:609` — build-path callers of `ClassifyDatas`

### Additional context

**The sibling partition strategy already routes this way.** `GNOIMIPartition`
computes all query-to-centroid inner products with `matmul`, a thin wrapper over
`BlasFunction::Sgemm` (`src/algorithm/ivf/gno_imi_partition.cpp:39-56`), on both
its build path (`:377-378`) and its search path (`:249-255`), then picks top-k
from the resulting matrix using precomputed centroid norms. So exhaustive
centroid routing is already an accepted pattern inside `src/algorithm/ivf/`; the
two IVF partition strategies are simply inconsistent about it today, and
`IVFNearestPartition` is the one paying the per-call overhead.

That also suggests the natural implementation: reuse `BlasFunction::Sgemm` over a
batch of queries rather than looping over per-centroid dot products. The scan
numbers in the table above come from a straightforward loop calling
`FP32ComputeIP` once per centroid, which is what we could measure; a GEMM-based
scan of the kind `GNOIMIPartition` already uses should push the crossover
further right, but we have not measured that.

**A scan is also exact, where the current router is not.** Routing searches with
`ef_search = max(10, buckets_per_data * 1.2)`
(`src/algorithm/ivf/ivf_nearest_partition.cpp:105-107`). Querying one index and
comparing the returned bucket order against distances computed directly through
`GetCentroid`, the order disagreed for 27 of 100 queries at `buckets_count=200`
under l2 and 34 of 100 under cosine. That is expected behaviour for an
approximate search, but it means bucket assignment today is approximate even
though nothing about IVF requires it to be.

**Scope note.** This is a separate root cause from #2824. That issue is about
`block_memory_io` reserving a 128 MiB block per datacell, which inflates the
router's construction cost and memory; this one is about the per-call cost of
routing through the public API. Skipping the router happens to remove both, which
is why the two are cross-referenced, but they are fixable independently.

_Drafted with AI assistant: Qoder:claude-sonnet-4.6_

LHT129:
## Summary

Add an additive `use_route_graph` IVF partition parameter. It defaults to `true` to preserve the existing routing behavior. When set to `false`, IVF keeps the trained centroids and routes vectors by an exact batched centroid scan instead of constructing and querying a routing HGraph.

## Changes

- add `use_route_graph` parameter parsing, serialization, external parameter mapping, and a backward-compatible default
- skip routing HGraph/datacell construction when centroid scan routing is selected
- compute query-to-centroid inner products with `BlasFunction::Sgemm` and select deterministic exact top-k buckets
- preserve L2, IP, and cosine distance ordering semantics
- add a self-describing centroid-scan serialization layout while keeping the legacy route-graph layout loadable with either reader configuration
- cover exact routing, untrained behavior, parameter mapping, serialization round trips, and cross-configuration loading

## Testing

- `make release`
- `make fmt`
- `make lint`
- `./build/tests/unittests '[ut][IVFNearestPartition]' --reporter compact`
- `./build/tests/unittests '[ut][IVFPartitionStrategyParameters]' --reporter compact`
- `./build/tests/unittests '[ut][IVFParameter]' --reporter compact`

Fixes #2826

### DIFF
--- a/src/algorithm/ivf/ivf_nearest_partition.cpp
+++ b/src/algorithm/ivf/ivf_nearest_partition.cpp
@@ -19,17 +19,54 @@
 
 #include <atomic>
 #include <cstdlib>
+#include <limits>
 
 #include "algorithm/hgraph/hgraph.h"
 #include "algorithm/inner_index_interface.h"
+#include "container_types.h"
 #include "impl/allocator/safe_allocator.h"
+#include "impl/blas/blas_function.h"
 #include "impl/cluster/kmeans_cluster.h"
 #include "inner_string_params.h"
 #include "query_context.h"
+#include "simd/fp32_simd.h"
 #include "utils/util_functions.h"
 #include "vsag_exception.h"
 
 namespace vsag {
+
+namespace {
+
+constexpr uint64_t K_CENTROID_SCAN_LAYOUT_MARKER = 0x565341475343414EULL;  // "VSAGSCAN"
+
+constexpr BucketIdType INVALID_BUCKET_ID = static_cast<BucketIdType>(-1);
+
+// Max-heap used to keep the k smallest routing keys: the largest retained key is at top() and
+// is discarded when a better candidate arrives. bucket_id provides a deterministic tie-break.
+using ScanRouteHeap = std::priority_queue<std::pair<float, InnerIdType>,
+                                          Vector<std::pair<float, InnerIdType>>,
+                                          std::less<>>;
+
+// C = A * B^T (same shape as GNOIMIPartition's batched centroid routing).
+void
+compute_centroid_dots(const float* A, const float* B, float* C, int64_t M, int64_t N, int64_t K) {
+    BlasFunction::Sgemm(BlasFunction::ColMajor,
+                        BlasFunction::Trans,
+                        BlasFunction::NoTrans,
+                        static_cast<int32_t>(N),
+                        static_cast<int32_t>(M),
+                        static_cast<int32_t>(K),
+                        1.0F,
+                        B,
+                        static_cast<int32_t>(K),
+                        A,
+                        static_cast<int32_t>(K),
+                        0.0F,
+                        C,
+                        static_cast<int32_t>(N));
+}
+
+}  // namespace
 
 static constexpr const char* SEARCH_PARAM_TEMPLATE_STR = R"(
 {{
@@ -43,8 +80,12 @@ IVFNearestPartition::IVFNearestPartition(BucketIdType bucket_count,
                                          const IndexCommonParam& common_param,
                                          IVFPartitionStrategyParametersPtr param)
     : IVFPartitionStrategy(common_param, bucket_count),
-      ivf_partition_strategy_param_(std::move(param)) {
-    this->factory_router_index(common_param);
+      ivf_partition_strategy_param_(std::move(param)),
+      use_route_graph_(this->ivf_partition_strategy_param_->use_route_graph),
+      common_param_(common_param) {
+    if (this->use_route_graph_) {
+        this->factory_router_index(common_param);
+    }
 }
 
 void
@@ -84,7 +125,32 @@ IVFNearestPartition::Train(const DatasetPtr dataset) {
         }
     }
 
-    auto build_result = this->route_index_ptr_->Build(centroids);
+    if (this->use_route_graph_) {
+        const auto failed_ids = this->route_index_ptr_->Build(centroids);
+        if (not failed_ids.empty()) {
+            throw VsagException(ErrorType::INTERNAL_ERROR,
+                                "failed to build all IVF routing centroids");
+        }
+    } else {
+        // Persist the trained centroids (normalized for cosine) so the scan routing path can
+        // assign buckets without creating a routing HGraph.
+        this->centroids_.resize(this->bucket_count_ * this->dim_);
+        memcpy(
+            this->centroids_.data(), data.data(), this->bucket_count_ * this->dim_ * sizeof(float));
+        this->norms_.resize(this->bucket_count_);
+        if (metric_type_ == MetricType::METRIC_TYPE_L2SQR) {
+            for (int i = 0; i < bucket_count_; ++i) {
+                this->norms_[i] = FP32ComputeIP(this->centroids_.data() + i * dim_,
+                                                this->centroids_.data() + i * dim_,
+                                                dim_) /
+                                  2.0F;
+            }
+        } else {
+            // IP/COSINE distances are defined as `1 - dot`; with normalized cosine centroids
+            // the constant term is 1.0 for every centroid.
+            std::fill(this->norms_.begin(), this->norms_.end(), 1.0F);
+        }
+    }
     this->is_trained_ = true;
 }
 
@@ -93,9 +159,12 @@ IVFNearestPartition::ClassifyDatas(const void* datas,
                                    int64_t count,
                                    BucketIdType buckets_per_data,
                                    QueryContext* ctx) const {
+    if (not this->use_route_graph_) {
+        return this->classify_datas_by_scan(datas, count, buckets_per_data, ctx);
+    }
     std::mutex dist_cmp_reduce_mutex;
     uint32_t dist_cmp = 0;
-    Vector<BucketIdType> result(buckets_per_data * count, -1, this->allocator_);
+    Vector<BucketIdType> result(buckets_per_data * count, INVALID_BUCKET_ID, this->allocator_);
     auto task = [&](int64_t i) {
         auto query = Dataset::Make();
         query->Dim(this->dim_)
@@ -158,12 +227,50 @@ IVFNearestPartition::ClassifyDatas(const void* datas,
 void
 IVFNearestPartition::Serialize(StreamWriter& writer) {
     IVFPartitionStrategy::Serialize(writer);
-    this->route_index_ptr_->Serialize(writer);
+    if (this->use_route_graph_) {
+        // Keep the route-graph layout byte-for-byte compatible with the legacy format.
+        this->route_index_ptr_->Serialize(writer);
+    } else {
+        StreamWriter::WriteObj(writer, K_CENTROID_SCAN_LAYOUT_MARKER);
+        StreamWriter::WriteVector(writer, this->centroids_);
+        StreamWriter::WriteVector(writer, this->norms_);
+    }
 }
 void
 IVFNearestPartition::Deserialize(LvalueOrRvalue<StreamReader> reader) {
     IVFPartitionStrategy::Deserialize(reader);
-    this->route_index_ptr_->Deserialize(reader);
+    // Peek the layout marker without consuming bytes so legacy (marker-less) route-graph
+    // indices remain loadable with any reader configuration.
+    const auto cursor = reader->GetCursor();
+    uint64_t layout = 0;
+    StreamReader::ReadObj(reader, layout);
+    reader->Seek(cursor);
+    if (layout == K_CENTROID_SCAN_LAYOUT_MARKER) {
+        StreamReader::ReadObj(reader, layout);
+        StreamReader::ReadVector(reader, this->centroids_);
+        StreamReader::ReadVector(reader, this->norms_);
+        const bool invalid_trained_layout =
+            this->is_trained_ and (this->centroids_.size() != this->bucket_count_ * this->dim_ or
+                                   this->norms_.size() != this->bucket_count_);
+        const bool invalid_untrained_layout =
+            not this->is_trained_ and (not this->centroids_.empty() or not this->norms_.empty());
+        if (invalid_trained_layout or invalid_untrained_layout) {
+            throw VsagException(ErrorType::INVALID_BINARY,
+                                "invalid IVF centroid scan routing layout");
+        }
+        this->route_index_ptr_.reset();
+        this->use_route_graph_ = false;
+    } else {
+        // Route-graph layout (the legacy marker-less format). Lazily create the routing HGraph
+        // when the reader was configured with use_route_graph=false.
+        if (this->route_index_ptr_ == nullptr) {
+            this->factory_router_index(this->common_param_);
+        }
+        this->route_index_ptr_->Deserialize(reader);
+        this->centroids_.clear();
+        this->norms_.clear();
+        this->use_route_graph_ = true;
+    }
 }
 void
 IVFNearestPartition::factory_router_index(const IndexCommonParam& common_param) {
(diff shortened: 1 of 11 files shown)


src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -22,15 +22,51 @@
...
+
+// Deterministic max-heap over (key, bucket_id): ties break by bucket_id so that equal routing
+// keys always produce the same assignment.
+using ScanRouteHeap = std::priority_queue<std::pair<float, InnerIdType>,
+                                          Vector<std::pair<float, InnerIdType>>,
+                                          std::less<std::pair<float, InnerIdType>>>;
+
+// C = A * B^T (same shape as GNOIMIPartition's batched centroid routing).
+void
+matmul(const float* A, const float* B, float* C, int64_t M, int64_t N, int64_t K) {
```
LHT129: [suggestion] The `matmul` helper name is too generic for a function that wraps a specific BLAS Sgemm call used for centroid routing. Consider renaming to something more descriptive like `compute_centroid_dots` or `batch_centroid_inner_product` to make its purpose clear at the call site.

Current:
```cpp
matmul(query_data, this->centroids_.data(), dots.data(), count, bucket_count_, dim_);
```

Suggested:
```cpp
compute_centroid_dots(query_data, this->centroids_.data(), dots.data(), count, bucket_count_, dim_);
```

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -175,12 +273,99 @@ IVFNearestPartition::GetCentroid(BucketIdType bucket_id, Vector<float>& centroid
...
+
+    // Batch dot products query x centroids^T via Sgemm (same shape as GNOIMIPartition).
+    Vector<float> dots(count * bucket_count_, allocator_);
+    matmul(query_data, this->centroids_.data(), dots.data(), count, bucket_count_, dim_);
+
+    auto task = [&](int64_t i) {
+        Vector<std::pair<float, InnerIdType>> heap_storage(this->allocator_);
+        ScanRouteHeap heap(std::less<std::pair<float, InnerIdType>>{}, std::move(heap_storage));
+        const auto* dots_i = dots.data() + i * bucket_count_;
+        for (BucketIdType b = 0; b < bucket_count_; ++b) {
```
LHT129, line 350: [note] The `ScanRouteHeap` comment says "Deterministic max-heap" but the actual purpose is to select the top-k smallest distances using a max-heap (largest element on top, popped when heap exceeds k). Consider clarifying the comment to explain the inversion: e.g. "Max-heap used to keep the k smallest distances; largest distance sits at top() and is popped when heap exceeds k."

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -22,15 +22,51 @@
...
 #include "utils/util_functions.h"
 #include "vsag_exception.h"
 
 namespace vsag {
 
+namespace {
+
+static constexpr uint64_t kCentroidScanLayoutMarker = 0x565341475343414EULL;  // "VSAGSCAN"
+
+static constexpr BucketIdType INVALID_BUCKET_ID = static_cast<BucketIdType>(-1);
```
LHT129: [note] `INVALID_BUCKET_ID` is defined here but the existing `ClassifyDatas` method (the route-graph path) still uses the literal `-1` for initialization at line 165. Consider using `INVALID_BUCKET_ID` consistently in both routing paths for clarity.

LHT129 (commented):
Thanks for this PR. The implementation is clean and well-structured. The existing inline comments (from self-review and Copilot) already cover the notable points:

1. **`matmul` naming** — consider a more descriptive name like `compute_centroid_dots`.
2. **`INVALID_BUCKET_ID` consistency** — the route-graph path still uses literal `-1`; consider using the named constant.
3. **`ScanRouteHeap` comment clarity** — the "max-heap for k-smallest" inversion could be explained more clearly.
4. **Untrained serialization validation** — the scan layout currently rejects empty centroids/norms for untrained indexes, making it non-roundtrippable before `Train()`.
5. **`dist_cmp` uint32_t truncation** — `distance_evaluations` is computed as uint64_t but truncated to uint32_t for `stats->dist_cmp`.
6. **Heap reservation** — the per-query heap could reserve `k` upfront to reduce allocations in the hot path.

The test coverage is comprehensive across all three metric types, untrained behavior, serialization round-trips, and cross-configuration loading. No blocking issues found.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: Switched both routing paths to INVALID_BUCKET_ID consistently. Fixed.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -149,12 +213,46 @@ IVFNearestPartition::ClassifyDatas(const void* datas,
...
+    reader->Seek(cursor);
+    if (layout == kCentroidScanLayoutMarker) {
+        StreamReader::ReadObj(reader, layout);
+        StreamReader::ReadVector(reader, this->centroids_);
+        StreamReader::ReadVector(reader, this->norms_);
+        if (this->centroids_.size() != this->bucket_count_ * this->dim_ ||
+            this->norms_.size() != this->bucket_count_) {
+            throw VsagException(ErrorType::INVALID_BINARY,
+                                "invalid IVF centroid scan routing layout");
+        }
```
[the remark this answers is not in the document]
LHT129: Untrained serialization is now round-trippable: empty centroids_/norms_ are accepted when is_trained_==false, with a dedicated test covering it. Fixed.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -84,7 +124,28 @@ IVFNearestPartition::Train(const DatasetPtr dataset) {
         }
     }
 
-    auto build_result = this->route_index_ptr_->Build(centroids);
+    if (this->use_route_graph_) {
+        auto build_result = this->route_index_ptr_->Build(centroids);
+    } else {
```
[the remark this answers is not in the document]
LHT129: Added Build() failure check: non-empty failed_ids now throws INTERNAL_ERROR before marking trained. Fixed.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -175,12 +273,99 @@ IVFNearestPartition::GetCentroid(BucketIdType bucket_id, Vector<float>& centroid
...
+    }
+
+    // Batch dot products query x centroids^T via Sgemm (same shape as GNOIMIPartition).
+    Vector<float> dots(count * bucket_count_, allocator_);
+    matmul(query_data, this->centroids_.data(), dots.data(), count, bucket_count_, dim_);
+
+    auto task = [&](int64_t i) {
+        Vector<std::pair<float, InnerIdType>> heap_storage(this->allocator_);
+        ScanRouteHeap heap(std::less<std::pair<float, InnerIdType>>{}, std::move(heap_storage));
+        const auto* dots_i = dots.data() + i * bucket_count_;
```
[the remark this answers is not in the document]
LHT129: Added heap_storage.reserve(k) before constructing ScanRouteHeap. Fixed.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: Renamed to compute_centroid_dots per suggestion. Fixed in latest commit.

src/algorithm/ivf/ivf_nearest_partition.cpp
LHT129: Expanded comment to clarify that the max-heap keeps the k smallest keys, with the largest retained key at top() and bucket_id as deterministic tie-break. Fixed.

src/algorithm/ivf/ivf_nearest_partition.cpp
```diff
@@ -175,12 +273,99 @@ IVFNearestPartition::GetCentroid(BucketIdType bucket_id, Vector<float>& centroid
...
+        for (auto& item : futures) {
+            item.get();
+        }
+    }
+
+    if (ctx != nullptr and ctx->stats != nullptr) {
+        const auto distance_evaluations =
+            static_cast<uint64_t>(count) * static_cast<uint64_t>(bucket_count_);
+        ctx->stats->dist_cmp.fetch_add(static_cast<uint32_t>(distance_evaluations),
+                                       std::memory_order_relaxed);
```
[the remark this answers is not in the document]
LHT129: Clamped dist_cmp increment to uint32_t::max() to avoid wrap-around. Fixed.

LHT129 (commented):
review

LHT129 (commented):
[suggestion] The `compute_centroid_dots` helper (src/algorithm/ivf/ivf_nearest_partition.cpp:54) uses generic parameter names `A`, `B`, `C`. Since this is a domain-specific file-local helper used in exactly one place, consider renaming the parameters to reflect their roles:

```cpp
void
compute_centroid_dots(const float* queries, const float* centroids, float* dots, int64_t num_queries, int64_t num_centroids, int64_t dim);
```

This would make the call site self-documenting. Current names `M`, `N`, `K` are conventional for BLAS but the function name already conveys the domain purpose.

LHT129 (commented):
**Review Summary**

This PR adds a `use_route_graph` IVF partition parameter that allows bypassing the routing HGraph construction and instead routes vectors via an exact batched centroid scan using Sgemm. The implementation is well-structured with proper serialization, backward compatibility, and comprehensive test coverage.

**Key observations from the latest commit (0d21677d):**

Most issues raised in earlier review rounds have been addressed:
- `matmul` renamed to `compute_centroid_dots` for clarity
- `ScanRouteHeap` comment expanded to explain the max-heap semantics
- `INVALID_BUCKET_ID` used consistently across both routing paths
- Untrained serialization round-trip now supported
- `dist_cmp` increment clamped to `uint32_t::max()` to prevent wrap-around
- `heap_storage.reserve(k)` added before constructing `ScanRouteHeap`
- `Build()` failure now throws `INTERNAL_ERROR` before marking trained

The self-describing serialization layout (centroid-scan marker vs. legacy route-graph) is well-designed and the cross-configuration loading test provides good coverage.

One remaining open item from earlier reviews: the full `count * bucket_count_` dot-product matrix allocation in `classify_datas_by_scan` could be memory-intensive for large datasets, though this is a design trade-off rather than a correctness issue.

LHT129 (commented):
Thanks for this PR. I have reviewed the full diff and the updated code after the previous round of review comments.

The implementation is solid:
- The `use_route_graph` parameter is correctly plumbed through external param mapping, serialization, and the partition strategy.
- The centroid-scan routing path correctly handles L2SQR, IP, and COSINE distance semantics.
- The serialization format is self-describing with the layout marker, and backward-compatible with legacy route-graph indices.
- Cross-configuration deserialization (scan→graph and graph→scan) is properly handled and tested.
- The `INVALID_BUCKET_ID` sentinel is used consistently across both routing paths.
- Statistics accumulation in the scan path correctly reports distance evaluations.

All previous review comments have been addressed in the latest commit. No blocking issues found.

LHT129 (commented):
I have reviewed the latest commit (5814474a) for this PR. The changes from the previous review rounds have all been properly addressed:

- `matmul` renamed to `compute_centroid_dots` for clarity
- `ScanRouteHeap` comment clarified to explain the max-heap inversion
- `INVALID_BUCKET_ID` used consistently in both routing paths
- `dist_cmp` overflow clamped to `uint32_t::max()`
- Untrained deserialization validation added for centroid-scan layout round-tripping

The implementation is clean and well-structured. The serialization format is self-describing via the layout marker, and cross-configuration loading (scan-layout index loaded by a route-graph reader, and vice versa) is handled correctly.

One existing open concern (raised by Copilot) about the transient `count * bucket_count_` dot-product matrix allocation in `classify_datas_by_scan` remains. For very large ingestion batches this could cause memory pressure, though it is inherent to the batched Sgemm approach and can be addressed in a follow-up if it becomes a practical issue.

No blocking issues found in this revision.

LHT129 (commented):
I have reviewed the latest commit (ed561f6) and confirmed that all issues raised in the previous review round have been addressed:

- `matmul` renamed to `compute_centroid_dots` for clarity
- `ScanRouteHeap` comment expanded to explain the max-heap inversion
- `INVALID_BUCKET_ID` now used consistently in both routing paths
- Untrained deserialization validation correctly handles `is_trained_` state
- `dist_cmp` overflow clamped to `uint32_t::max()`

The remaining open item from Copilot about large-batch memory allocation in `classify_datas_by_scan` (the full `count * bucket_count_` dot-product matrix) is a valid performance/memory consideration but is not a correctness issue. The current approach is consistent with how `GNOIMIPartition` handles batched centroid routing. Tiling can be considered as a follow-up optimization if profiling shows it is necessary for production workloads.

No new issues found in this revision. The code is well-structured, properly handles all three distance metrics (L2SQR, IP, COSINE), maintains backward compatibility with legacy serialized indices, and has thorough test coverage including cross-configuration deserialization.

wxyucs (approved):
lgtm

PR #2858, opened 2026-09-04
feat: add parallel PiPNN graph construction

### LINKED ISSUE #2847: [feat](hgraph): add PiPNN as a parallel graph construction mode
### Affected area

hgraph

### Motivation

HGraph currently supports NSW and ODescent graph construction. The existing configuration and
build dispatch are documented in `src/inner_string_params.h:145`,
`src/algorithm/hgraph/hgraph_build.cpp:157`, and
`docs/docs/en/src/indexes/hgraph.md:17`.

NSW relies on incremental insertion and graph search. ODescent provides batch construction, but
there is still room to reduce construction time on large datasets and highly parallel machines.
PiPNN is an offline graph construction algorithm designed for multicore CPUs. It uses overlapping
partitions, bulk distance computation within partitions, bounded HashPrune candidate reservoirs,
and final pruning to reduce the random-access search cost during construction.

A preliminary VSAG PoC on SIFT1M (one million 128-dimensional vectors, FP32/L2, Recall@10 at least
0.993) produced the following build times:

| Build threads | PiPNN | NSW | ODescent | vs. NSW | vs. ODescent |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 16 | 14.10 s | 70.49 s | 58.03 s | about 5.00x | about 4.12x |
| 32 | 8.52 s | 43.28 s | 43.24 s | about 5.08x | about 5.08x |
| 48 | 7.18 s | 36.26 s | 39.97 s | about 5.05x | about 5.57x |

These numbers are preliminary results from one machine and one 1M-point dataset. They do not
replace formal validation on larger datasets and additional configurations.

### Proposed solution

Add PiPNN as a third opt-in HGraph construction mode:

```json
{
  "index_param": {
    "graph": {
      "graph_type": "pipnn"
    }
  }
}
```

The core implementation should:

1. Recursively partition the dataset and use overlap or fanout to increase the probability that
   nearby points meet in at least one leaf partition.
2. Generate candidate edges with cache-friendly bulk distance computation inside leaf partitions.
3. Use bounded HashPrune candidate reservoirs to control construction memory.
4. Apply final neighbor pruning and write the HGraph bottom layer.
5. Reuse the existing HGraph route layers and search path.
6. Parallelize data preparation, recursive partitioning, leaf processing, candidate pruning, and
   route-layer construction while avoiding nested BLAS oversubscription.

The exact first-version compatibility scope is **TBD**. This includes supported data types,
distance metrics, quantizers, incremental build, filtering, reorder, cache-assisted build, and
serialization compatibility. This issue first establishes PiPNN as a third HGraph builder and
validates the core build-to-search path. The initial release boundary should then be decided from
implementation complexity, measured performance, and the existing HGraph compatibility model.

The current SIFT1M FP32/L2 experiment is only preliminary performance evidence. It does not mean
that the first version will support only FP32/L2 and does not define the final support boundary.
The default must remain `graph_type: "nsw"`.

Acceptance and performance validation should include:

- Tune PiPNN, NSW, and ODescent independently, then compare construction time only after they reach
  the same Recall@10 target.
- At that recall target, report query QPS and latency for every generated graph; construction speed
  alone is insufficient.
- Test 16, 32, 48, and a higher thread count appropriate for the machine. A single-thread run is
  not required for the initial evaluation.
- Add at least one materially larger, locally available real-world dataset in addition to SIFT1M.
- Report peak construction memory, CPU utilization, and parallel scaling efficiency, with
  particular attention to scaling beyond 32 threads.
- Fix and document BLAS thread settings, construction parameters, search parameters, datasets, and
  hardware information.
- Provide a reproducible VSAG benchmark configuration.

### Alternatives considered

1. Continue tuning NSW. It retains the widest HGraph compatibility but does not eliminate the
   search cost of incremental insertion.
2. Continue tuning ODescent. It already provides batch construction, but its construction time was
   higher than the preliminary PiPNN PoC at the same recall target.
3. Implement PiPNN as a separate index type. This would unnecessarily duplicate HGraph storage,
   routing, and search behavior.
4. Copy an external implementation. An independent implementation is preferred so it follows
   VSAG's allocator, concurrency, ownership, and licensing conventions.

additive only (no break)

### References

- Tobias Rubel, Richard Wen, Laxman Dhulipala, Lars Gottesbüren, Rajesh Jayaram, and Jakub Łącki,
  “PiPNN: Ultra-Scalable Graph-Based Nearest Neighbor Indexing,” arXiv:2602.21247, 2026:
  https://arxiv.org/abs/2602.21247
- PiPNN paper text: https://arxiv.org/html/2602.21247v3
- Paper license, CC BY 4.0: https://creativecommons.org/licenses/by/4.0/
- Existing HGraph graph-type constants: `src/inner_string_params.h:145`
- Existing HGraph build dispatch: `src/algorithm/hgraph/hgraph_build.cpp:157`
- Existing HGraph construction documentation: `docs/docs/en/src/indexes/hgraph.md:17`
- Existing `build_thread_count` documentation: `docs/docs/en/src/indexes/hgraph.md:80`

### Additional context

The feature is intended as an independent implementation based on the cited paper. It should not
copy paper figures, substantial paper text, or third-party source code whose license has not been
reviewed. The paper page identifies the paper as CC BY 4.0, and this issue provides its authors,
title, and source link. Any future reuse of external code requires a separate code-license review;
potential patent considerations should also be checked before a production release.

Performance reported by the paper is research motivation, not a VSAG performance claim. VSAG
acceptance should be based exclusively on reproducible measurements at matched search quality.

## Related issues

_None found._

---

jac0626:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2847

## What Changed

- Add an independent, parallel graph builder based on [PiPNN: Ultra-Scalable Graph-Based Nearest Neighbor Indexing](https://arxiv.org/abs/2602.21247), with deterministic overlapping partitioning, bounded candidate reservoirs, and robust pruning.
- Add `graph_type: "pipnn"` to HGraph. PiPNN builds the bottom graph and reuses the existing route-layer, storage, query, filter, reorder, incremental `Add`, removal, and serialization paths.
- Add PiPNN to Pyramid root construction. For `root_graph_type: "multi_layer"`, PiPNN batch-builds both the root bottom graph and every sparse root routing layer; descendant path-node graphs continue to use ODescent.
- Re-run the applicable existing NSW batch-Build matrix with PiPNN instead of relying on feature-specific smoke tests. The matrix covers HGraph and Pyramid search, filters, duplicate handling, attributes, extra info, raw vectors, quantization/reorder, single- and multi-layer routing, multi-hierarchy isolation, removal, Tune, serialization/streaming, clone/export, allocator, IO, memory reporting, timeout, threshold, reverse edges, and concurrent operations.
- Fix compatibility gaps exposed by that matrix: validate extra-info atomically, preserve Pyramid duplicate-label/input-row/path mappings, expand duplicate representatives from finite entry points in both searchers, accept the existing `build_thread_count=0` serial behavior, and keep non-finite-distance vertices connected as worst-ranked PiPNN candidates.

## Test Evidence

- [x] `make fmt`
- [x] Production files checked with `clang-tidy-15`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
OPENBLAS_NUM_THREADS=1 ./build/tests/functests '[hgraph][pipnn]' --reporter compact --durations yes
All tests passed (3,034,607 assertions in 46 test cases)

OPENBLAS_NUM_THREADS=1 ./build/tests/functests '[pyramid][pipnn]' --reporter compact --durations yes
All tests passed (484,783 assertions in 48 test cases)

OPENBLAS_NUM_THREADS=1 ./build/tests/unittests '[pipnn],[BasicSearcher],[ParallelSearcher]' --reporter compact
All tests passed (1,663 assertions in 24 test cases)

OPENBLAS_NUM_THREADS=1 build/tests/unittests -d yes "" --allow-running-no-tests ""
All tests passed (85,355,878 assertions in 858 test cases)

OPENBLAS_NUM_THREADS=1 build/tools/eval/eval_performance /tmp/pipnn_case.yaml
SIFT1M 48/96-thread NSW, ODescent, and PiPNN graphs were query-validated at
Recall@10 >= 0.993. tools/eval/pipnn_parallel.yaml contains the reproducible matrix.
```

The complete compatibility matrix intentionally excludes Add-only/Train-only tests, ODescent-specific behavior, cache-assisted build, deduplicated vector storage, and sparse vectors because they do not exercise PiPNN batch construction or are explicitly unsupported PiPNN inputs. Dedicated query-validated matrices cover L2, inner product, and cosine across HGraph storage/quantization combinations and Pyramid single-/multi-layer roots.

## Compatibility Impact

- API/ABI compatibility: Additive configuration value only. The default remains `graph_type: "nsw"`; no public C++ API or serialized bytes are added.
- Behavior changes: `pipnn` is opt-in. Initial PiPNN construction requires dense FP32 input and supports the existing L2, inner-product, and cosine metrics. HGraph and Pyramid continue to support the existing persistent quantizers and reorder paths, including RaBitQ + SQ8 reorder. Cache-assisted build and deduplicated vector storage are rejected for PiPNN.

- Performance impact: Improved initial construction at matched search quality. On the local 48-core/96-thread Intel Xeon Platinum 8163, SIFT1M FP32 L2, `M=32`, and `OPENBLAS_NUM_THREADS=1`:

| Build threads | Builder | Build (s) | Recall@10 | QPS | Avg latency (ms) |
| ---: | --- | ---: | ---: | ---: | ---: |
| 48 | NSW | 38.02 | 0.99367 | 1805 | 0.551 |
| 48 | ODescent | 41.01 | 0.99340 | 1259 | 0.791 |
| 48 | PiPNN | 6.43 | 0.99358 | 1779 | 0.559 |
| 96 | NSW | 33.98 | 0.99340 | 1817 | 0.547 |
| 96 | ODescent | 37.27 | 0.99329 | 1241 | 0.802 |
| 96 | PiPNN | 5.14 | 0.99358 | 1795 | 0.554 |

  At 48 threads, PiPNN is 5.9x faster than NSW and 6.4x faster than ODescent; at 96 threads, it is 6.6x and 7.3x faster respectively. Only SIFT1M was locally available, so larger-scale validation is intentionally left for follow-up infrastructure runs.
- Concurrency/thread-safety impact: PiPNN parallelizes partition assignment, leaf candidate generation, and final pruning through the existing `SafeThreadPool`. Graph writes remain serialized. The implementation is deterministic across build thread counts.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/indexes/{hgraph,pyramid}.md`

## Risk and Rollback

- Risk level: medium
- Rollback plan: Users can select the unchanged `nsw` or `odescent` builders; the implementation can be reverted without a serialized-format migration.

## Checklist


src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1053 @@
...
+    return hash;
+}
+
+void
+PiPNNPipeline::insert_candidate(uint32_t source, uint32_t target, float distance) {
+    if (source == target or not std::isfinite(distance)) {
+        return;
+    }
+
+    auto& state = reservoir_states_[source];
```
LHT129: [suggestion] `reservoir_states_[source]` is read on line 787 before the `PointLockGuard` is acquired on line 792. In the parallel build path (overlapping leaves from `fanout > 1`), another thread may concurrently modify `state.size` or `state.farthest`, making the initial read of `state` a TOCTOU race.

The practical impact is low since `state.size` only grows and a stale `farthest` at worst causes an extra comparison, but for strict correctness the state reference should be obtained after acquiring the lock.

Suggested fix: move `auto& state = ...` and `auto* row = ...` to after the `PointLockGuard` declaration.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1053 @@
...
+                const uint64_t begin = next.fetch_add(block_size, std::memory_order_relaxed);
+                if (begin >= total) {
+                    return;
+                }
+                task(begin, std::min(begin + block_size, total));
+            }
+        }));
+    }
+
+    std::exception_ptr first_exception;
```
LHT129, line 1050: [note] When multiple workers throw exceptions in `parallel_for`, only the first one is captured and rethrown (lines 979-991). The remaining exceptions are silently discarded, which could mask secondary failures during debugging.

Consider collecting all exceptions (e.g., into a vector) and reporting them together, or at least logging subsequent exceptions before discarding them.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1053 @@
...
+
+        for (uint64_t point = 0; point < item.points.size(); ++point) {
+            for (uint64_t selected = 0; selected < fanout; ++selected) {
+                clusters[assignments[point * fanout + selected]].emplace_back(item.points[point]);
+            }
+        }
+        return clusters;
+    }
+
+    for (uint64_t begin = 0; begin < item.points.size(); begin += PARTITION_STRIPE_SIZE) {
```
LHT129, line 632: [suggestion] In the sequential `assign_to_leaders` path (lines 575-633), `point_values`, `dots`, and `candidates` vectors are allocated inside the stripe loop and reallocated on every iteration. Since the stripe size is fixed (`PARTITION_STRIPE_SIZE` = 256 except for the last stripe), these could be allocated once before the loop and reused via `.clear()` to reduce allocator pressure.

The parallel path (lines 495-572) avoids this because each `parallel_for` worker gets its own stack-local vectors.

LHT129 (commented):
## Code Review Summary

This is a substantial and well-structured PR adding parallel PiPNN graph construction to VSAG. The implementation is clean, well-tested, and follows the project conventions. Below are the findings from the review.

### Issues found

1. **[suggestion] TOCTOU in `insert_candidate`** (`src/impl/pipnn/pipnn_graph_builder.cpp:787`): `reservoir_states_[source]` is read before the `PointLockGuard` is acquired on line 792. In the parallel build path (overlapping leaves from `fanout > 1`), another thread may concurrently modify `state.size` or `state.farthest`. The practical impact is low, but for strict correctness the state reference should be obtained after acquiring the lock.

2. **[note] Exception handling in `parallel_for`** (`src/impl/pipnn/pipnn_graph_builder.cpp:979-991`): When multiple workers throw exceptions, only the first one is rethrown; subsequent exceptions are silently discarded. Consider logging them before discarding.

3. **[suggestion] Repeated allocations in sequential `assign_to_leaders`** (`src/impl/pipnn/pipnn_graph_builder.cpp:575-633`): `point_values`, `dots`, and `candidates` are reallocated on every stripe iteration. Since the stripe size is fixed, these could be allocated once before the loop and reused.

4. **[note] Unused variable in ODescent path** (`src/algorithm/hgraph/hgraph_build.cpp:210`): `attr_sets` is fetched but only consumed in the `use_pipnn` branch. In the ODescent path it is unused, which may trigger a compiler warning.

5. **[note] Extra-info / attr_sets only populated for PiPNN** (`src/algorithm/hgraph/hgraph_build.cpp:281-288`): The `extra_infos_` and `attr_filter_index_` insertions are guarded by `use_pipnn`, meaning the ODescent batch-build path will skip populating them. (Already noted in a prior review comment.)

### Positive observations

- Comprehensive test coverage: unit tests for the builder, HGraph integration tests (build, search, RaBitQ+SQ8 reorder, serialization, extra info, duplicate labels, parameter validation), and Pyramid integration.
- Good input validation throughout (`require_argument`, `Validate`, constructor checks).
- Clean separation of concerns: `PiPNNGraphBuilder` owns only graph construction; labels, storage, route layers remain in the index layer.
- Deterministic output verified across thread counts.
- Well-documented with English and Chinese docs, plus a reproducible eval workload.

LHT129 (commented):
## Code Review Summary

This PR adds a parallel PiPNN graph construction feature (2656 additions across 26 files). Overall this is a well-structured and substantial contribution. Below are the review findings organized by severity.

---

### [critical] Reservoir size validation gap in `PiPNNGraphBuilderParameter::Validate()`

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `Validate()` method

The validation checks `std::max(reservoir_size, max_degree) <= std::numeric_limits<uint16_t>::max()` which ensures the reservoir fits in uint16_t. However, `prepare()` computes `reservoir_size_ = std::max(parameter_.reservoir_size, graph->MaximumDegree())` — this means if `graph->MaximumDegree()` exceeds the validated `max_degree` parameter (e.g., if the graph was pre-configured differently), the actual `reservoir_size_` could be larger than validated. The `Validate()` is called with `graph->MaximumDegree()` from `Build()`, so this is actually consistent. But the `prepare()` method independently re-computes `reservoir_size_` using `graph_->MaximumDegree()` which could differ from the validated value if the graph state changes between `Build()` and `prepare()`. Consider using the already-validated `parameter_.reservoir_size` consistently or validating against the actual graph max_degree at prepare time.

### [critical] Potential integer overflow in `reservoir_count` computation

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `prepare()` method

```cpp
const uint64_t reservoir_count = checked_product(ids_.size(), reservoir_size_, "reservoir");
require_argument(reservoir_count <= reservoirs_.max_size(), ...);
```

`reservoirs_.max_size()` returns the allocator's theoretical max, which could be very large. The subsequent `reservoirs_.resize(reservoir_count)` could still fail with `std::bad_alloc` if the allocator cannot fulfill the request. While `checked_product` catches multiplication overflow, there's no practical upper bound check for the reservoir memory. For 1B points with reservoir_size=64, this would request 512 GB just for reservoirs. Consider adding a practical upper bound or documenting the expected scale limits.

### [suggestion] `parallel_for` in `partition()` may deadlock with nested parallelism

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `partition()` method

The `partition()` method calls `parallel_for(work.size(), 1, ...)` which dispatches work items to the thread pool. Inside each work item, `split_work_item()` may call `assign_to_leaders()` with `allow_inner_parallelism=true`, which in turn calls `parallel_for()` again — potentially exhausting the thread pool if all threads are already busy with outer `parallel_for` tasks. The code mitigates this by checking `all_items_are_small` before enabling outer parallelism, and `assign_to_leaders` has its own `MIN_PARALLEL_PARTITION_POINTS` threshold. However, the interaction between the two levels of parallelism is subtle and could lead to thread starvation. Consider documenting the nesting strategy or using a dedicated inner thread pool.

### [suggestion] `relative_hash` coincident flag uses 16-bit comparison that may miss near-coincident vectors

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `relative_hash()` method

```cpp
coincident = coincident and target_value == source_value;
```

Floating-point equality comparison for hash plane projections. Due to floating-point rounding, vectors that are numerically identical may produce slightly different dot products with the random hyperplanes, causing the coincident flag to not be set. This means two identical vectors could hash to different buckets. The `group_duplicate_rows()` function already handles exact duplicates before the pipeline runs, so this only affects near-duplicates. Consider using an epsilon-based comparison or documenting this as intentional.

### [suggestion] `robust_prune` recomputes pairwise distances instead of reusing reservoir distances

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `robust_prune()` method

```cpp
const float distance = pair_distance(source, candidate);
```

The reservoir already stores BF16-quantized distances for each candidate. `robust_prune` recomputes full float32 distances via `pair_distance()` which is an O(dim) operation per candidate. For high-dimensional vectors this could be expensive. The reservoir distances are BF16 and may have precision loss, but they could be used for the initial sort, with full-precision recomputation only for the alpha-pruning step. This is a performance consideration, not a correctness issue.

### [suggestion] `prepare_sketches` uses fixed seed 42 for random hyperplanes

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `prepare_sketches()` method

```cpp
std::mt19937_64 random(42);
```

The fixed seed ensures deterministic builds across runs, which is good. However, it means all PiPNN indexes built with this code will use identical hash planes. If an adversary knows this seed, they could construct adversarial vectors that cause worst-case hash collisions. Consider making the seed configurable or deriving it from a runtime entropy source while still logging it for reproducibility.

### [note] `PiPNNPipeline` stores `rows_` as a const reference — lifetime is safe but fragile

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`, `PiPNNPipeline` class

```cpp
const Vector<const float*>& rows_;
```

The pipeline is stack-allocated inside `PiPNNGraphBuilder::Build()`, and `Build()` is called synchronously. The `rows` vector in all callers (`build_by_batch_graph`, pyramid `Build`) is alive for the duration of the `Build()` call. So the reference is valid. However, this pattern is fragile — a future refactor that makes `Build()` asynchronous would create a use-after-free. Consider adding a comment documenting the lifetime requirement.

### [note] Duplicate `require_argument` checks between `PiPNNGraphBuilder::Build()` and `PiPNNPipeline::prepare()`

**File:** `src/impl/pipnn/pipnn_graph_builder.cpp`

The `Build()` method validates `ids_sequence.size() == rows.size()` and `ids_sequence.size() <= std::numeric_limits<uint32_t>::max()` before constructing the pipeline, and `prepare()` repeats the same checks. This is defensive but redundant for the non-duplicate path. For the duplicate path, `Build()` does additional validation (unique IDs, graph capacity) that `prepare()` also does. Consider consolidating to avoid the duplication.

### [note] `hgraph_fast_build.cpp` skips PiPNN without explanation

**File:** `src/algorithm/hgraph/hgraph_fast_build.cpp`

```cpp
if (graph_type_ == GRAPH_TYPE_VALUE_PIPNN) {
    return std::nullopt;
}
```

The fast/optimized build path is skipped for PiPNN. This is correct since PiPNN has its own parallel build strategy, but a brief comment explaining why would help future readers understand this is intentional, not an oversight.

### [note] Test coverage is comprehensive

**Files:** `*_test.cpp`, `tests/test_hgraph.cpp`, `tests/test_pyramid.cpp`

The test suite covers: adjacency invariants, deterministic builds across thread counts, single-point edge cases, parallel execution, duplicate vector reachability, HGraph integration (build, search, serialize, add, remove, attribute filtering, extra info, RaBitQ+SQ8), Pyramid integration (single/multi-layer root, duplicate path semantics, path prefix/descendant/sibling), and parameter validation. This is thorough.

---

**Overall assessment:** The implementation is solid, well-tested, and properly integrated into both HGraph and Pyramid indexes. The main concerns are around the nested parallelism strategy and the reservoir memory bounds. The code follows VSAG coding standards and the architecture is cleanly separated.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    initial.resize(ids_.size());
+    for (uint64_t i = 0; i < ids_.size(); ++i) {
+        initial[i] = static_cast<uint32_t>(i);
+    }
+
+    Leaves finished(allocator_);
+    if (initial.size() <= parameter_.max_leaf_size) {
+        finished.emplace_back(std::move(initial));
+        return finished;
+    }
```
LHT129, line 408: [suggestion] The nested parallel_for calls in partition() could exhaust the thread pool. The guard all_items_are_small mitigates this but the interaction is subtle. Consider adding a comment explaining the nesting strategy.

src/algorithm/hgraph/hgraph_fast_build.cpp
```diff
@@ -108,6 +108,10 @@ HGraphBuildTaskGuard::~HGraphBuildTaskGuard() {
```
LHT129, line 108: [note] PiPNN is excluded from the optimized build path. A brief comment explaining why (PiPNN has its own parallel build strategy) would help future readers.

LHT129 (commented):
Inline review comments

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    float distance = 0.0F;
+    for (uint64_t dim = 0; dim < dimensions_; ++dim) {
+        const float delta = left[dim] - right[dim];
+        distance += delta * delta;
+    }
+    return normalized_distance(distance);
+}
+
+Vector<InnerIdType>
+PiPNNPipeline::robust_prune(uint32_t source, const ReservoirEntry* row, uint16_t size) const {
```
LHT129, line 934: [suggestion] robust_prune recomputes full float32 pairwise distances via pair_distance() which is O(dim) per candidate. The reservoir already stores BF16 distances that could be used for the initial sort, with full-precision recomputation only for the alpha-pruning step.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    for (uint64_t local_id = 0; local_id < ids_.size(); ++local_id) {
+        const auto id = ids_[local_id];
+        require_argument(id < graph_capacity, "PiPNN input ID is outside the graph capacity");
+        require_argument(seen_ids.emplace(id).second, "PiPNN input IDs must be unique");
+        require_argument(rows_[local_id] != nullptr, "PiPNN input rows must not be null");
+    }
+
+    reservoir_size_ =
+        std::max(parameter_.reservoir_size, static_cast<uint64_t>(graph_->MaximumDegree()));
+    const uint64_t reservoir_count = checked_product(ids_.size(), reservoir_size_, "reservoir");
```
LHT129, line 336: [suggestion] prepare_sketches uses fixed seed 42 for random hyperplanes. This ensures deterministic builds but means all PiPNN indexes use identical hash planes. Consider making the seed configurable or deriving it from a runtime entropy source.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1121 @@
...
+          allocator_(allocator),
+          graph_(std::move(graph)),
+          thread_pool_(thread_pool),
+          thread_count_(thread_count),
+          dimensions_(dimensions),
+          rows_(rows),
+          ids_(allocator),
+          norms_(allocator),
+          sketches_(allocator),
+          reservoirs_(allocator),
```
LHT129, line 208: [note] PiPNNPipeline stores rows_ as a const reference. The lifetime is safe since Build() is synchronous, but this pattern is fragile. Consider adding a comment documenting the lifetime requirement.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: Fixed in 4f99c8a4: the sequential partition path now allocates point_values, dots, and candidates once and resizes or clears them for each stripe.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: Documented in 010a7b78. Build is synchronous, and the internal pipeline borrows the caller-owned rows vector only for the pipeline lifetime.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -241,6 +278,18 @@ HGraph::build_by_odescent(const DatasetPtr& data) {
         if (temporary_sq8_build_data != nullptr) {
             temporary_sq8_build_data->InsertVector(get_data(data, i), inner_id);
         }
+        if (use_pipnn and this->extra_infos_ != nullptr) {
+            const auto* extra_info =
+                extra_infos == nullptr ? nullptr : extra_infos + i * extra_info_size_;
+            this->extra_infos_->InsertExtraInfo(extra_info, inner_id);
+        }
```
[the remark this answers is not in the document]
jac0626: Thanks. I checked the pre-PR implementation on upstream/main: build_by_odescent already omitted extra-info population, so this PR does not regress that path. The use_pipnn guard limits the newly added metadata writes to the new PiPNN path. Changing the existing ODescent contract is unrelated scope, so I am leaving it unchanged here.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: Fixed in 4f99c8a4: the reservoir state and row bindings now appear after PointLockGuard, making the protected access boundary explicit.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: The fixed seed is intentional: it makes graph construction reproducible across thread counts and supports the determinism contract test. Making it configurable would add a new public construction parameter without a demonstrated quality or performance need, so it is outside this first integration.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: The BF16 reservoir distance is an admission key, while robust pruning uses the full-precision source distance both for ordering and for the alpha comparison. Sorting by BF16 would change tie/order behavior, and recomputing full precision during pruning would still perform essentially the same O(dim) work unless selection reaches max_degree early. Without same-quality benchmark evidence, I am keeping the quality-preserving full-precision path.

src/algorithm/hgraph/hgraph_fast_build.cpp
jac0626: Documented in 010a7b78. PiPNN owns a one-shot parallel graph build and therefore uses the normal flatten build path instead of the optimized NSW/ODescent session.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: Documented in 010a7b78. The partitioner uses only one thread-pool level at a time: outer work-item fan-out passes allow_inner_parallelism=false, while the sequential outer path permits stripe-level fan-out for a large item.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: The implementation intentionally drains every future and retains the first exception so that no worker can outlive the captured stack state. The public build operation can surface only one exception; aggregating or logging later worker exceptions would not change correctness and could emit duplicate diagnostics, so I am keeping the drain-then-rethrow behavior.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -241,6 +278,18 @@ HGraph::build_by_odescent(const DatasetPtr& data) {
...
+        if (use_pipnn and attr_sets != nullptr and this->use_attribute_filter_) {
+            this->attr_filter_index_->Insert(attr_sets[i], inner_id);
+        }
```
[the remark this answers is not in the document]
jac0626: Thanks. I verified that upstream/main already skips attribute-index population in the ODescent batch path; this PR did not introduce that behavior. The new use_pipnn branch populates attributes for PiPNN without changing the existing ODescent contract, whose expansion should be handled separately.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -254,27 +303,62 @@ HGraph::build_by_odescent(const DatasetPtr& data) {
...
             odescent_param_, build_data, allocator_, this->thread_pool_.get());
         odescent_builder.Build();
         odescent_builder.SaveGraph(bottom_graph_);
     }
+
+    auto route_odescent_param = this->odescent_param_;
+    if (route_odescent_param == nullptr) {
+        route_odescent_param = std::make_shared<ODescentParameter>();
+    }
     for (auto& route_graph_id : route_graph_ids) {
```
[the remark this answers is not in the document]
jac0626: Fixed in 010a7b78. Each PiPNN route graph now receives a copy of the configured ODescent parameters before route-specific max_degree and block_size tuning. A regression test verifies that the configured shared parameters remain unchanged. The existing ODescent path is intentionally preserved.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1123 @@
...
+    }
+    if (coincident) {
+        hash |= COINCIDENT_HASH_FLAG;
+    }
+    return hash;
+}
+
+void
+PiPNNPipeline::insert_candidate(uint32_t source, uint32_t target, float distance) {
+    if (source == target or not std::isfinite(distance)) {
```
LHT129: [suggestion] In `insert_candidate`, when an incoming candidate has a different hash from an existing entry, the code replaces the *first* entry with a differing hash rather than the *worst* one among all entries with differing hashes. This means a relatively good entry could be evicted while a worse entry with yet another hash remains in the reservoir.

Consider scanning all entries to find the worst one among those whose hash differs from the incoming candidate, then replacing that worst entry. The linear scan cost is bounded by `reservoir_size` (default 64).

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -30,6 +30,7 @@
 #include "impl/heap/standard_heap.h"
```
LHT129, line 30: [suggestion] The `build_by_batch_graph` function in `hgraph_build.cpp` mixes ODescent and PiPNN logic through a boolean `use_pipnn` parameter. The function is now ~130 lines with interleaved conditional branches for two distinct algorithms. Consider extracting the PiPNN path into a dedicated `build_by_pipnn` method to keep each algorithm self-contained and avoid accidental coupling between the two code paths.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1125 @@
...
+    for (uint64_t local_id = 0; local_id < ids.size(); ++local_id) {
+        const auto key = std::string_view(reinterpret_cast<const char*>(rows[local_id]),
+                                          static_cast<std::size_t>(row_bytes));
+        const auto [it, inserted] = representatives.emplace(key, ids[local_id]);
+        if (inserted) {
+            representative_ids.emplace_back(ids[local_id]);
+            representative_rows.emplace_back(rows[local_id]);
+        } else {
+            duplicates.emplace_back(it.value(), ids[local_id]);
+        }
```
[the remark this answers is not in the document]
LHT129: tsl::robin_map iterators DO have a .value() method (defined in robin_hash.h line 249-253, returning value_type& which is std::pair<Key, T>&). The code at line 111 (it.value()) compiles and works correctly — it is equivalent to it->second but more explicit. The existing review comment claiming this will not compile is incorrect.

LHT129 (commented):
Reviewed the PiPNN parallel graph construction feature (PR #2858, 2698 additions across 26 files). The existing 25 review comments are thorough and cover all substantive issues I identified — including extra_info/attribute filter handling, odescent_param mutation (fixed), TOCTOU race in reservoir access (fixed), sequential partition allocation (fixed), nested parallelism, BF16 distance precision, fixed random seed, exception handling, and the mixed build_by_batch_graph logic.

One correction: the comment at src/impl/pipnn/pipnn_graph_builder.cpp:112 claiming that tsl::robin_map iterators lack .value() is incorrect — tsl::robin_map iterators DO have .value() (defined in robin_hash.h, returning value_type&). The code at line 111 compiles and works correctly.

No other new substantive issues found. The implementation is well-structured with appropriate validation, thread safety, and test coverage.

src/algorithm/hgraph/hgraph_build.cpp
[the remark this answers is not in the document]
jac0626: Fixed the PiPNN crash in 466cbc09. build_by_pipnn now validates that the extra-info payload is present and has the configured row size before any shared batch-build mutation; the insertion path therefore only dereferences a validated pointer. Unit regressions cover missing and mismatched payloads, and the unchanged stock extra-info helper also passes in the full PiPNN matrix. The ODescent omission predates this PR and remains outside this change, as noted in the earlier thread.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: I rechecked the scan: entries with a different hash take the continue branch and are not replaced. Replacement inside the loop happens only for an existing entry with the same non-coincident hash. If no such entry exists, the candidate is appended while capacity remains, or compared with and replaces the global farthest entry when full. Therefore the reported first-different-hash eviction does not occur, so no code change is needed.

src/impl/pipnn/pipnn_graph_builder.cpp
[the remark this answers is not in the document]
jac0626: Confirmed locally: this UnorderedMap is backed by tsl::robin_map, whose iterator exposes value(), and the code compiles under the project toolchain. The PiPNN unit matrix and both functional matrices pass, so no source change is required.

src/algorithm/hgraph/hgraph_build.cpp
jac0626: I kept the shared batch orchestration intentionally. ID allocation, duplicate-label filtering, persistent-code insertion, metadata insertion, route-node assignment, and deferred-code finalization are common stateful steps; only bottom-graph construction and PiPNN-specific validation differ, and build_by_pipnn remains the dedicated wrapper for those boundaries. Splitting the whole method would duplicate these invariants and make ODescent/PiPNN drift more likely. The expanded matrix now runs all applicable stock batch-Build contracts against PiPNN, so I am keeping the single-source orchestration.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1126 @@
...
+            if (incoming_key < std::make_tuple(entry.distance, ids_[entry.neighbor], entry.hash)) {
+                const bool was_farthest = index == state.farthest;
+                entry = ReservoirEntry{target, hash, distance_key};
+                if (was_farthest) {
+                    update_farthest(source);
+                }
+            }
+            return;
+        }
+        if (entry.hash != hash or (hash & COINCIDENT_HASH_FLAG) != 0) {
```
LHT129, line 867: [suggestion] When the incoming candidate has a coincident hash (`hash & COINCIDENT_HASH_FLAG != 0`), the loop at line 847 skips all existing entries (both those with different hashes via `entry.hash != hash` and those with the same coincident hash via the `COINCIDENT_HASH_FLAG` check). This means coincident-hash candidates can only enter the reservoir via the farthest-replacement path (lines 860-868), never by replacing an existing coincident-hash entry at the same position. If this is intentional (e.g., to preserve diversity by not letting coincident points crowd out others), a brief comment explaining the rationale would help future readers.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -254,27 +307,65 @@ HGraph::build_by_odescent(const DatasetPtr& data) {
...
+                for (auto& id : route_graph_id) {
+                    id = bottom_graph_->GetGroupId(id);
+                }
+                std::sort(route_graph_id.begin(), route_graph_id.end());
+                route_graph_id.erase(std::unique(route_graph_id.begin(), route_graph_id.end()),
+                                     route_graph_id.end());
+            }
+        }
+    } else {
         odescent_param_->max_degree = bottom_graph_->MaximumDegree();
```
LHT129, line 336: [note] At line 335, `odescent_param_->max_degree` is mutated in-place for the ODescent bottom-graph path. The route-graph path was already fixed (lines 347-349) to use a per-route copy for PiPNN, but the ODescent path still mutates the shared parameter. This is pre-existing behavior and the ODescent path does not re-enter `build_by_batch_graph`, so it is functionally safe. However, the asymmetry between the bottom-graph and route-graph mutation patterns may surprise future readers. A brief comment noting that the bottom-graph mutation is intentional (single-use, not re-entered) would help.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -241,6 +283,17 @@ HGraph::build_by_odescent(const DatasetPtr& data) {
         if (temporary_sq8_build_data != nullptr) {
             temporary_sq8_build_data->InsertVector(get_data(data, i), inner_id);
         }
+        if (use_pipnn and this->extra_infos_ != nullptr) {
+            const auto* extra_info = extra_infos + i * extra_info_size_;
```
LHT129, line 287: [note] `extra_info_size_` is used at line 287 without the `this->` prefix, while lines 196 and 253 in the same function use `this->extra_info_size_`. The bare name resolves correctly since it is a member variable, but the inconsistency may confuse readers about whether a local shadow is in play. Consider using `this->` for consistency with the surrounding code.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -702,7 +702,7 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
     if (check_func(ep) and is_result_distance_eligible<mode>(dist, inner_search_param)) {
         top_candidates->Push(dist, ep);
     }
-    if (not std::isfinite(dist) and inner_search_param.consider_duplicate and
+    if (std::isfinite(dist) and inner_search_param.consider_duplicate and
```
LHT129, line 705: [note] The `not std::isfinite(dist)` → `std::isfinite(dist)` change at the entry-point duplicate expansion guard flips the condition from "expand duplicates when distance is non-finite" to "expand duplicates when distance is finite". This is the correct semantic for PiPNN (and any builder): duplicate IDs should be pushed into the result set only when the entry point has a real, finite distance. The original `not std::isfinite` guard would have expanded duplicates for NaN/inf distances, which would pollute the candidate set with entries at meaningless distances. This also affects the existing NSW/ODescent path — the fix is correct for all builders, but reviewers should verify that no existing test relies on the old (buggy) behavior of expanding duplicates for non-finite distances.

src/impl/searcher/parallel_searcher.cpp
```diff
@@ -189,7 +189,7 @@ ParallelSearcher::search_impl(const GraphInterfacePtr& graph,
     if (check_func(ep) and is_result_distance_eligible<mode>(dist, inner_search_param)) {
         top_candidates->Push(dist, ep);
     }
-    if (not std::isfinite(dist) and inner_search_param.consider_duplicate and
+    if (std::isfinite(dist) and inner_search_param.consider_duplicate and
```
LHT129, line 192: [note] Same `not std::isfinite(dist)` → `std::isfinite(dist)` fix as in basic_searcher.cpp. The parallel searcher has a slightly different code path — it does not gate on `not use_custom_distance` here (line 193), unlike the basic searcher (line 706). This is consistent with the parallel searcher's existing behavior where custom-distance duplicate scoring is handled separately via the `score_duplicates` lambda, but worth noting the asymmetry for future maintainers.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -579,26 +580,55 @@ Pyramid::run_parallel_insertions(
...
-    int64_t data_num = base->GetNumElements();
+Pyramid::build_by_batch_graph(const DatasetPtr& base) {
+    const int64_t input_count = base->GetNumElements();
     const auto* data_vectors = base->GetFloat32Vectors();
     const auto* data_ids = base->GetIds();
     const auto* source_ids = base->GetSourceID();
+    Vector<int64_t> input_indices(allocator_);
+    input_indices.reserve(static_cast<uint64_t>(input_count));
+    std::vector<int64_t> failed_ids;
+    if (graph_type_ == GRAPH_TYPE_VALUE_PIPNN) {
```
LHT129, line 591: [suggestion] In `build_by_batch_graph`, the `input_indices` vector filters out duplicate labels when `graph_type_ == GRAPH_TYPE_VALUE_PIPNN`, but this deduplication happens only for PiPNN. The ODescent path (the `else` branch on line 602) does not deduplicate labels — it passes all input indices through. This means `build_by_batch_graph` now has different duplicate-label semantics depending on `graph_type_`. The original `build_by_odescent` had no deduplication, so the PiPNN path is adding new behavior. Consider whether the ODescent path should also deduplicate for consistency, or add a comment explaining why PiPNN needs it but ODescent doesn't.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -607,14 +637,95 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
...
+                                            allocator_,
+                                            this->thread_pool_.get(),
+                                            this->build_thread_count_);
+            auto local_odescent_param = std::make_shared<ODescentParameter>(*odescent_param_);
+            local_odescent_param->alpha = hierarchy->alpha;
+            ODescent odescent_builder(local_odescent_param,
+                                      codes,
+                                      allocator_,
+                                      this->thread_pool_.get(),
+                                      true,
```
LHT129, line 668: [note] The PiPNN path in `build_by_batch_graph` creates a new `ODescentParameter` copy for each route graph level (`std::make_shared<ODescentParameter>(*route_odescent_param)` on line 667), while the ODescent path mutates the shared `route_odescent_param` in-place. This was already flagged in an existing review comment and fixed in commit 010a7b78. The fix correctly isolates route-level parameter mutations for PiPNN while preserving the original behavior for ODescent. Good.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -607,14 +637,95 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
...
             }
         }
     }
     auto codes = construction_codes();
 
-    if (thread_pool_ != nullptr && hierarchies_.size() > 1) {
+    if (graph_type_ == GRAPH_TYPE_VALUE_PIPNN) {
+        Vector<const float*> rows(allocator_);
+        rows.reserve(static_cast<uint64_t>(data_num));
+        for (const auto input_index : input_indices) {
```
LHT129, line 650: [suggestion] The `PiPNNGraphBuilder` is created once per hierarchy (line 650) and reused for the base graph, all node-level graphs, and all route graphs. This means all graphs within a hierarchy share the same builder configuration (alpha, thread count). However, `PiPNNGraphBuilder` is stateless — it creates a new `PiPNNPipeline` for each `Build()` call, so this is safe. Consider adding a brief comment noting that the builder is intentionally reusable and stateless to avoid confusion for future readers.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2009,12 +2120,15 @@ Pyramid::Build(const DatasetPtr& base) {
         logger::warn(
             "[pyramid_build_cache] duplicate source_id or label; falling back to cold build");
     }
-    populate_hierarchy_trees(base);
     if (graph_type_ == GRAPH_TYPE_VALUE_NSW) {
+        populate_hierarchy_trees(base);
         return this->Add(base);
     }
+    if (graph_type_ != GRAPH_TYPE_VALUE_PIPNN) {
```
LHT129, line 2128: [note] In the PiPNN path, `populate_hierarchy_trees` is called inside `build_by_batch_graph` (line 600) with the filtered `input_indices`, while for NSW the call remains in `Build()` (line 2124) without filtering. For ODescent, `populate_hierarchy_trees` is also called in `Build()` (line 2127) without filtering. This is correct because PiPNN needs the filtered indices to avoid inserting duplicate labels into the hierarchy trees, but the asymmetry in call sites makes the control flow harder to follow. The refactoring is clean, but a comment in `Build()` explaining why PiPNN defers `populate_hierarchy_trees` into `build_by_batch_graph` would help.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1126 @@
...
+    const auto incoming_key = std::make_tuple(distance_key, ids_[target], hash);
+    PointLockGuard lock(point_locks_.get(), source);
+    auto& state = reservoir_states_[source];
+    auto* row = reservoirs_.data() + static_cast<uint64_t>(source) * reservoir_size_;
+
+    if (state.size == reservoir_size_) {
+        const auto& farthest = row[state.farthest];
+        const auto farthest_key =
+            std::make_tuple(farthest.distance, ids_[farthest.neighbor], farthest.hash);
+        if (incoming_key >= farthest_key) {
```
LHT129, line 850: [suggestion] In `insert_candidate`, when a candidate has a different hash from an existing reservoir entry, the code skips it via `continue` (line ~830). This means the first entry that occupies a hash slot permanently blocks all later candidates with different hashes from being considered for that slot, even if the later candidate has a better (lower) distance. This is the intended reservoir behavior (first-come-first-served per hash bucket), but it means insertion order matters within `build_leaf`. Since `build_leaf` iterates over leaf points in their natural order, this is deterministic but could produce suboptimal neighbor selection if early-arriving points have poor distances. Consider whether a distance-based replacement policy (replace if new distance < existing distance, regardless of hash) would improve graph quality.

src/algorithm/hgraph/hgraph_pipnn_test.cpp
```diff
@@ -0,0 +1,303 @@
+// Copyright 2024-present the vsag project
```
LHT129, line 1: [note] The test coverage is thorough: `pipnn_graph_builder_test.cpp` covers parameter validation, determinism, single-point edge case, parallel leaf construction, duplicate row grouping, and empty input. `hgraph_pipnn_test.cpp` covers build+search, route parameter preservation, duplicate-label semantics, parameter validation, serialization round-trip, and extra-info preservation. The test matrix extensions in `test_hgraph.cpp` and `test_pyramid.cpp` add PiPNN to the existing parameterized test suites. One area not explicitly tested: the interaction between `support_duplicate_` and PiPNN in the HGraph path (the `GetGroupId` remapping in `build_by_batch_graph` lines 321-330). The Pyramid tests cover duplicate support via `support_duplicate_` in the route level sampling path, but the HGraph path's duplicate remapping of `entry_point_id_` and `route_graph_ids` after PiPNN build is only exercised indirectly through the existing test matrix. Consider adding a focused test that builds with duplicates and verifies the entry point and route graph IDs are correctly remapped.

src/impl/pipnn/pipnn_graph_builder.cpp
```diff
@@ -0,0 +1,1169 @@
...
+    }
+    const float norm_product = norms_[lhs] * norms_[rhs];
+    if (norm_product <= 0.0F or not std::isfinite(norm_product)) {
+        return 1.0F;
+    }
+    const float similarity = std::clamp(dot / norm_product, -1.0F, 1.0F);
+    return sanitize_distance(1.0F - similarity);
+}
+
+Vector<InnerIdType>
```
LHT129, line 933: [suggestion] `state.size` is read at line 933 before `PointLockGuard` acquires the lock at line 930. While the per-source lock ensures mutual exclusion in practice, reading shared mutable state outside the critical section is technically a data race. If the compiler reorders the read relative to the lock acquisition, or if a future change introduces concurrent access to the same source, this could lead to subtle bugs.

Move the `state.size` check inside the lock scope, after the `PointLockGuard` declaration.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -176,12 +185,34 @@ HGraph::Build(const DatasetPtr& data) {
...
+    PiPNNGraphBuilderParameter pipnn_parameter;
+    if (use_pipnn) {
+        pipnn_parameter.alpha = this->alpha_;
+        pipnn_parameter.Validate(bottom_graph_->MaximumDegree());
+    }
     auto total = data->GetNumElements();
     const auto* labels = data->GetIds();
     const auto* vectors = get_data(data);
     const auto* extra_infos = data->GetExtraInfos();
+    const auto* attr_sets = data->GetAttributeSets();
```
LHT129, line 215: [suggestion] `attr_sets` is declared unconditionally at line 215 but is only consumed inside the `use_pipnn` branch at line 290. Consider moving the declaration closer to its use site (inside the `if (use_pipnn)` block or right before line 290) to reduce scope and make the dependency clearer.

src/impl/searcher/basic_searcher.cpp
LHT129, line 705: [good] Good catch fixing the duplicate expansion condition from `not std::isfinite(dist)` to `std::isfinite(dist)`. The old condition would only expand duplicates when the distance was non-finite (inf/nan), which is the opposite of the intended behavior — duplicates should be expanded when a valid finite distance is found.

src/impl/searcher/parallel_searcher.cpp
LHT129, line 192: [good] Same `not std::isfinite` → `std::isfinite` fix as in basic_searcher.cpp. Good catch.

tools/eval/monitor/memory_peak_monitor.cpp
```diff
@@ -29,25 +29,67 @@ MemoryPeakMonitor::MemoryPeakMonitor(const std::string& name)
...
+    this->Stop();
+}
+
+void
+MemoryPeakMonitor::sample() {
+    std::lock_guard<std::mutex> lock(record_mutex_);
     uint64_t val1, val2;
     this->infile_ >> val1 >> val2;
     this->infile_.clear();
     this->infile_.seekg(0, std::ios::beg);
```
LHT129, line 45: [suggestion] The sampling thread and the external Record() caller both invoke sample() which reads from the same ifstream. While record_mutex_ serializes these calls, the ifstream seekg/read operations are not atomic across threads — a concurrent Record() call could interleave seekg with the sampling thread's extraction operator, producing a garbled page count. Consider using a separate ifstream for the sampling thread, or reading /proc/self/statm into a string and parsing it under the lock.

tools/eval/monitor/duration_monitor.cpp
```diff
@@ -17,22 +17,43 @@
...
 DurationMonitor::DurationMonitor() : Monitor("duration_monitor") {
 }
 
 void
 DurationMonitor::Start() {
+    getrusage(RUSAGE_SELF, &start_usage_);
     cur_time_ = Clock::now();
 }
 void
 DurationMonitor::Stop() {
```
LHT129, line 38: [note] The DurationMonitor::Start() calls getrusage(RUSAGE_SELF, &start_usage_) without checking the return value. On Linux, getrusage can fail (e.g., under restrictive seccomp profiles), leaving start_usage_ zero-initialized but producing silently incorrect CPU duration values. Consider checking the return value and setting a flag to skip CPU metrics on failure.

src/datacell/flatten_datacell.h
```diff
@@ -306,6 +306,19 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::BatchInsertVector(const void* vectors,
         }
         layout_->WriteRange(cur_count, codes.data, count);
     } else {
+        bool ids_are_contiguous = count > 0;
```
LHT129, line 358: [note] The contiguous-IDs optimization in BatchInsertVector adds an O(count) scan for every non-contiguous batch. Since the caller already knows whether the IDs are contiguous (e.g., PiPNN's ParallelBatchInsertVector always passes contiguous IDs), consider adding a separate BatchInsertContiguousVector method or a boolean parameter to avoid the scan overhead on known-contiguous paths.

src/datacell/flatten_build_utils.h
```diff
@@ -0,0 +1,75 @@
...
+            try {
+                futures.emplace_back(thread_pool->GeneralEnqueue(std::move(insert_batch)));
+            } catch (...) {
+                first_exception = std::current_exception();
+                break;
+            }
+        } else {
+            insert_batch();
+        }
+    }
```
LHT129, line 60: [suggestion] When thread_pool->GeneralEnqueue throws (line 60), the loop breaks and drains already-queued futures. However, if an enqueued task throws, the loop at line 68 drains remaining futures but only keeps the first exception. This is the same pattern used in pipnn_graph_builder.cpp and is intentional, but consider adding a brief comment noting that subsequent exceptions are intentionally discarded.

src/impl/odescent/odescent_graph_builder.cpp
```diff
@@ -214,12 +215,51 @@ void
...
+    if (thread_pool_ != nullptr and data_num_ >= 4096) {
+        Vector<std::atomic<uint32_t>> reverse_counts(static_cast<uint64_t>(data_num_), allocator_);
+        auto initialize_counts = [&](int64_t start, int64_t end) {
+            for (int64_t i = start; i < end; ++i) {
+                reverse_counts[i].store(0, std::memory_order_relaxed);
+            }
+        };
+        parallelize_task(initialize_counts);
+
+        auto count_edges = [&](int64_t start, int64_t end) {
```
LHT129, line 227: [note] The parallel add_reverse_edges path uses atomic exchange-then-fill pattern (allocate_rows resets counts to 0 via exchange, fill_edges uses fetch_add to index). This is correct but the 4096 threshold for enabling parallelism is hardcoded. Consider making this threshold configurable or deriving it from the thread pool configuration, since the optimal crossover point depends on both data_num and thread_count.

src/quantization/scalar_quantization/scalar_quantizer.cpp
```diff
@@ -85,14 +85,42 @@ fill_codes(uint8_t* codes, uint8_t value, uint8_t value_bit_size, uint64_t index
...
 ScalarQuantizer<metric, bit>::EncodeOneImpl(const float* data, uint8_t* codes) const {
+    Vector<float> normalized(this->allocator_);
+    if constexpr (metric == MetricType::METRIC_TYPE_COSINE) {
+        normalized.resize(this->dim_);
+    }
+    return EncodeOneWithScratch(data, codes, normalized.data());
+}
+
+template <MetricType metric, int bit>
+bool
```
LHT129, line 96: [note] EncodeBatchImpl allocates a single normalized buffer for cosine and reuses it across all rows via EncodeOneWithScratch. This avoids per-call allocation overhead. However, for non-cosine metrics the normalized buffer is allocated but never used (the if constexpr branch is dead). Consider using if constexpr to skip the allocation entirely for non-cosine metrics to avoid the unnecessary Vector allocation.

LHT129 (commented):
## Code Review: feat: add parallel PiPNN graph construction

Thank you for this substantial contribution. The PiPNN implementation is well-structured with clear separation of concerns, thorough validation, and good test coverage. Below are findings organized by category.

### Correctness

**1. `group_duplicate_rows` uses `std::string_view` with raw byte comparison (pipnn_graph_builder.cpp:126-140)**

Using `std::string_view` constructed from raw float bytes as a hash key relies on the exact binary representation of the floats. IEEE 754 has two representations of zero (+0 and -0) that compare equal mathematically but differ in memory, and NaN values never compare equal to themselves. If the input vectors are normalized or computed through different code paths, this could miss genuine duplicates or falsely treat near-identical vectors as distinct. Consider whether the caller guarantees exact byte-identical vectors, or whether a tolerance-based comparison is needed.

**2. `insert_candidate` hash collision logic may silently drop non-coincident candidates (internal/build.cpp:1085-1095)**

When an incoming candidate shares the same hash with an existing entry and the hash is not coincident, the incoming candidate replaces the existing one only if it is closer. However, two different hash-plane configurations can produce the same hash value for non-coincident points. In this case, the replacement logic silently drops one of the two distinct hash-plane regions. This is by design per the PiPNN algorithm (one candidate per hash region), but the comment should clarify this is intentional hash-region deduplication, not a bug.

**3. `build_by_batch_graph` in pyramid.cpp uses `input_indices` to build `rows` but `odescent_builder` receives `data_vectors` directly (pyramid.cpp:672-696)**

When `graph_type_ == GRAPH_TYPE_VALUE_PIPNN`, duplicate labels are filtered into `failed_ids` and excluded from `input_indices`. The `rows` vector is built from `input_indices` (filtered), but `odescent_builder` is constructed with the full `data_vectors` and `data_num` (which is `input_indices.size()`). This is correct because `data_num` matches the filtered count, but the `odescent_builder` is only used for non-level-0 routing when `support_duplicate_` is false, so it never actually uses `data_vectors` for duplicate-filtered data. This is subtle and worth a comment.

### Code Style

**4. `using Leaf = Vector<uint32_t>; using Leaves = Vector<Leaf>;` at file scope in an anonymous namespace (internal/build.cpp:88-89)**

The VSAG style guide prefers explicit types over type aliases for readability. These aliases are used extensively (50+ occurrences), so they improve readability here, but the project convention should be checked.

**5. `int` vs `int64_t` in `odescent_graph_builder.cpp` (lines 217-263)**

The new parallel `add_reverse_edges` and `count_in_edges` code uses `int64_t` consistently, which is correct. However, the existing code in `repair_no_in_edge` (line 332) still uses `int` for loop variables. While pre-existing, the new code correctly uses `int64_t`.

**6. `NOLINTNEXTLINE(modernize-use-equals-default)` on `ReservoirEntry` constructor (internal/build.cpp:79)**

The comment explains why the default is suppressed (fully assigned before readable), but this is a common pattern that should be documented in the commit message rather than relying on a lint suppression.

### Tests

**7. Good test coverage overall**

The test suite covers:
- Determinism across multiple builds
- Single-point edge case
- Parallel execution determinism
- Identical vector fallback leaves
- Duplicate row registration
- Parameter validation
- All three supported metrics (L2, IP, cosine)
- HGraph integration with all parameters, RaBitQ, serialization, extra info, remove, duplicates
- Pyramid integration with multi_layer root, path search, duplicates, RaBitQ
- Existing tests parametrized with PiPNN via `HGRAPH_PR_DAILY_CASE_WITH_PIPNN`

**8. Missing test: `build_by_batch_graph` with `extra_infos` validation in Pyramid**

The HGraph `build_by_pipnn` validates extra info presence and size (hgraph_build.cpp:210-215), but the Pyramid `build_by_batch_graph` does not appear to perform the same validation for PiPNN builds. The Pyramid path relies on `populate_hierarchy_trees` which may handle this differently, but a test confirming extra info validation in Pyramid PiPNN builds would be valuable.

### Performance

**9. `prepare_sketches` allocates and fills hyperplanes every build (internal/build.cpp:425-468)**

The random projection hyperplanes are generated fresh on every `Build` call. For repeated builds (e.g., tuning, benchmarking), this is fine. But if the builder is reused, consider caching the hyperplanes.

**10. `parallel_for` creates `std::future` vector on every call (internal/build.cpp:1370-1390)**

The `parallel_for` helper allocates a `Vector<std::future<void>>` on every invocation. For the inner `build_leaf` loop, this is called once per leaf, which could be many small allocations. Consider a pre-allocated future pool or using a simpler atomic counter with a barrier.

### Security

**11. No command injection or buffer overflow issues found**

The code properly validates all inputs, uses `checked_product` for overflow-safe multiplication, and validates array bounds. No external input is passed to shell commands.

### Maintainability

**12. `build_by_batch_graph` in hgraph_build.cpp has code duplication with `build_by_odescent`**

The `build_by_batch_graph` method was extracted to share code between ODescent and PiPNN, but `build_by_odescent` still exists as a thin wrapper. Consider whether `build_by_odescent` can be removed entirely and its callers updated to use `build_by_batch_graph(data, false)`.

**13. `hgraph_pipnn_test.cpp` reimplements `MakeCommonParam` instead of using the existing test fixture**

The new test file creates its own `MakePiPNNCommonParam` helper that duplicates the existing `MakeCommonParam` pattern. Consider consolidating these helpers into the shared test fixture.

### Summary

The implementation is solid, well-tested, and follows the project conventions. The issues identified are minor — the most significant being the potential float comparison edge case in `group_duplicate_rows` and the missing extra info validation in Pyramid PiPNN builds. No blocking issues found.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: This is intentional, and no code change is needed. A coincident sketch is excluded only from same-hash collision replacement; it is still appended while capacity remains and participates in the global farthest replacement when full. The current control flow already makes both outcomes explicit, so I avoided adding a restatement comment.

src/algorithm/hgraph/hgraph_build.cpp
jac0626: Confirmed this is pre-existing, single-use ODescent behavior and is not re-entered. I kept it unchanged to avoid expanding this PiPNN PR with a comment-only edit to an unaffected path.

src/algorithm/hgraph/hgraph_build.cpp
jac0626: Confirmed there is no local shadow and the member resolves unambiguously. I kept the existing expression unchanged because this is style-only and does not affect the contract.

src/impl/searcher/basic_searcher.cpp
jac0626: Restored the finite-distance duplicate expansion after the rebase in ffb4cc79. This was required: the Pyramid duplicate CI case reproduced a missing alias at the entry point. The full local Pyramid Duplicate Test now passes all 54,740 assertions.

src/impl/searcher/parallel_searcher.cpp
jac0626: Restored the same finite-distance condition in ParallelSearcher in ffb4cc79. The concurrent Pyramid duplicate coverage and the HGraph PiPNN duplicate matrix both pass locally.

src/algorithm/pyramid/pyramid.cpp
jac0626: Kept the paths intentionally asymmetric. ODescent retains its existing input-label behavior; PiPNN needs a unique dense local-id sequence shared by the label table, path tree, row array, and graph builder. Changing ODescent here would alter pre-existing observable behavior outside this PR.

src/algorithm/pyramid/pyramid.cpp
jac0626: Confirmed. The per-route copy remains in the current head and preserves the configured shared parameter.

src/algorithm/pyramid/pyramid.cpp
jac0626: Confirmed the builder is stateless because each Build call owns a fresh pipeline. I did not add a comment because the stack-local pipeline in Build is the direct ownership evidence and an additional comment would only restate the implementation.

src/algorithm/pyramid/pyramid.cpp
jac0626: Confirmed. PiPNN keeps populate_hierarchy_trees next to the filtered input_indices that define its local-id mapping; NSW and ODescent retain their original unfiltered call site. I kept the code self-describing rather than adding a control-flow comment.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: No change is needed here: a different hash does not block the incoming candidate. The loop continues across different hashes; after the scan, the candidate is appended when capacity remains or replaces the global farthest entry when full. Distance replacement within the loop applies only to an actual same-direction hash collision.

src/algorithm/hgraph/hgraph_pipnn_test.cpp
jac0626: The observable remapping contract is already covered by HGraph PiPNN preserves duplicate-vector search quality, PR HGraph Duplicate PiPNN, compressed serialization, and duplicate-build tests. I ran the complete HGraph PiPNN duplicate tag locally: 487,947 assertions across four cases passed. I kept that existing matrix instead of adding an internal-state-only test.

src/impl/pipnn/pipnn_graph_builder.cpp
jac0626: Fixed in the current implementation: PointLockGuard is acquired before reservoir_states_[source] is referenced and before state.size is read. The current PiPNN unit suite passes.

src/algorithm/hgraph/hgraph_build.cpp
jac0626: Kept the declaration with the other Dataset payload pointers. Moving only this pointer would be style-only, while moving the lookup into the per-row branch would enlarge the hot-loop expression without changing behavior.

src/impl/searcher/basic_searcher.cpp
jac0626: Confirmed and retained after rebase in ffb4cc79. The finite-distance expansion is now covered by the previously failing Pyramid duplicate test.

src/impl/searcher/parallel_searcher.cpp
jac0626: Confirmed and retained after rebase in ffb4cc79 for the parallel path as well.

tools/eval/monitor/memory_peak_monitor.cpp
jac0626: The stream operations cannot interleave because record_mutex_ is held across extraction, clear, seek, and max update in sample(), and Record() also delegates to sample(). I did add the required read validation in ffb4cc79 so a failed or malformed statm read cannot update the peak from uninitialized values; eval_dataset_test passes.

tools/eval/monitor/duration_monitor.cpp
jac0626: I kept this unchanged. RUSAGE_SELF with valid local storage has no normal failure path on the supported Linux evaluation target; a seccomp profile that rejects the required syscall cannot provide this metric. Adding conditional result fields would create a new output schema without a supported reproducer.

src/datacell/flatten_datacell.h
jac0626: I kept the current API. BatchInsertVector must determine whether the generic id array is contiguous before selecting WriteRange; adding a second public-style method or mode flag would expand the interface. No measurement shows this linear integer scan is material relative to batch encoding and storage writes.

src/datacell/flatten_build_utils.h
jac0626: Confirmed the first-exception behavior is intentional and matches the surrounding future-drain pattern. The code keeps the first failure while draining every submitted task; I avoided a comment that only restates that control flow.

src/impl/odescent/odescent_graph_builder.cpp
jac0626: I kept the threshold unchanged. Making it configurable would add a new tuning surface without same-workload crossover measurements, and this review provides no evidence that 4096 is a current bottleneck.

src/quantization/scalar_quantization/scalar_quantizer.cpp
jac0626: No heap allocation occurs for non-cosine instantiations: the allocator-aware Vector remains empty because resize is compiled only for the cosine branch. The object itself is stack-local, so an additional branch would not remove a backing allocation.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -84,6 +84,13 @@ HGraph::HGraph(const HGraphParameterPtr& hgraph_param, const vsag::IndexCommonPa
         throw VsagException(ErrorType::INVALID_ARGUMENT,
                             "HGraph deduplicate_storage only supports dense vectors");
     }
+    if (this->graph_type_ == GRAPH_TYPE_VALUE_PIPNN) {
+        if (common_param.repr_ != RecordRepr::DENSE or
+            common_param.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
+            throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                "HGraph PiPNN only supports dense float32 indexes");
+        }
```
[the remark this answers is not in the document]
jac0626: Fixed in ffb4cc79. The non-optimized batch path now publishes fused node codes after labels are installed, in parallel batches, before PiPNN reads the graph. Added a focused fused RaBitQ plus PiPNN build/distance/search regression; all seven assertions pass locally.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -624,14 +662,96 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
...
+                                      codes,
+                                      allocator_,
+                                      this->thread_pool_.get(),
+                                      true,
+                                      data_vectors,
+                                      data_num);
+            GraphBuildFunc build_graph =
+                [&](GraphInterfacePtr& graph, const Vector<InnerIdType>& ids, uint32_t level) {
+                    if (level == 0) {
+                        graph->SetMaxCapacity(static_cast<InnerIdType>(data_num));
```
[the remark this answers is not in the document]
jac0626: Fixed in the current head. build_by_batch_graph calls Pyramid::resize(data_num) before graph construction, and Pyramid::resize invokes resize_graph on every hierarchy root, which resizes the bottom graph storage, duplicate tracker, and removal metadata as applicable. The multi-layer PiPNN root tests pass locally.

src/algorithm/pyramid/pyramid.h
```diff
@@ -75,6 +75,11 @@ class Pyramid : public InnerIndexInterface {
           persist_source_id_(pyramid_param->persist_source_id),
           store_paths_(pyramid_param->store_paths),
           cache_(std::make_unique<PyramidBuildCache>(common_param.allocator_.get())) {
+        if (graph_type_ == GRAPH_TYPE_VALUE_PIPNN and
+            common_param.data_type_ != DataTypes::DATA_TYPE_FLOAT) {
```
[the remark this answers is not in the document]
jac0626: Fixed in ffb4cc79 by matching HGraph and requiring both RecordRepr::DENSE and float32 for Pyramid PiPNN. Added a focused non-dense rejection test, included in the passing PiPNN unit suite.

LHT129 (approved):
LGTM

Issue #2812, opened 2026-08-31
[feat](ivf): offload IVF index build to an optional CUDA backend


seuclkzsd:
| Field | Value |
|---|---|
| Affected area | `ivf` |
| API / ABI compatibility impact | `additive only (no break)` |

---

## Motivation

Building an IVF index is CPU bound, and today users pay for index quality with
build time. Measured with `tools/eval/` on SIFT1M (1M base, 10k queries,
topk=10, 48 threads, `fp32`):

| buckets_count | `ivf_train_type` | build (s) | Recall@10 |
|---|---|---|---|
| 16384 | random | 66.16 | 0.9471 |
| 16384 | kmeans | **729.04** | 0.9723 |

k-means training costs 11x the build time of random training to buy 2.5 points
of recall, and it gets worse as `buckets_count` grows, because
`train_sample_count = max(65536, 64 * buckets_count)`
(`src/algorithm/ivf/ivf_parameter.cpp:88-97`) grows with it.

Two hot spots dominate, in different regimes:

- **k-means training.** `select_initial_centroids_kmeans_plus_plus`
  (`src/impl/cluster/kmeans_cluster.cpp:383`) sweeps every point once per
  centroid, single-threaded, so seeding alone is `O(k * count * dim)` — roughly
  nine tenths of the build at `buckets_count = 16384`.
- **Assigning the base to buckets.** `IVFNearestPartition::ClassifyDatas`
  (`src/algorithm/ivf/ivf_nearest_partition.cpp:92`) sends each vector through a
  graph search over the centroids. That shape suits search-time routing, which
  the same function also serves; at build time it asks for the single nearest
  centroid over the whole base, which is a dense matrix product. It is about
  four fifths of the build at `buckets_count = 1000`.

Existing knobs only move the cost around: `random` training trades 1.2 to 2.5
points of recall, and lowering `buckets_count` trades search performance.

## Proposed solution

An optional CUDA backend for the build path, off by default.

A self-contained `src/gpu/` module behind a new `ENABLE_CUDA` build option
(default `OFF`). With the option off a stub with the same symbols is compiled, so
callers link unconditionally and no `#ifdef` reaches the rest of the tree. The
interface takes and returns host memory only. Points are streamed in chunks, so
the base does not have to fit in VRAM.

Four places would offload: k-means++ seeding, nearest-centroid assignment
(cuBLAS SGEMM plus a fused argmin, so the k×n distance matrix is consumed on the
[...]

PR #2860, opened 2026-09-04
build(gpu): add an optional CUDA backend target, off by default

seuclkzsd:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [x] CI/Build/Infra

## Linked Issue

- Part of: #2812

## What Changed

Groundwork for offloading parts of the IVF index build to a CUDA device. This
PR adds the plumbing only: a build option, a target, and device discovery. No
code calls into the backend, so behaviour is unchanged in both configurations.

- `ENABLE_CUDA` (default `OFF`, requires CUDA 11.8+) and
  `VSAG_CUDA_ARCHITECTURES` (default Turing through Hopper, SASS for each and
  PTX for the newest)
- `src/gpu/` with `CudaAvailable()` and `CudaSelectDevice()`, a CUDA
  implementation and a stub exporting the same symbols
- `vsag_gpu` linked into `VSAG_DEP_LIBS`; `gpu_test` registered with
  `tests/CMakeLists.txt`, the Makefile module list and the topology check
- tests for both configurations, with the device-dependent assertions
  skipping rather than compiling out on a machine without one
- `src/gpu/` registered as a Codecov component; clang-format covers `.cu`

Scope is limited to the build system so that whether a second compile language
belongs in the tree can be reviewed on its own, separately from the kernels that
follow.

**A stub rather than `#ifdef` at the call sites.** With the option off, a stub
exporting the same symbols is compiled in place of the CUDA translation unit, so
callers link unconditionally and ask at runtime whether a device is present. No
call site carries a preprocessor branch, and no CUDA type appears in the header,
so translation units that include it do not need the toolkit.

**An ordinal that does not exist is refused rather than replaced.** On a machine
whose devices are shared, substituting a different card would consume a resource
the caller did not ask for, and without any indication that it had happened. A
negative ordinal is refused before the runtime is touched at all.

**The CUDA runtime is linked statically.** The backend's contract is that a
CUDA-enabled library still loads on a host without a device and reports none.
The shared runtime would make `libcudart.so` a hard dependency of the result.
The language is enabled before the toolkit is searched for, so nvcc, headers
and libraries come from one install; the host pass uses the project's C++
compiler and standard. `VSAG_ENABLE_CUDA` and the toolkit's include path are
private to the backend's own target, so no other translation unit sees either.

## Test Evidence

- [x] `make fmt`
- [ ] `make lint`
- [x] Other (described below)

`make fmt` was run with clang-format 15.0.7 and left the tree unchanged; the
script now covers `src/**/*.cu` as well. `make lint` was not run: clang-tidy 15
is not installed here.

Both configurations were configured, built, and their tests run.

```text
# ENABLE_CUDA=OFF (default)
$ cmake -S . -B build-off -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON
$ cmake --build build-off --target unittests
$ ./build-off/tests/unittests "[gpu_backend]"
All tests passed (4 assertions in 3 test cases)

# ENABLE_CUDA=ON
$ cmake -S . -B build-on -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON \
        -DENABLE_CUDA=ON -DCMAKE_CUDA_COMPILER=/usr/local/cuda-12.4/bin/nvcc
-- CUDA backend: ON (toolkit 12.4.131, arch 75-real;80-real;86-real;89-real;90)
$ cmake --build build-on --target unittests
$ ./build-on/tests/unittests "[gpu_backend]"          # 7 devices visible
All tests passed (9 assertions in 3 test cases)
$ CUDA_VISIBLE_DEVICES= ./build-on/tests/unittests "[gpu_backend]"
test cases: 3 | 2 passed | 1 skipped

# Same symbols either way, which is what lets callers link unconditionally
$ nm -C build-off/src/gpu/libvsag_gpu.a | grep 'T vsag::gpu'
T vsag::gpu::CudaAvailable()
T vsag::gpu::CudaSelectDevice(int)
$ nm -C build-on/src/gpu/libvsag_gpu.a  | grep 'T vsag::gpu'
T vsag::gpu::CudaAvailable()
T vsag::gpu::CudaSelectDevice(int)
```

With the option on, availability is checked against `cudaGetDeviceCount` and
the ordinal equal to that count is refused, so the case passes with zero devices
or several. The case that selects a device goes to the last ordinal first, so
with more than one device the binding is a switch `cudaGetDevice` can confirm;
on a machine with none it skips. The stub's cases are guarded on
`VSAG_ENABLE_CUDA` and compile only when the option is off. `libvsag.so` built
with the option on has no dynamic dependency on `libcudart`.

Environment: Ubuntu 20.04.5, GCC 9.4.0, CMake 3.20.6, CUDA 12.4.131, RTX 3090.

Building with the option on needs `nvcc` on `PATH` or `CMAKE_CUDA_COMPILER` set,
as CMake requires for any CUDA project.

## Compatibility Impact

- API/ABI compatibility: none. No public header changes.
- Behavior changes: none. No code calls the backend; with `ENABLE_CUDA=OFF` the
  stub reports no device.

- Performance impact: none. No code path calls into the backend.
- Concurrency/thread-safety impact: none. `CudaSelectDevice` binds the calling
  thread, which is CUDA's per-thread context model.

## Documentation Impact

- [x] Docs updated

`ENABLE_CUDA` and `VSAG_CUDA_ARCHITECTURES` are listed in `DEVELOPMENT.md` and
in `docs/docs/{en,zh}/src/development/building.md`; `gpu` is added to the
maintained module list in `docs/agents/build-and-test.md`. Runtime knobs and
their documentation come with the PR that first offloads work.

## Risk and Rollback

- Risk level: low. Under the default option the change is one static library of
  two functions that nothing calls.
- Rollback plan: revert the commit. No target depends on it.

## Checklist


LHT129 (commented):
Reviewed the full diff across all 9 files. This is a clean, well-designed skeleton PR.

**What works well:**

- The stub pattern (compile `cuda_backend_stub.cpp` in place of `cuda_backend.cu` when `ENABLE_CUDA=OFF`) is elegant — callers link unconditionally and decide at runtime, avoiding `#ifdef` at every call site.
- The header exposes no CUDA types, so includers do not need the toolkit.
- `CudaSelectDevice` correctly refuses out-of-range ordinals rather than silently substituting a different device.
- `VSAG_ENABLE_CUDA` is a PUBLIC definition on `vsag_gpu`, so the test guard (`#ifndef VSAG_ENABLE_CUDA`) works correctly in both configurations.
- The test covers the stub contract well: no device reported, and valid/invalid/negative ordinals all refused.
- Build system integration is careful: `gpu_test` is added to `--whole-archive`/`-force_load` on both Linux and macOS, and the CUDA compile options use generator expressions to avoid leaking into CXX compilation.
- All files follow VSAG coding standards (4-space indent, `.cpp` suffix, `vsag::gpu` namespace, trailing newlines).

No issues found worth flagging. LGTM.

src/gpu/cuda_backend.cu
```diff
@@ -0,0 +1,40 @@
...
+    int count = 0;
+    // A machine with no driver reports an error rather than a count of zero.
+    if (cudaGetDeviceCount(&count) != cudaSuccess) {
+        return false;
+    }
+    return count > 0;
+}
+
+bool
+CudaSelectDevice(int32_t device_id) {
```
LHT129, line 32: [note] `CudaSelectDevice` calls `cudaSetDevice` without first saving the current device via `cudaGetDevice`. If `cudaSetDevice` fails, the calling thread may be left in an indeterminate device state — the CUDA runtime documentation does not guarantee that a failed `cudaSetDevice` leaves the previous device intact.

This is low-risk for the skeleton since callers that get `false` are expected to fall back to the CPU path and never touch CUDA again, but it is worth keeping in mind when the backend grows real kernels that may be called after a failed device selection.

If future callers need to attempt device selection and then proceed with CUDA work on a fallback device, consider saving and restoring the original device:

```cpp
int prev = 0;
cudaGetDevice(&prev);
if (cudaSetDevice(device_id) != cudaSuccess) {
    cudaSetDevice(prev);  // best-effort restore
    return false;
}
```

seuclkzsd: Thanks. I measured the behaviours this rests on, on CUDA 12.4.131 / driver
550.54.15 / RTX 3090.

A failed cudaSetDevice does not move the current device on this runtime, though
the documentation does not guarantee that:

    cudaSetDevice(7)  -> no error                current device: 7
    cudaSetDevice(99) -> invalid device ordinal  current device: 7

The restore is not free. cudaGetDevice creates no context, but cudaSetDevice
binds the primary context immediately, and the restore is a cudaSetDevice. On a
thread that has none yet, a failed CudaSelectDevice(3) would restore to device 0
and take memory there. On a shared machine that is the case this function exists
to avoid.

A failed call also latches an error that later successful calls do not clear.
Four succeeded here and the stale error was still readable:

    cudaSetDevice(99)  -> invalid device ordinal
    cudaMalloc         -> no error
    cudaMemset         -> no error
    launch + sync      -> no error
    cudaGetLastError() -> invalid device ordinal

Harmless in this PR, since false sends the caller to the CPU path. It matters
once an entry point checks cudaGetLastError() after a launch: an earlier failed
probe on the same thread reads as a failure of that launch.

So I would leave this PR as it is. The ordinal is validated against
cudaGetDeviceCount first, and false means CPU path. The save/restore (guarded so
it creates no context that was not there) and the error clearing belong in the
PR that adds the first kernels, where both change behaviour. Happy to fold them
in here instead if you prefer.

Roxanne0321:
I validated the current head (`34e99462`) on Ubuntu 22.04 with CUDA 11.8 / driver 535.98 / an RTX 2080 Ti.

The CUDA-enabled configuration and build succeeded, including `vsag_gpu` and `unittests_gpu`. I also ran a small runtime smoke test against the built library: `CudaAvailable()` reported the GPU, device 0 could be selected, and negative/out-of-range ordinals were rejected.

There are two test-integration gaps worth addressing in this PR:

1. **The CUDA-ON test target is currently empty.** The two `[gpu_backend]` cases are compiled only under `#ifndef VSAG_ENABLE_CUDA`, so running the CUDA-enabled binary reports `No tests ran`. This can make a CUDA build look green without exercising the CUDA implementation. Please keep at least one CUDA-enabled test case, for example:
   - always verify that a negative ordinal is rejected;
   - after obtaining the device count, verify that `device_count` is rejected;
   - when a device is available, select device 0 and verify the selected device with `cudaGetDevice()`;
   - if the runner has no GPU, skip only the device-dependent assertions instead of compiling out the entire test suite.

2. **`make test-module MODULE=gpu` is rejected by the Makefile whitelist.** `tests/CMakeLists.txt` registers `gpu`, but `UNITTEST_MODULES` in the root `Makefile` does not include it. Adding `gpu` there would make the repository's standard module-test entry point work for this new module.

For completeness, I also ran the full Release suite with CUDA disabled. Its six failures were reproduced individually on the exact base commit (`467ad0f8`), so they do not appear to be regressions from this PR.

LHT129 (commented):
Reviewed the full diff (297 additions across 18 files). This is a well-structured PR that adds CUDA backend plumbing with no behavioral change in the default configuration.

**What I checked:**
- **Correctness**: `CudaAvailable()` and `CudaSelectDevice()` logic is sound — negative ordinals are refused before touching the runtime, `cudaGetDeviceCount` errors are handled, and ordinal bounds are validated.
- **Stub design**: The stub exporting identical symbols is a clean approach that avoids `#ifdef` at every call site. Callers link unconditionally and decide at runtime.
- **Build system**: CUDA language is enabled before toolkit search, host compiler is pinned to the project C++ compiler, `cudart_static` keeps `libcudart.so` off the shared library, and `VSAG_ENABLE_CUDA` is private to the backend target.
- **Tests**: Both configurations are tested — stub tests verify no device is reported, real backend tests verify agreement with the runtime and device binding. The skip-on-no-device pattern is correct.
- **Documentation**: `DEVELOPMENT.md`, both `building.md` files, and `build-and-test.md` are updated consistently. The `gpu` module is registered in the Makefile, test topology check, and Codecov.
- **Tooling**: `.cu` is added to `format-cpp.sh` and `check-struct-names.py`.

**No blocking issues found.** The earlier discussion about `cudaSetDevice` failure state (LHT129's note and the author's reply) was resolved with a well-reasoned conclusion that save/restore belongs in the PR that adds the first kernels.

This PR is ready from a review perspective.

wxyucs (approved):
lgtm

PR #2861, opened 2026-09-04
perf(pyramid): reuse RaBitQ split build session

### LINKED ISSUE #2783: Pyramid does not reuse the optimized RaBitQ split build session
## Problem

HGraph can build a RaBitQ split index through the quantizer's optimized scalar-code build representation. Pyramid instead creates an ODescent-specific temporary SQ8 datacell when split storage is enabled.

This duplicates build-representation policy inside ODescent and prevents Pyramid, IVF, and other index implementations from reusing the optimized build lifecycle exposed by the underlying datacell. It is especially noticeable for MRLE + RaBitQ split configurations.

## Expected behavior

The temporary build representation should be owned by the flatten/datacell abstraction. Index implementations should open an optional build session, use the normal distance APIs while the session is active, and commit it after graph construction. Implementations that do not provide an optimized representation should keep their existing behavior.

Pyramid should use the same RaBitQ split scalar-code build session as HGraph without duplicating quantizer-specific logic in ODescent.

## Validation

Validate build completion, search, flat-node promotion, incremental add, serialization, and optimized-build failure cleanup.

LightWant:
## Summary

- expose an optional RAII optimized-build session through `FlattenInterface`
- let Pyramid use the RaBitQ split scalar-code representation during ODescent construction
- remove ODescent-specific temporary SQ8 build datacell policy
- restore persisted split codes when an optimized session is opened for incremental Add

## Performance

Release build on the 50,000-vector, 1536-dimensional OpenAI angular dataset, using Pyramid + ODescent, MRLE dim 768, RaBitQ split 3+5, and 16 build threads:

| Version | Median build time | Throughput |
| --- | ---: | ---: |
| `origin/main` | 17.437 s | 2867 vectors/s |
| This PR | 10.744 s | 4654 vectors/s |

This is a 1.62x speedup and a 38.38% reduction in build time across three runs. Index memory remained 265.37 MiB.

Search quality with `ef_search=100` and 1,000 queries remained equivalent:

| Version | Median Recall@10 | Median single-thread QPS |
| --- | ---: | ---: |
| `origin/main` | 0.7541 | 1400.77 |
| This PR | 0.7543 | 1394.27 |

## Validation

- Release build
- `RaBitQSplitDataCell optimized scalar-code build`
- `Pyramid MRLE split promotes flat nodes without raw vectors`
- `[MRLE][pyramid]`
- `Pyramid MRLE RaBitQ Split`
- `HGraph optimized build accepts MRLE RaBitQ split with raw-vector storage`
- `ODescent Build Test`

Fixes: #2783

### DIFF
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -593,6 +593,10 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
         }
     }
 
+    auto codes = construction_codes();
+    FlattenOptimizedBuildContext build_context{thread_pool_, build_thread_count_};
+    auto build_session = codes->CreateOptimizedBuildSession(build_context);
+
     base_codes_->BatchInsertVector(data_vectors, data_num);
     if (has_precise_reorder()) {
         precise_codes_->BatchInsertVector(data_vectors, data_num);
@@ -612,22 +616,12 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
             }
         }
     }
-    auto codes = construction_codes();
-
     if (thread_pool_ != nullptr && hierarchies_.size() > 1) {
-        auto build_flatten = ODescent::CreateBuildFlatten(codes, data_vectors, data_num);
         Vector<std::future<void>> futures(allocator_);
         for (const auto& [hname, h_ptr] : hierarchies_) {
             auto* hierarchy = h_ptr.get();
-            futures.push_back(thread_pool_->GeneralEnqueue([&, codes, build_flatten, hierarchy]() {
-                ODescent builder(odescent_param_,
-                                 codes,
-                                 allocator_,
-                                 nullptr,
-                                 true,
-                                 data_vectors,
-                                 data_num,
-                                 build_flatten);
+            futures.push_back(thread_pool_->GeneralEnqueue([&, codes, hierarchy]() {
+                ODescent builder(odescent_param_, codes, allocator_, nullptr, true);
                 hierarchy->root->Build(builder);
             }));
         }
@@ -635,16 +629,13 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
             f.get();
         }
     } else {
-        ODescent graph_builder(odescent_param_,
-                               codes,
-                               allocator_,
-                               this->thread_pool_.get(),
-                               true,
-                               data_vectors,
-                               data_num);
+        ODescent graph_builder(odescent_param_, codes, allocator_, this->thread_pool_.get(), true);
         for (const auto& [hname, h_ptr] : hierarchies_) {
             h_ptr->root->Build(graph_builder);
         }
+    }
+    if (build_session != nullptr) {
+        build_session->Commit();
     }
     cur_element_count_ = data_num;
     return {};
@@ -1543,8 +1534,14 @@ Pyramid::ExportModel(const IndexCommonParam& param) const {
 
 std::vector<int64_t>
 Pyramid::Add(const DatasetPtr& base) {
+    auto codes = construction_codes();
+    FlattenOptimizedBuildContext build_context{thread_pool_, build_thread_count_};
+    auto build_session = codes->CreateOptimizedBuildSession(build_context);
     auto batch = prepare_add_batch(base);
     insert_add_batch(base, batch);
+    if (build_session != nullptr) {
+        build_session->Commit();
+    }
     return batch.failed_ids;
 }
 
--- a/src/algorithm/pyramid/pyramid.cpp
+++ b/src/algorithm/pyramid/pyramid.cpp
@@ -2241,19 +2241,23 @@ Pyramid::add_to_path(Hierarchy& hierarchy,
         if (depth != static_cast<int>(path_slices.size())) {
             child = node->GetChild(path_slices[depth], true);
         }
-        if (no_build_level_index < static_cast<int>(hierarchy.no_build_levels.size()) &&
-            depth == hierarchy.no_build_levels[no_build_level_index]) {
-            node = child;
+        const bool skip_level =
+            no_build_level_index < static_cast<int>(hierarchy.no_build_levels.size()) &&
+            depth == hierarchy.no_build_levels[no_build_level_index];
+        if (skip_level) {
             ++no_build_level_index;
-            continue;
+        } else {
+            add_one_point(hierarchy,
+                          node,
+                          inner_id,
+                          vector,
+                          0,
+                          false,
+                          depth == 0 ? sampled_root_level : std::numeric_limits<int>::min());
         }
-        add_one_point(hierarchy,
-                      node,
-                      inner_id,
-                      vector,
-                      0,
-                      false,
-                      depth == 0 ? sampled_root_level : std::numeric_limits<int>::min());
+        if (child == nullptr) {
+            break;
+        }
         node = child;
     }
 }
--- a/src/datacell/flatten_interface.h
+++ b/src/datacell/flatten_interface.h
@@ -25,6 +25,7 @@
 #include "basic_types.h"
 #include "flatten_datacell_parameter.h"
 #include "flatten_interface_parameter.h"
+#include "flatten_optimized_build_interface.h"
 #include "hash_types.h"
 #include "impl/runtime_parameter.h"
 #include "index_common_param_fwd.h"
@@ -106,6 +107,24 @@ public:
 
     virtual ComputerInterfacePtr
     FactoryComputer(const void* query) = 0;
+
+    // Starts an optional temporary build-code representation. The returned session aborts on
+    // destruction unless Commit() succeeds. While it is alive, all regular distance APIs are
+    // required to address the temporary representation transparently.
+    virtual FlattenBuildSessionPtr
+    CreateOptimizedBuildSession(const FlattenOptimizedBuildContext&) {
+        return nullptr;
+    }
+
+    [[nodiscard]] virtual bool
+    IsOptimizedBuildActive() const {
+        return false;
+    }
+
+    virtual ComputerInterfacePtr
+    FactoryComputerForBuild(const void* query, InnerIdType) {
+        return this->FactoryComputer(query);
+    }
 
     virtual void
     Train(const void* data, uint64_t count) = 0;
--- a/src/datacell/rabitq_split_datacell.h
+++ b/src/datacell/rabitq_split_datacell.h
@@ -75,6 +75,29 @@ public:
         uint64_t code_sum_{0};
     };
 
+    class BuildSession final : public FlattenBuildSession {
+    public:
+        explicit BuildSession(RaBitQSplitDataCell* owner) : owner_(owner) {
+        }
+
+        ~BuildSession() override {
+            if (owner_ != nullptr) {
+                owner_->AbortOptimizedBuild();
+            }
+        }
+
+        void
+        Commit() override {
+            if (owner_ != nullptr) {
+                owner_->FinalizeOptimizedBuild();
+                owner_ = nullptr;
+            }
+        }
+
+    private:
+        RaBitQSplitDataCell* owner_;
+    };
+
     RaBitQSplitDataCell() = default;
 
     explicit RaBitQSplitDataCell(const QuantizerParamPtr& quantization_param,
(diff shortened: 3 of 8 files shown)


src/datacell/rabitq_split_datacell.h
```diff
@@ -75,6 +75,29 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
         uint64_t code_sum_{0};
```
LHT129, line 75: [note] test comment

LHT129 (commented):
## Review Summary

This PR introduces a well-designed RAII session pattern (`FlattenBuildSession`) that wraps the existing `BeginOptimizedBuild`/`FinalizeOptimizedBuild`/`AbortOptimizedBuild` lifecycle. The changes correctly:

1. **Simplify ODescent**: Remove the ad-hoc SQ8 build datacell and the `CreateBuildFlatten` static method. ODescent now uses the same `FlattenInterface` for both build-time and regular distance computations, with the optimized build path transparently handled inside `RaBitQSplitDataCell`.
2. **Improve exception safety**: The `BuildSession` destructor calls `AbortOptimizedBuild()`, ensuring cleanup on exception paths.
3. **Support incremental builds**: The rehydration logic in `BeginOptimizedBuild` restores scalar codes from persisted split codes, enabling incremental `Add()` calls to reuse the optimized build path.

A few observations below.

src/datacell/rabitq_split_datacell.h
```diff
@@ -75,6 +75,29 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
...
     };
 
+    class BuildSession final : public FlattenBuildSession {
```
LHT129, line 78: [note] The `BuildSession` stores a raw `owner_` pointer to the `RaBitQSplitDataCell`. This is safe in practice because the session is always scoped within the data cell's lifetime (created via `CreateOptimizedBuildSession` and held as a local variable). However, if someone were to store the `FlattenBuildSessionPtr` beyond the data cell's lifetime, it would result in a use-after-free. Consider adding a brief comment noting this ownership invariant on `CreateOptimizedBuildSession`.

src/datacell/rabitq_split_datacell.h
```diff
@@ -75,6 +75,29 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
...
+        explicit BuildSession(RaBitQSplitDataCell* owner) : owner_(owner) {
+        }
+
+        ~BuildSession() override {
+            if (owner_ != nullptr) {
+                owner_->AbortOptimizedBuild();
+            }
+        }
+
+        void
```
LHT129, line 89: [note] In `Commit()`, `owner_->FinalizeOptimizedBuild()` is called before `owner_ = nullptr`. If `FinalizeOptimizedBuild()` throws, `owner_` remains non-null, so the `BuildSession` destructor will call `AbortOptimizedBuild()` to clean up. This is correct exception-safety behavior. However, note that `FinalizeOptimizedBuild()` may have partially written to `x_bit_layout_` and `supplement_layout_` before throwing — those partial writes are not rolled back by `AbortOptimizedBuild()`. In practice this is acceptable since the exception path indicates a fatal error, but worth being aware of.

src/datacell/flatten_interface.h
```diff
@@ -25,6 +25,7 @@
 #include "basic_types.h"
 #include "flatten_datacell_parameter.h"
 #include "flatten_interface_parameter.h"
+#include "flatten_optimized_build_interface.h"
 #include "hash_types.h"
```
LHT129, line 29: [suggestion] The three new virtual methods (`CreateOptimizedBuildSession`, `IsOptimizedBuildActive`, `FactoryComputerForBuild`) added to `FlattenInterface` duplicate the interface already defined in `FlattenOptimizedBuildInterface`. Since `RaBitQSplitDataCell` inherits from both, and `pyramid.cpp` now calls these through `FlattenInterface` (not `FlattenOptimizedBuildInterface`), the old `FlattenOptimizedBuildInterface` methods (`BeginOptimizedBuild`, `FinalizeOptimizedBuild`, `AbortOptimizedBuild`, `IsOptimizedBuildActive`, `FactoryComputerForBuild`) are now only used internally by `RaBitQSplitDataCell`. Consider whether `FlattenOptimizedBuildInterface` should be deprecated or whether the new session API should fully replace it in a follow-up PR.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -593,6 +593,10 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
         }
     }
```
LHT129, line 595: [note] In `Pyramid::Add()`, the `codes` variable is obtained via `construction_codes()` but is only used to create the build session. Unlike `build_by_odescent()`, the ODescent builder is not invoked here — the session is opened, data is inserted via `insert_add_batch`, and then the session is committed. The session's purpose here is to keep the scalar-code representation alive during the insert so that any distance computations during incremental graph updates use the optimized path. This is correct, but the `codes` variable name might be slightly misleading since it is not passed to any builder. Consider renaming or adding a brief comment.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2244,19 +2241,23 @@ Pyramid::add_to_path(Hierarchy& hierarchy,
...
             ++no_build_level_index;
-            continue;
+        } else {
+            add_one_point(hierarchy,
+                          node,
+                          inner_id,
+                          vector,
+                          0,
+                          false,
+                          depth == 0 ? sampled_root_level : std::numeric_limits<int>::min());
```
LHT129, line 2256: [note] The refactored `add_to_path` loop introduces a behavioral change: when `skip_level` is true and `child == nullptr`, the new code executes `break` to exit the loop, whereas the old code executed `node = child` (setting node to nullptr) followed by `continue`, which would dereference nullptr in the next iteration.

Old code path for skip_level:
```cpp
node = child;           // could be nullptr
++no_build_level_index;
continue;               // next iteration dereferences nullptr → UB
```

New code path for skip_level:
```cpp
++no_build_level_index;
if (child == nullptr) {
    break;              // cleanly exits the loop
}
node = child;
```

The new behavior is safer (prevents UB) but also changes the loop termination condition. In the old code, a nullptr child on a skip level would crash; in the new code, it silently exits early. If `GetChild(path, true)` can never return nullptr (since the second argument requests creation), then this is purely a safety improvement with no functional difference. Please confirm this assumption holds for all callers.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2244,19 +2241,23 @@ Pyramid::add_to_path(Hierarchy& hierarchy,
...
+        }
+        if (child == nullptr) {
+            break;
```
LHT129, line 2259: [note] The `add_to_path` refactoring introduces a `if (child == nullptr) break;` guard that was absent in the original code. In the old version, when `skip_level` was true on the last iteration (where `depth == path_slices.size()` and `child` is `nullptr`), `node = child` would set `node` to null, and the next loop iteration would crash on `node->GetChild(...)`. This change is actually a bug fix for a latent null-pointer dereference, not a behavioral regression.

Suggested: No action needed — this is a correct fix.

src/datacell/rabitq_split_datacell.h
```diff
@@ -490,6 +521,23 @@ class RaBitQSplitDataCell : public FlattenInterface, public FlattenOptimizedBuil
             build_codes->Resize(this->max_capacity_);
             code_sums->resize(this->max_capacity_, 0);
         }
+
+        // A session may also be used for a non-empty incremental build. Rehydrate the scalar
+        // representation from persisted split codes so old and new IDs share exactly the same
+        // pairwise-distance path without requiring raw vectors.
```
LHT129, line 527: [note] The rehydration loop in `BeginOptimizedBuild` iterates over all `total_count_` IDs one by one, calling `GetCodesById` and `UnpackScalarCode` for each. For large indexes (e.g. millions of vectors) during incremental `Add`, this O(N) scan could add noticeable latency to session creation. This is consistent with the existing per-ID iteration pattern in `FinalizeOptimizedBuild`, and incremental adds typically involve smaller datasets, so this is acceptable for now. If large-scale incremental builds become common, consider batching the I/O reads.

LHT129 (commented):
I have completed a thorough review of this PR. The existing inline comments (8 total) already cover the notable observations. Here is a summary:

**Architecture**: The RAII `FlattenBuildSession` / `BuildSession` pattern is a clean improvement over the previous raw `BeginOptimizedBuild`/`FinalizeOptimizedBuild`/`AbortOptimizedBuild` lifecycle. The session API exposed through `FlattenInterface` (with safe defaults returning `nullptr`) lets callers like Pyramid opt in without `dynamic_pointer_cast`.

**Correctness**: The `add_to_path` refactoring fixes a latent null-pointer dereference — when `GetChild` returns `nullptr` on a skip-level iteration, the old code would crash on the next loop iteration. The new `break` guard is correct.

**Exception safety**: `BuildSession::Commit()` sets `owner_ = nullptr` only after `FinalizeOptimizedBuild()` succeeds. If `FinalizeOptimizedBuild()` throws, the destructor calls `AbortOptimizedBuild()`. Callers in `pyramid.cpp` rely on stack unwinding to trigger the destructor, which is correct RAII.

**Performance**: The 1.62x build speedup is achieved by eliminating the temporary SQ8 datacell that ODescent previously created, instead reusing the RaBitQ split scalar-code representation directly. The O(N) rehydration loop in `BeginOptimizedBuild` for incremental adds is acceptable given the typical scale of incremental operations.

**Test coverage**: The updated test covers the new session API, the abort-via-destructor path, and incremental build with mixed old/new codes. The existing "enqueue failure" test still uses the internal `FlattenOptimizedBuildInterface` directly, which is fine since that interface remains the internal implementation.

No blocking issues found.

LHT129 (commented):
This is a well-structured PR that delivers a clean 1.62x build speedup by replacing the ODescent-specific SQ8 build workaround with a general-purpose RAII session API on `FlattenInterface`.

**Summary of changes:**

1. **New `FlattenBuildSession` RAII class** (`flatten_optimized_build_interface.h`): A `unique_ptr`-managed session that calls `AbortOptimizedBuild()` on destruction unless `Commit()` succeeds. This is the right abstraction for scoped build optimizations.

2. **`FlattenInterface` gains three virtual methods** (`CreateOptimizedBuildSession`, `IsOptimizedBuildActive`, `FactoryComputerForBuild`) with sensible defaults (return nullptr / false / delegate to `FactoryComputer`). This lets callers like Pyramid use the optimized path through the base interface without `dynamic_pointer_cast`.

3. **`RaBitQSplitDataCell` implements the session API** via an inner `BuildSession` class and a new `CreateOptimizedBuildSession` override. The `BeginOptimizedBuild` method now also rehydrates scalar codes from persisted split codes for incremental builds — a key correctness feature verified by the new test case.

4. **Pyramid uses the session in both `build_by_odescent()` and `Add()`**: The session is opened before data insertion and committed afterward, keeping the scalar-code representation alive for all distance computations during graph construction.

5. **ODescent is simplified**: The `CreateBuildFlatten` static method, the `build_flatten_interface_` member, and the `build_vectors_`/`build_vector_count_` fields are all removed. `prepare_build_flatten()` now simply points `distance_flatten_interface_` at the main flatten interface.

6. **`add_to_path` refactoring**: The loop logic is restructured with an explicit `skip_level` boolean and a `child == nullptr` guard, which also fixes a latent null-pointer dereference on the skip-level path.

7. **`UnpackScalarCode` visibility**: Moved from private to public in `RaBitQuantizer` so `RaBitQSplitDataCell` can call it during rehydration.

**Test coverage:** The existing test is updated to use the new session API, and a new incremental-build sub-test verifies that rehydrated old codes and newly encoded codes produce consistent distances after commit.

The existing inline comments from the previous review round cover the notable design points (raw pointer lifetime, exception safety, interface duplication, rehydration performance). No new blocking issues found.

src/quantization/rabitq_quantization/rabitq_quantizer.h
```diff
@@ -203,6 +203,10 @@ class RaBitQuantizer : public Quantizer<RaBitQuantizer<metric>> {
                               uint8_t* one_bit_code,
                               uint8_t* supplement_code) const;
 
+    // Restores the scalar build representation from a persisted full code.
+    [[nodiscard]] uint64_t
```
LHT129, line 207: [note] `UnpackScalarCode` is moved from `private` to `public` in `RaBitQuantizer` so that `RaBitQSplitDataCell::BeginOptimizedBuild` can call it during the rehydration loop. This exposes an internal implementation detail that was previously only used by `EncodeOneInternal`. The method is `const` and does not mutate quantizer state, so the exposure is safe, but it does expand the public API surface. Consider whether this method should instead be exposed through a friend declaration or a more restricted access mechanism if the quantizer's public API is meant to remain minimal.

Current:
```cpp
public:
    [[nodiscard]] uint64_t
    UnpackScalarCode(const uint8_t* codes, uint8_t* scalar_code) const;
```

Alternative: keep it private and declare `RaBitQSplitDataCell` as a friend, or expose it through `BottomQuantizerAccessor` which already mediates access to the bottom quantizer.

PR #2864, opened 2026-09-05
feat(eval): report query-level recall coverage

### LINKED ISSUE #2841: [feat](eval_performance): report query-level recall coverage
### Affected area

eval_performance

### Motivation

`eval_performance` currently records per-query recall internally and reports average recall plus selected recall percentiles. It also reports latency, QPS, and aggregate/average search statistics such as `dist_cmp` and `hops`.

Average recall does not express the quality contract needed by adaptive search. For example, the same average can describe a workload where nearly every query reaches a target or one where easy queries compensate for a substantial failing tail. Recall percentiles help inspect the distribution, but they do not directly report the fraction of queries that individually reach an arbitrary declared recall target.

The tool should report query-level recall coverage so fixed and adaptive policies can be compared under the same explicit quality constraint.

### Proposed solution

Add an optional recall target to search evaluation. A proposed YAML/CLI name is `recall_target`, constrained to `[0, 1]`.

For `N` measured queries, report:

```text
query_coverage(R) = count(q where recall@k(q) >= R) / N
```

When recall is represented as an integer number of correct results over `k`, this is equivalent to requiring at least `ceil(R * k)` correct results per query.

Reuse the per-query values already produced by `RecallMonitor`, preserving the existing distance-based or ID-based recall semantics selected by the evaluation configuration. A proposed result shape is:

```json
{
  "query_coverage": {
    "recall_target": 0.9,
    "reached_queries": 950,
    "query_count": 1000,
    "rate": 0.95
  }
}
```

When no target is configured, keep the current output unchanged. Add focused coverage for:

- a workload whose average recall and target coverage lead to different conclusions;
- exact boundary behavior at `ceil(R * k)`;
- both configured recall modes where their existing semantics differ;
- repeated/concurrent query execution and empty/invalid configuration handling;
- CLI/YAML parsing and user documentation.

### Alternatives considered

- **Infer coverage from the existing recall percentiles.** The current fixed percentile set cannot recover exact coverage for an arbitrary target, and it discards the success count needed for statistical calibration.
- **Export all per-query recall values and compute coverage externally.** Raw export can still be useful, but every controller comparison would then need a separate implementation of a core quality metric.
- **Continue selecting by average recall.** This allows easy-query overperformance to hide queries below the declared target.

additive only (no break)

### References

- `tools/eval/eval_config.h:42-59` — current search and metric configuration.
- `tools/eval/case/search_eval_case.cpp:142-152` — current recall monitor registration.
- `tools/eval/monitor/recall_monitor.cpp:78-157` — per-query recall recording and current average/percentile outputs.
- `tools/eval/case/search_eval_case.cpp:434-564` — existing aggregate and average search-work statistics.
- https://vsag.io/docs/en/resources/eval.html

### Additional context

Part of #2840.

This Issue adds a supplementary metric; it does not replace the OSPP project's calibrated average-recall comparison, select a controller, or claim that any dynamic policy meets a target. Confidence bounds or calibration procedures can consume `reached_queries` and `query_count` later without changing the definition of the reported empirical coverage.

## Related issues

_None found._

---
_Drafted with AI assistant: Codex:gpt-5.6-sol_

Utopia-V:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2841

## What Changed

- Add an optional `recall_target` setting for CLI and YAML search evaluation, validated as a finite value in `[0, 1]` for `knn` and `knn_filter` searches.
- Report `query_coverage` in JSON with the target, reached query executions, recorded query executions, and empirical coverage rate.
- Reuse the existing distance-based or ID-based per-query recall semantics, including the discrete `recall@k` boundary.
- Keep existing output unchanged when no target is configured.
- Add focused configuration, semantic, compatibility, repeated-query, and multi-threaded integration coverage.
- Document the option and output in the English and Chinese eval guides and YAML template.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details (Debug build):

```text
cmake --build build --target eval_dataset_test eval_performance --parallel 8
  passed

./build/tools/eval/eval_dataset_test --rng-seed 12345
  All tests passed (165 assertions in 18 test cases)

clang-format 15 --dry-run --Werror on the modified C++ files
  passed

git diff --check
  passed
```

Configuration tests exercise CLI/YAML values at search construction; the in-memory integration test also rejects an invalid target without going through a config loader. Full repository tests, coverage collection, and sanitizers were not run locally for this revision.

## Compatibility Impact

- API/ABI compatibility: No installed VSAG API or ABI change; the affected types are internal to `eval_performance`.
- Behavior changes: Adds opt-in CLI/YAML configuration and an additive JSON object. Without `recall_target`, existing output is unchanged. Configuring the target for `range` or `range_filter` is rejected because those modes do not currently produce per-query `recall@k`.

- Performance impact: Production index search paths are unchanged. Coverage adds one comparison and conditional increment to the existing recall pass; configuring only coverage activates the recall pass needed to compute it. Recall processing remains outside latency/QPS measurement.
- Concurrency/thread-safety impact: The reached-query counter is updated under the existing recall-record mutex. Integration coverage includes repeated queries executed with multiple OpenMP workers.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/resources/eval.md` and `tools/eval/eval_template.yaml`

## Risk and Rollback

- Risk level: low
- Rollback plan: Revert this PR to remove the option, JSON metric, tests, and documentation together.

## Checklist

- [x] I have updated docs for the behavior change

### DIFF
--- a/tools/eval/case/search_eval_case.cpp
+++ b/tools/eval/case/search_eval_case.cpp
@@ -111,6 +111,7 @@ SearchEvalCase::SearchEvalCase(const std::string& dataset_path,
     } else if (search_mode == "range_filter") {
         this->search_type_ = SearchType::RANGE_FILTER;
     }
+    config_.Validate();
     this->init_monitor();
 }
 
@@ -140,9 +141,12 @@ SearchEvalCase::init_latency_monitor() {
 
 void
 SearchEvalCase::init_recall_monitor() {
-    if (config_.enable_recall or config_.enable_percent_recall) {
-        auto recall_monitor = std::make_shared<RecallMonitor>(
-            this->dataset_ptr_->GetNumberOfQuery(), config_.use_id_based_recall);
+    if (config_.enable_recall or config_.enable_percent_recall or
+        config_.recall_target.has_value()) {
+        auto recall_monitor =
+            std::make_shared<RecallMonitor>(this->dataset_ptr_->GetNumberOfQuery(),
+                                            config_.use_id_based_recall,
+                                            config_.recall_target);
         if (config_.enable_recall) {
             recall_monitor->SetMetrics("avg_recall");
         }
--- a/tools/eval/eval_config.cpp
+++ b/tools/eval/eval_config.cpp
@@ -15,6 +15,9 @@
 
 #include "./eval_config.h"
 
+#include <cmath>
+#include <stdexcept>
+
 #include "./common.h"
 
 namespace vsag::eval {
@@ -26,6 +29,72 @@ check_and_get_value(const YAML::Node& node, const std::string& key, T& value) {
         value = node[key].as<T>();
     }
 };
+
+void
+EvalConfig::AddCliArguments(argparse::ArgumentParser& parser) {
+    // index
+    parser.add_argument<std::string>("--datapath", "-d")
+        .required()
+        .help("The hdf5 file path for eval");
+    parser.add_argument<std::string>("--type", "-t")
+        .required()
+        .choices("build", "search")
+        .help(R"(The eval method to select, choose from {"build", "search"})");
+    parser.add_argument<std::string>("--index_name", "-n")
+        .required()
+        .help("The name of index for create index");
+    parser.add_argument<std::string>("--create_params", "-c")
+        .required()
+        .help("The param for create index");
+    parser.add_argument<std::string>("--index_path", "-i")
+        .default_value("/tmp/performance/index")
+        .help("The index path for load or save");
+    parser.add_argument<std::string>("--search_params", "-s")
+        .default_value("")
+        .help("The param for search");
+    parser.add_argument<std::string>("--search_mode")
+        .default_value("knn")
+        .choices("knn", "range", "knn_filter", "range_filter")
+        .help(
+            "The mode supported while use 'search' type,"
+            " choose from {\"knn\", \"range\", \"knn_filter\", \"range_filter\"}");
+    parser.add_argument("--delete-index-after-search")
+        .default_value(false)
+        .help("Delete index after search");
+    parser.add_argument("--topk")
+        .default_value(10)
+        .help("The topk value for knn search or knn_filter search")
+        .scan<'i', int>();
+    parser.add_argument("--range")
+        .default_value(0.5F)
+        .help("The range value for range search or range_filter search")
+        .scan<'f', float>();
+    parser.add_argument("--search-query-count")
+        .default_value(100000)
+        .help("The number of queries to run for search performance evaluation")
+        .scan<'i', uint64_t>();
+
+    // metrics
+    parser.add_argument("--disable_recall")
+        .default_value(false)
+        .help("Disable average recall eval");
+    parser.add_argument("--disable_percent_recall")
+        .default_value(false)
+        .help("Disable percent recall eval, include p0, p10, p30, p50, p70, p90");
+    parser.add_argument("--recall_target")
+        .help(
+            "Report query coverage at this recall target in [0, 1] for knn and knn_filter searches")
+        .scan<'g', double>();
+    parser.add_argument("--disable_qps").default_value(false).help("Disable qps eval");
+    parser.add_argument("--disable_tps").default_value(false).help("Disable tps eval");
+    parser.add_argument("--disable_memory").default_value(false).help("Disable memory eval");
+    parser.add_argument("--disable_latency")
+        .default_value(false)
+        .help("Disable average latency eval");
+    parser.add_argument("--disable_percent_latency")
+        .default_value(false)
+        .help("Disable percent latency eval, include p50, p80, p90, p95, p99");
+}
 
 EvalConfig
 EvalConfig::Load(argparse::ArgumentParser& parser) {
@@ -42,6 +111,9 @@ EvalConfig::Load(argparse::ArgumentParser& parser) {
     config.top_k = parser.get<int>("--topk");
     config.radius = parser.get<float>("--range");
     config.search_query_count = parser.get<uint64_t>("--search-query-count");
+    if (parser.is_used("--recall_target")) {
+        config.recall_target = parser.get<double>("--recall_target");
+    }
 
     config.delete_index_after_search = parser.get<bool>("--delete-index-after-search");
 
@@ -67,6 +139,7 @@ EvalConfig::Load(argparse::ArgumentParser& parser) {
         config.enable_percent_latency = false;
     }
 
+    config.Validate();
     return config;
 }
 
@@ -94,6 +167,9 @@ EvalConfig::Load(YAML::Node& yaml_node, const EvalJob& global_options) {
     check_and_get_value<int>(yaml_node, "topk", config.top_k);
     check_and_get_value<float>(yaml_node, "range", config.radius);
     check_and_get_value<uint64_t>(yaml_node, "search_query_count", config.search_query_count);
+    if (yaml_node["recall_target"].IsDefined()) {
+        config.recall_target = yaml_node["recall_target"].as<double>();
+    }
 
     check_and_get_value<bool>(
         yaml_node, "delete_index_after_search", config.delete_index_after_search);
@@ -143,6 +219,7 @@ EvalConfig::Load(YAML::Node& yaml_node, const EvalJob& global_options) {
         disable = false;
     }
 
+    config.Validate();
     return config;
 }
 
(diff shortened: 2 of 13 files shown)


tools/eval/case/search_eval_case.cpp
```diff
@@ -111,6 +111,7 @@ SearchEvalCase::SearchEvalCase(const std::string& dataset_path,
     } else if (search_mode == "range_filter") {
         this->search_type_ = SearchType::RANGE_FILTER;
     }
+    config_.Validate();
     this->init_monitor();
```
LHT129, line 115: [suggestion] `Validate()` is called both in `SearchEvalCase` constructor (`search_eval_case.cpp:115`) and in `validate_search()` (`evaluator.cpp:54`). For the `EvaluateSearch` in-memory path, this results in a redundant double validation. Consider removing the call from `SearchEvalCase` constructor since `validate_search` already covers it for the in-memory path, or remove it from `validate_search` and ensure the file-based path also calls `Validate()`.

The current code is not incorrect (Validate is idempotent), but the double call is unnecessary and could confuse future readers.

tools/eval/monitor/recall_monitor.cpp
```diff
@@ -61,8 +63,17 @@ get_id_recall(const int64_t* neighbors,
...
+    : Monitor("recall_monitor"),
+      use_id_based_recall_(use_id_based_recall),
+      recall_target_(recall_target) {
+    if (recall_target_.has_value() &&
+        (not std::isfinite(recall_target_.value()) || recall_target_.value() < 0.0 ||
+         recall_target_.value() > 1.0)) {
+        throw std::invalid_argument("recall_target must be finite and in [0, 1]");
+    }
     if (max_record_counts > 0) {
         this->recall_records_.reserve(max_record_counts);
```
LHT129, line 71: [suggestion] The `recall_target` range validation (`std::isfinite`, `[0, 1]`) is duplicated in both `RecallMonitor` constructor (`recall_monitor.cpp:78-82`) and `EvalConfig::Validate()` (`eval_config.cpp:249-255`). If the validation rules change in the future, both locations need to be updated. Consider keeping the validation only in `EvalConfig::Validate()` and removing it from the `RecallMonitor` constructor, since `EvalConfig::Validate()` is always called before a `RecallMonitor` is constructed with the target value.

tools/eval/case/search_eval_case.cpp
Utopia-V: Addressed in a53f27f5. I removed `config.Validate()` from `validate_search()`. `SearchEvalCase` remains the common validation point for file-based and direct factory construction, while `EvaluateSearch` keeps only its in-memory-specific checks.

tools/eval/monitor/recall_monitor.cpp
Utopia-V: Addressed in a53f27f5. I removed the duplicate numeric validation and its now-unused includes from `RecallMonitor`; current production construction reaches `EvalConfig::Validate()` through `SearchEvalCase`.

tools/eval/monitor/recall_monitor.cpp
```diff
@@ -155,4 +174,12 @@ RecallMonitor::cal_recall_rate(double rate) {
...
     return recall_records_[pos];
 }
+
+void
+RecallMonitor::record_recall(double recall) {
+    recall_records_.emplace_back(recall);
+    if (recall_target_.has_value() && recall >= recall_target_.value()) {
+        ++reached_queries_;
+    }
+}
```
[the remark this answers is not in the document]
Utopia-V: I am keeping the direct comparison intentionally. Coverage is defined over the existing per-query recall value, which is already a `double`, against the configured `double` target. Adding an epsilon would change the contract by allowing recall values below the declared target to pass. A naive `ceil(target * k)` also has its own decimal-multiplication boundary problem and would require changing the existing recall interface to retain hit counts. The focused test covers the inclusive discrete boundary (`4/5` against `0.8`). If a tolerance is desired, it should be introduced as an explicit metric-semantic decision rather than an implementation hardening.

LHT129 (commented):
Reviewed the latest commit a53f27f5. The previous review comments from LHT129 and Copilot have been addressed:

- Duplicate `recall_target` validation removed from `RecallMonitor` constructor — validation now lives solely in `EvalConfig::Validate()`, called through `SearchEvalCase` constructor.
- `config.Validate()` removed from `validate_search()` in `evaluator.cpp` — the `SearchEvalCase` constructor remains the single validation entry point.
- The floating-point comparison discussion (epsilon vs. direct `>=`) has been resolved by the author with a reasoned explanation that the discrete boundary test (`4/5` against `0.8`) covers the intended semantics.

The code is well-structured, tests are comprehensive (CLI parsing, YAML parsing, mode rejection, boundary semantics, distance-based vs ID-based recall, unconfigured path), and the feature is properly documented in both English and Chinese docs. No further issues to raise.

tools/eval/eval_config.cpp
```diff
@@ -168,4 +246,19 @@ EvalConfig::CheckKeyAndType(YAML::Node& yaml_node) {
     check_and_get_value<bool>(yaml_node, "disable_percent_latency");
 }
 
+void
+EvalConfig::Validate() const {
+    if (not recall_target.has_value()) {
```
LHT129: [suggestion] `EvalConfig::Validate()` accepts `recall_target` for `--type build` runs because it only checks `search_mode` (which defaults to `"knn"`), not `action_type`. When a user configures `--type build --recall_target 0.9`, validation passes silently but the target is never used — the build path never instantiates a `RecallMonitor`. This contradicts the documented contract that `recall_target` is a search-evaluation option.

Consider adding an `action_type` guard:

```cpp
if (action_type != "search") {
    throw std::invalid_argument(
        "recall_target is supported only for search evaluations");
}
```

This would catch the misconfiguration at parse time rather than silently ignoring it.

tools/eval/eval_config.cpp
```diff
@@ -67,6 +139,7 @@ EvalConfig::Load(argparse::ArgumentParser& parser) {
         config.enable_percent_latency = false;
     }
```
LHT129, line 140: [note] `Validate()` is now called in both `EvalConfig::Load()` (both CLI and YAML paths in `eval_config.cpp:140,222`) and `SearchEvalCase` constructor (`search_eval_case.cpp:115`). The previous review cycle removed the call from `validate_search()` to avoid a double call for the in-memory path, but the double call still exists — just shifted to `Load()` + `SearchEvalCase()`. Since `Validate()` is idempotent this is harmless, but it may confuse future readers who see validation happening twice. Consider keeping `Validate()` in only one place: either in `EvalConfig::Load()` (gate at config construction) or in `SearchEvalCase` (gate at usage), but not both.

tools/eval/query_coverage_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <catch2/catch_approx.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_string.hpp>
+#include <string>
```
LHT129, line 18: [note] The include `<catch2/matchers/catch_matchers_string.hpp>` is not used in this test file — no `Catch::Matchers::*` string matchers appear in any assertion. Removing it would keep the include list minimal.

tools/eval/monitor/recall_monitor.cpp
```diff
@@ -81,6 +85,16 @@ RecallMonitor::GetResult() {
     for (auto& metric : metrics_) {
         this->cal_and_set_result(metric, result);
     }
+    if (recall_target_.has_value()) {
```
LHT129, line 88: [note] `GetResult()` reads `reached_queries_` and `recall_records_.size()` without holding `record_mutex_`. In the current usage pattern, `GetResult()` is only called after `Stop()` (i.e., after all `Record()` calls complete), so there is no data race. However, if `GetResult()` were ever called concurrently with `Record()` (e.g., from a monitoring/metrics polling thread), this would be a data race on `reached_queries_` and `recall_records_`. Consider documenting this invariant or adding a lock in `GetResult()` for defense-in-depth.

tools/eval/monitor/recall_monitor.h
```diff
@@ -16,13 +16,16 @@
 #pragma once
 
 #include <chrono>
```
LHT129, line 18: [note] The `#include <chrono>` added in this header is unused — no `std::chrono` types or functions appear in `recall_monitor.h`. The include appears to be a leftover from a previous iteration that had validation logic in the constructor. Removing it would keep the header minimal.

PR #2882, opened 2026-09-08
docs: fix site links and stale guidance

wxyucs:
## Summary

- fix mdBook overview links, API anchors, and stale repository paths across English and Chinese docs
- align per-search allocator guidance and the HGraph allocator example with current ownership behavior
- clarify the historical roadmap and provide useful links from the English blog landing page
- add a dependency-free documentation validator and PR workflow to prevent regressions

## Validation

- `node .github/scripts/check-docs.mjs`
- `node --check .github/scripts/check-docs.mjs`
- parsed `.github/workflows/docs-check.yml` with Ruby YAML
- built English and Chinese docs and blogs with mdBook 0.4.51 (172 HTML pages)
- `git diff --check origin/main...HEAD`

`clang-format-15` is not installed in the local environment. The C++ example change only removes manual deallocation calls and adds comments; it does not alter code layout.


.github/scripts/check-docs.mjs
```diff
@@ -0,0 +1,209 @@
...
+        return { line: fenceMarker === null ? line : '', lineNumber: index + 1 };
+    });
+}
+
+function headingSlug(text) {
+    return text
+        .replace(/<[^>]+>/g, '')
+        .replace(/!?\[([^\]]*)\]\([^)]*\)/g, '$1')
+        .replace(/`([^`]*)`/g, '$1')
+        .trim()
```
LHT129, line 48: [suggestion] The `headingSlug` function uses `toLowerCase()` which is a no-op for CJK characters, but the regex `[^\p{Letter}\p{Number}_-]` will strip characters like `·` (middle dot, U+00B7) and other punctuation that mdBook preserves in slugs. mdBook's slug algorithm also collapses consecutive hyphens and strips leading/trailing hyphens, which this implementation does not do.

For example, a heading like `## Foo -- Bar` would produce slug `foo----bar` here, while mdBook produces `foo--bar`. This could cause false negatives in anchor validation.

.github/scripts/check-docs.mjs
```diff
@@ -0,0 +1,209 @@
...
+                            `${file}:${lineNumber}: inline README.md link renders as missing ` +
+                                'README.html; link to the directory instead'
+                        );
+                    }
+                    const target = resolveMarkdownTarget(file, filePart);
+                    if (!existsSync(target)) {
+                        errors.push(`${file}:${lineNumber}: missing local link target ${rawTarget}`);
+                        continue;
+                    }
+                    if (anchor === '' || !target.endsWith('.md')) {
```
LHT129, line 120: [suggestion] The `README.md` link check at line 120 flags `filePart.endsWith('README.md')` for all non-SUMMARY.md files. However, `SUMMARY.md` itself may legitimately reference `README.md` files (e.g., `[Overview](README.md)` in a subdirectory listing). The current check exempts `SUMMARY.md` but there may be other valid cases — for instance, a `README.md` in the repo root that references `docs/docs/en/src/README.md` as a relative link. Consider whether this check should be scoped more narrowly or made into a warning rather than an error.

.github/scripts/check-docs.mjs
```diff
@@ -0,0 +1,209 @@
...
+    return walk(directory).filter(file => file.endsWith('.md'));
+}
+
+function markdownLinesOutsideFences(file) {
+    const lines = readFileSync(file, 'utf8').split(/\r?\n/);
+    let fenceMarker = null;
+
+    return lines.map((line, index) => {
+        const fence = line.match(/^\s*(`{3,}|~{3,})/);
+        if (fence !== null) {
```
LHT129, line 31: [suggestion] The fence detection in `markdownLinesOutsideFences` matches only the first character of the fence marker (`fence[1][0]`). This means a block opened with ```` ``` ```` (3 backticks) could be incorrectly closed by ```` ~~~~ ```` (3+ tildes) since both have length 3 and the function only checks the character type, not the full marker. While uncommon in practice, the GitHub Flavored Markdown spec requires the closing fence to use the same character and at least as many characters as the opening fence. Consider tracking the full fence string for stricter matching.

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,36 @@
+name: Docs Check
+
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
```
LHT129, line 6: [note] The workflow triggers on changes to `examples/**`, `include/**`, `src/**`, and `tests/**` in addition to `docs/**`. This is because the `checkReferencedRepositoryPaths` function validates backticked file paths like `` `src/algorithm/hgraph/hgraph.cpp` `` that appear in documentation. However, this means the docs check will run on every C++ source change, which may be noisy. Consider whether the path trigger should be narrowed to only `docs/**` plus the checker script and workflow files themselves, since source-path references in docs change infrequently and are already validated at doc-change time.

docs/docs/en/src/advanced/search_allocator.md
```diff
@@ -117,15 +112,16 @@ arena.reset();              // drops every per-query buffer at once
```
LHT129, line 112: [note] The updated allocator table drops the explicit rows for `KnnSearch(query, k, parameters_str)` and `RangeSearch(...)` that were present in the previous version. While the new text above the table mentions that "dedicated `RangeSearch` overloads do not accept an allocator" and the `KnnSearch(query, k, parameters_str)` row was arguably redundant (it had no per-search allocator hook), removing these rows makes the table less self-contained. A reader scanning only the table may miss that `KnnSearch(query, k, parameters_str)` uses the index allocator. Consider keeping these rows for completeness or adding a footnote.

docs/README.md
```diff
@@ -46,3 +46,15 @@ published documentation.
...
 The remaining top-level markdown files in `docs/` are design notes that have not been fully
 merged into the website source yet. Keep new user-facing documentation in `docs/docs/{en,zh}/src/`.
+
+## Validation
+
+Before submitting documentation changes, run:
+
+```bash
+node .github/scripts/check-docs.mjs
+```
```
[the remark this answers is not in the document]
wxyucs: Thanks. This is a non-blocking command-invocation documentation clarification. Deferring the wording-only change and resolving this thread without code changes or another CI run.

LHT129 (approved):
LGTM

Issue #2884, opened 2026-09-08
[improve](testing): restore trustworthy C++ coverage and ratchet main to 90%


wxyucs:
### Kind of improvement

test

### Affected module

not applicable (cross-cutting C++ coverage)

### Affected area

area/testing, area/ci

### Summary

At [`9f35d891`](https://github.com/antgroup/vsag/commit/9f35d891233ab3cfc129f853de8c0305d426ecb6), Codecov reports **85.61% line coverage**: 36,012 covered of 42,061 measured lines, with 6,049 misses in 437 files. The project status fails its 90% target, while patch coverage passes at 88.37% (152/172 lines). With the current denominator, reaching 90% requires 1,843 more covered lines.

This commit is not the regression: its parent was 85.54% (35,932/42,003), so the factory-sharding change improved project coverage by 0.07 percentage points and added focused tests. The low result is cumulative.

The current misses are concentrated in maintained code:

| Area | Measured | Missed | Coverage |
| --- | ---: | ---: | ---: |
| `src/algorithm` | 19,017 | 3,161 | 83.38% |
| `src/datacell` | 7,194 | 958 | 86.68% |
| `src/impl` | 4,476 | 724 | 83.82% |
| MCI-related files plus `clique_datacell.cpp` | 3,454 | 1,265 | 63.38% |
| RaBitQ-named files | 4,498 | 757 | 83.17% |
| HGraph/IVF serialization | 1,195 | 339 | 71.63% |
| SIMQ-named files | 1,126 | 292 | 74.07% |

[...]

PR #2887, opened 2026-09-08
ci: make C++ coverage repository-scoped

wxyucs:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [x] CI/Build/Infra

## Issue #2884 phases

Refs: #2884

Implemented in this first phase:

- Phase 1 measurement integrity: capture only tracked maintained `src/` and `include/` sources, narrowly exclude generated `src/version.h` and vendored `expected.hpp`, normalize repository-relative paths, retain branch records through every LCOV stage and the upload input, and reject empty, branchless, absolute, untracked, or out-of-scope traces.
- Phase 1 instrumentation and reproducibility: attach coverage privately to analyzer, BLAS, cluster, reorder, and SIMD production targets; link the coverage runtime explicitly at final test executables; verify every compiled production translation unit; and use/report seed `424242` in coverage CI.
- Phase 2 ownership: assign every maintained reported source to exactly one Codecov component and disable server-side default path rewriting so local and uploaded source populations stay aligned.
- Phase 3 initial signal: add a lightweight PR configuration check using the existing change detector. It validates LCOV, CMake target scope, component ownership, workflow, and seed semantics without running `make cov` or duplicating the multi-hour suite.

Deferred #2884 phases:

- A real pre-merge patch/project coverage upload and regression gate. This needs a resource-aware suite design; this PR does not depend on still-open PR #2856.
- MCI legacy enumerator removal or active-path coverage. The no-argument path is internal and unused by current production call sites, but removing the mechanically imported implementation belongs in a separate algorithm-focused change.
- Maintained algorithm/error-path coverage additions, three consecutive full runs at 90%, branch ratcheting, and platform-specific coverage jobs for conditional backends.

## Validation

- `python3 -m unittest tests/scripts/coverage_diagnostics_test.py`: 15 tests passed.
- `make test-cmake`: passed all six helper suites.
- Codecov API configuration validation: valid.
- `make cov`: passed; the compile database verified 243/243 maintained production commands carry `--coverage`.
- Focused seeded coverage run for allocator, SIMD, BLAS, cluster, reorder, and analyzer paths: 85 test cases and 30,604,097 assertions passed.
- Repository-scoped LCOV collection: 495 tracked source files, 52,040 measured lines, and 80,305 branch records. The focused shard covered 10,698 lines and 7,599 branches; this is evidence for collection semantics, not a full-suite coverage claim.
- Separate non-coverage build of all newly instrumented targets: passed, with no `--coverage` flags.
- Workflow YAML parse, shell syntax checks, diff checks, and public-text hygiene checks: passed.
- Exact clang-format/tidy 15 were unavailable locally; CI is authoritative for those checks.

## Compatibility and risk

- Public API/ABI: unchanged.
- Normal builds: unchanged because `coverage_config` remains empty unless `ENABLE_COVERAGE=ON`.
- Runtime performance: unchanged outside coverage builds.
- Rollback: revert the single commit to restore the previous collection and CI configuration.

### DIFF
--- a/docs/agents/build-and-test.md
+++ b/docs/agents/build-and-test.md
@@ -58,9 +58,9 @@ versions for consistent formatting and diagnostics. CI will fail otherwise.
   `make test` builds and runs the unit + functional suite but does **not**
   enable coverage instrumentation. To produce the coverage report locally use
   the `make cov` flow (which configures the build with `ENABLE_COVERAGE=ON`)
-  followed by running the test binaries and
-  `scripts/coverage/collect_cpp_coverage.sh`; the equivalent CI coverage
-  workflow enforces the 90% threshold. This threshold is based on the
-  unit-test suite; functional tests under `tests/` and Python code are not
-  currently included in the coverage metric unless explicitly documented
-  otherwise.
+  followed by `VSAG_TEST_SEED=424242 bash scripts/testing/test_parallel_bg.sh`
+  and `bash scripts/coverage/collect_cpp_coverage.sh`. The C++ report combines the
+  non-daily unit and functional suites, but measures only maintained production
+  sources under `src/` and `include/`; it excludes the generated
+  `src/version.h` and vendored `include/vsag/expected.hpp`. Python sources and
+  platform paths not compiled by the coverage job are not part of this metric.
--- a/scripts/coverage/collect_cpp_coverage.sh
+++ b/scripts/coverage/collect_cpp_coverage.sh
@@ -7,41 +7,52 @@ SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="$( cd "${SCRIPTS_DIR}/../../" && pwd )"
 
 COVERAGE_DIR="${ROOT_DIR}/coverage"
+COVERAGE_RAW_FILE="${COVERAGE_DIR}/coverage.raw.info"
+COVERAGE_SCOPED_FILE="${COVERAGE_DIR}/coverage.scoped.info"
+COVERAGE_FILE="${COVERAGE_DIR}/coverage.info"
 if [ -d "${COVERAGE_DIR}" ]; then
     rm -rf "${COVERAGE_DIR:?}"/*
 else
     mkdir -p "${COVERAGE_DIR}"
 fi
 
-# Exclude system headers before lcov runs consistency checks. The remove step below keeps the
-# same filter for compatibility with existing trace files.
+# Limit capture to repository-owned production sources before lcov runs consistency checks.
+# The later extract is a defense-in-depth check and normalizes paths for local and Codecov use.
 lcov --rc branch_coverage=1 \
      --rc geninfo_unexecuted_blocks=1 \
      --parallel 8 \
-     --directory . \
+     --directory "${ROOT_DIR}" \
      --capture \
      --exclude '/usr/*' \
-     --substitute "s#${ROOT_DIR}/##g" \
+     --include "${ROOT_DIR}/src/*" \
+     --include "${ROOT_DIR}/include/*" \
      --ignore-errors mismatch,mismatch \
      --ignore-errors count,count \
-     --output-file ${COVERAGE_DIR}/coverage.info
-lcov --remove ${COVERAGE_DIR}/coverage.info \
-     '/usr/*' \
-     'build/*' \
-     'tests/*' \
-     '*/expected.hpp' \
-     '*_test.cpp' \
-     '*/avx512.cpp' \
+     --output-file "${COVERAGE_RAW_FILE}"
+
+# These two files are not maintained production sources: version.h is generated at build time,
+# and expected.hpp is a vendored public compatibility header.
+lcov --rc branch_coverage=1 \
+     --remove "${COVERAGE_RAW_FILE}" \
+     "${ROOT_DIR}/src/version.h" \
+     "${ROOT_DIR}/include/vsag/expected.hpp" \
      --ignore-errors inconsistent,inconsistent \
      --ignore-errors unused,unused \
-     --output-file ${COVERAGE_DIR}/coverage.info
-lcov --list ${COVERAGE_DIR}/coverage.info \
-     --ignore-errors inconsistent,inconsistent,child
+     --output-file "${COVERAGE_SCOPED_FILE}"
 
-pushd "${COVERAGE_DIR}"
-coverages=$(ls coverage.info)
-if [ ! "$coverages" ];then
-    echo "no coverage file"
-    exit 0
-fi
-popd
+lcov --rc branch_coverage=1 \
+     --extract "${COVERAGE_SCOPED_FILE}" \
+     'src/*' \
+     'include/*' \
+     --substitute "s#${ROOT_DIR}/##g" \
+     --ignore-errors inconsistent,inconsistent \
+     --ignore-errors unused,unused \
+     --output-file "${COVERAGE_FILE}"
+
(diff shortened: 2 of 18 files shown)


LHT129 (commented):
I reviewed this PR thoroughly across all 18 changed files. The changes are well-structured and the PR description is comprehensive.

- **`collect_cpp_coverage.sh`**: The three-stage pipeline (capture → remove exclusions → extract with path normalization) is clean and correctly scopes coverage to maintained production sources only. The intermediate file cleanup and final trace verification via `verify_cpp_coverage.py` are good defensive measures.

- **`verify_cpp_coverage.py`**: Well-designed verification script with two modes (trace validation and instrumentation validation). The trace mode checks for empty files, absolute paths, non-production sources, untracked/generated sources, and missing branch records — all the right invariants.

- **`codecov.yml`**: The expanded component definitions provide proper ownership mapping for every source directory. The `disable_default_path_fixes: true` combined with the repository-relative path normalization in the collector ensures local and uploaded source populations stay aligned.

- **CMakeLists.txt changes**: Adding `coverage_config` as PRIVATE to analyzer, blas, cluster, reorder targets, and fixing simd from INTERFACE to PRIVATE is correct. The test executables already link `coverage_config` PRIVATE, so the coverage runtime propagates correctly through the static library chain.

- **`coverage_diagnostics_test.py`**: The expanded test suite (358 new lines) provides good coverage of the collector pipeline, CMake scope verification, verifier modes, checker semantics, parallel test runner seed propagation, Codecov component ownership, and workflow configuration checks.

- **`pr-ci.yml`**: The new `coverage-config` job is a lightweight pre-merge check that validates configuration without running the full multi-hour coverage suite — a good design choice.

No blocking issues found. This is a solid CI/infra improvement.

docs/agents/build-and-test.md
```diff
@@ -58,9 +58,9 @@ versions for consistent formatting and diagnostics. CI will fail otherwise.
...
   enable coverage instrumentation. To produce the coverage report locally use
   the `make cov` flow (which configures the build with `ENABLE_COVERAGE=ON`)
-  followed by running the test binaries and
-  `scripts/coverage/collect_cpp_coverage.sh`; the equivalent CI coverage
-  workflow enforces the 90% threshold. This threshold is based on the
-  unit-test suite; functional tests under `tests/` and Python code are not
-  currently included in the coverage metric unless explicitly documented
-  otherwise.
+  followed by `VSAG_TEST_SEED=424242 scripts/testing/test_parallel_bg.sh` and
+  `scripts/coverage/collect_cpp_coverage.sh`. The C++ report combines the
```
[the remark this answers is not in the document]
wxyucs: Good catch. I updated the agent guide to invoke both the seeded test runner and coverage collector through `bash`, matching CI and the contributor/developer guides. The focused 15-test coverage diagnostics suite still passes. Addressed in bbcb1d1c.

scripts/coverage/verify_cpp_coverage.py
```diff
@@ -0,0 +1,162 @@
...
+import sys
+
+
+PRODUCTION_SOURCE_SUFFIXES = {".c", ".cc", ".cpp", ".cxx"}
+
+
+def fail(message: str) -> None:
+    print(f"coverage verification failed: {message}", file=sys.stderr)
+    raise SystemExit(1)
+
```
LHT129, line 17: [suggestion] Using `raise SystemExit(1)` directly is non-idiomatic Python. The standard way to exit with a non-zero code is `sys.exit(1)`, which internally raises `SystemExit`. While functionally equivalent, the direct raise form is unusual and may confuse readers who expect `sys.exit()`.

Suggested change:
```python
def fail(message: str) -> None:
    print(f"coverage verification failed: {message}", file=sys.stderr)
    sys.exit(1)
```

scripts/coverage/verify_cpp_coverage.py
```diff
@@ -0,0 +1,162 @@
...
+    trace_parser = subparsers.add_parser("trace")
+    trace_parser.add_argument("trace_file", type=Path)
+    trace_parser.add_argument(
+        "--source-root", type=Path, default=Path(__file__).resolve().parents[2]
+    )
+
+    instrumentation_parser = subparsers.add_parser("instrumentation")
+    instrumentation_parser.add_argument("compile_commands", type=Path)
+    instrumentation_parser.add_argument(
+        "--source-root", type=Path, default=Path(__file__).resolve().parents[2]
```
LHT129, line 148: [note] The `"--coverage" not in arguments` check on line 148 uses a simple substring match across the entire argument list. This could produce a false positive if a source file path happens to contain the substring `--coverage` (e.g., `/path/to/--coverage-test/foo.cpp`). In practice this is unlikely with the current project structure, but a more precise check would be `"--coverage" not in arguments` changed to verify it appears as a standalone flag rather than as part of another token.

This is low-risk given the controlled build environment, but worth noting for robustness.

scripts/coverage/collect_cpp_coverage.sh
```diff
@@ -7,41 +7,52 @@ SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-if [ ! "$coverages" ];then
-    echo "no coverage file"
-    exit 0
-fi
-popd
+lcov --rc branch_coverage=1 \
+     --extract "${COVERAGE_SCOPED_FILE}" \
+     'src/*' \
+     'include/*' \
+     --substitute "s#${ROOT_DIR}/##g" \
```
LHT129, line 47: [note] The intermediate files `coverage.raw.info` and `coverage.scoped.info` are cleaned up with `rm -f` at the end of the script. If the script fails between creating these files and reaching the cleanup line (e.g., if `verify_cpp_coverage.py` exits non-zero), the intermediate files will be left on disk. This is harmless for CI (ephemeral runners) but could cause confusion during local development if a subsequent run picks up stale intermediate files.

Consider using a `trap` to clean up on exit, or moving the `rm -f` earlier (e.g., right after the final `lcov --extract` succeeds):
```bash
trap "rm -f \"${COVERAGE_RAW_FILE}\" \"${COVERAGE_SCOPED_FILE}\"" EXIT
```

tests/scripts/coverage_diagnostics_test.py
```diff
@@ -117,16 +151,336 @@ def test_excludes_system_headers_during_capture(self):
...
         self.assertIn("--remove", calls[1])
-        self.assertIn("/usr/*", calls[1])
+        self.assertTrue(any(value.endswith("/src/version.h") for value in calls[1]))
+        self.assertTrue(any(value.endswith("/include/vsag/expected.hpp") for value in calls[1]))
+        self.assertNotIn("*/avx512.cpp", calls[1])
+
+        self.assertIn("--extract", calls[2])
+        extract_index = calls[2].index("--extract")
+        self.assertEqual(calls[2][extract_index + 2 : extract_index + 4], ["src/*", "include/*"])
+        self.assertIn("--substitute", calls[2])
```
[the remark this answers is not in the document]
wxyucs: Verified against the exact LCOV v2.3 pinned by `coverage.yml`. LCOV does not use pathname-glob semantics for these filters: its [`transform_pattern()`](https://github.com/linux-test-project/lcov/blob/v2.3/lib/lcovutil.pm#L1609-L1640) converts `*` to the Perl regex `(.*)`, which matches `/`. An isolated v2.3 capture with `--include <repo>/src/*`, followed by `--extract ... src/*`, retained both `src/top.c` and `src/nested/deep.c`. The existing diagnostics test intentionally asserts these arguments, and all 15 diagnostics tests pass unchanged. Therefore recursive globs are not needed (and `**` adds no recursive semantics in lcov); no code or test change is justified.

LHT129 (approved):
LGTM

PR #2615, opened 2026-08-04
fix(hgraph): backport duplicate handling to v0.18

jac0626:
## Summary

- Validate v0.18 `LabelTable` duplicate records before publishing them during native deserialization.
- Add `max_duplicates_per_group` to standard, parallel, and iterator KNN search (`-1` unlimited, `0` no members, positive values cap accepted members per group).
- Preserve iterator duplicate members across pages and include them in final drain.
- Align entrypoint, filter-before-count, and `min_distance` behavior between BasicSearcher and ParallelSearcher.
- Wait for v0.18 parallel-search workers before their stack-owned queues and query state are destroyed.
- Append and restore v0.14 duplicate records sequentially when the existing `support_duplicate` setting is enabled.

## Scope

This is a direct adaptation to the v0.18 `LabelTable` architecture. It does not backport the graph-level duplicate tracker refactor (#1797) or duplicate vector storage (#2244). The native serialization format is unchanged, and `max_duplicates_per_group` applies only to KNN; RangeSearch keeps its existing expansion behavior.

The v0.14 extension deliberately uses the existing `support_duplicate` setting on both write and read. Legacy business indexes were produced without duplicate support and remain readable because no extension is expected. Writer and reader settings must match: pre-patch v0.14 blobs produced with `support_duplicate=true` are not supported, and the business owner confirmed that no such persisted indexes exist. Duplicate records are appended and consumed in order; this PR does not add `Length()`, `Seek()`, footer probing, or generic StreamReader behavior.

The deserialization behavior follows #2258, and the search behavior follows and hardens #2264 for multi-page Iterator search. Development started from exact tag `v0.18.10` (`39afdfa27ed91b3b892c014a6cd172bdb25cdd6e`) and is now rebased onto `upstream/0.18` at `09aec7736fc010ca9d68d0d8451c118fcb1f5ba1`.

## Red-green validation

The regression tests first reproduced:

- acceptance of out-of-range and overlapping duplicate records;
- invalid negative duplicate limits being accepted;
- entrypoint/parallel/filter inconsistencies;
- duplicate groups consuming Iterator `ef_search` and losing later groups across pages;
- pending members missing from final drain;
- a normal-return parallel worker lifetime crash.

All corresponding tests pass after the implementation.

## Test plan

- [x] `make fmt` with clang-format 15
- [x] clang-tidy 15 on modified production and test code
- [x] Debug build with tests using all 96 logical CPUs
- [x] Focused LabelTable duplicate tests: 8 cases, 32 assertions
[...]

PR #2891, opened 2026-09-08
[bug](hgraph): check remaining duplicate build candidates

### LINKED ISSUE #2889: [bug](hgraph): check remaining build candidates for duplicate codes
### Affected area
hgraph

### Interface
cpp

### VSAG version
Reproduced on the v0.18-based PR #2615 at `306bc89c4afe946b8acc4e605177800557b716e7`.
The nearest-only check also exists on current `main` (`30c217bc`) and `1.0` (`ce2a985d`); runtime validation on those branches is pending.

### Summary
With `support_duplicate=true` and `duplicate_distance_threshold=0`, HGraph only checks the nearest build candidate for matching codes. SQ cosine can rank a different code ahead of an identical code, so an existing matching candidate is missed and the new vector becomes another graph node. This can increase neighbor pressure and, in a reproduced case, cause loss of reachability and recall.

### Minimal reproducer
Use a two-node graph (0 linked to 1 and vice versa) with real SQ8 cosine storage:

```cpp
// dim=2; train SQ8 on these four vectors, fixing both intervals to [-1, 1].
const std::vector<float> training = {1, 0, -1, 0, 0, 1, 0, -1};
// Store A, B, and a new A at IDs 0, 1, 2. Only IDs 0 and 1 are in the graph.
const std::vector<float> vectors = {0.599F, 0.801F, 0.6F, 0.8F, 0.599F, 0.801F};
// Search with query=vectors.data()+4, ep=0, ef=2, topk=2,
// find_duplicate=true, duplicate_query_id=2, duplicate_distance_threshold=0.
```

In the regression test, `CompareVectors(2, 0)` is true and `CompareVectors(2, 1)` is false, but query distance to B is 0.003141463 versus 0.007838547 to A. Both candidates are returned; the current duplicate ID is -1 instead of 0.

### Environment
```text
- OS: Ubuntu 22.04
- vsag version: v0.18.10-11-g306bc89c-dirty
```

### Expected behavior
When the nearest candidate does not match, check the remaining returned build candidates and use a matching candidate if present. Preserve the existing nearest-match fast path and positive distance-threshold behavior.

### Actual behavior
The nearest nonmatching candidate ends duplicate detection despite another matching candidate being available.

A separate private 200-row, 1024-dimensional fixture uses a fixed SQ8 model, cosine, single-threaded Add, M=64 and ef_construction=300. Checking the remaining candidates, without changing distance or pruning code, gives:

| Metric | Before | After (3 runs) |
| --- | ---: | ---: |
| Graph nodes | 88 | 3 |
| Duplicate children | 112 | 197 |
| Correct-group incoming edges from other groups after Add 178 | 0 | 2 |
| Correct results at k=100, ef_search=64 | 0/100 | 100/100 |
| Correct results at k=100, ef_search=1000 | 0/100 | 100/100 |

The regression test fails before the change and passes afterward. The private fixture's fixed-result validation also fails before and passes afterward. Customer data is not attached.

### Additional context
Scope: check only existing build candidates; do not change quantization, pruning, serialization, or concurrency semantics. This does not guarantee finding representatives outside the candidate set or fix every graph-connectivity issue.

Delivery:
- [ ] main: #2891; preserve stored-query and raw-query comparison paths.
- [ ] 1.0: #2892.
- [ ] 0.18: #2615, updated and rebased onto upstream/0.18 (09aec773).

### References
- https://github.com/antgroup/vsag/blob/306bc89c4afe946b8acc4e605177800557b716e7/src/impl/searcher/basic_searcher.cpp#L490-L508
- https://github.com/antgroup/vsag/blob/30c217bc3ec0bb7395d5dea843fbee094365e8b6/src/impl/searcher/basic_searcher.cpp#L900-L926
- https://github.com/antgroup/vsag/blob/ce2a985d563042a161bd421b05e6946ba9d64c0c/src/impl/searcher/basic_searcher.cpp#L531-L549

## Related issues
- #2574: duplicate expansion for entrypoint and iterator search (related work, not this build-candidate defect).
- Duplicate scan for `duplicate SQ cosine`: no matches.

---

jac0626:
## Change Type

- [x] Bug fix

## Linked Issue

Fixes: #2889

## What Changed

- At threshold zero, check the nearest build candidate first, then the remaining returned candidates until matching codes are found.
- Keep positive distance-threshold behavior unchanged. No quantization, pruning, serialization, or concurrency changes.
- Add a deterministic 2D SQ8 cosine regression and update the English/Chinese parameter documentation. No customer data is included.

- Preserve both stored-query and raw-query comparisons on main. The regression covers both paths; the existing stored-ID test continues to check that stored codes take precedence over raw input.

## Test Evidence

- Red: both stored-query and raw-query cases returned duplicate ID -1 despite the matching ID 0 being in the candidates.
- Green: all 7 cases in basic_searcher_test.cpp passed (50 assertions).
- clang-format 15 and clang-tidy 15 checks on changed code passed. clang-tidy was run separately per translation unit using its matching compile command after a multi-input invocation crashed.
- Release build of the current main base, using the locally built OpenBLAS 0.3.23 dependency.
- Full-suite and coverage checks are left to CI.

No API/ABI or persisted-format changes. No additional candidate comparisons when duplicate detection is disabled, the positive threshold path is selected, or the nearest code already matches. A mismatch can add a bounded scan of the returned candidates; build-performance impact has not been benchmarked. This does not guarantee finding representatives outside that set or repair existing graphs.

## Related delivery

- 0.18: #2615.
- 1.0: #2892.

## Risk and Rollback

Localized build-grouping behavior change. Revert this commit to restore nearest-only detection; no format migration is needed.

### DIFF
--- a/src/impl/searcher/basic_searcher.cpp
+++ b/src/impl/searcher/basic_searcher.cpp
@@ -916,12 +916,21 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
         } else {
             const bool has_stored_query =
                 inner_search_param.duplicate_query_id < flatten->TotalCount();
-            const bool is_duplicate =
-                has_stored_query
-                    ? flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)
-                    : flatten->CompareRawVectorWithId(query, min_index);
-            if (is_duplicate) {
+            auto matches_query = [&](InnerIdType id) {
+                return has_stored_query
+                           ? flatten->CompareVectors(inner_search_param.duplicate_query_id, id)
+                           : flatten->CompareRawVectorWithId(query, id);
+            };
+            if (matches_query(min_index)) {
                 inner_search_param.duplicate_id = min_index;
+            } else {
+                // Quantization can rank a different code ahead of an identical one.
+                for (uint32_t i = 0; i < top_candidates->Size(); ++i) {
+                    if (data[i].second != min_index && matches_query(data[i].second)) {
+                        inner_search_param.duplicate_id = data[i].second;
+                        break;
+                    }
+                }
             }
         }
     }
--- a/docs/docs/en/src/indexes/hgraph.md
+++ b/docs/docs/en/src/indexes/hgraph.md
@@ -82,7 +82,7 @@ most users need; the exhaustive list is in [Index Parameters](../resources/index
 | `build_thread_count` | int | `100` | Threads used to parallelise build |
 | `support_duplicate` | bool | `false` | Enable duplicate-ID detection on insert |
 | `deduplicate_storage` | bool | `false` | Share vector storage between duplicates; requires `support_duplicate: true` |
-| `duplicate_distance_threshold` | float | `0.0` | Duplicate-detection distance threshold. When greater than `0`, deduplicate by the nearest candidate distance; when `0`, fall back to the current code `memcmp` check |
+| `duplicate_distance_threshold` | float | `0.0` | Duplicate-detection distance threshold. When greater than `0`, deduplicate by the nearest candidate distance; when `0`, compare codes with the nearest candidate first, then check the remaining returned build candidates until a match is found |
 | `support_remove` | bool | `false` | Enable graph delete-tracking metadata used by mark-remove recovery paths |
 | `support_force_remove` | bool | `false` | Enable `RemoveMode::FORCE_REMOVE` and its extra synchronization on the built index |
 | `store_raw_vector` | bool | `false` | Keep the raw vector in addition to the quantized copy (useful for `cosine`) |
--- a/docs/docs/zh/src/indexes/hgraph.md
+++ b/docs/docs/zh/src/indexes/hgraph.md
@@ -76,7 +76,7 @@ auto result = index->KnnSearch(
 | `build_thread_count` | int | `100` | 构建阶段并发线程数 |
 | `support_duplicate` | bool | `false` | 是否在插入时做重复 ID 检测 |
 | `deduplicate_storage` | bool | `false` | 让重复向量共享存储；需同时设置 `support_duplicate: true` |
-| `duplicate_distance_threshold` | float | `0.0` | 重复判定距离阈值。大于 `0` 时按最近候选的距离判重；等于 `0` 时退化为当前编码 `memcmp` 判重 |
+| `duplicate_distance_threshold` | float | `0.0` | 重复判定距离阈值。大于 `0` 时按最近候选的距离判重；等于 `0` 时先比较最近候选的编码，不匹配则继续检查构建搜索返回的其余候选，直到找到匹配 |
 | `support_remove` | bool | `false` | 是否启用 mark-remove 恢复路径所需的图删除追踪元数据 |
 | `support_force_remove` | bool | `false` | 是否启用 `RemoveMode::FORCE_REMOVE` 及其额外同步 |
 | `store_raw_vector` | bool | `false` | 除量化副本外再保留原始向量（`cosine` 场景有用） |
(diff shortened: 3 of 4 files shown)


LHT129 (commented):
[review] This PR fixes a real bug where SQ quantization can rank a different code ahead of an identical one, causing duplicate detection to miss the true match when `duplicate_distance_threshold` is 0.

**What changed:**
- `basic_searcher.cpp`: When the nearest candidate by quantized distance does not match the query code, the fix scans the remaining top candidates until a matching code is found. The positive-threshold path is untouched.
- `basic_searcher_test.cpp`: Adds a deterministic SQ8 cosine regression test that exercises both stored-query and raw-query paths. Also updates an existing FP32 test expectation to reflect the corrected behavior.
- Documentation: English and Chinese parameter docs updated to describe the new fallback behavior.

**Review notes:**
- The logic is correct and minimal. The fallback loop only activates when `duplicate_distance_threshold == 0` and the nearest candidate does not match — both conditions are checked before the scan.
- The new test is well-constructed: it deterministically fixes SQ8 intervals, avoids random training data, and covers both code paths via `GENERATE`.
- The changed test assertion in the existing "keeps nearest owner" test correctly reflects the new behavior: when a stored query code is provided, it remains authoritative even when the nearest quantized candidate differs.
- One existing review comment flags a `uint32_t` loop counter mismatch with `top_candidates->Size()` (which returns `uint64_t`). This is a real concern for large result sets and should be addressed.

LHT129 (commented):
Reviewed the changes in this PR. The fix correctly handles the case where quantization ranks a different code ahead of an identical one at threshold zero, by checking remaining candidates when the nearest code does not match.

**Correctness**: The `matches_query` lambda correctly dispatches between stored-query (`CompareVectors`) and raw-query (`CompareRawVectorWithId`) paths, and the fallback loop correctly scans all returned candidates beyond `min_index`. The `duplicate_id` is only set when a match is found, preserving the existing "no match → unchanged" semantics.

**Tests**: The new SQ8 cosine regression test is well-constructed — it sets up a realistic quantization scenario where B ranks closer to A's raw query than A's own code, and verifies both stored-query and raw-query paths via `GENERATE`. The updated expectation in the existing duplicate threshold test (`run_search({0.3F, 0.0F}, 0.0F, 0) == 0`) correctly reflects that the stored query code is now authoritative when checking other candidates.

**Documentation**: Both EN and ZH docs accurately describe the new behavior.

The `uint32_t` loop counter vs `uint64_t Size()` mismatch on line 928 has already been noted by Copilot. No other issues found.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -916,12 +916,21 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
...
-            const bool is_duplicate =
-                has_stored_query
-                    ? flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)
-                    : flatten->CompareRawVectorWithId(query, min_index);
-            if (is_duplicate) {
+            auto matches_query = [&](InnerIdType id) {
+                return has_stored_query
+                           ? flatten->CompareVectors(inner_search_param.duplicate_query_id, id)
+                           : flatten->CompareRawVectorWithId(query, id);
+            };
```
LHT129, line 923: [suggestion] The loop starts at `i = 0` but `data[0]` is always `min_index` (set at line 907). The `data[i].second != min_index` guard will skip it, but starting from `i = 1` would be slightly clearer and avoid one unnecessary iteration.

```cpp
for (uint32_t i = 1; i < top_candidates->Size(); ++i) {
    if (matches_query(data[i].second)) {
        inner_search_param.duplicate_id = data[i].second;
        break;
    }
}
```

PR #2892, opened 2026-09-08
[bug](hgraph): backport duplicate candidate checks to 1.0

jac0626:
## Change Type

- [x] Bug fix

## Linked Issue

Fixes: #2889

## What Changed

- At threshold zero, check the nearest build candidate first, then the remaining returned candidates until matching codes are found.
- Keep positive distance-threshold behavior unchanged. No quantization, pruning, serialization, or concurrency changes.
- Add a deterministic 2D SQ8 cosine regression and update the English/Chinese parameter documentation. No customer data is included.

## Test Evidence

- Red: the matching candidate is present, but the original code returns duplicate ID -1 instead of 0.
- Green: 6 relevant cases passed (143888 assertions), including the regression, nearest/positive-threshold checks, normal search, and iterator drain.
- clang-format 15 and clang-tidy 15 checks on changed code passed.
- Built against current upstream/1.0 (ce2a985d) in Release mode.
- Full-suite and coverage checks are left to CI.

### Existing Release test failure

The existing "BasicSearcher traverses through a non-finite-distance bridge" case fails two assertions in this Release configuration. I restored the original production source at ce2a985d, rebuilt with identical settings, and reproduced exactly the same failure (4/6 assertions pass). The same failure remains with this fix; it is not introduced here. No unrelated non-finite-distance or compiler-flag fix is included.

No API/ABI or persisted-format changes. No additional candidate comparisons when duplicate detection is disabled, the positive threshold path is selected, or the nearest code already matches. A mismatch can add a bounded scan of the returned candidates; build-performance impact has not been benchmarked. This does not guarantee finding representatives outside that set or repair existing graphs.

## Related delivery

- 0.18: #2615.
- main: #2891.

## Risk and Rollback

Localized build-grouping behavior change. Revert this commit to restore nearest-only detection; no format migration is needed.

### DIFF
--- a/src/impl/searcher/basic_searcher.cpp
+++ b/src/impl/searcher/basic_searcher.cpp
@@ -543,9 +543,20 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
             if (min_distance <= inner_search_param.duplicate_distance_threshold) {
                 inner_search_param.duplicate_id = min_index;
             }
-        } else if (inner_search_param.duplicate_query_id < flatten->TotalCount() &&
-                   flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)) {
-            inner_search_param.duplicate_id = min_index;
+        } else if (inner_search_param.duplicate_query_id < flatten->TotalCount()) {
+            if (flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)) {
+                inner_search_param.duplicate_id = min_index;
+            } else {
+                // Quantization can rank a different code ahead of an identical one.
+                for (uint32_t i = 0; i < top_candidates->Size(); ++i) {
+                    if (data[i].second != min_index &&
+                        flatten->CompareVectors(inner_search_param.duplicate_query_id,
+                                                data[i].second)) {
+                        inner_search_param.duplicate_id = data[i].second;
+                        break;
+                    }
+                }
+            }
         }
     }
 
--- a/docs/docs/en/src/indexes/hgraph.md
+++ b/docs/docs/en/src/indexes/hgraph.md
@@ -73,7 +73,7 @@ most users need; the exhaustive list is in [Index Parameters](../resources/index
 | `base_pq_dim` | int | `1` | Number of PQ subspaces. When using `pq` / `pqfs`, set this explicitly instead of relying on the default. |
 | `build_thread_count` | int | `100` | Threads used to parallelise build |
 | `support_duplicate` | bool | `false` | Enable duplicate-ID detection on insert |
-| `duplicate_distance_threshold` | float | `0.0` | Duplicate-detection distance threshold. When greater than `0`, deduplicate by the nearest candidate distance; when `0`, fall back to the current code `memcmp` check |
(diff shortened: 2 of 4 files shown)


LHT129 (commented):
test review

src/impl/searcher/basic_searcher.cpp
```diff
@@ -543,9 +543,20 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
             if (min_distance <= inner_search_param.duplicate_distance_threshold) {
                 inner_search_param.duplicate_id = min_index;
             }
-        } else if (inner_search_param.duplicate_query_id < flatten->TotalCount() &&
-                   flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)) {
-            inner_search_param.duplicate_id = min_index;
+        } else if (inner_search_param.duplicate_query_id < flatten->TotalCount()) {
+            if (flatten->CompareVectors(inner_search_param.duplicate_query_id, min_index)) {
+                inner_search_param.duplicate_id = min_index;
+            } else {
```
LHT129, line 549: [note] The fallback loop iterates from `i = 0` and skips `min_index` via the `data[i].second != min_index` guard, which is correct. However, since `data[0].second` is always `min_index` (by construction from the preceding min-distance scan), starting the loop at `i = 1` would avoid one guaranteed-to-fail comparison per invocation. This is a micro-optimization and the current code is fine as-is — just noting it for awareness.

LHT129 (commented):
Reviewed the backport of duplicate candidate checks to 1.0.

**Changes reviewed:**
- `src/impl/searcher/basic_searcher.cpp`: The fix correctly handles the case where SQ8 cosine quantization ranks a different code ahead of an identical one. When `duplicate_distance_threshold == 0`, it first checks the nearest candidate, then falls back to scanning remaining returned candidates. The `data[i].second != min_index` guard correctly avoids re-checking the already-failed candidate.
- `src/impl/searcher/basic_searcher_test.cpp`: The regression test is well-constructed — it sets up a deterministic 2D SQ8 cosine scenario where quantization makes a different vector (B) appear closer than the identical vector (A), then verifies the duplicate is still found.
- `docs/docs/en/src/indexes/hgraph.md` and `docs/docs/zh/src/indexes/hgraph.md`: Parameter documentation updates accurately describe the new behavior.

**Existing note:** @LHT129 already posted a `[note]` about the loop starting at `i=0` instead of `i=1` (since `data[0].second` is always `min_index`). This is a valid micro-optimization observation but not blocking.

No critical issues found. The fix is minimal, focused, and well-tested.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -543,9 +543,20 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
...
+                // Quantization can rank a different code ahead of an identical one.
+                for (uint32_t i = 0; i < top_candidates->Size(); ++i) {
```
LHT129, line 551: [note] Correction to the earlier comment: `data[0]` is **not** always `min_index`. `top_candidates` is `StandardHeap<true, false>` — a max-heap where `data[0]` (the heap front) is the element with the **largest** distance. The preceding scan explicitly searches the entire array to find `min_index` (the element with the smallest distance), which could be at any position. Therefore starting the fallback loop at `i = 0` (rather than `i = 1`) is correct and necessary — `data[0]` may not be `min_index` and should not be skipped.

Issue #2771, opened 2026-08-25
[improve](build): cache prebuilt third-party dependencies for cold builds


wxyucs:
### Kind of improvement

chore

### Affected area

build

### Summary

Provide a reusable binary cache or prebuilt dependency layer for third-party libraries used by clean local and CI builds.

The current workflow caches downloaded archives and uses ccache for many project translation units, but cold builds may still compile dependencies such as ANTLR4 from source. OpenBLAS also falls back to a serialized source build when a suitable system installation is unavailable.

Cache dependency install directories using keys that include the dependency pin, compiler, architecture, ABI, build type, and relevant options. Alternatively, publish a maintained dependency image or package for supported CI toolchains.

### Expected benefit

Reduce cold-build latency and variance, avoid repeatedly compiling unchanged third-party code, and make offline builds more predictable.

### Benchmark evidence

Not required for the initial cache infrastructure change. CI should report cache hit state and dependency preparation/build wall time.

### References

- `extern/antlr4/antlr4.cmake:21`
- `extern/openblas/openblas.cmake:170`
- `extern/openblas/openblas.cmake:196`
- `extern/openblas/openblas.cmake:207`
- `.github/workflows/pr-ci.yml:94`
- `.github/workflows/pr-ci.yml:99`

### Additional context

Cache correctness is more important than hit rate. Incompatible compilers, ABIs, sanitizer settings, or architectures must not share binary artifacts.

## Related issues

_None found._

---
_Drafted with AI assistant: Codex:gpt-5_

wxyucs:
# Issue #2771: dependency install-cache investigation

**Investigation baseline:** `origin/main` at `a5cec27cd3a2436fd1ed734c5bee74722392dac3` (2026-09-03).  
**Scope:** investigation and proposed design only; no implementation is included.  
**Evidence labels:** **Verified** means observed in repository source, public CI/history, or the contained probe described below. **Inference** means a conclusion from that evidence. **Proposed** describes a future design, not current behavior.

[...]

PR #2899, opened 2026-09-08
ci: make dependency build metrics trustworthy

### LINKED ISSUE #2893: [improve](ci): establish trustworthy dependency build metrics
## Parent

- #2771

## Scope

Establish trustworthy dependency source-build coverage and measurements before introducing reusable binary dependency caches.

- Make the daily x86 OpenBLAS source-build guard explicit instead of allowing `AUTO` to select the system installation.
- Correct dependency classification and duplicate counting in the build metrics collector, including `.ci-fetchcontent` and ExternalProject stages.
- Separate true no-op, warm-cache, and clean-build measurements.
- Add dependency selection, preparation time, cache state, payload size, and fallback information to the GitHub Actions job summary.
- Ensure pull-request workflows restore source archives without writing untrusted cache entries.

## Acceptance criteria

- [ ] The daily source guard demonstrably selects bundled OpenBLAS.
- [ ] ANTLR4, HDF5, FetchContent dependencies, and multi-output Ninja edges are classified once and covered by regression tests.
- [ ] Job summaries clearly distinguish source, system, and cache paths and their timings.
- [ ] PR workflows cannot save dependency cache entries.
- [ ] Existing system dependency, offline archive, and source fallback behavior remains intact.

This is the first prerequisite phase from the investigation in #2771.

wxyucs:
Measure dependency preparation and clean, warm-cache, and no-op builds separately, with per-dependency resolution, timing, local sizes, stages, and fallbacks. Deduplicate multi-output Ninja edges; select bundled OpenBLAS in the daily x86 guard and restore source archives without saving PR cache entries.

JSON schema version 3 reports aggregate dependency `local_bytes`, excluding shared `BUILD_INFO_DIR` metadata, host installations, and separately measured archives. Read configured CMake cache paths before environment fallbacks and normalize system dependency policies as CMake does.

Use the configured generator to select Ninja statistics, and preserve paths containing spaces in debug build recipes, including aggregate and module tests. HDF5 minimization and binary install-prefix caching remain separate work.

Validation: all 33 focused collector/workflow tests pass. Real Make/CMake fixtures cover generator fallback, clean/warm/no-op builds, default and overridden dependency paths, and aggregate/module test execution with custom paths and filter/shard arguments. Targeted regressions fail before the three fixes and pass afterward. `git diff --check` passes. Rebased onto current main without conflicts. Full C++ tests, documentation validation, and exact format/lint 15 checks remain with CI; new-head results are pending.

Previous-head CircleCI job 18154 failed in `(PR) HGraph Search with Dirty Vector`: the result-count assertion reported 5 versus 10. This is outside the dependency metrics changes; no unrelated test changes or manual rerun were made.

Closes: #2893
Parent: #2771


scripts/collect_build_metrics.py
```diff
@@ -141,7 +335,225 @@ def parse_ninja_log(
...
+        elif dependency in FETCHCONTENT_DIRECTORY_NAMES and fetchcontent_dir is not None:
+            directory_name = FETCHCONTENT_DIRECTORY_NAMES[dependency]
+            source_bytes = directory_size(fetchcontent_dir / f"{directory_name}-src")
+            build_bytes = directory_size(fetchcontent_dir / f"{directory_name}-build")
+            if fetchcontent_dir.name == ".ci-fetchcontent":
+                details["preparation"] = "pinned FetchContent checkout"
+            else:
+                details["preparation"] = "FetchContent source tree"
+        else:
+            details["preparation"] = "host dependency"
```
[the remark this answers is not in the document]
wxyucs: Fixed in a4563f70. Unclassified entries now report their actual resolution as unclassified preparation instead of implying a host dependency, with a focused regression for the source case.

scripts/collect_build_metrics.py
```diff
@@ -126,13 +285,48 @@ def parse_ninja_log(
...
-        category = classify_edge(output, link_outputs)
+        if len(fields) >= 5 and fields[4]:
+            edge_key: tuple[Any, ...] = (start_ms, end_ms, fields[4])
+        else:
+            edge_key = (start_ms, end_ms, output)
+        record = edge_records.setdefault(
+            edge_key, {"start_ms": start_ms, "end_ms": end_ms, "outputs": []}
+        )
+        if output not in record["outputs"]:
+            record["outputs"].append(output)
```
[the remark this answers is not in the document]
wxyucs: Fixed in 3a911528. Deduplication now joins only adjacent records with the same start, end, restat value, and command hash. Real multi-output groups were verified consecutive, and a regression preserves separate nonconsecutive collisions.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +121,124 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+    if dependency is not None:
+        return {"category": "dependency_build", "dependency": dependency, "stage": "build"}
+    normalized_link_outputs = {normalize_output(item) for item in link_outputs or set()}
+    if normalized in normalized_link_outputs:
+        return {"category": "link", "dependency": None, "stage": None}
     if normalized.endswith((".a", ".so", ".dylib", ".dll", ".exe")):
-        return "link"
-    return "other"
+        return {"category": "link", "dependency": None, "stage": None}
+    return {"category": "other", "dependency": None, "stage": None}
```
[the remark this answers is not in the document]
wxyucs: Fixed in 3a911528. Link outputs are normalized once per Ninja log parse and the prepared set is reused for every edge; normalization behavior is covered by a regression.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +121,124 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+        "dependency_compile": 0,
+        "test_compile": 1,
+        "production_compile": 2,
+        "other_compile": 3,
+        "dependency_build": 4,
+        "link": 5,
+        "build_maintenance": 6,
+        "other": 7,
+    }
+    return min(details, key=lambda detail: priority[str(detail["category"])])
```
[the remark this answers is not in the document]
wxyucs: Fixed in 3a911528. ExternalProject precedence now applies only when an output is positively recognized by the stamp/build-info parser. A regression verifies a regular dependency build output cannot override dependency-compile priority.

LHT129 (commented):
I have reviewed this PR thoroughly. The changes are well-structured and the code quality is high.

**Summary of what this PR does:**
- Forces daily x86 guard to explicitly select bundled OpenBLAS (`-DVSAG_USE_SYSTEM_OPENBLAS=OFF`)
- Classifies FetchContent and ExternalProject work per dependency with deduplication of multi-output Ninja edges
- Measures clean, warm-cache, and verified no-op builds separately
- Reports dependency resolution, preparation/cache timing, sizes, stages, and fallbacks
- Switches PR CI to `actions/cache/restore@v4` (restore-only, preventing PR jobs from saving cache entries)
- Adds comprehensive tests for ninja log parsing, dependency resolution, workflow guards, and no-op verification

**Key observations:**

1. The `classify_outputs()` function uses `EXTERNAL_PROJECT_STAGES` to prioritize ExternalProject stamp stages over other classifications. Since `"build"` is in `EXTERNAL_PROJECT_STAGES` and `classify_output()` assigns `stage="build"` to generic dependency build outputs (not just ExternalProject stamps), this could cause a multi-output edge containing both a dependency `.o` and a non-stamp build artifact to be classified as `dependency_build` instead of `dependency_compile`. This was already flagged by Copilot and is worth addressing to ensure accurate per-dependency stage attribution.

2. The `else` branch in `build_dependency_report()` now correctly reports `"{resolution} preparation not classified"` instead of the previous misleading `"host dependency"` fallback. The test `test_unclassified_source_dependency_is_not_reported_as_host_dependency` confirms this fix.

3. The workflow changes are correct: `daily_test.yml` now explicitly passes `-DVSAG_USE_SYSTEM_OPENBLAS=OFF` via `EXTRA_DEFINED`, and `pr-ci.yml` uses `actions/cache/restore@v4` to prevent PR jobs from saving cache entries.

No blocking issues found. The one classification edge case noted above is a suggestion-level concern.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +345,225 @@ def parse_ninja_log(
...
+    return total
+
+
+def read_cmake_cache(path: Path) -> dict[str, str]:
+    if not path.is_file():
+        return {}
+    values = {}
+    for line in path.read_text(encoding="utf-8", errors="replace").splitlines():
+        if not line or line.startswith(("#", "//")) or "=" not in line or ":" not in line:
+            continue
```
LHT129, line 474: [suggestion] `read_cmake_cache` splits on the first `=` in each line, but CMake cache values can themselves contain `=` (e.g. `CMAKE_CXX_FLAGS:STRING=-DFOO=bar`). The `split("=", 1)` would capture only `-DFOO` and drop `=bar`. Additionally, the `":" not in line` guard rejects any line whose value contains a colon, which is legitimate in CMake cache entries.

Consider using a more robust parse: split on the first `:` to separate type+name from the value, then split on `=` only within the type+name portion. For example:

```python
type_name, _, value = line.partition("=")
if ":" not in type_name:
    continue
_, name = type_name.split(":", 1)
values[name] = value
```

scripts/collect_build_metrics.py
```diff
@@ -141,7 +345,225 @@ def parse_ninja_log(
...
+            source_bytes = directory_size(fetchcontent_dir / f"{directory_name}-src")
+            build_bytes = directory_size(fetchcontent_dir / f"{directory_name}-build")
+            if fetchcontent_dir.name == ".ci-fetchcontent":
+                details["preparation"] = "pinned FetchContent checkout"
+            else:
+                details["preparation"] = "FetchContent source tree"
+        else:
+            details["preparation"] = f"{details['resolution']} preparation not classified"
+
+        policy = cmake_cache.get(
```
LHT129, line 543: [suggestion] In `build_dependency_report`, the `else` branch (unclassified dependencies) sets `preparation` to `"{resolution} preparation not classified"` but leaves `fallback` as `"none reported"` from `setdefault`. This is inconsistent: when a dependency is unclassified, the fallback should reflect that the classification path is unknown rather than implying no fallback exists.

Consider setting `details["fallback"]` to something like `"classification path unknown; metrics may be incomplete"` in the `else` branch to make the report self-consistent.

wxyucs: Fixed in 09fc5e70. When an unclassified dependency has no reported fallback, the report now says that its classification path is unknown and metrics may be incomplete; the existing regression asserts both preparation and fallback text.

scripts/collect_build_metrics.py
```diff
@@ -247,13 +669,25 @@ def capture_build(self, name: str) -> None:
         )
         captured_log = self.output_dir / f"{name}.ninja_log"
         if ninja_log.is_file():
-            shutil.copy2(ninja_log, captured_log)
+            added_records = ninja_log_delta(before_records, read_ninja_log_records(ninja_log))
+            captured_log.write_text(
+                "# ninja log v5\n" + "\n".join(added_records) + ("\n" if added_records else ""),
+                encoding="utf-8",
+            )
```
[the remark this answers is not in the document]
wxyucs: Fixed in 09fc5e70. Captured deltas now preserve the version header from the active Ninja log and omit it if none exists. The no-op regression uses a synthetic v7 header to ensure it is not hardcoded.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +345,225 @@ def parse_ninja_log(
...
+    elif "Building OpenBLAS from source" in text:
+        resolutions.setdefault(
+            "openblas",
+            {
+                "resolution": "source",
+                "pin": "v0.3.24",
+                "source_origin": "default",
+                "fallback": "system dependency was disabled or unavailable",
+            },
+        )
```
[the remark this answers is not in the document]
wxyucs: Fixed in 09fc5e70. The fallback parser now reports an unknown OpenBLAS pin rather than embedding a version; normal source configurations still obtain the exact pin from the emitted pin-qualified third-party override message. A focused regression covers the missing-message case.

scripts/collect_build_metrics.py
wxyucs: Fixed in 09fc5e70. Splitting on the first  already preserved later delimiters, but the type guard did inspect the whole line and could admit an untyped entry whose value contained . The parser now partitions first and validates  only in the declaration; a regression covers values containing both  and  plus the malformed case.

scripts/collect_build_metrics.py
```diff
@@ -17,6 +18,59 @@
...
+    "mkdir",
+    "download",
+    "update",
+    "patch",
+    "configure",
+    "build",
+    "install",
+    "test",
+    "done",
+}
```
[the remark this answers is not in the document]
wxyucs: Fixed in 09fc5e70. ExternalProject stages are now an ordered tuple, so suffix matching and stage precedence remain deterministic.

scripts/collect_build_metrics.py
wxyucs: Clarification: splitting on the first equals sign already preserved later delimiters, but the type guard inspected the whole line and could admit an untyped entry whose value contained a colon. The parser now partitions first and validates the colon only in the declaration; a regression covers both delimiters and the malformed case.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +121,126 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+    external_stage = next(
+        (
+            classify_output(output, normalized_link_outputs)
+            for output in outputs
+            if external_project_stage(output) is not None
+        ),
+        None,
+    )
+    if external_stage is not None:
+        return external_stage
```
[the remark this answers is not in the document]
wxyucs: Fixed in 09fc5e70. classify_output now records positive ExternalProject-stage identification in the existing detail, and classify_outputs reuses those details instead of reclassifying each output.

scripts/collect_build_metrics.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import argparse
+from collections import Counter
 import datetime as dt
 import json
```
LHT129, line 9: [suggestion] The self-mapping entries in `DEPENDENCY_ALIASES` (`"thread_pool": "thread_pool"`, `"yaml_cpp": "yaml_cpp"`, `"roaringbitmap": "roaringbitmap"`) are redundant. `canonical_dependency` already returns the normalized name unchanged when no alias is found (line 126: `DEPENDENCY_ALIASES.get(normalized, normalized)`). Removing these entries would make the intent clearer — only keep mappings that actually transform a name.

```python
DEPENDENCY_ALIASES = {
    "nlohmann_json": "json",
    "nlohmann-json": "json",
    "thread-pool": "thread_pool",
    "yaml-cpp": "yaml_cpp",
}
```

wxyucs: Fixed in c878cfc2. The three identity mappings were removed; only aliases that transform a normalized dependency name remain.

LHT129 (commented):
I have completed a thorough review of the latest commit (c878cfc2). The previous review comments from Copilot and LHT129 have all been addressed:

- The identity mappings in `DEPENDENCY_ALIASES` have been removed (only transforming aliases remain).
- `classify_output` now records ExternalProject-stage identification in the detail dict, and `classify_outputs` reuses those details.
- Unclassified entries now report their actual resolution as "unclassified preparation" instead of implying a host dependency.
- The `read_cmake_cache` function now uses `partition("=")` to preserve value delimiters.

The overall changes look solid:

- **CI workflows**: The daily x86 guard now explicitly passes `-DVSAG_USE_SYSTEM_OPENBLAS=OFF`, making the intent unambiguous. The PR CI splits dependency preparation into separate FetchContent and ExternalProject phases with proper timing, and uses `actions/cache/restore@v4` to prevent PR jobs from saving cache entries.
- **Metrics collection**: The ninja log parser now deduplicates multi-output edges, classifies outputs per-dependency with stage-level granularity, and separates build maintenance edges from real build work. The no-op build verification is robust.
- **Test coverage**: Comprehensive tests cover edge deduplication, ExternalProject stage parsing, dependency resolution, no-op verification, ninja log delta handling, and workflow regression checks.

No blocking issues found.

LHT129 (commented):
Reviewed the latest commit c878cfc2. All previously flagged issues in `collect_build_metrics.py` have been addressed:

- `DEPENDENCY_ALIASES` self-mapping entries removed
- `EXTERNAL_PROJECT_STAGES` is now an ordered tuple
- `classify_outputs` correctly delegates ExternalProject-stage detection to `classify_output`
- `normalized_link_outputs` is computed once per `parse_ninja_log` call
- Edge deduplication uses adjacent-record merging with command hash
- OpenBLAS pin is no longer hardcoded
- Ninja log header is preserved from the active log
- `read_cmake_cache` type guard is correct
- Unclassified dependency fallback reporting is accurate

The workflow changes in `pr-ci.yml` and `daily_test.yml` look correct: the daily x86 guard explicitly sets `-DVSAG_USE_SYSTEM_OPENBLAS=OFF`, and the PR CI uses `actions/cache/restore@v4` with `restore-keys` fallback preserved. The new `test_dependency_workflows.py` regression tests provide good coverage for these invariants.

No new issues found in this revision.

LHT129 (commented):
[suggestion] In `parse_dependency_resolutions` (scripts/collect_build_metrics.py:420-421), `resolutions["fmt"] = ...` uses direct assignment which overwrites any previously parsed Third-party override for fmt. If fmt were configured as a source dependency with a specific pin via the override mechanism, but the configure output also mentions "Using pre-existing fmt::fmt target", the resolution would be incorrectly reported as "system". Use `resolutions.setdefault("fmt", {...})` instead (as done for OpenBLAS at line 409-418) so that an explicit Third-party override takes precedence over the heuristic string match. The same issue applies to `resolutions["mkl"]` at the following block (line 427-428).

scripts/collect_build_metrics.py
```diff
@@ -141,7 +343,238 @@ def parse_ninja_log(
...
+            "pin": "system",
+            "source_origin": "system",
+            "fallback": "bundled source fallback available when policy is AUTO",
+        }
+    elif "Building OpenBLAS from source" in text:
+        resolutions.setdefault(
+            "openblas",
+            {
+                "resolution": "source",
+                "pin": "unknown",
```
LHT129, line 415: [suggestion] The `pin` field is hardcoded to `"unknown"` when OpenBLAS is built from source (both in the `elif "Building OpenBLAS from source"` branch and the `"Intel MKL is not supported on this architecture"` fallback). However, the version is explicitly pinned to `v0.3.24` in `extern/openblas/openblas.cmake` via the download URL and `vsag_resolve_thirdparty_override`. Reporting `"unknown"` undermines the trustworthiness goal of this PR — the pin is deterministic and known.

Consider extracting the version from the CMake configure output (e.g., parsing the `vsag_resolve_thirdparty_override (OPENBLAS v0.3.24 ...)` line that already appears in configure logs) or, at minimum, hardcoding `"v0.3.24"` here to match the CMake recipe. The same applies to the MKL-fallback branch at line ~439.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +343,238 @@ def parse_ninja_log(
...
+    for match in pattern.finditer(text):
+        dependency = canonical_dependency(match.group(1))
+        resolutions[dependency] = {
+            "resolution": "source",
+            "pin": match.group(2).strip(),
+            "source_origin": match.group(3).strip(),
+            "override_variable": match.group(4).strip(),
+            "fallback": match.group(5).strip().strip("; ") or "none reported",
+        }
+
```
LHT129, line 402: [note] The `parse_dependency_resolutions` function relies on substring matching against CMake configure output (e.g., `"Using system OpenBLAS as BLAS backend"`, `"Building OpenBLAS from source"`). These strings are not part of a stable API — they come from CMake `message()` calls in `extern/openblas/openblas.cmake` and could be reworded or removed without notice. If any of these messages change, the resolution detection silently falls through to `"unknown"` without any warning.

A defensive improvement would be to log a warning when none of the expected resolution markers are found for a dependency that appears in the Ninja log (i.e., it was actually built but its resolution path is undetected). This would make the metrics self-diagnosing when upstream CMake messages drift.

LHT129:
We are preparing a separate agent-hive task to move the PR ASan cold/warm/no-op benchmark into a manual/scheduled workflow and retain a single normal PR build. The maintainer explicitly selected #2899 as the prerequisite; we will wait for its merge and will not duplicate the collector changes here.

Read-only review of effe97d27828396afde9f16096082d9934ecd780 found two edge cases worth covering before consumers rely on verified no-op:

1. In `capture_build`, if `.ninja_log` is absent but the build command succeeds, the fallback creates empty metrics with `available=True`, and `true_noop` becomes true solely from exit code + zero edges. Missing telemetry should be distinguishable from a verified unchanged existing log; please add a missing-log regression and avoid claiming verified no-op without valid evidence.
2. `--build-dir` controls log discovery, deletion and clean, but the configure/build make commands do not propagate `DEBUG_BUILD_DIR`. A nondefault path therefore configures/builds `./build` while looking elsewhere for its log, interacting with (1). Either wire the directory consistently or reject unsupported nondefault values, with a regression.

Our migration will keep the default build directory and preserve schema v2 / dependency-cache metadata. We also note that `true_noop=False` is currently reported rather than a failing exit; we will document that distinction rather than claim the benchmark enforces it.

No changes to this PR branch have been made by the follow-up task.

LHT129 (approved):
LGTM

scripts/collect_build_metrics.py
wxyucs: Acknowledged as a non-blocking observability enhancement. Unknown is explicit rather than a fabricated pin; duplicating the recipe version here risks drift. Pin extraction can be addressed separately.

scripts/collect_build_metrics.py
wxyucs: Acknowledged as non-blocking diagnostic hardening. The current fallback explicitly reports unknown. Additional warnings can be considered separately without rerunning this PR for optional polish.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +343,238 @@ def parse_ninja_log(
...
+        if not separator or ":" not in declaration:
+            continue
+        name, _ = declaration.split(":", 1)
+        values[name] = value
+    return values
+
+
+def local_dependency_path(root: Path, value: str | None, default: Path) -> Path | None:
+    candidate = Path(value).resolve() if value else default.resolve()
+    if candidate == root or root in candidate.parents:
```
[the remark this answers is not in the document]
wxyucs: Fixed in b82b955c. Dependency paths must resolve to strict descendants of the repository. Regression coverage rejects the root, a symlink back to it, and paths outside it, while preserving valid dependency subdirectories. Nonexistent descendants remain allowed because they can be populated later; they do not cause root traversal.

scripts/collect_build_metrics.py
```diff
@@ -247,13 +681,28 @@ def capture_build(self, name: str) -> None:
...
+                read_compile_commands(self.build_dir / "compile_commands.json"),
+                read_link_outputs(self.build_dir),
+            )
+        else:
+            phase["ninja"] = empty_ninja_metrics(available=phase["exit_code"] == 0)
+        phase["measurement_mode"] = measurement_mode
+        phase["ninja_edges"] = phase["ninja"]["edge_count"]
+        phase["build_edges"] = phase["ninja"]["build_edge_count"]
+        if measurement_mode == "no-op":
+            phase["true_noop"] = phase["exit_code"] == 0 and phase["build_edges"] == 0
```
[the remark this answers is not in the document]
wxyucs: Fixed in b82b955c. Missing Ninja logs now set available=false, expose null phase edge counts (n/a in Markdown), and cannot verify a true no-op. Stale captured logs are removed before capture. Regressions cover successful and failed builds with missing logs, stale captures, and the existing unchanged-log case that still verifies a no-op.

wxyucs:
Updated in b82b955c with focused regressions for both substantive review findings. Repository-root dependency paths are rejected, and absent Ninja logs now report unavailable edge counts without verifying a no-op. A stale captured log cannot supply evidence for a later measurement.

The separate build-directory review finding is also fixed: configure and build both receive the selected directory, with coverage for a custom path containing spaces. The two optional OpenBLAS comments remain resolved without further changes.

Validation: 23 focused collector/workflow tests, 21 script regression tests, and 4 linter unit tests passed. Fresh CI has been dispatched; previous ASan, unit, functional, Python-build, TSAN, and format passes belong to the earlier head.

The earlier Python failure is the repeated SINDI `statistics["complete"]` assertion tracked in #2907. Compatibility failed at index creation because the `v0.14.1_hgraph` build-parameter JSON was empty or unreadable; that input comes from the compatibility fixture, and this PR does not change its reader. No matching tracker was found, so this remains an external-fixture blocker with the underlying missing/empty-file cause unconfirmed. Neither failure was blindly rerun.

LHT129 (commented):
Reviewed the latest commit (b82b955c). The changes look solid:

- `collect_build_metrics.py`: The dependency classification, edge deduplication, and ninja log delta logic are well-structured and correctly handle edge cases (missing logs, stale captures, multi-output edges, build-maintenance filtering). The `local_dependency_path` fix properly rejects root and out-of-tree paths. The `cache_state` classification logic in `pr-ci.yml` correctly distinguishes exact-hit, fallback-hit, and miss based on `actions/cache/restore@v4` output semantics.
- `pr-ci.yml`: The split of FetchContent preparation and ExternalProject archive restore into separately timed steps is clean. The restore-only cache policy is correctly enforced.
- `daily_test.yml`: The explicit `-DVSAG_USE_SYSTEM_OPENBLAS=OFF` override correctly guards the source-build path.
- Tests: Comprehensive coverage for classification, deduplication, no-op verification, dependency resolution parsing, cmake cache parsing, and workflow regression checks.

No blocking issues found. The earlier review comments have been addressed in this revision.

scripts/collect_build_metrics.py
```diff
@@ -17,6 +18,56 @@
...
+    "cpuinfo": "cpuinfo",
+    "fmt": "fmt",
+    "httplib": "httplib",
+    "json": "nlohmann_json",
+    "roaringbitmap": "roaringbitmap",
+    "tabulate": "tabulate",
+    "thread_pool": "thread_pool",
+    "tsl": "tsl",
+    "yaml_cpp": "yaml-cpp",
+}
```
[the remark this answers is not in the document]
wxyucs: Fixed in cb786730. Added the pybind11 FetchContent directory mapping. The regression follows normalized Ninja outputs into the dependency report and checks preparation classification plus exact nonzero source/build sizes for both _deps and .ci-fetchcontent layouts. It failed on both layouts before the fix; all 24 focused collector and dependency-workflow tests now pass. Previous fixes are preserved.

LHT129 (commented):
## Review Summary

This PR substantially improves the trustworthiness of dependency build metrics by:

1. **Separating dependency preparation timing** into FetchContent source preparation, ExternalProject archive cache restore, and total preparation phases — each measured independently.
2. **Classifying dependency edges per-dependency and per-stage** (mkdir, download, configure, build, etc.) in the Ninja log parser, with proper deduplication of multi-output edges.
3. **Adding dependency resolution tracking** that parses CMake configure output to determine whether each dependency was resolved from source or system, with fallback reporting.
4. **Measuring dependency storage sizes** (FetchContent source/build trees, ExternalProject source/build/install trees, archive cache).
5. **Making the daily x86 guard explicit** about `-DVSAG_USE_SYSTEM_OPENBLAS=OFF` instead of relying on implicit defaults.
6. **Switching PR CI to `actions/cache/restore@v4`** (restore-only) to avoid polluting the cache with PR builds.
7. **Adding comprehensive tests** (24 new test methods across two test files) covering classification, deduplication, delta calculation, no-op verification, and workflow regression checks.

### Observations

- The `classify_outputs` priority scheme correctly handles multi-output Ninja edges by preferring ExternalProject stage classification over generic categories.
- The `ninja_log_delta` approach for capturing only new records between build phases is sound and avoids re-parsing the entire log.
- The `local_dependency_path` function correctly prevents directory size traversal from escaping the repository root.
- The `read_cmake_cache` parser properly handles values containing `=` signs by using `partition("=")`.
- All previously raised review comments have been addressed or acknowledged with clear rationale.

No blocking issues found. The changes are well-tested and the implementation is thorough.

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured and the code quality is high.

**Summary of changes:**
- `daily_test.yml`: Explicitly sets `-DVSAG_USE_SYSTEM_OPENBLAS=OFF` to guard the bundled source-build path, replacing the previous implicit default behavior. This is a clear improvement for trustworthiness.
- `pr-ci.yml`: Separates FetchContent preparation and ExternalProject archive restore into independently timed steps. Switches from `actions/cache@v4` (which saves cache entries) to `actions/cache/restore@v4` (restore-only), preventing PR jobs from polluting the shared cache. Passes new metrics parameters (cache state, matched key, write policy, split timings) to the collector script.
- `scripts/collect_build_metrics.py`: Major refactor that introduces per-dependency classification, stage tracking, deduplication of multi-output Ninja edges, dependency resolution parsing, storage sizing, and a structured dependency report. The `classify_output` / `classify_outputs` approach correctly prioritizes external project stages over generic categories.
- `tests/scripts/test_collect_build_metrics.py`: Comprehensive test additions covering dependency path validation, missing Ninja log handling, edge classification, deduplication, pybind11 FetchContent layouts, unclassified source dependencies, cmake cache parsing, and measurement ordering.
- `tests/scripts/test_dependency_workflows.py`: New regression tests verifying workflow structure (coverage vs. metrics test separation, explicit OpenBLAS disable, restore-only cache policy, hash validation in ExternalProject recipes).

**Review findings:**
- No critical issues found.
- No blocking suggestions.
- All previously raised review comments have been addressed by the author.
- Test coverage is thorough (24 focused tests + 21 regression discovery tests + 4 linter tests all passing).
- The `local_dependency_path` guard correctly rejects repository-root traversal and symlink aliases.
- The `ninja_log_delta` correctly handles both append and compaction scenarios.
- Unclassified dependencies are explicitly reported as "source preparation not classified" rather than misreported as host dependencies.

This PR looks ready to merge once CI passes on the current head (cb786730)."

scripts/collect_build_metrics.py
```diff
@@ -17,6 +18,57 @@
 from typing import Any
 
 
+DEPENDENCY_ALIASES = {
+    "nlohmann_json": "json",
+    "nlohmann-json": "json",
```
LHT129, line 23: [suggestion] In `parse_ninja_log`, when `fields[4]` is an empty string, the edge deduplication key falls back to `(start_ms, end_ms, fields[2], output)` instead of `(start_ms, end_ms, fields[2], fields[4])`. This is correct in practice because empty hash strings are not meaningful for deduplication, but the intent would be clearer if the condition explicitly checked for a non-empty hash rather than relying on falsy string evaluation:

```python
if len(fields) >= 5 and fields[4] != "":
    edge_key = (start_ms, end_ms, fields[2], fields[4])
else:
    edge_key = (start_ms, end_ms, fields[2], output)
```

This makes the intent explicit and avoids any ambiguity if ninja ever emits a hash field that is falsy for other reasons.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +119,127 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+    for marker in (".ci-fetchcontent", "_deps"):
+        if marker not in components:
+            continue
+        marker_index = components.index(marker)
+        if marker_index + 1 >= len(components):
+            continue
+        directory = components[marker_index + 1]
+        match = re.fullmatch(r"(.+)-(?:build|src|subbuild)", directory)
+        if match:
+            return canonical_dependency(match.group(1))
```
LHT129, line 159: [suggestion] In `parse_ninja_log`, when `fields[4]` is an empty string, the edge deduplication key falls back to `(start_ms, end_ms, fields[2], output)` instead of `(start_ms, end_ms, fields[2], fields[4])`. This is correct in practice because empty hash strings are not meaningful for deduplication, but the intent would be clearer if the condition explicitly checked for a non-empty hash rather than relying on falsy string evaluation:

```python
if len(fields) >= 5 and fields[4] != "":
    edge_key = (start_ms, end_ms, fields[2], fields[4])
else:
    edge_key = (start_ms, end_ms, fields[2], output)
```

This makes the intent explicit and avoids any ambiguity if ninja ever emits a hash field that is falsy for other reasons.

scripts/collect_build_metrics.py
```diff
@@ -126,13 +287,55 @@ def parse_ninja_log(
             start_ms, end_ms = int(fields[0]), int(fields[1])
         except ValueError:
             continue
+        raw_log_records += 1
         output = fields[3]
-        duration = max(0, end_ms - start_ms) / 1000
-        category = classify_edge(output, link_outputs)
+        if len(fields) >= 5 and fields[4]:
```
LHT129, line 292: [suggestion] In `parse_ninja_log`, when `fields[4]` is an empty string, the edge deduplication key falls back to `(start_ms, end_ms, fields[2], output)` instead of `(start_ms, end_ms, fields[2], fields[4])`. This is correct in practice because empty hash strings are not meaningful for deduplication, but the intent would be clearer if the condition explicitly checked for a non-empty hash rather than relying on falsy string evaluation:

```python
if len(fields) >= 5 and fields[4] != "":
    edge_key = (start_ms, end_ms, fields[2], fields[4])
else:
    edge_key = (start_ms, end_ms, fields[2], output)
```

This makes the intent explicit and avoids any ambiguity if ninja ever emits a hash field that is falsy for other reasons.

scripts/collect_build_metrics.py
wxyucs: Thanks. This is optional stylistic polish: the field is a string, and the existing truthiness check correctly distinguishes an empty hash. Keeping the working implementation unchanged to avoid unnecessary churn.

scripts/collect_build_metrics.py
wxyucs: Thanks. This is optional stylistic polish: the field is a string, and the existing truthiness check correctly distinguishes an empty hash. Keeping the working implementation unchanged to avoid unnecessary churn.

scripts/collect_build_metrics.py
wxyucs: Thanks. This is optional stylistic polish: the field is a string, and the existing truthiness check correctly distinguishes an empty hash. Keeping the working implementation unchanged to avoid unnecessary churn.

scripts/collect_build_metrics.py
```diff
@@ -232,28 +666,47 @@ def ccache(self, *arguments: str) -> dict[str, Any]:
...
-        ninja_log.unlink(missing_ok=True)
+        before_header = read_ninja_log_header(ninja_log)
+        before_records = read_ninja_log_records(ninja_log)
         self.ccache("--zero-stats")
         phase = self.run_logged(
             name,
             [
                 "make",
                 "build-asan",
+                f"DEBUG_BUILD_DIR={shlex.quote(str(self.build_dir))}",
```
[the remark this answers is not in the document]
wxyucs: Fixed in 32651eb1. End-to-end verification found a nuance: the previous collector did handle spaces because Make's recipe shell interpreted the embedded quotes. The reproducible failure was a direct raw `DEBUG_BUILD_DIR` override: CMake received a split path and the build failed with `Unknown argument build`.

The collector now passes raw argv values, and Make quotes the build-directory variable consistently, including its default and clean recipe. Added real Make/CMake/Ninja regressions for default and space-containing paths and the full configure → clean build → clean target → warm-cache build → verified no-op sequence. All 26 focused collector/workflow tests, 25 other Python regressions, and six focused CMake checks pass. Full CI is pending on the new head.

LHT129 (commented):
Reviewed the latest commit (32651eb). This PR adds trustworthy dependency build metrics with proper separation of FetchContent preparation, ExternalProject cache restore, per-dependency edge classification, deduplication, and storage reporting. All previously raised review comments have been addressed (local_dependency_path strictness, unclassified preparation fallback, openblas pin). The 26 new collector tests and 4 workflow regression tests provide solid coverage. No new issues found.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +344,238 @@ def parse_ninja_log(
...
+    return resolutions
+
+
+def directory_size(path: Path | None) -> int:
+    if path is None or not path.exists():
+        return 0
+    if path.is_file():
+        return path.stat().st_size
+    total = 0
+    for root, _, files in os.walk(path, followlinks=False):
```
LHT129, line 454: [note] The `directory_size` function at line 454 uses `root` as the `os.walk` loop variable, which shadows the `root` parameter name used in `local_dependency_path` (line 480). While these are separate module-level functions and there is no actual bug, reusing the same name for different purposes within the same module can confuse readers. Consider renaming the loop variable to `dirpath` (the conventional `os.walk` name) for clarity.

```python
# Current:
for root, _, files in os.walk(path, followlinks=False):
    candidate = Path(root) / filename

# Suggested:
for dirpath, _, files in os.walk(path, followlinks=False):
    candidate = Path(dirpath) / filename
```

wxyucs: These names are local to separate functions and do not shadow each other. This is optional naming polish with no correctness issue, so the implementation is intentionally unchanged.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +344,238 @@ def parse_ninja_log(
...
+        install_bytes = 0
+        if details["resolution"] == "system":
+            details["preparation"] = "host dependency"
+        elif dependency in EXTERNAL_PROJECT_DEPENDENCIES:
+            dependency_dir = build_dir / dependency
+            source_bytes = directory_size(dependency_dir / "source")
+            install_bytes = directory_size(dependency_dir / "install")
+            build_bytes = max(
+                0, directory_size(dependency_dir) - source_bytes - install_bytes
+            )
```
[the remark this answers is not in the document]
wxyucs: Fixed in 518c28415ebdd1711e27b391fd4813c95f8bf907. The misleading source_bytes/build_bytes split is replaced by local_bytes in JSON schema version 3. ExternalProject prefixes are counted once, including source, nested build, install and metadata files; FetchContent sums sibling source/build trees. English/Chinese docs and Markdown explain the scope. Regression fixtures cover HDF5, OpenBLAS and ANTLR4 layouts without double counting; all 23 collector and 4 workflow tests passed.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +119,127 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+    if normalized_link_outputs and normalized in normalized_link_outputs:
+        return {"category": "link", "dependency": None, "stage": None}
     if normalized.endswith((".a", ".so", ".dylib", ".dll", ".exe")):
-        return "link"
-    return "other"
+        return {"category": "link", "dependency": None, "stage": None}
+    return {"category": "other", "dependency": None, "stage": None}
+
+
+def classify_edge(output: str, link_outputs: set[str] | None = None) -> str:
```
LHT129, line 217: [note] `classify_edge` is defined but never called in this file or anywhere in the repository. It appears to be a backward-compatibility wrapper around the new `classify_output` function, but since the old call site (removed in this PR) was the only usage, this function is now dead code.

Consider removing it to avoid confusion for future readers who might mistake it for the primary classification API when `classify_output` / `classify_outputs` are the actual entry points used by `parse_ninja_log`.

LHT129 (commented):
Reviewed the latest changes (SHA: 518c2841). All previously identified issues have been addressed through the review rounds:

- **Deduplication**: Adjacent-record-only merging with regression coverage (3a911528)
- **Link output normalization**: Pre-normalized once per parse, reused across edges (3a911528)
- **ExternalProject precedence**: Only applied when positively identified by stamp/build-info parser (3a911528)
- **classify_outputs recomputation**: Now reuses `details` instead of reclassifying (09fc5e70)
- **CMake cache parsing**: Uses `partition` first, validates `:` only in declaration (09fc5e70)
- **OpenBLAS pin**: Reports `"unknown"` instead of hardcoded version (09fc5e70)
- **Ninja log header**: Preserves original version header from active log (09fc5e70)
- **EXTERNAL_PROJECT_STAGES**: Ordered tuple for deterministic iteration (09fc5e70)
- **Unclassified fallback**: Reports classification-path-unknown message (09fc5e70)
- **Identity aliases**: Removed redundant self-mappings (c878cfc2)
- **local_dependency_path**: Strict descendant check, rejects root (b82b955c)
- **Missing ninja log**: Sets `available=False` instead of reporting `0` edges (b82b955c)

No new issues of substance found. The remaining minor code patterns (redundant `normalize_output` inside `is_build_maintenance_output`, duplicate `external_project_stage` call chain in `classify_output`) are negligible performance nits not worth blocking.

PR is mergeable. No approval — leaving final decision to maintainers.

scripts/collect_build_metrics.py
wxyucs: This is optional cleanup of a compatibility wrapper, not a correctness issue. Keeping it unchanged in this review iteration to avoid unrelated churn; the actionable generator and storage-scope findings are being handled separately.

scripts/collect_build_metrics.py
```diff
@@ -232,28 +660,47 @@ def ccache(self, *arguments: str) -> dict[str, Any]:
...
+        before_records = read_ninja_log_records(ninja_log)
         self.ccache("--zero-stats")
         phase = self.run_logged(
             name,
             [
                 "make",
                 "build-asan",
+                f"DEBUG_BUILD_DIR={self.build_dir}",
                 f"COMPILE_JOBS={self.args.jobs}",
                 "CMAKE_BUILD_ARGS=-d stats",
```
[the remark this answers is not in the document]
wxyucs: Fixed in 9073806b. Build arguments now use the generator recorded in CMakeCache.txt: only Ninja generators receive -d stats. A real Make/CMake regression makes Ninja unavailable and verifies configure, clean, warm-cache, and no-op collection succeed with unavailable Ninja metrics and no false verified-no-op claim. It fails against the previous collector; all 24 collector tests and four workflow tests pass.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +344,232 @@ def parse_ninja_log(
...
+        details["stages"] = ninja_dependencies.get(dependency, {}).get("stages", {})
+
+        local_bytes = 0
+        if details["resolution"] == "system":
+            details["preparation"] = "host dependency"
+        elif dependency in EXTERNAL_PROJECT_DEPENDENCIES:
+            dependency_dir = build_dir / dependency
+            # Source trees can contain build artifacts; measure the whole prefix once.
+            local_bytes = directory_size(dependency_dir)
+            details["preparation"] = "verified ExternalProject archive"
```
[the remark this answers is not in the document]
wxyucs: Fixed in 9073806b by clarifying the measured scope. Verified the dependency recipes and existing build artifacts: TMP_DIR/STAMP_DIR use shared BUILD_INFO_DIR outside the dependency prefixes. Markdown, English/Chinese guides, and the PR description now explicitly exclude those temporary files, stamps, logs, and metadata from both local_bytes and external_build_bytes. Regression fixtures include shared metadata and verify prefix-only measurement; rendering checks require the exclusion text. Documentation validation passes after rebasing onto main with #2950.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +119,127 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+        suffix = f"-{stage}"
+        if basename.endswith(suffix) and len(basename) > len(suffix):
+            dependency = canonical_dependency(basename[: -len(suffix)])
+            return dependency, stage
+    return None
+
+
+def dependency_from_output(output: str) -> str | None:
+    normalized = normalize_output(output)
+    stage = external_project_stage(normalized)
```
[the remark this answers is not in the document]
wxyucs: This is optional normalization cleanup with no reported incorrect result. Keeping the current interfaces and behavior unchanged rather than expanding this correctness fix.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +119,127 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
+        return {"category": "link", "dependency": None, "stage": None}
     if normalized.endswith((".a", ".so", ".dylib", ".dll", ".exe")):
-        return "link"
-    return "other"
+        return {"category": "link", "dependency": None, "stage": None}
+    return {"category": "other", "dependency": None, "stage": None}
+
+
+def classify_edge(output: str, link_outputs: set[str] | None = None) -> str:
+    normalized_link_outputs = {normalize_output(item) for item in link_outputs or set()}
```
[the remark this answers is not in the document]
wxyucs: This is a speculative optimization of a compatibility helper, not an observed parsing bottleneck. No correctness issue is identified; leaving it unchanged.

Makefile
```diff
@@ -93,18 +93,18 @@ help:                    ## Show the help.
...
 
 .PHONY: test
 test:                    ## Build and run unit tests.
-	cmake ${VSAG_CMAKE_ARGS} -B${DEBUG_BUILD_DIR} -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=OFF -DENABLE_CCACHE=ON -DENABLE_TESTS=ON
-	cmake --build ${DEBUG_BUILD_DIR} --parallel ${COMPILE_JOBS}
+	cmake ${VSAG_CMAKE_ARGS} -B"${DEBUG_BUILD_DIR}" -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=OFF -DENABLE_CCACHE=ON -DENABLE_TESTS=ON
+	cmake --build "${DEBUG_BUILD_DIR}" --parallel ${COMPILE_JOBS}
 	./build/tests/unittests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
 	./build/tests/functests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
 	./build/tests/eval_monitor_test -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
```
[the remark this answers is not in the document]
wxyucs: Fixed in fde6bfdd. `make test` now quotes and uses `DEBUG_BUILD_DIR` for all three test executables.

The real Make/CMake regression runs executable fixtures in the default directory, a custom relative directory, and relative/absolute directories containing spaces, and checks that the filter and shard arguments reach all three binaries unchanged. The three custom-path cases fail before the fix; all 25 collector tests and four workflow tests pass after it.

tests/scripts/test_collect_build_metrics.py
```diff
@@ -20,6 +23,60 @@
 
 
 class BuildMetricsTest(unittest.TestCase):
+    def test_dependency_paths_require_strict_repository_descendants(self) -> None:
+        with tempfile.TemporaryDirectory() as directory:
+            root = Path(directory).resolve()
+            default = root / "_deps"
+            alias = root / "root-alias"
+            alias.symlink_to(root, target_is_directory=True)
```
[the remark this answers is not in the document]
wxyucs: The documented C++ development scope is Linux (Ubuntu 20.04+/CentOS 7+) and macOS 14+ arm64, with CI coverage on Linux and macOS. Windows without symlink privileges is outside that supported scope. This fixture creates a directory symlink inside its own temporary directory, and the focused test passes.

There is no concrete failure on a supported platform here. Retaining the symlink assertion preserves coverage of an alias that resolves to the repository root; no capability skip is added for this optional portability suggestion.

tests/scripts/test_dependency_workflows.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertIn("uses: actions/cache/restore@v4", pr_ci)
+        self.assertNotIn("uses: actions/cache@v4", pr_ci)
+        self.assertNotIn("uses: actions/cache/save@v4", pr_ci)
+        self.assertIn("dependency-cache-write-policy restore-only", pr_ci)
+
+    def test_external_archives_keep_hash_validation_and_pin_overrides(self) -> None:
+        for dependency in ("antlr4", "hdf5", "openblas"):
+            recipe = (ROOT / f"extern/{dependency}/{dependency}.cmake").read_text(
+                encoding="utf-8"
+            )
```
LHT129, line 63: [suggestion] The `test_external_archives_keep_hash_validation_and_pin_overrides` test reads `extern/{dependency}/{dependency}.cmake` directly without checking file existence first. If a dependency is removed or renamed, the test will crash with `FileNotFoundError` rather than producing a clear assertion failure. Consider wrapping the `read_text` call with `self.assertTrue(recipe.is_file(), ...)` before reading.

```python
recipe = ROOT / f"extern/{dependency}/{dependency}.cmake"
self.assertTrue(recipe.is_file(), f"Recipe not found: {recipe}")
```

tests/scripts/test_dependency_workflows.py
```diff
@@ -0,0 +1,61 @@
...
+
+ROOT = Path(__file__).resolve().parents[2]
+
+
+def job_text(workflow: Path, job_name: str) -> str:
+    text = workflow.read_text(encoding="utf-8")
+    match = re.search(
+        rf"(?ms)^  {re.escape(job_name)}:\n(?P<body>.*?)(?=^  [a-zA-Z0-9_-]+:\n|\Z)", text
+    )
+    if match is None:
```
LHT129, line 17: [note] The `job_text` regex uses `[a-zA-Z0-9_-]` as the character class for matching the next job name. While this covers all current VSAG workflow job names, YAML mapping keys can technically contain a wider set of characters. If a future job name includes characters outside this class (e.g., a job with a name containing `.` or a parent key like `runs-on`), the regex would fail to terminate the body capture correctly. This is a minor robustness concern for the current codebase.

tests/scripts/test_collect_build_metrics.py
```diff
@@ -62,14 +347,301 @@ def test_rendered_summary_uses_real_newlines(self) -> None:
...
+                environment["PATH"] = str(bindir) + os.pathsep + environment["PATH"]
+                with patch.dict(os.environ, environment, clear=True):
+                    self.assertEqual(collector.collect(), 0)
+                self.assertEqual(METRICS.read_cmake_cache(
+                    collector.build_dir / "CMakeCache.txt")["CMAKE_GENERATOR"], "Unix Makefiles")
+                self.assertTrue((collector.build_dir / "artifact.txt").is_file())
+                for phase in collector.phases:
+                    self.assertEqual(phase["exit_code"], 0)
+                    if phase["name"] in ("clean_build", "warm_cache_build", "noop_build"):
+                        self.assertIn("CMAKE_BUILD_ARGS=", phase["command"])
```
LHT129, line 654: [note] The `test_collection_supports_unix_makefiles_fallback` test creates a fake `ninja` script that always exits with code 1. This works correctly but means the test is coupled to the specific error-handling path where `ninja` is unavailable. If the underlying code changes how it detects or reports a missing `ninja` (e.g., checking for the binary before attempting to run it), this test would silently stop exercising the fallback path. Consider also asserting on the specific error message or return code to make the coupling explicit.

LHT129 (commented):
[suggestion] The `test` target was updated to use `"${DEBUG_BUILD_DIR}/tests/..."` for test binary paths (supporting spaces in build directories), but the `test_asan` and `test_tsan` targets in the same Makefile still use hardcoded `./build/tests/...` paths. These targets would break when `DEBUG_BUILD_DIR` is set to a path containing spaces or a non-default value. Consider updating them consistently:

```makefile
test_asan: asan
	"${DEBUG_BUILD_DIR}/tests/unittests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
	"${DEBUG_BUILD_DIR}/tests/functests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}

test_tsan: tsan
	"${DEBUG_BUILD_DIR}/tests/unittests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
	"${DEBUG_BUILD_DIR}/tests/functests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
```

Similarly, the `test-module` target still uses unquoted `${DEBUG_BUILD_DIR}` on lines 120-122, which would also break with spaces.

LHT129 (commented):
test review

LHT129 (commented):
[suggestion] The `test` target was updated to use `"${DEBUG_BUILD_DIR}/tests/..."` for test binary paths (supporting spaces in build directories), but the `test_asan` and `test_tsan` targets in the same Makefile still use hardcoded `./build/tests/...` paths. These targets would break when `DEBUG_BUILD_DIR` is set to a path containing spaces or a non-default value. Consider updating them consistently:

```makefile
test_asan: asan
	"${DEBUG_BUILD_DIR}/tests/unittests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
	"${DEBUG_BUILD_DIR}/tests/functests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}

test_tsan: tsan
	"${DEBUG_BUILD_DIR}/tests/unittests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
	"${DEBUG_BUILD_DIR}/tests/functests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
```

Similarly, the `test-module` target still uses unquoted `${DEBUG_BUILD_DIR}` on lines 120-122, which would also break with spaces.

LHT129 (commented):
Reviewed the full diff (8 files, +1421/-99) covering the schema v3 metrics collector, CI workflow timing split, Makefile space-safe paths, documentation, and 25+ new tests.

**Summary**: The changes are well-structured and address the stated goals of making dependency build metrics trustworthy. Prior review rounds have already resolved the main concerns (unclassified dependency fallback, path traversal guard, OpenBLAS pin accuracy). No blocking issues found.

**What this PR does well**:
- `Makefile`: Proper quoting of `${DEBUG_BUILD_DIR}` throughout all targets fixes builds with space-containing paths. The `test` target now correctly uses the variable instead of hardcoding `./build/`.
- `pr-ci.yml`: Splitting the dependency preparation timer into FetchContent preparation and archive cache restore phases gives accurate per-phase timing. Using `actions/cache/restore@v4` (restore-only) correctly avoids polluting the PR cache with source archives.
- `daily_test.yml`: Explicit `-DVSAG_USE_SYSTEM_OPENBLAS=OFF` via `EXTRA_DEFINED` is more robust than relying on the default, since `AUTO` would pick up the system installation in the CI image.
- `collect_build_metrics.py`: Schema v3 with `local_bytes` correctly handles in-source builds (HDF5, OpenBLAS) by measuring the whole prefix once. Ninja edge deduplication, build maintenance filtering, and generator-aware stats (`-d stats` only for Ninja) are all solid.
- `test_collect_build_metrics.py` / `test_dependency_workflows.py`: Comprehensive test coverage including real Make/CMake/Ninja fixtures, Unix Makefiles fallback, space-safe paths, and workflow regression guards.

**Minor observations (non-blocking)**:
- The `external_project_stage` function uses `f"/{normalized}"` as a prefix trick for the `in` check on `.vsag-build-info`. This works correctly but is slightly unusual — a comment noting the intent would help future readers.
- The `parse_dependency_resolutions` regex and string matching depend on specific CMake log output formats. The tests cover the current format, but if those log messages change upstream, the resolution parsing would silently degrade. This is inherent to log parsing and not a defect in this PR.

No changes requested. This looks ready to merge once CI passes.

scripts/collect_build_metrics.py
```diff
@@ -141,7 +344,233 @@ def parse_ninja_log(
...
+            details["preparation"] = f"{details['resolution']} preparation not classified"
+            if details["fallback"] == "none reported":
+                details["fallback"] = "classification path unknown; metrics may be incomplete"
+
+        policy = cmake_cache.get(
+            f"VSAG_USE_SYSTEM_{dependency.upper()}", cmake_cache.get("VSAG_USE_SYSTEM_DEPS", "AUTO")
+        )
+        if dependency == "yaml_cpp":
+            policy = cmake_cache.get("VSAG_USE_SYSTEM_YAML_CPP", policy)
+        details["system_policy"] = policy or cmake_cache.get("VSAG_USE_SYSTEM_DEPS", "AUTO")
```
LHT129: [suggestion] The `fallback` field is unconditionally overwritten for `system`+`ON` and `source`+`OFF` policy combinations, which discards the resolution-specific fallback information parsed from configure output (e.g., "bundled source fallback available when policy is AUTO" for system OpenBLAS). When policy is explicitly ON, the overwritten value "none; system dependency required" is accurate for the current configuration but loses the context of what the alternative would be.

Consider preserving the original fallback when it provides additional context, for example:

```python
if details["resolution"] == "system" and details["system_policy"] == "ON":
    if details["fallback"] == "none reported":
        details["fallback"] = "none; system dependency required"
elif details["resolution"] == "source" and details["system_policy"] == "OFF":
    if details["fallback"] == "none reported":
        details["fallback"] = "system dependency disabled; source hash validation required"
```

This would keep the richer fallback descriptions from `parse_dependency_resolutions` when available while still providing sensible defaults for unclassified dependencies.

tests/scripts/test_dependency_workflows.py
wxyucs: This is optional diagnostic polish. A missing recipe already fails with its path; no correctness change is needed.

tests/scripts/test_dependency_workflows.py
wxyucs: This is a hypothetical future naming concern rather than a defect in the current controlled workflow fixture. Keeping the scoped implementation.

tests/scripts/test_collect_build_metrics.py
wxyucs: This is optional future-proofing without an observed failure in the current fallback regression. No additional code churn is needed.

Makefile
```diff
@@ -11,7 +11,7 @@ endif
 CMAKE_INSTALL_PREFIX ?= "/usr/local/"
 COMPILE_JOBS ?= 6
 CMAKE_BUILD_ARGS ?=
-DEBUG_BUILD_DIR ?= "./build/"
+DEBUG_BUILD_DIR ?= ./build/
 RELEASE_BUILD_DIR ?= "./build-release/"
 PERF_RELEASE_BUILD_DIR ?= "./build-release-perf/"
```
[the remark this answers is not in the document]
wxyucs: Unifying other build-directory variables is optional cleanup outside this debug-path change. No concrete regression in those unchanged recipes is identified; retaining scope.

scripts/collect_build_metrics.py
wxyucs: The explicit ON/OFF policy description intentionally reports the current configuration rather than an alternative AUTO policy. Keeping this accurate scoped behavior.

.github/workflows/pr-ci.yml
```diff
@@ -117,23 +117,40 @@ jobs:
...
+          elapsed = (now - int(os.environ["STARTED_NS"])) / 1_000_000_000
+          print(f"seconds={elapsed:.3f}")
+          print(f"cache_started_ns={now}")
+          PY
+      - name: Restore ExternalProject Archives
+        id: dependency-archive-cache
+        uses: actions/cache/restore@v4
         with:
           path: .ci-downloads
           key: external-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('extern/**/*.cmake') }}
```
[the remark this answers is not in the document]
wxyucs: Verified the actual configuration chain and the successful CI artifact: the ASan job already exports `VSAG_THIRDPARTY_DOWNLOAD_DIR` at job scope to `.ci-downloads`; Make forwards it as `-DDOWNLOAD_DIR`, and the collector inherits the same variable. Without that override, both CMake and the collector default to `build/.vsag-downloads`.

The [CI metrics run](https://github.com/antgroup/vsag/actions/runs/34819479530) records the `.ci-downloads` configure argument and `archive_bytes = 41084997` (cache miss, followed by dependency downloads). There is no directory mismatch requiring a production change.

Added focused regressions in 960fdb78 using the real Makefile and CMake options: verify both default and CI paths, count restored plus downloaded archive bytes, and exclude an unrelated default directory when overridden. A workflow regression requires the export at job scope. All 31 focused tests pass; normal CI is running for the new head.

scripts/collect_build_metrics.py
```diff
@@ -67,31 +119,127 @@ def read_compile_commands(path: Path) -> dict[str, str]:
...
-def classify_edge(output: str, link_outputs: set[str] | None = None) -> str:
-    normalized = output.replace("\\", "/").lower()
+def classify_output(
+    output: str, normalized_link_outputs: set[str] | None = None
+) -> dict[str, str | None]:
+    normalized = normalize_output(output)
+    stage = external_project_stage(normalized)
+    dependency = dependency_from_output(normalized)
+    if is_build_maintenance_output(normalized):
+        return {"category": "build_maintenance", "dependency": None, "stage": None}
```
LHT129, line 186: [suggestion] `local_dependency_path` resolves relative `value` paths against the process CWD (`Path(value).resolve()`) rather than against the `root` parameter. When `VSAG_FETCHCONTENT_BASE_DIR` or `VSAG_THIRDPARTY_DOWNLOAD_DIR` is set to a relative path, this could resolve to an unexpected location if the process CWD differs from the repository root, causing the function to return `None` and silently skip size measurements.

Consider resolving relative paths against `root`:

```python
def local_dependency_path(root: Path, value: str | None, default: Path) -> Path | None:
    candidate = (root / value).resolve() if value else default.resolve()
    if root.resolve() in candidate.parents:
        return candidate
    return None
```

This ensures relative paths are always interpreted relative to the repository root, matching the intent of the containment check on the next line.

wxyucs: No production mismatch found: Collector sets self.root = Path.cwd().resolve(), never changes the process directory, and launches Make with cwd=self.root. Therefore relative environment paths already resolve against the same repository root during collection. Keeping this optional helper change out of scope; no code or test-only changes were added for this suggestion.

Makefile
```diff
@@ -93,21 +93,21 @@ help:                    ## Show the help.
...
-	cmake ${VSAG_CMAKE_ARGS} -B${DEBUG_BUILD_DIR} -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=OFF -DENABLE_CCACHE=ON -DENABLE_TESTS=ON
-	cmake --build ${DEBUG_BUILD_DIR} --parallel ${COMPILE_JOBS}
-	./build/tests/unittests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
-	./build/tests/functests -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
-	./build/tests/eval_monitor_test -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
+	cmake ${VSAG_CMAKE_ARGS} -B"${DEBUG_BUILD_DIR}" -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=OFF -DENABLE_CCACHE=ON -DENABLE_TESTS=ON
+	cmake --build "${DEBUG_BUILD_DIR}" --parallel ${COMPILE_JOBS}
+	"${DEBUG_BUILD_DIR}/tests/unittests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
+	"${DEBUG_BUILD_DIR}/tests/functests" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
+	"${DEBUG_BUILD_DIR}/tests/eval_monitor_test" -d yes ${UT_FILTER} --allow-running-no-tests ${UT_SHARD}
```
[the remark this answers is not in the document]
wxyucs: Fixed in 55d4e5dd: quote DEBUG_BUILD_DIR in the three test-module configure/build/executable commands. A real Make/CMake regression verifies default, custom relative, and relative/absolute paths containing spaces, plus CASE and SHARD forwarding. Both space-containing cases fail before the fix and pass afterward.

scripts/collect_build_metrics.py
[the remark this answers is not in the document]
wxyucs: Fixed in 55d4e5dd: uppercase the effective policy after per-dependency/global fallback selection, matching vsag_get_system_dep_policy. Regression coverage includes mixed-case global values, empty overrides, and mixed-case per-dependency overrides, checking normalized policy and required/disabled fallback descriptions.

scripts/collect_build_metrics.py
```diff
@@ -263,26 +713,65 @@ def collect(self) -> int:
...
+        fetchcontent_dir = local_dependency_path(
+            self.root,
+            os.environ.get("VSAG_FETCHCONTENT_BASE_DIR"),
+            self.build_dir / "_deps",
+        )
+        archive_dir = local_dependency_path(
+            self.root,
+            os.environ.get("VSAG_THIRDPARTY_DOWNLOAD_DIR"),
+            self.build_dir / ".vsag-downloads",
+        )
```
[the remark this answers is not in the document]
wxyucs: Fixed in 55d4e5dd: read FETCHCONTENT_BASE_DIR and DOWNLOAD_DIR from CMakeCache.txt first, retaining environment/default fallbacks when those entries are absent. The real Make/CMake fixture covers defaults, the CI environment override, and explicit cache overrides with a conflicting archive environment path; it verifies configured source/archive byte counts. The cache-override case fails before the fix. All 33 focused tests pass.

PR #2900, opened 2026-09-08
perf(pyramid): accelerate build cache restoration

LHT129:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

## What Changed

- Restore Pyramid build-cache rows through a reusable old-to-new inner-ID table instead of translating every neighbor through source-ID strings.
- Decode and validate cached rows in parallel blocks, then commit graph rows deterministically after a barrier for compressed/sparse graph safety.
- Preserve complete single-layer cache hits without running `ef_construction / 3` refinement on every hit; continue refining multi-layer hits because routing overlays are not included in the cache.
- Add graph-membership hit/miss and restored-edge statistics, regression tests, and synchronized English/Chinese documentation.

## Test Evidence

- [x] `make fmt`
- [x] `make lint`
- [x] `make test`
- [x] `make cov`, run tests, and collect coverage
- [x] Other (describe below)

Test details:

```text
clang-format 15 dry-run and git diff --check: passed
clang-tidy 15 on modified production sources: passed
BuildCache unit tests: 35 assertions passed
PyramidBuildCache unit tests: 11 assertions passed
Pyramid cache functional tests: 108 assertions passed
Changed source coverage: build_cache.cpp 93.5%, pyramid_build_cache.cpp 94.8%

A pre-existing unrelated test failure was reproduced on origin/main (fe17d6d):
Pyramid threshold filters before flat-node pruning

Production-like POC, 1M x 768, 99% cache hit, max_degree=64:
cold build: 4136.737 s
warm build: 1138.379 s (3.63x faster)
cold/warm host-root recall@10: 0.694 / 0.706
cold/warm host-path recall@10: 0.936 / 0.935
cold/warm date-path recall@10: 0.920 / 0.908
```

## Compatibility Impact

- API/ABI compatibility: none; the new raw cache-row lookup is internal.
- Behavior changes: complete cache-hit rows in single-layer Pyramid graphs are retained without refining every hit. Multi-layer routing behavior is unchanged.

- Performance impact: improved. The production-like 1M benchmark is 3.63x faster at 99% cache hit rate.
- Concurrency/thread-safety impact: cache rows are decoded in parallel but graph mutations occur deterministically after all decode workers finish.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/en/src/indexes/pyramid.md`, `docs/docs/zh/src/indexes/pyramid.md`

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert commit `9113baa`; this restores the previous refine-all warm-build behavior.

## Checklist

- [x] Issue link is not required for `kind/improvement`


src/algorithm/pyramid/pyramid_build_cache.cpp
```diff
@@ -85,6 +85,26 @@ PyramidBuildCache::GetGraphCache(const std::string& hierarchy_name,
...
 
+uint64_t
+PyramidBuildCache::CountMatchedSourceIds(const std::string* source_ids, uint64_t count) const {
+    UnorderedSet<std::string> cached_source_ids(allocator_);
+    for (const auto& [key, graph_cache] : graph_caches_) {
+        if (graph_cache == nullptr) {
+            continue;
+        }
+        for (const auto& source_id : graph_cache->source_ids_) {
+            cached_source_ids.emplace(source_id);
```
LHT129, line 97: [suggestion] `CountMatchedSourceIds` builds a full `UnorderedSet<std::string>` of every cached source ID across all graph caches by copying each string via `emplace`. For large production deployments with many hierarchies and millions of vectors, this one-time set construction can be noticeable (both in memory and time). Since this method is called once per `Build()` it is not a hot path, but consider whether the set could be built once when the cache is imported and reused, or whether the overlap check could be done with a cheaper approximate method (e.g., reservoir sampling a subset of input IDs).

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2805,10 +2887,108 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
...
+                                         nullptr);
+                            for (uint64_t index = 0; index < filtered.size(); ++index) {
+                                merged->Push(distances[index], filtered[index]);
+                            }
+                        }
+                        while (not candidates->Empty()) {
+                            const auto candidate = candidates->Top();
+                            candidates->Pop();
+                            if (candidate.second != inner_id &&
+                                seen.emplace(candidate.second).second) {
```
LHT129, line 2980: [question] `build_cache_restored_edges_` is only incremented in the fast-path commit block (when `use_fast_path` is true). The `refine_block` lambda in the single-layer root repair path also restores edges but does not count them. Is this intentional? If the refine path restores a significant number of edges, the counter will under-report the total restored edges.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2805,10 +2887,108 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
...
+                        const auto inner_id = node_hit_ids[offset];
+                        const auto* vector = data_vectors + dim_ * inner_id;
+                        Vector<InnerIdType> existing(allocator_);
+                        graph_node->graph_->GetNeighbors(inner_id, existing);
+
+                        InnerSearchParam search_param;
+                        search_param.ef = refine_ef;
+                        search_param.topk = static_cast<int64_t>(refine_ef);
+                        search_param.search_mode = KNN_SEARCH;
+                        search_param.hops_limit = 10000;
```
LHT129: [note] The `hops_limit = 10000` in the single-layer root repair `refine_block` lambda is a hardcoded constant that limits graph traversal during the neighbor refinement search. Consider whether this limit should be derived from `ef_construction` or the graph's maximum degree to avoid silently truncating searches on large graphs.

If 10000 is a well-tested upper bound that covers all practical use cases, a brief comment explaining the rationale would help future readers understand why this value was chosen.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2727,62 +2734,103 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
...
+                        }
+                        std::sort(new_neighbors.begin(), new_neighbors.end());
+                        new_neighbors.erase(std::unique(new_neighbors.begin(), new_neighbors.end()),
+                                            new_neighbors.end());
+
+                        if (new_neighbors.empty()) {
+                            node_missed_ids[miss_count.fetch_add(1, std::memory_order_relaxed)] =
+                                inner_id;
+                            continue;
                         }
```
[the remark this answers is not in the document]
LHT129: Fixed in 282d1d4. Restored rows are now clamped to the current graph maximum degree using batched distance computation plus deterministic partial sorting before commit. Added a regression test that exports a max_degree=32 cache and warms a max_degree=8 Pyramid.

src/algorithm/pyramid/pyramid.h
```diff
@@ -584,6 +584,9 @@ class Pyramid : public InnerIndexInterface {
     float build_cache_hit_rate_{-1.0F};     // cache hit rate from last cache-based build
     uint64_t build_cache_hit_nodes_{0};     // number of nodes with cache hit
     uint64_t build_cache_missed_nodes_{0};  // number of nodes without cache hit
+    uint64_t build_cache_hit_memberships_{0};
+    uint64_t build_cache_missed_memberships_{0};
+    uint64_t build_cache_restored_edges_{0};
```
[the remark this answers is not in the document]
LHT129: Fixed in 282d1d4 by documenting graph-membership hit/miss semantics and that restored_edges counts outgoing edges loaded before refinement.

src/algorithm/pyramid/pyramid_build_cache.cpp
```diff
@@ -85,6 +85,26 @@ PyramidBuildCache::GetGraphCache(const std::string& hierarchy_name,
...
+        }
+    }
```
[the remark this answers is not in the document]
LHT129: Fixed in 282d1d4 by reserving the input batch size before collecting cached source IDs.

src/algorithm/pyramid/pyramid.cpp
LHT129: Fixed in 282d1d4. The low-cost root repair now uses an unlimited hop cap, consistent with construction semantics, instead of the hardcoded 10000 limit.

src/algorithm/pyramid/pyramid.cpp
LHT129: This is intentional: build_cache_restored_edges counts edges loaded directly from imported cache rows before refinement. The root repair output is newly selected refinement output, not a directly restored cache edge. The field comment in 282d1d4 now makes this scope explicit.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2727,62 +2734,103 @@ Pyramid::build_with_cache(const DatasetPtr& base) {
...
+                        old_to_new[old_inner] = current->second;
                     }
+                }
 
+                Vector<Vector<InnerIdType>> restored_neighbors(
+                    node_member_ids.size(), Vector<InnerIdType>(allocator_), allocator_);
+                node_hit_ids.resize(node_member_ids.size());
+                node_missed_ids.resize(node_member_ids.size());
+                std::atomic<uint64_t> hit_count{0};
+                std::atomic<uint64_t> miss_count{0};
```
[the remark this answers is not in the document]
LHT129: The concern is valid. At the validated 1M production-like run the current staging peaked at about 6.53 GiB and remained within the target envelope. A compact flat row buffer would add substantial complexity to deterministic parallel commit, so I propose keeping this implementation for this PR and tracking further memory compaction separately if larger-scale benchmarks require it.

src/algorithm/pyramid/pyramid_build_cache.cpp
LHT129: Reserved the input batch size in 282d1d4 to avoid repeated rehashing. Persisting another global lookup inside PyramidBuildCache or using approximate sampling would change cache memory or eligibility semantics; since this is one pre-build pass, I propose leaving further optimization to a follow-up.

LHT129 (commented):
I have reviewed the latest commit (282d1d4c) for PR #2900. The changes look solid and well-structured.

**Summary of changes:**
- Replaced per-neighbor source-ID translation with a reusable `old_to_new` inner-ID lookup table, eliminating repeated string-based lookups in the hot path.
- Parallelized cache-row decoding into block-parallel workers with deterministic post-barrier graph commits.
- Added `FindNeighborInnerIds()` to `BuildCache` for direct inner-ID access, avoiding the string round-trip through `GetNeighbors()`.
- Added `CountMatchedSourceIds()` with an 80% overlap threshold to gate cache-based warm builds.
- Added batch `Query()` distance computation in `connect_cached_graph_point()` replacing one-at-a-time `ComputePairVectors()`.
- Added graph-membership and restored-edge statistics to `GetStats()`.
- Added regression tests including a max-degree clamping test and updated existing cache tests.

**Existing Copilot comments (all addressed in 282d1d4c):**
1. Max-degree clamping: now uses `partial_sort`-based distance truncation when `new_neighbors.size() > max_degree`.
2. Memory overhead of `restored_neighbors`: author acknowledged and deferred to follow-up.
3. Missing comments on new stats fields: comments added in pyramid.h.
4. Missing `reserve()` in `CountMatchedSourceIds`: `cached_source_ids.reserve(count)` added.

**No new issues found.** The parallel decode with deterministic commit pattern is correct, the atomic indexing for hit/miss arrays is safe, and the read-only shared state (old_to_new, node_ids, graph_cache) is properly handled across threads. The 3.63x speedup at 99% cache hit rate is a significant improvement.

LHT129 (commented):
I have reviewed the latest commit (282d1d4c) and the overall PR. Here is my assessment:

**Summary**: This is a well-structured performance optimization for Pyramid build cache restoration. The key changes — reusable old-to-new inner-ID table, parallel block-based decoding, batched distance computation in `connect_cached_graph_point`, and the 80% overlap gate — are all correctly implemented with good test coverage.

**What was addressed in 282d1d4c**:
- Max-degree clamping for restored rows with distance-based truncation + regression test
- Reserve for `CountMatchedSourceIds` to reduce rehashing
- Inline comments on the new stats fields (`build_cache_hit_memberships_`, `build_cache_missed_memberships_`, `build_cache_restored_edges_`)

**Remaining minor items** (already flagged by Copilot, non-blocking):
- `CountMatchedSourceIds` could benefit from an early return for `count == 0` and reserving based on the actual cached source count rather than the input batch size. These are micro-optimizations for a one-time pre-build pass and do not affect correctness.

**Correctness**: The parallel decode-then-sequential-commit pattern is safe. The `old_to_new` table is read-only during parallel execution. Atomic `fetch_add` on `hit_count`/`miss_count` with relaxed ordering is sufficient since the barrier (`drain_futures`) provides the necessary synchronization before the resize. The `connect_cached_graph_point` signature change is consistently applied to all call sites.

**Test coverage**: The new max-degree clamping test, `CountMatchedSourceIds` unit test, updated stats assertions, and the disjoint-source-ID cold-build fallback test all provide good coverage of the changed paths.

No blocking issues found.

src/algorithm/pyramid/pyramid_build_cache.cpp
```diff
@@ -85,6 +85,27 @@ PyramidBuildCache::GetGraphCache(const std::string& hierarchy_name,
     return it->second.get();
 }
 
+uint64_t
+PyramidBuildCache::CountMatchedSourceIds(const std::string* source_ids, uint64_t count) const {
+    UnorderedSet<std::string> cached_source_ids(allocator_);
+    cached_source_ids.reserve(count);
```
LHT129, line 91: [suggestion] `CountMatchedSourceIds` reserves based on `count` (the input dataset size), but the set is populated with all cached source IDs from every graph cache. When the input is much larger than the total cached source IDs, this over-reserves memory. When the total cached source IDs exceed the input size, rehashing may still occur.

Consider computing the total cached source ID count first (e.g., by summing `graph_cache->source_ids_.size()` across all non-null entries in `graph_caches_`), and reserving based on that total instead of `count`.

```cpp
uint64_t total_cached = 0;
for (const auto& [key, graph_cache] : graph_caches_) {
    if (graph_cache != nullptr) {
        total_cached += graph_cache->source_ids_.size();
    }
}
cached_source_ids.reserve(total_cached);
```

src/algorithm/pyramid/pyramid_build_cache.cpp
```diff
@@ -85,6 +85,27 @@ PyramidBuildCache::GetGraphCache(const std::string& hierarchy_name,
     return it->second.get();
 }
 
+uint64_t
```
LHT129, line 88: [suggestion] `CountMatchedSourceIds` builds the full `cached_source_ids` set even when `count == 0`. The unit test exercises this path (`CountMatchedSourceIds(source_ids, 0)`). Consider adding an early return to avoid unnecessary allocations:

```cpp
if (count == 0) {
    return 0;
}
```

LHT129 (commented):
[note] This PR is in good shape overall. The code changes are well-structured, test coverage is comprehensive (unit tests for `BuildCache::FindNeighborInnerIds`, `PyramidBuildCache::CountMatchedSourceIds`, and functional tests for degree clamping and stats reporting), and the performance improvement (3.63x faster warm build at 99% cache hit) is well-documented.

One remaining concern already flagged by other reviewers: `CountMatchedSourceIds` in `pyramid_build_cache.cpp:91` reserves based on the input `count` parameter, but the unordered set is populated with all cached source IDs from every graph cache. When the input dataset is much larger than the total cached entries, this over-reserves memory. Computing the total cached count first (summing `graph_cache->source_ids_.size()` across all non-null entries) and reserving based on that total would be more precise. This is a minor optimization and not a blocker.

wxyucs (approved):
lgtm

PR #2902, opened 2026-09-08
fix(hgraph): prevent stale entry point after force-remove tail nodes

### LINKED ISSUE #2901: fix: stale entry point after force-remove causes null-pointer crash during Add
## Description

force-removing elements from an HGraph index can leave the entry point pointing past the valid ID range. A subsequent Add reads that stale entry point during its bottom-graph probe, acquires a null codes lease from the flatten storage, and passes the null pointer into SIMD distance functions (AVX512/Neon load), causing a SIGSEGV.

## Root Cause

1. find_new_entry_point() only searches route_graphs_ for a replacement. When the deleted node is the sole upper-level node, all route graphs are popped without finding a successor.
2. When the deleted node is the swap-tail (swap_id == inner_id), move_id is skipped and the stale entry_point_id_ persists.
3. shrink_to_fit() shrinks storage but does not update max_capacity_, so the next Add resize() incorrectly thinks there is enough capacity.
4. FlattenDataCell::query() does not validate the lease before calling ComputeDist, allowing null pointers into SIMD loads.

## Reproduction

Build an HGraph index with support_force_remove=true, force-remove elements from the tail, then Add new ones. The first call to flatten_datacell.h:399 Acquire(idx[i]) returns an empty lease when the stale entry point references an out-of-range ID.

## Affected Code

- src/algorithm/hgraph/hgraph_modify.cpp: find_new_entry_point, force_remove_one, shrink_to_fit
- src/datacell/flatten_datacell.h: query() batch4 and single-point paths

LHT129:
## Summary

fix(hgraph): prevent stale entry point after force-remove tail nodes

find_new_entry_point only searched route_graphs_ for a replacement entry point. When the deleted node was the sole upper-level node and also the swap-tail, entry_point_id_ remained stale and pointed past the live range. A subsequent Add used that stale ID in its bottom-graph probe, reading out-of-range codes, which crashed in SIMD load.

## Changes

- find_new_entry_point: fall back to bottom_graph_ when route_graphs_ are exhausted without a successor
- force_remove_one: retry find_new_entry_point after graph state is updated, guarding the swap_id == inner_id skip-move case
- shrink_to_fit: synchronise max_capacity_ so subsequent Add resize correctly grows flatten and graph storage
- flatten_datacell query: check lease validity before SIMD distance evaluation and throw READ_ERROR on null instead of crashing
- test_hgraph_remove: add regression tests for tail-force-remove followed by Add
- test_hgraph_stale_entry_point: add deterministic functional and concurrent stress coverage

## Files Changed

- src/algorithm/hgraph/hgraph_modify.cpp
- src/datacell/flatten_datacell.h
- tests/test_hgraph_remove.cpp
- tests/test_hgraph_stale_entry_point.cpp (new)

## Testing

- All 6 new regression tests pass on x86 (AVX512) and ARM (Neon)
- Concurrent Add/force-remove stress test passes
- No SIGSEGV on previously crashing seed values

Fixes: #2901

### DIFF
--- a/src/algorithm/hgraph/hgraph_modify.cpp
+++ b/src/algorithm/hgraph/hgraph_modify.cpp
@@ -81,6 +81,22 @@ HGraph::find_new_entry_point() {
         }
         route_graphs_.pop_back();
     }
+
+    // No upper-level successor found -- fall back to the bottom graph.
+    // Pick any remaining node as the new entry point so the index stays
+    // usable for subsequent Add / search operations.
+    if (not find_new_ep and this->bottom_graph_ != nullptr and
+        this->bottom_graph_->TotalCount() > 0) {
+        auto total = this->total_count_.load();
+        for (InnerIdType candidate = 0; candidate < total; ++candidate) {
+            if (candidate != this->entry_point_id_ and
+                this->bottom_graph_->CheckIdExists(candidate)) {
+                this->entry_point_id_ = candidate;
+                find_new_ep = true;
+                break;
+            }
+        }
+    }
 }
 
 void
@@ -194,6 +210,13 @@ HGraph::force_remove_one(int64_t label) {
     bool was_mark_removed = false;
     {
         std::unique_lock lock(this->label_lookup_mutex_);
+        // After move_id (or when swap_id == inner_id so move_id is skipped),
+        // entry_point_id_ may still equal the deleted inner_id when
+        // find_new_entry_point failed to locate an upper-level successor.
+        // Retry with the bottom graph visible now that graph state is updated.
+        if (this->entry_point_id_ == inner_id and this->total_count_.load() > 1) {
+            this->find_new_entry_point();
+        }
         was_mark_removed = this->label_table_->IsRemoved(inner_id);
         this->label_table_->ForceRemove(label, inner_id);
         if (swap_id != inner_id) {
@@ -228,6 +251,8 @@ HGraph::shrink_to_fit() {
         route_graph->ShrinkToFit(total_count);
     }
     label_table_->ShrinkToFit(total_count);
+
+    this->max_capacity_.store(total_count);
 }
 
 void
--- a/src/datacell/flatten_datacell.h
+++ b/src/datacell/flatten_datacell.h
@@ -385,6 +385,10 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::query(float* result_dists,
         auto lease2 = this->layout_->Acquire(idx[i + 1]);
         auto lease3 = this->layout_->Acquire(idx[i + 2]);
         auto lease4 = this->layout_->Acquire(idx[i + 3]);
+        if (not lease1 or not lease2 or not lease3 or not lease4) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                "failed to acquire codes for batch distance evaluation");
+        }
         computer->ComputeDistsBatch4(lease1.Data(),
                                      lease2.Data(),
                                      lease3.Data(),
@@ -396,6 +400,10 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::query(float* result_dists,
     }
     for (; i < id_count; ++i) {
         auto lease = this->layout_->Acquire(idx[i]);
+        if (not lease) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                "failed to acquire codes for id " + std::to_string(idx[i]));
+        }
         computer->ComputeDist(lease.Data(), result_dists + i);
     }
     if (ctx != nullptr and ctx->stats != nullptr and ctx->track_distance_evaluations)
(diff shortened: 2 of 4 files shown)


tests/test_hgraph_stale_entry_point.cpp
```diff
@@ -0,0 +1,424 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <atomic>
+#include <catch2/catch_test_macros.hpp>
+#include <future>
+#include <limits>
```
LHT129, line 18: [suggestion] Several headers included in this new test file appear unused:

- `<future>` — no `std::future`, `std::promise`, or `std::async` usage
- `<limits>` — no `std::numeric_limits` usage
- `<sstream>` — no `std::stringstream` usage
- `<unordered_set>` — no `std::unordered_set` usage

Removing them keeps the include list minimal and avoids unnecessary compilation overhead.

src/datacell/flatten_datacell.h
```diff
@@ -385,6 +385,10 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::query(float* result_dists,
         auto lease2 = this->layout_->Acquire(idx[i + 1]);
         auto lease3 = this->layout_->Acquire(idx[i + 2]);
         auto lease4 = this->layout_->Acquire(idx[i + 3]);
+        if (not lease1 or not lease2 or not lease3 or not lease4) {
+            throw VsagException(ErrorType::READ_ERROR,
```
LHT129, line 389: [suggestion] The `memset(result_dists, 0, sizeof(float) * id_count)` on line 390 runs before the lease validity checks added below. If a lease is null and the exception is thrown, the memset work is wasted. Consider moving the lease acquisition and validation before the memset, or only zeroing after successful acquisition.

tests/test_hgraph_stale_entry_point.cpp
```diff
@@ -0,0 +1,424 @@
...
+    bool found_new = false;
+    const auto* result_ids = search_result.value()->GetIds();
+    for (int64_t j = 0; j < search_result.value()->GetDim(); ++j) {
+        if (result_ids[j] == new_id) {
+            found_new = true;
+            break;
+        }
+    }
+    REQUIRE(found_new);
+}
```
LHT129: [note] The new file is missing a trailing newline at the end. The VSAG coding standard requires files to end with a blank line.

LHT129 (commented):
Overall this is a well-targeted fix for a real crash bug (#2901). The changes are surgical and address the root cause at three levels:

**Core fixes (hgraph_modify.cpp):**
1. `find_new_entry_point` fallback to bottom graph when route_graphs are exhausted — correct and necessary.
2. Retry `find_new_entry_point` inside `label_lookup_mutex_` when `swap_id == inner_id` — correctly handles the skip-move case.
3. `shrink_to_fit` synchronizing `max_capacity_` — critical for subsequent `Add` to allocate enough storage.

**Defensive hardening (flatten_datacell.h):**
- Lease validity checks before SIMD distance evaluation prevent null-pointer crashes and turn them into clean `READ_ERROR` exceptions.

**Test coverage:**
- Comprehensive regression tests in both `test_hgraph_remove.cpp` and the new `test_hgraph_stale_entry_point.cpp` covering tail force-remove, full clear, degenerate cases, and concurrent stress.

Left a few minor inline comments (unused includes, trailing newline, memset ordering). None are blocking.

src/algorithm/hgraph/hgraph_modify.cpp
```diff
@@ -194,6 +210,13 @@ HGraph::force_remove_one(int64_t label) {
     bool was_mark_removed = false;
     {
         std::unique_lock lock(this->label_lookup_mutex_);
+        // After move_id (or when swap_id == inner_id so move_id is skipped),
```
LHT129, line 213: [note] The comment says "After move_id (or when swap_id == inner_id so move_id is skipped)", but this retry of `find_new_entry_point()` actually executes **before** `move_id` (line 224) and before `ForceRemove` (line 219). The retry happens after `graph_force_remove_one` has cleaned up the deleted node from the bottom and route graphs, which is what makes the fallback in `find_new_entry_point` succeed on the second attempt. Consider updating the comment to reflect the actual ordering, e.g.:

```
// After graph_force_remove_one cleaned up the deleted node from all graphs
// (or when swap_id == inner_id so the node was never moved),
// entry_point_id_ may still equal the deleted inner_id when
// find_new_entry_point failed to locate an upper-level successor.
// Retry now that graph state is updated and the bottom graph fallback
// can find a valid replacement.
```

LHT129 (commented):
All review suggestions acknowledged — unused includes, Options restore guard, test timing signal, batch4 error message detail, and comment wording will be addressed in a follow-up PR. None are blocking.

wxyucs (approved):
lgtm

Issue #2888, opened 2026-09-08
[improve](lite): implement and optimize VSAG Lite for lightweight vector search


nw9663644-eng:
## Description

This issue tracks the OSPP 2026 project **VSAG Lite Implementation and Optimization** (project ID: **26c520035**).

The goal is to design and implement a lightweight VSAG variant for small-scale vector search and resource-constrained deployments. It should support essential insertion, deletion, update, search, persistence, and loading capabilities while reducing unnecessary features and dependencies.

Development will proceed incrementally: establish a minimal BruteForce-based implementation first, then introduce graph search, selected quantization methods, and memory/loading optimizations.

This is a project-level tracking issue. The first implementation PR is not expected to deliver the entire roadmap.

## Background / Motivation

VSAG provides comprehensive indexing, persistence, and extensibility capabilities for general-purpose, high-performance vector search.

For edge RAG, small local knowledge bases, and mobile or embedded applications, the full feature set may introduce unnecessary binary size, startup overhead, and memory consumption.

A lightweight implementation should address:

- Index selection and data organization for small-scale workloads.
- Memory fragmentation and space reuse under deletion and updates.
- Efficient index loading and reduced data copying.
- A smaller, clearly documented public API.
- Compatibility with existing VSAG APIs where practical.
- Measurable resource savings without compromising basic correctness or retrieval quality.

### Functional Requirements

- [ ] Provide a VSAG Lite implementation that can be built independently or enabled through configuration.
- [ ] Support vector insertion, deletion, updates, and search.
- [ ] Support index persistence and loading.
- [ ] Preserve data consistency after deletion, updates, and save/load operations.
- [ ] Provide configuration options, unit tests, and usage documentation.
- [ ] Pass relevant builds and tests.

### Architecture Requirements

- [ ] Document retained capabilities and excluded modules.
- [ ] Explain index selection and intended workload sizes.
- [ ] Describe data layout, memory management, and deletion/update handling.
- [ ] Describe persistence, loading, and strategies for reducing data copies.
- [ ] Define API compatibility boundaries and unsupported capabilities.
- [ ] Reuse suitable VSAG components while minimizing changes to existing Full indexes.

### Evaluation Requirements

- [ ] Provide reproducible benchmark scripts, configurations, and raw results.
- [ ] Cover at least two small datasets or two dataset-scale configurations.
- [ ] Compare Full and Lite under documented, comparable configurations.
[...]

PR #2904, opened 2026-09-08
feat(lite): add standalone BruteForce and FP32 graph index

nw9663644-eng:
## Change Type
- [x] New feature
- [x] Improvement/Refactor
- [x] Documentation
- [x] CI/Build/Infra

## Linked Issue
Related to #2888. This PR adds the Lite library and FP32 graph functionality; it does not close the broader OSPP issue because public quantization and mmap remain follow-up work.

## What Changed
This PR builds standalone VSAG Lite without linking the Full VSAG runtime. An Index starts with exact FP32 squared-L2 BruteForce search and can explicitly call BuildGraph(max_degree, ef_search) to use a single-layer approximate FP32 graph. Add, Update, Remove, Search, Save and Load remain available after the transition. BuildGraph publishes a replacement only after a successful build.

Search has an external-ID filter overload: a callback returns true to allow an ID in results. Graph search may traverse rejected nodes to preserve connectivity but never returns them. FP32 distance uses runtime SIMD dispatch on supported x86 CPUs, with the generic path for short dimensions. BruteForce v1 snapshots remain byte-identical; graph snapshots use a separate v2 format and Load accepts both.

### Directory layout
| Path | Role |
| --- | --- |
| `include/vsag/lite/index.h` | Public Lite API and filter contract |
| `src/lite/index.cpp` | API dispatch and v1/v2 persistence |
| `src/lite/bruteforce_backend.cpp` | Exact backend and CRUD |
| `src/lite/graph_backend.cpp` | FP32 graph construction, CRUD, filtered search and v2 state |
| `src/lite/fp32_distance*.cpp` | Standalone runtime distance dispatch |
| `src/lite/*_test.cpp` | Unit, scale, graph, filter and snapshot tests |
| `lite/CMakeLists.txt`, `lite/example/` | Independent library build, install and consumer example |
| `lite/README.md`, `docs/docs/{en,zh}/src/development/lite_first.md` | Architecture and user guides |

Benchmark runners and SQ8/FP16 candidates are in dependent PR #2926, not in this feature branch.

## Validation
Validated head: `ade31f1b494f31b71f82e646471793831a29d281` on Ubuntu 22.04 x86_64, GCC 11.4.0. Fresh Debug, Release, ASan+UBSan and coverage builds each passed all three CTest suites (unit, 100k scale and graph). clang-format-15, the repository Lite CI clang-tidy-15 scope, YAML parsing and git diff checks passed. The seven standalone Lite implementation files reached 548/606 covered lines (90.43%). Installed shared and static consumers passed BuildGraph, filtered Search and v2 Save/Load; the shared library does not depend on Full `libvsag.so`.

The 100k case is a functional smoke test, not a general performance claim. Online GitHub Actions for this head are not claimed as passed here.

## Compatibility and Limits
The default Full VSAG build/API is unchanged. Lite calls require external synchronization. Graph search is approximate; transition temporarily holds flat and graph state. The v2 graph snapshot is not a Full VSAG snapshot and has no checksum or crash-safe replacement guarantee. Quantization, mmap and automatic BruteForce-to-Graph switching are outside this PR.

## Follow-up
PR #2926 contains reproducible baseline/Full-Lite/SIFT experiments and internal SQ8/FP16 candidates. A public quantized backend remains contingent on quality, resource and snapshot validation.

### DIFF
--- a/lite/CMakeLists.txt
+++ b/lite/CMakeLists.txt
@@ -9,17 +9,24 @@ get_filename_component(VSAG_ROOT "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUTE)
 option(ENABLE_TESTS "Build Lite tests with the repository's Catch2" OFF)
 option(ENABLE_COVERAGE "Instrument Lite sources for gcov" OFF)
 
-add_library(vsag_lite SHARED "${VSAG_ROOT}/src/lite/index.cpp")
+set(VSAG_LITE_SOURCES "${VSAG_ROOT}/src/lite/index.cpp")
+add_library(vsag_lite_objects OBJECT ${VSAG_LITE_SOURCES})
+add_library(vsag_lite SHARED $<TARGET_OBJECTS:vsag_lite_objects>)
+add_library(vsag_lite_static STATIC $<TARGET_OBJECTS:vsag_lite_objects>)
 add_library(vsag::lite ALIAS vsag_lite)
+set_target_properties(vsag_lite_objects PROPERTIES POSITION_INDEPENDENT_CODE ON)
 set_target_properties(vsag_lite PROPERTIES OUTPUT_NAME vsag-lite EXPORT_NAME lite)
-target_compile_features(vsag_lite PUBLIC cxx_std_17)
-target_include_directories(vsag_lite PUBLIC
-    $<BUILD_INTERFACE:${VSAG_ROOT}/include>
-    $<INSTALL_INTERFACE:include>
-    PRIVATE "${VSAG_ROOT}/src")
-target_compile_options(vsag_lite PRIVATE -Wall -Wextra -Wpedantic)
+set_target_properties(vsag_lite_static PROPERTIES OUTPUT_NAME vsag-lite)
+foreach(target vsag_lite_objects vsag_lite vsag_lite_static)
+    target_compile_features(${target} PUBLIC cxx_std_17)
+    target_include_directories(${target} PUBLIC
+        $<BUILD_INTERFACE:${VSAG_ROOT}/include>
+        $<INSTALL_INTERFACE:include>
+        PRIVATE "${VSAG_ROOT}/src")
+endforeach()
+target_compile_options(vsag_lite_objects PRIVATE -Wall -Wextra -Wpedantic)
 if(ENABLE_COVERAGE)
-    target_compile_options(vsag_lite PRIVATE --coverage -O0 -g)
+    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
     target_link_options(vsag_lite PRIVATE --coverage)
 endif()
 
@@ -33,6 +40,7 @@ endif()
 include(GNUInstallDirs)
 install(TARGETS vsag_lite EXPORT vsag-lite-config
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(TARGETS vsag_lite_static ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 install(FILES "${VSAG_ROOT}/include/vsag/lite/index.h" DESTINATION include/vsag/lite)
 # Header-only error contract dependencies; no Full runtime is linked.
 install(FILES
--- a/lite/CMakeLists.txt
+++ b/lite/CMakeLists.txt
@@ -1,0 +1,45 @@
+cmake_minimum_required(VERSION 3.18)
+project(vsag_lite VERSION 0.1.0 LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+# A separate entry point deliberately avoids the Full build's dependency closure.
+get_filename_component(VSAG_ROOT "${CMAKE_CURRENT_LIST_DIR}/.." ABSOLUTE)
+option(ENABLE_TESTS "Build Lite tests with the repository's Catch2" OFF)
+option(ENABLE_COVERAGE "Instrument Lite sources for gcov" OFF)
+
+add_library(vsag_lite SHARED "${VSAG_ROOT}/src/lite/index.cpp")
+add_library(vsag::lite ALIAS vsag_lite)
+set_target_properties(vsag_lite PROPERTIES OUTPUT_NAME vsag-lite EXPORT_NAME lite)
+target_compile_features(vsag_lite PUBLIC cxx_std_17)
+target_include_directories(vsag_lite PUBLIC
+    $<BUILD_INTERFACE:${VSAG_ROOT}/include>
+    $<INSTALL_INTERFACE:include>
+    PRIVATE "${VSAG_ROOT}/src")
+target_compile_options(vsag_lite PRIVATE -Wall -Wextra -Wpedantic)
+if(ENABLE_COVERAGE)
+    target_compile_options(vsag_lite PRIVATE --coverage -O0 -g)
+    target_link_options(vsag_lite PRIVATE --coverage)
+endif()
+
+if(ENABLE_TESTS)
+    include("${VSAG_ROOT}/cmake/VSAGThirdPartyOverride.cmake")
+    include("${VSAG_ROOT}/extern/catch2/catch2.cmake")
+    add_executable(lite_tests "${VSAG_ROOT}/src/lite/index_test.cpp")
+    target_link_libraries(lite_tests PRIVATE vsag::lite Catch2::Catch2WithMain)
+endif()
+
+include(GNUInstallDirs)
+install(TARGETS vsag_lite EXPORT vsag-lite-config
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(FILES "${VSAG_ROOT}/include/vsag/lite/index.h" DESTINATION include/vsag/lite)
+# Header-only error contract dependencies; no Full runtime is linked.
+install(FILES
+    "${VSAG_ROOT}/include/vsag/errors.h"
+    "${VSAG_ROOT}/include/vsag/expected.hpp"
+    "${VSAG_ROOT}/include/vsag/options.h"
+    "${VSAG_ROOT}/include/vsag/allocator.h"
+    "${VSAG_ROOT}/include/vsag/logger.h"
+    DESTINATION include/vsag)
+install(EXPORT vsag-lite-config NAMESPACE vsag:: DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/vsag-lite)
--- a/src/lite/index.cpp
+++ b/src/lite/index.cpp
@@ -1,0 +1,310 @@
+// Copyright 2024-present the vsag project
+// SPDX-License-Identifier: Apache-2.0
+#include "vsag/lite/index.h"
+
+#include <algorithm>
+#include <cmath>
+#include <cstring>
+#include <istream>
+#include <limits>
+#include <new>
+#include <ostream>
+#include <queue>
+#include <stdexcept>
+#include <unordered_map>
+
+#include "simd/kernels/compute_l2.h"
+#include "simd/traits/simd_traits_generic.h"
+
+namespace vsag::lite {
+namespace {
+
+auto
+failure(ErrorType type, const char* message) {
+    return tl::unexpected(Error(type, message));
+}
+
+tl::expected<void, Error>
+validate(const float* data, uint64_t actual, uint64_t expected) {
+    if (actual != expected) {
+        return failure(ErrorType::DIMENSION_NOT_EQUAL, "dimension mismatch");
+    }
+    if (data == nullptr) {
+        return failure(ErrorType::INVALID_ARGUMENT, "null vector");
+    }
+    for (uint64_t i = 0; i < actual; ++i) {
+        if (not std::isfinite(data[i])) {
+            return failure(ErrorType::INVALID_ARGUMENT, "non-finite vector");
+        }
+    }
+    return {};
+}
+
+bool
+better(const Neighbor& a, const Neighbor& b) {
+    return a.distance < b.distance or (a.distance == b.distance and a.id < b.id);
+}
+
+// Explicit byte encoding: never serialize native containers, pointers, or struct padding.
+void
+write(std::ostream& out, uint64_t value, uint64_t bytes = 8) {
+    for (uint64_t i = 0; i < bytes; ++i) {
+        out.put(static_cast<char>((value >> (8 * i)) & 255));
+    }
+}
+
+uint64_t
+read(std::istream& in, uint64_t bytes = 8) {
+    uint64_t value = 0;
+    for (uint64_t i = 0; i < bytes; ++i) {
+        auto byte = in.get();
+        if (byte == std::char_traits<char>::eof()) {
+            throw std::ios_base::failure("truncated snapshot");
+        }
+        value |= static_cast<uint64_t>(static_cast<unsigned char>(byte)) << (8 * i);
+    }
+    return value;
+}
+
+constexpr uint64_t K_HEADER_BYTES = 48;
+constexpr char K_MAGIC[] = "VSAGLT01";
+static_assert(sizeof(float) == 4 and std::numeric_limits<float>::is_iec559);
+
+}  // namespace
+
+struct Index::Impl {
+    explicit Impl(uint64_t dimension) : dim(dimension) {
+    }
+    uint64_t dim;
+    std::vector<float> vectors;
+    std::vector<int64_t> ids;
+    std::unordered_map<int64_t, uint64_t> slots;
+};
+
+Index::Index(uint64_t dim) : impl_(std::make_unique<Impl>(dim)) {
+}
+Index::~Index() = default;
+
+tl::expected<std::unique_ptr<Index>, Error>
+Index::Create(uint64_t dim) {
+    if (dim == 0 or dim > std::vector<float>().max_size()) {
+        return failure(ErrorType::INVALID_ARGUMENT, "invalid dimension");
+    }
+    try {
+        return std::unique_ptr<Index>(new Index(dim));
+    } catch (const std::bad_alloc&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "create allocation failed");
+    }
+}
+
+uint64_t
+Index::Size() const {
+    return impl_->ids.size();
+}
+
+uint64_t
+Index::Dim() const {
+    return impl_->dim;
+}
+
+tl::expected<void, Error>
+Index::Add(int64_t id, const float* vector, uint64_t dim) {
+    auto valid = validate(vector, dim, Dim());
+    if (not valid) {
+        return tl::unexpected(valid.error());
+    }
+    if (impl_->slots.count(id) != 0) {
+        return failure(ErrorType::INVALID_ARGUMENT, "duplicate ID");
+    }
+    if (Size() >= impl_->vectors.max_size() / dim or Size() >= impl_->ids.max_size()) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "index capacity exceeded");
+    }
+    try {
+        // Reserve geometrically, before publishing any logical record. reserve(n+1) each time
+        // would turn a sequence of single-vector inserts into quadratic copying.
+        auto grow = [](auto& values, uint64_t required) {
+            if (required > values.capacity()) {
+                const uint64_t maximum = values.max_size();
+                const uint64_t capacity = values.capacity();
+                const uint64_t doubled = capacity > maximum / 2 ? maximum : capacity * 2;
+                values.reserve(std::max(required, doubled));
+            }
+        };
+        grow(impl_->vectors, (Size() + 1) * dim);
+        grow(impl_->ids, Size() + 1);
+        impl_->slots.emplace(id, Size());
+        // Scalar insertion cannot allocate after the successful reserves above.
+        impl_->vectors.insert(impl_->vectors.end(), vector, vector + dim);
+        impl_->ids.push_back(id);
+        return {};
+    } catch (const std::bad_alloc&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "add allocation failed");
+    } catch (const std::length_error&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "add capacity exceeded");
+    }
+}
+
+tl::expected<void, Error>
+Index::Update(int64_t id, const float* vector, uint64_t dim) {
+    auto valid = validate(vector, dim, Dim());
+    if (not valid) {
+        return tl::unexpected(valid.error());
+    }
+    auto slot = impl_->slots.find(id);
+    if (slot == impl_->slots.end()) {
+        return failure(ErrorType::INVALID_ARGUMENT, "missing ID");
+    }
+    std::copy_n(vector, dim, impl_->vectors.data() + slot->second * dim);
+    return {};
+}
+
+bool
+Index::Remove(int64_t id) {
+    auto found = impl_->slots.find(id);
+    if (found == impl_->slots.end()) {
+        return false;
+    }
+    const uint64_t slot = found->second;
+    const uint64_t last = Size() - 1;
+    // Same dense-slot principle as Full BruteForce::Remove(FORCE_REMOVE).
+    // This is deliberately not a graph-compatible stable-slot contract.
+    if (slot != last) {
+        std::copy_n(
+            impl_->vectors.data() + last * Dim(), Dim(), impl_->vectors.data() + slot * Dim());
+        impl_->ids[slot] = impl_->ids[last];
+        impl_->slots.find(impl_->ids[slot])->second = slot;
+    }
+    impl_->slots.erase(found);
+    impl_->ids.pop_back();
+    impl_->vectors.resize(last * Dim());
+    return true;
+}
+
+tl::expected<std::vector<Neighbor>, Error>
+Index::Search(const float* query, uint64_t dim, uint64_t k) const {
+    auto valid = validate(query, dim, Dim());
+    if (not valid) {
+        return tl::unexpected(valid.error());
+    }
+    try {
+        k = std::min(k, Size());
+        if (k == 0) {
+            return std::vector<Neighbor>{};
+        }
+        std::priority_queue<Neighbor, std::vector<Neighbor>, decltype(&better)> heap(&better);
+        for (uint64_t slot = 0; slot < Size(); ++slot) {
+            const auto distance = simd::ComputeL2SqrImpl<simd::SimdTraits<simd::GenericTag>>(
+                query, impl_->vectors.data() + slot * dim, dim);
+            Neighbor next{impl_->ids[slot], distance};
+            if (heap.size() < k) {
+                heap.push(next);
+            } else if (better(next, heap.top())) {
+                heap.pop();
+                heap.push(next);
+            }
+        }
+        std::vector<Neighbor> result(heap.size());
+        for (uint64_t i = result.size(); i > 0; --i) {
+            result[i - 1] = heap.top();
+            heap.pop();
+        }
+        return result;
+    } catch (const std::bad_alloc&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "search allocation failed");
+    } catch (const std::length_error&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "search capacity exceeded");
+    }
+}
+
+tl::expected<void, Error>
+Index::Save(std::ostream& output) const {
+    if (Dim() > (UINT64_MAX - 8) / 4 or Size() > (UINT64_MAX - K_HEADER_BYTES) / (8 + 4 * Dim())) {
+        return failure(ErrorType::INVALID_ARGUMENT, "snapshot size overflow");
+    }
+    try {
+        output.write(K_MAGIC, 8);
+        write(output, 1);  // Format version.
+        write(output, Dim());
+        write(output, Size());
+        write(output, Size() * (8 + 4 * Dim()));
+        write(output, 1);  // FP32 squared-L2 representation.
+        for (auto id : impl_->ids) {
+            write(output, static_cast<uint64_t>(id));
+        }
+        for (float value : impl_->vectors) {
+            uint32_t bits;
+            std::memcpy(&bits, &value, 4);
+            write(output, bits, 4);
+        }
+        if (not output) {
+            return failure(ErrorType::READ_ERROR, "snapshot write failed");
+        }
+        return {};
+    } catch (const std::ios_base::failure&) {
+        return failure(ErrorType::READ_ERROR, "snapshot write failed");
+    }
+}
+
+tl::expected<std::unique_ptr<Index>, Error>
+Index::Load(std::istream& input) {
+    try {
+        const auto start = input.tellg();
+        input.seekg(0, std::ios::end);
+        const auto end = input.tellg();
+        if (start == std::streampos(-1) or end == std::streampos(-1) or end < start) {
+            return failure(ErrorType::INVALID_BINARY, "a seekable input is required");
+        }
+        input.seekg(start);
+        const auto available = static_cast<uint64_t>(end - start);
+        char magic[8];
+        input.read(magic, 8);
+        if (not input or std::memcmp(magic, K_MAGIC, 8) != 0 or read(input) != 1) {
+            return failure(ErrorType::INVALID_BINARY, "invalid magic or version");
+        }
+        const auto dim = read(input);
+        const auto count = read(input);
+        const auto payload = read(input);
+        const auto representation = read(input);
+        if (dim == 0 or dim > (UINT64_MAX - 8) / 4 or representation != 1 or
+            count > (UINT64_MAX - K_HEADER_BYTES) / (8 + 4 * dim) or
+            payload != count * (8 + 4 * dim) or available != K_HEADER_BYTES + payload) {
+            return failure(ErrorType::INVALID_BINARY, "invalid snapshot layout");
+        }
+        auto created = Create(dim);
+        if (not created) {
+            return tl::unexpected(created.error());
+        }
+        auto& data = *(*created)->impl_;
+        // Bounds validated against the actual input length before any payload allocation.
+        data.ids.reserve(count);
+        data.slots.reserve(count);
+        for (uint64_t slot = 0; slot < count; ++slot) {
+            const uint64_t bits = read(input);
+            const int64_t id = bits <= INT64_MAX ? static_cast<int64_t>(bits)
+                                                 : -1 - static_cast<int64_t>(UINT64_MAX - bits);
+            if (not data.slots.emplace(id, slot).second) {
+                return failure(ErrorType::INVALID_BINARY, "duplicate snapshot ID");
+            }
+            data.ids.push_back(id);
+        }
+        data.vectors.reserve(count * dim);
+        for (uint64_t i = 0; i < count * dim; ++i) {
+            auto bits = static_cast<uint32_t>(read(input, 4));
+            float value;
+            std::memcpy(&value, &bits, 4);
+            if (not std::isfinite(value)) {
+                return failure(ErrorType::INVALID_BINARY, "non-finite snapshot vector");
+            }
+            data.vectors.push_back(value);
+        }
+        return std::move(*created);
+    } catch (const std::ios_base::failure&) {
+        return failure(ErrorType::INVALID_BINARY, "snapshot read failed");
+    } catch (const std::bad_alloc&) {
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "load allocation failed");
+    } catch (const std::length_error&) {
+        return failure(ErrorType::INVALID_BINARY, "snapshot exceeds container capacity");
+    }
+}
+
+}  // namespace vsag::lite
--- a/src/lite/index.cpp
+++ b/src/lite/index.cpp
@@ -238,7 +238,7 @@ Index::Save(std::ostream& output) const {
         write(output, Size());
         write(output, Size() * (8 + 4 * Dim()));
         write(output, 1);  // FP32 squared-L2 representation.
-        for (const auto id : impl_->ids) {
+        for (const auto& id : impl_->ids) {
             write(output, static_cast<uint64_t>(id));
         }
         for (float value : impl_->vectors) {
@@ -290,8 +290,8 @@ Index::Load(std::istream& input) {
         data.slots.reserve(count);
         for (uint64_t slot = 0; slot < count; ++slot) {
             const uint64_t bits = read(input);
-            const int64_t id = bits <= INT64_MAX ? static_cast<int64_t>(bits)
-                                                 : -1 - static_cast<int64_t>(UINT64_MAX - bits);
+            int64_t id;
+            std::memcpy(&id, &bits, sizeof(id));
             if (not data.slots.emplace(id, slot).second) {
                 return failure(ErrorType::INVALID_BINARY, "duplicate snapshot ID");
             }
--- a/src/lite/index.cpp
+++ b/src/lite/index.cpp
@@ -45,6 +45,13 @@ better(const Neighbor& a, const Neighbor& b) {
     return a.distance < b.distance or (a.distance == b.distance and a.id < b.id);
 }
 
+struct NeighborWorseFirst {
+    bool
+    operator()(const Neighbor& a, const Neighbor& b) const {
+        return better(a, b);
+    }
+};
+
 // Explicit byte encoding: never serialize native containers, pointers, or struct padding.
 void
 write(std::ostream& out, uint64_t value, uint64_t bytes = 8) {
@@ -126,14 +133,17 @@ Index::Add(int64_t id, const float* vector, uint64_t dim) {
             if (required > values.capacity()) {
                 const uint64_t maximum = values.max_size();
                 const uint64_t capacity = values.capacity();
-                const uint64_t doubled = capacity > maximum / 2 ? maximum : capacity * 2;
+                const uint64_t doubled =
+                    capacity == 0 ? 1 : (capacity > maximum / 2 ? maximum : capacity * 2);
                 values.reserve(std::max(required, doubled));
             }
         };
-        grow(impl_->vectors, (Size() + 1) * dim);
-        grow(impl_->ids, Size() + 1);
-        impl_->slots.emplace(id, Size());
-        // Scalar insertion cannot allocate after the successful reserves above.
+        const uint64_t slot = Size();
+        grow(impl_->vectors, (slot + 1) * dim);
+        grow(impl_->ids, slot + 1);
+        // Publish the map entry only after every operation that can reallocate a vector has
+        // succeeded. The following scalar inserts cannot throw after the reserves above.
+        impl_->slots.emplace(id, slot);
         impl_->vectors.insert(impl_->vectors.end(), vector, vector + dim);
         impl_->ids.push_back(id);
         return {};
@@ -172,7 +182,7 @@ Index::Remove(int64_t id) {
         std::copy_n(
             impl_->vectors.data() + last * Dim(), Dim(), impl_->vectors.data() + slot * Dim());
         impl_->ids[slot] = impl_->ids[last];
-        impl_->slots.find(impl_->ids[slot])->second = slot;
+        impl_->slots.at(impl_->ids[slot]) = slot;
     }
     impl_->slots.erase(found);
     impl_->ids.pop_back();
@@ -191,7 +201,7 @@ Index::Search(const float* query, uint64_t dim, uint64_t k) const {
         if (k == 0) {
             return std::vector<Neighbor>{};
         }
-        std::priority_queue<Neighbor, std::vector<Neighbor>, decltype(&better)> heap(&better);
+        std::priority_queue<Neighbor, std::vector<Neighbor>, NeighborWorseFirst> heap;
         for (uint64_t slot = 0; slot < Size(); ++slot) {
             const auto distance = simd::ComputeL2SqrImpl<simd::SimdTraits<simd::GenericTag>>(
                 query, impl_->vectors.data() + slot * dim, dim);
@@ -228,7 +238,7 @@ Index::Save(std::ostream& output) const {
         write(output, Size());
         write(output, Size() * (8 + 4 * Dim()));
         write(output, 1);  // FP32 squared-L2 representation.
-        for (auto id : impl_->ids) {
+        for (const auto id : impl_->ids) {
             write(output, static_cast<uint64_t>(id));
         }
         for (float value : impl_->vectors) {
--- a/.github/workflows/lite.yml
+++ b/.github/workflows/lite.yml
@@ -1,0 +1,72 @@
+name: Standalone Lite
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: ["main", "0.*"]
+    paths:
+      - "include/vsag/lite/**"
+      - "src/lite/**"
+      - "lite/**"
+      - ".github/workflows/lite.yml"
+
+concurrency:
+  group: standalone-lite-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  build-test-install:
+    name: Build, test, and install
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install required tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format-15 clang-tidy-15 ninja-build
+      - name: Check formatting
+        run: |
+          clang-format-15 --dry-run --Werror \
+            include/vsag/lite/index.h \
+            src/lite/index.cpp \
+            src/lite/index_test.cpp \
+            lite/example/main.cpp
+      - name: Configure Debug tests
+        run: |
+          cmake -S lite -B build-lite-first-debug -G Ninja \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+            -DENABLE_TESTS=ON
+      - name: Run clang-tidy
+        run: |
+          clang-tidy-15 src/lite/index.cpp src/lite/index_test.cpp \
+            -p build-lite-first-debug \
+            --header-filter='.*/include/vsag/lite/.*'
+      - name: Build and test Debug
+        run: |
+          cmake --build build-lite-first-debug -j2
+          ./build-lite-first-debug/lite_tests '[lite]'
+          ./build-lite-first-debug/lite_tests '[lite-scale]'
+      - name: Build and install Release library
+        run: |
+          cmake -S lite -B build-lite-first-release -G Ninja \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_INSTALL_PREFIX="$PWD/install-lite-first" \
+            -DENABLE_TESTS=OFF
+          cmake --build build-lite-first-release -j2
+          cmake --install build-lite-first-release
+          if ldd build-lite-first-release/libvsag-lite.so | grep -q 'libvsag\.so'; then
+            echo 'Standalone Lite unexpectedly depends on the Full library' >&2
+            exit 1
+          fi
+      - name: Build installed consumer
+        run: |
+          cmake -S lite/example -B build-lite-consumer -G Ninja \
+            -DCMAKE_PREFIX_PATH="$PWD/install-lite-first"
+          cmake --build build-lite-consumer -j2
+          ./build-lite-consumer/lite_example "$RUNNER_TEMP/lite-example.snapshot"
--- a/.github/workflows/lite.yml
+++ b/.github/workflows/lite.yml
@@ -60,6 +60,8 @@ jobs:
             -DENABLE_TESTS=OFF
           cmake --build build-lite-first-release -j2
           cmake --install build-lite-first-release
+          test -f build-lite-first-release/libvsag-lite.a
+          test -f "$PWD/install-lite-first/lib/libvsag-lite.a"
           if ldd build-lite-first-release/libvsag-lite.so | grep -q 'libvsag\.so'; then
             echo 'Standalone Lite unexpectedly depends on the Full library' >&2
             exit 1
@@ -70,3 +72,11 @@ jobs:
             -DCMAKE_PREFIX_PATH="$PWD/install-lite-first"
           cmake --build build-lite-consumer -j2
           ./build-lite-consumer/lite_example "$RUNNER_TEMP/lite-example.snapshot"
+      - name: Build installed static consumer
+        run: |
+          c++ -std=c++17 \
+            -I"$PWD/install-lite-first/include" \
+            lite/example/main.cpp \
+            "$PWD/install-lite-first/lib/libvsag-lite.a" \
+            -o build-lite-static-consumer
+          ./build-lite-static-consumer "$RUNNER_TEMP/lite-static-example.snapshot"
--- a/docs/docs/en/src/development/lite_first.md
+++ b/docs/docs/en/src/development/lite_first.md
@@ -1,0 +1,97 @@
+# Independent VSAG Lite v0.1
+
+This experimental entry point starts with exact FP32 squared-L2 BruteForce and
+selective source reuse, rather than conditionally compiling the Full library.
+It currently targets Linux x86_64, C++17 and the repository compiler baseline.
+The default Full build is unchanged. Run these commands from the repository root:
+
+```sh
+cmake -S lite -B build-lite-first -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON
+cmake --build build-lite-first -j2
+./build-lite-first/lite_tests '[lite]'
+./build-lite-first/lite_tests '[lite-scale]'
+cmake --install build-lite-first --prefix "$PWD/install-lite-first"
+cmake -S lite/example -B build-lite-consumer -DCMAKE_PREFIX_PATH="$PWD/install-lite-first"
+cmake --build build-lite-consumer -j2
+./build-lite-consumer/lite_example "$PWD/new-example.lite"
+```
+
+Unlike the Full `make` targets, `cmake -S lite` is a separate dependency-isolated
+entry point. Tests use the existing pinned Catch2 v3.7.1 configuration. For an
+offline build, pass `-DFETCHCONTENT_SOURCE_DIR_CATCH2=/path/to/catch2-v3.7.1`.
+Tests are invoked directly through Catch2; no parallel custom testing framework
+or CTest suite is introduced. With tests disabled, no Catch2 download is needed.
+
+## API and ownership
+
+`vsag::lite::Index::Create(dim)` creates a fixed-dimension index. `Add`, `Update`,
+`Search`, `Save`, and static `Load` use the existing `tl::expected` / `vsag::Error`
+contract without calling Full's global logger. `Remove` returns a boolean.
+Inputs must contain the specified number of finite float32 values. Calls must
+be externally serialized. Results own their memory; internal slots are not exposed.
+
+- Single-record Add rejects duplicate IDs; Update rejects absent IDs.
+- Remove returns false for absent IDs. A removed external ID can be inserted again.
+- Add failure leaves logical records unchanged, though reserved capacity may grow.
+- Search returns `min(k, Size())` entries sorted by squared-L2 then ascending ID.
+  Valid queries with k=0 or an empty index return an empty result.
+- FP32 accumulation can overflow to positive infinity for extreme finite inputs;
+  those distances tie and are ordered by ID. This is not arbitrary-precision L2.
+- NaN/Inf inputs and dimension mismatches are rejected even for k=0.
+- Memory allocation failures are translated to Error where caught; constructing
+  an error itself may still allocate. No hard real-time or no-throw OOM guarantee.
+
+## Data organization and reuse
+
+The private implementation owns contiguous FP32 records, external IDs, and an
+ID-to-slot map. BruteForce scans use the official generic `ComputeL2SqrImpl`
+kernel. Physical removal uses the same last-record-to-hole principle as Full
+BruteForce, but retains capacity for reuse. It does not promise immediate RSS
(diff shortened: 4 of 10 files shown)


src/lite/index.cpp
```diff
@@ -0,0 +1,310 @@
...
+        return failure(ErrorType::NO_ENOUGH_MEMORY, "index capacity exceeded");
+    }
+    try {
+        // Reserve geometrically, before publishing any logical record. reserve(n+1) each time
+        // would turn a sequence of single-vector inserts into quadratic copying.
+        auto grow = [](auto& values, uint64_t required) {
+            if (required > values.capacity()) {
+                const uint64_t maximum = values.max_size();
+                const uint64_t capacity = values.capacity();
+                const uint64_t doubled = capacity > maximum / 2 ? maximum : capacity * 2;
```
LHT129: [note] The geometric growth in `grow` degenerates to exact allocation when `capacity == 0` (the initial state): `capacity * 2 == 0`, so `reserve(std::max(required, 0))` is equivalent to `reserve(required)`. This means the first insertion reserves exactly the needed space rather than a geometrically-grown amount. For a single-vector insert this is harmless, but if the first batch insert is large, subsequent inserts will trigger additional reallocations sooner than expected.

Consider using `std::max(required, std::max(capacity * 2, static_cast<uint64_t>(1)))` or initializing with a small non-zero default capacity so that the geometric doubling takes effect from the first growth.

src/lite/index.cpp
```diff
@@ -0,0 +1,310 @@
...
+        impl_->ids[slot] = impl_->ids[last];
+        impl_->slots.find(impl_->ids[slot])->second = slot;
+    }
+    impl_->slots.erase(found);
+    impl_->ids.pop_back();
+    impl_->vectors.resize(last * Dim());
+    return true;
+}
+
+tl::expected<std::vector<Neighbor>, Error>
```
LHT129, line 193: [note] The `std::priority_queue` comparator type is `decltype(&better)` where `better` is a function in an anonymous namespace. This works correctly but the type is a plain function pointer `bool(*)(const Neighbor&, const Neighbor&)`. If `better` is ever changed to a lambda or functor (e.g. to capture state), this declaration would break.

Consider using `decltype([](const Neighbor& a, const Neighbor& b) { return better(a, b); })` or `std::function<bool(const Neighbor&, const Neighbor&)>` if future flexibility is desired. Not a blocker for the current implementation.

src/lite/index.cpp
```diff
@@ -0,0 +1,310 @@
...
+    auto valid = validate(vector, dim, Dim());
+    if (not valid) {
+        return tl::unexpected(valid.error());
+    }
+    auto slot = impl_->slots.find(id);
+    if (slot == impl_->slots.end()) {
+        return failure(ErrorType::INVALID_ARGUMENT, "missing ID");
+    }
+    std::copy_n(vector, dim, impl_->vectors.data() + slot->second * dim);
+    return {};
```
LHT129, line 168: [suggestion] In `Remove`, the line `impl_->slots.find(impl_->ids[slot])->second = slot;` dereferences the result of `find` without checking for `end()`. While the lookup is guaranteed to succeed because `impl_->ids[slot]` was just copied from `impl_->ids[last]` which has a valid entry in `slots`, this invariant is non-obvious and fragile under future refactoring.

Consider using `impl_->slots.at(impl_->ids[slot]) = slot;` which would throw a detectable exception if the invariant is ever violated, or add an assertion:
```cpp
auto it = impl_->slots.find(impl_->ids[slot]);
assert(it != impl_->slots.end());
it->second = slot;
```

src/lite/index.cpp
```diff
@@ -0,0 +1,310 @@
...
+            return failure(ErrorType::INVALID_BINARY, "a seekable input is required");
+        }
+        input.seekg(start);
+        const auto available = static_cast<uint64_t>(end - start);
+        char magic[8];
+        input.read(magic, 8);
+        if (not input or std::memcmp(magic, K_MAGIC, 8) != 0 or read(input) != 1) {
+            return failure(ErrorType::INVALID_BINARY, "invalid magic or version");
+        }
+        const auto dim = read(input);
```
LHT129, line 274: [note] The `int64_t` two's-complement decoding in `Load`:
```cpp
const int64_t id = bits <= INT64_MAX ? static_cast<int64_t>(bits)
                                     : -1 - static_cast<int64_t>(UINT64_MAX - bits);
```
relies on `UINT64_MAX - bits` fitting in `int64_t` when `bits > INT64_MAX`, which is correct because the range is `[0, INT64_MAX]`. However, the encoding in `Save` uses `static_cast<uint64_t>(id)` for negative `id` values, which is implementation-defined in C++17 (only defined as two's-complement in C++20). While all target platforms use two's-complement, this could be documented explicitly in the snapshot format spec or guarded with a `static_assert`:
```cpp
static_assert(static_cast<uint64_t>(static_cast<int64_t>(1)) == 1);
```
Not a blocker since the documentation already states two's-complement encoding.

src/lite/index.cpp
```diff
@@ -0,0 +1,310 @@
...
+        write(output, Size());
+        write(output, Size() * (8 + 4 * Dim()));
+        write(output, 1);  // FP32 squared-L2 representation.
+        for (auto id : impl_->ids) {
+            write(output, static_cast<uint64_t>(id));
+        }
+        for (float value : impl_->vectors) {
+            uint32_t bits;
+            std::memcpy(&bits, &value, 4);
+            write(output, bits, 4);
```
LHT129, line 247: [suggestion] In `Save`, the loop `for (auto id : impl_->ids)` makes a copy of each `int64_t` element. While the copy is trivial (8 bytes), using `for (int64_t id : impl_->ids)` or `for (const auto& id : impl_->ids)` would be more idiomatic and avoid any potential confusion about whether a copy is intentional.

```cpp
for (int64_t id : impl_->ids) {
    write(output, static_cast<uint64_t>(id));
}
```

LHT129 (commented):
## Code Review for PR #2904

This is a well-structured PR that introduces a clean, minimal VSAG Lite baseline. The code quality is high overall. Below are the review findings:

### Suggestions

**1. `src/lite/index.cpp:1012` — Fragile `slots.emplace` ordering**

The `impl_->slots.emplace(id, Size())` call records the slot mapping before the vector data and ID are actually appended to their containers (lines 1014-1015). While the comment correctly notes that scalar insertion cannot allocate after the successful `reserve` calls, this ordering creates a fragile invariant: if any future code is inserted between the emplace and the push_back/insert, the `slots` map could become inconsistent with the actual data.

Consider moving `impl_->slots.emplace(id, Size())` to after `impl_->ids.push_back(id)` so that the slot mapping is only published once the data is fully committed. The slot value would then be `Size() - 1` instead of `Size()`.

**2. `src/lite/index.cpp:1052` — Unchecked `find()` dereference**

```cpp
impl_->slots.find(impl_->ids[slot])->second = slot;
```

The result of `find()` is dereferenced without checking for `end()`. While this is safe under the current invariants (the moved ID must exist in the map), an assertion or check would make the invariant explicit and catch potential future regressions. Consider adding `assert(found != impl_->slots.end())` or storing the iterator and checking it.

**3. `lite/CMakeLists.txt:353` — Static library target**

The library is only built as `SHARED`. For embedded or offline use cases where shared library deployment is undesirable, consider also providing a static library target (e.g., `add_library(vsag_lite_static STATIC ...)`).

### Positive Observations

- The snapshot format is well-designed with explicit little-endian byte encoding, versioning, and comprehensive validation (truncation, corruption, duplicate IDs, non-finite values, trailing bytes).
- The test coverage is thorough: CRUD correctness, independent reference comparison, malformed snapshot rejection, stream failure contracts, and a 100k-vector end-to-end smoke test.
- The `last-record-to-hole` removal strategy correctly reuses the proven Full BruteForce approach.
- Input validation is consistently applied across all API entry points.
- The CI workflow covers formatting, clang-tidy, debug/release builds, dependency isolation verification, and installed consumer validation.
- Error handling properly translates exceptions to `tl::expected` errors without leaking exceptions to callers.
- The documentation clearly communicates the experimental nature, limitations, and future extension points.

Overall this is a solid foundation for the VSAG Lite effort. The suggestions above are non-blocking improvements.

LHT129 (commented):
## Code Review for PR #2904

This is a well-structured PR that introduces a clean, minimal VSAG Lite baseline. The code quality is high overall. Below are the review findings:

### Suggestions

**1. src/lite/index.cpp:1012 — Fragile slots.emplace ordering**

The `impl_->slots.emplace(id, Size())` call records the slot mapping before the vector data and ID are actually appended to their containers (lines 1014-1015). While the comment correctly notes that scalar insertion cannot allocate after the successful `reserve` calls, this ordering creates a fragile invariant: if any future code is inserted between the emplace and the push_back/insert, the `slots` map could become inconsistent with the actual data.

Consider moving `impl_->slots.emplace(id, Size())` to after `impl_->ids.push_back(id)` so that the slot mapping is only published once the data is fully committed. The slot value would then be `Size() - 1` instead of `Size()`.

**2. src/lite/index.cpp:1052 — Unchecked find() dereference**

```cpp
impl_->slots.find(impl_->ids[slot])->second = slot;
```

The result of `find()` is dereferenced without checking for `end()`. While this is safe under the current invariants (the moved ID must exist in the map), an assertion or check would make the invariant explicit and catch potential future regressions. Consider adding `assert(found != impl_->slots.end())` or storing the iterator and checking it.

**3. lite/CMakeLists.txt:353 — Static library target**

The library is only built as `SHARED`. For embedded or offline use cases where shared library deployment is undesirable, consider also providing a static library target (e.g., `add_library(vsag_lite_static STATIC ...)`).

### Positive Observations

- The snapshot format is well-designed with explicit little-endian byte encoding, versioning, and comprehensive validation (truncation, corruption, duplicate IDs, non-finite values, trailing bytes).
- The test coverage is thorough: CRUD correctness, independent reference comparison, malformed snapshot rejection, stream failure contracts, and a 100k-vector end-to-end smoke test.
- The `last-record-to-hole` removal strategy correctly reuses the proven Full BruteForce approach.
- Input validation is consistently applied across all API entry points.
- The CI workflow covers formatting, clang-tidy, debug/release builds, dependency isolation verification, and installed consumer validation.
- Error handling properly translates exceptions to `tl::expected` errors without leaking exceptions to callers.
- The documentation clearly communicates the experimental nature, limitations, and future extension points.

Overall this is a solid foundation for the VSAG Lite effort. The suggestions above are non-blocking improvements.

LHT129 (commented):
## Code Review for PR #2904

This is a well-structured PR that introduces a clean, minimal VSAG Lite baseline.

### Suggestions

**1. src/lite/index.cpp:1012 - Fragile slots.emplace ordering**

The slots.emplace(id, Size()) call records the slot mapping before the vector data and ID are actually appended. Consider moving it after push_back so the slot mapping is only published once the data is fully committed.

**2. src/lite/index.cpp:1052 - Unchecked find() dereference**

The result of find() is dereferenced without checking for end(). An assertion would make the invariant explicit.

**3. lite/CMakeLists.txt - Static library target**

Consider also providing a static library target for embedded use cases.

### Positive Observations

Well-designed snapshot format, thorough test coverage, proper input validation, clean CI workflow, and clear documentation. Solid foundation for VSAG Lite.

LHT129:
## Detailed Review Notes

### 1. `src/lite/index.cpp:1012` — Fragile `slots.emplace` ordering [suggestion]

```cpp
impl_->slots.emplace(id, Size());  // slot recorded here
// ...
impl_->vectors.insert(impl_->vectors.end(), vector, vector + dim);  // data appended here
impl_->ids.push_back(id);  // ID appended here
```

The slot mapping is published via `slots.emplace` before the actual data is committed to `vectors` and `ids`. While the comment correctly notes that scalar insertion cannot allocate after successful `reserve`, this ordering creates a fragile invariant. If any future code is inserted between the emplace and the push_back/insert, the slots map could become inconsistent.

**Suggested fix:** Move `impl_->slots.emplace(id, Size())` to after `impl_->ids.push_back(id)`, using `Size() - 1` as the slot value.

### 2. `src/lite/index.cpp:1052` — Unchecked `find()` dereference [note]

```cpp
impl_->slots.find(impl_->ids[slot])->second = slot;
```

The result of `find()` is dereferenced without checking for `end()`. While this is safe under the current invariants (the moved ID must exist in the map), an explicit assertion would make the invariant explicit and catch potential future regressions.

### 3. `lite/CMakeLists.txt:353` — Static library target [suggestion]

The library is only built as `SHARED`. For embedded or offline use cases where shared library deployment is undesirable, consider also providing a static library target.

---

**Note:** Inline comments could not be posted due to GitHub API limitations with fork PRs (position/line resolution unavailable). The findings above reference specific source locations.

src/lite/index.cpp
nw9663644-eng: The zero-capacity case is now handled explicitly:

    const uint64_t doubled =
        capacity == 0 ? 1 : (capacity > maximum / 2 ? maximum : capacity * 2);

This avoids a zero doubled capacity while preserving the existing overflow protection. Subsequent growth continues geometrically. The focused Lite tests and the 100k-vector test both pass.

src/lite/index.cpp
nw9663644-eng: I replaced the unchecked `find()` dereference with:

    impl_->slots.at(impl_->ids[slot]) = slot;

This preserves the current dense-slot invariant while making any future invariant violation detectable instead of dereferencing `end()`.

src/lite/index.cpp
nw9663644-eng: I replaced the function-pointer comparator with a named `NeighborWorseFirst` function object. The priority queue no longer depends on `better` remaining a plain function, so the comparator declaration is more robust for future refactoring.

src/lite/index.cpp
```diff
@@ -0,0 +1,320 @@
...
+        write(output, Size());
+        write(output, Size() * (8 + 4 * Dim()));
+        write(output, 1);  // FP32 squared-L2 representation.
+        for (const auto id : impl_->ids) {
+            write(output, static_cast<uint64_t>(id));
+        }
+        for (float value : impl_->vectors) {
+            uint32_t bits;
+            std::memcpy(&bits, &value, 4);
+            write(output, bits, 4);
```
LHT129, line 247: [suggestion] The `for (const auto id : impl_->ids)` loop makes a copy of each `int64_t` element. While the copy is trivial (8 bytes), using `const auto&` would be more idiomatic and consistent with typical C++ best practices for range-based for loops over containers.

```cpp
for (const auto& id : impl_->ids) {
    write(output, static_cast<uint64_t>(id));
}
```

src/lite/index.cpp
```diff
@@ -0,0 +1,320 @@
...
+        auto& data = *(*created)->impl_;
+        // Bounds validated against the actual input length before any payload allocation.
+        data.ids.reserve(count);
+        data.slots.reserve(count);
+        for (uint64_t slot = 0; slot < count; ++slot) {
+            const uint64_t bits = read(input);
+            const int64_t id = bits <= INT64_MAX ? static_cast<int64_t>(bits)
+                                                 : -1 - static_cast<int64_t>(UINT64_MAX - bits);
+            if (not data.slots.emplace(id, slot).second) {
+                return failure(ErrorType::INVALID_BINARY, "duplicate snapshot ID");
```
LHT129, line 296: [note] The int64 two's-complement decoding in `Load` is correct but the formula is somewhat opaque:

```cpp
const int64_t id = bits <= INT64_MAX ? static_cast<int64_t>(bits)
                                     : -1 - static_cast<int64_t>(UINT64_MAX - bits);
```

A more direct approach using `memcpy` (which avoids undefined behavior from integer overflow) would be more readable and equally portable given the existing `static_assert` on IEEE 754 and `sizeof(float)`:

```cpp
int64_t id;
std::memcpy(&id, &bits, sizeof(id));
```

This is not a correctness issue — the current implementation works correctly for all `uint64_t` inputs. Just a readability suggestion for future maintenance.

src/lite/index.cpp
nw9663644-eng: Thank you for pointing this out. I rechecked the C++17 integral conversion rules.

The conversion from a signed integer to `uint64_t` is defined modulo 2^64. The decoding path also avoids an out-of-range unsigned-to-signed conversion: when `bits > INT64_MAX`, `UINT64_MAX - bits` is guaranteed to be within `[0, INT64_MAX]` before the signed cast.

The snapshot documentation already specifies two's-complement encoding, and the existing round-trip tests cover both `INT64_MIN` and `INT64_MAX`. Therefore, I kept the snapshot format unchanged.

src/lite/index.cpp
nw9663644-eng: The loop now uses:

    for (const auto id : impl_->ids)

This makes the read-only intent explicit while retaining inexpensive value semantics for `int64_t`.

src/lite/index.cpp
nw9663644-eng: The serialization loop now uses:

    for (const auto& id : impl_->ids)

This avoids copying each ID and follows the suggested range-based loop style.

src/lite/index.cpp
nw9663644-eng: I replaced the two's-complement reconstruction formula with an object-representation copy:

    int64_t id;
    std::memcpy(&id, &bits, sizeof(id));

This keeps the existing snapshot representation unchanged while making the decoding path more direct and readable. The existing boundary tests for `INT64_MIN` and `INT64_MAX` continue to pass.

src/lite/index.cpp
```diff
@@ -0,0 +1,320 @@
...
+        write(output, 1);  // Format version.
+        write(output, Dim());
+        write(output, Size());
+        write(output, Size() * (8 + 4 * Dim()));
+        write(output, 1);  // FP32 squared-L2 representation.
+        for (const auto& id : impl_->ids) {
+            write(output, static_cast<uint64_t>(id));
+        }
+        for (float value : impl_->vectors) {
+            uint32_t bits;
```
LHT129, line 245: [note] The `for (float value : impl_->vectors)` loop copies each float by value, while the `ids` loop above was already changed to `for (const auto& id : impl_->ids)`. For consistency, consider using `for (float value : impl_->vectors)` as-is (the copy is trivial at 4 bytes) or explicitly using `const auto&`:

```cpp
for (const auto& value : impl_->vectors) {
    uint32_t bits;
    std::memcpy(&bits, &value, 4);
    write(output, bits, 4);
}
```

Either approach is fine for `float`; this is just a consistency note.

LHT129 (commented):
I have completed a thorough review of this PR. The existing review comments from LHT129 have already covered the substantive issues, and the author has addressed most of them in the latest commits.

**Summary of what has been reviewed and addressed:**

1. Geometric growth in `grow` lambda (fixed: `capacity == 0 ? 1` case added)
2. Priority queue comparator type (fixed: replaced function pointer with `NeighborWorseFirst` functor)
3. Unchecked `find()` dereference in `Remove` (fixed: now uses `slots.at()`)
4. Two's-complement decoding in `Load` (fixed: replaced with `std::memcpy`)
5. Copy in Save ID loop (fixed: `const auto&` instead of `const auto`)
6. Float copy in Save vectors loop (noted: `for (float value : ...)` vs `const auto&`)

**Overall assessment:**

The code is well-structured, well-tested (21,844+ assertions, 90.37% line coverage, sanitizer-clean), and well-documented. The isolation from the Full VSAG build is clean, and the CI coverage (format, clang-tidy, debug/release builds, installation, dependency isolation, external consumer) is comprehensive. The API design is minimal and appropriate for a v0.1 baseline.

No blocking issues remain. The remaining `[note]` about `for (float value : impl_->vectors)` in `Save` is a minor consistency observation that does not affect correctness.

lite/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+        $<INSTALL_INTERFACE:include>
+        PRIVATE "${VSAG_ROOT}/src")
+endforeach()
+target_compile_options(vsag_lite_objects PRIVATE -Wall -Wextra -Wpedantic)
+if(ENABLE_COVERAGE)
+    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
+    target_link_options(vsag_lite PRIVATE --coverage)
+endif()
+
+if(ENABLE_TESTS)
```
LHT129, line 33: [suggestion] The `ENABLE_COVERAGE` option adds `--coverage` link flags only to the `vsag_lite` shared library target, but the test executable `lite_tests` does not receive the coverage link flag. This means that when running tests with coverage enabled, the test binary itself will not produce gcov notes, and coverage data for the object files may be incomplete.

Consider also adding the coverage link flag to the test target:

```cmake
if(ENABLE_COVERAGE)
    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
    target_link_options(vsag_lite PRIVATE --coverage)
    target_link_options(lite_tests PRIVATE --coverage)
endif()
```

Alternatively, link `lite_tests` against `vsag_lite_static` with coverage flags when coverage is enabled.

src/lite/index.cpp
```diff
@@ -0,0 +1,320 @@
...
+    const uint64_t slot = found->second;
+    const uint64_t last = Size() - 1;
+    // Same dense-slot principle as Full BruteForce::Remove(FORCE_REMOVE).
+    // This is deliberately not a graph-compatible stable-slot contract.
+    if (slot != last) {
+        std::copy_n(
+            impl_->vectors.data() + last * Dim(), Dim(), impl_->vectors.data() + slot * Dim());
+        impl_->ids[slot] = impl_->ids[last];
+        impl_->slots.at(impl_->ids[slot]) = slot;
+    }
```
LHT129, line 186: [note] The `impl_->slots.at(impl_->ids[slot])` call in `Remove()` will throw `std::out_of_range` if the slot-to-ID invariant is broken (e.g. due to memory corruption). While this should never happen in normal operation since every ID in `ids` must have been inserted via `Add()` which always populates `slots`, the exception is not caught by any handler in `Remove()`. The previous `find()->second` approach would have produced UB on a missing key.

This is a defensive improvement over the previous code and the invariant holds by construction, so this is just a note. If you want to be extra defensive, you could add a check:

```cpp
auto it = impl_->slots.find(impl_->ids[slot]);
if (it == impl_->slots.end()) {
    // invariant broken, attempt recovery
    impl_->slots.erase(found);
    impl_->ids.pop_back();
    impl_->vectors.resize(last * Dim());
    return true;
}
it->second = slot;
```

But this is likely over-engineering for v0.1 given the invariant holds by construction.

src/lite/index.cpp
```diff
@@ -0,0 +1,320 @@
...
+    std::unordered_map<int64_t, uint64_t> slots;
+};
+
+Index::Index(uint64_t dim) : impl_(std::make_unique<Impl>(dim)) {
+}
+Index::~Index() = default;
+
+tl::expected<std::unique_ptr<Index>, Error>
+Index::Create(uint64_t dim) {
+    if (dim == 0 or dim > std::vector<float>().max_size()) {
```
LHT129, line 97: [suggestion] The `Index::Create` dimension validation uses `std::vector<float>().max_size()` as the upper bound, which is the theoretical maximum number of elements a vector can hold, not a practical dimension limit. For a vector index, a dimension of e.g. 2^60 would pass this check but would be unusable in practice (a single vector would require exabytes of memory).

Consider adding a more practical upper bound, or at least documenting that extremely large dimensions are technically allowed but practically unusable. The `Save()` function already has a tighter check: `Dim() > (UINT64_MAX - 8) / 4`.

This is not a correctness issue since memory allocation would fail for such dimensions anyway, but a more meaningful limit would produce a clearer error message earlier.

LHT129 (commented):
Reviewed the latest commit d5e8d5fd (style: add const to Save loop variable).

The change is a trivial const-correctness improvement to the `Save()` function. The previous review feedback (geometric growth initialization, priority_queue comparator, slots.at() defensive access) has been addressed in earlier commits.

No new issues found in this update. The remaining open suggestion about `Index::Create` dimension upper bound (using `max_size()` vs a practical limit) is a low-priority improvement that can be addressed in a follow-up.

Overall the PR continues to look solid for a v0.1 experimental baseline.

LHT129 (commented):
Reviewed commit d5e8d5fd: the only change is adding `const` to the `Save()` loop variable (`for (const float value : impl_->vectors)`). This is a correct const-correctness improvement.

No new issues found in this update. The existing review feedback from earlier rounds has been addressed in previous commits.

LHT129 (approved):
LGTM

lite/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+target_compile_options(vsag_lite_objects PRIVATE -Wall -Wextra -Wpedantic)
+if(VSAG_LITE_HAS_X86_SIMD)
+    target_compile_definitions(vsag_lite_objects PRIVATE VSAG_LITE_HAS_X86_SIMD)
+endif()
+if(ENABLE_COVERAGE)
+    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
+    target_link_options(vsag_lite PRIVATE --coverage)
+endif()
+
+if(ENABLE_TESTS)
```
LHT129, line 57: [suggestion] The `ENABLE_COVERAGE` option adds `--coverage` compile flags to `vsag_lite_objects` and link flags to `vsag_lite`, but the test executable `lite_tests` does not receive the `--coverage` link flag. This means that when running tests with coverage enabled, the test binary itself will not produce gcov notes (.gcda files), and coverage data for the object files linked into the test executable will be incomplete.

To fix, add the coverage link flag to `lite_tests` as well:

```cmake
if(ENABLE_COVERAGE)
    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
    target_link_options(vsag_lite PRIVATE --coverage)
    target_link_options(lite_tests PRIVATE --coverage)
endif()
```

Alternatively, apply `--coverage` to `vsag_lite_objects` via `target_link_options` as well, since the object library is linked into both `vsag_lite` and `lite_tests`.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+
+    [[nodiscard]] float
+    distance(const float* left, const float* right) const {
+        return simd::ComputeL2SqrImpl<simd::SimdTraits<simd::GenericTag>>(left, right, Dim());
+    }
+
+    std::vector<uint64_t>
+    nearest(const float* vector,
```
LHT129, line 342: [suggestion] `GraphBackend::distance()` always uses `simd::SimdTraits<simd::GenericTag>`, while `BruteForceBackend::SearchImpl` uses `SelectFP32Distance()` to pick the fastest available SIMD kernel (AVX512/AVX2/SSE4.1) at runtime. This means graph search and construction (via `nearest()` and `link()`) do not benefit from SIMD acceleration even when the hardware supports it.

The graph backend calls `distance()` in three hot paths: `SearchImpl::visit`, `nearest` (which internally calls `Search`), and `link` pruning. Consider using the same `SelectFP32Distance()` dispatcher (or the existing `fp32_distance.h` infrastructure) so that graph operations also leverage available SIMD.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,447 @@
...
+        if (Size() == 0) {
+            return {};
+        }
+        const auto found = Search(vector, Dim(), std::min(Size(), std::max(count * 4, ef_search_)));
+        if (not found) {
+            throw std::bad_alloc();
+        }
+        std::vector<uint64_t> result;
+        result.reserve(count);
+        for (const auto& neighbor : *found) {
```
LHT129, line 354: [suggestion] `nearest()` calls `Search()` and throws `std::bad_alloc()` when the result is an error, regardless of the actual error type. If `Search` returns a `DIMENSION_NOT_EQUAL` or `INVALID_ARGUMENT` error, the code still throws `std::bad_alloc()` — which is misleading and causes the caller to report `NO_ENOUGH_MEMORY` instead of the real error.

Consider propagating the actual error instead:
```cpp
if (not result) {
    throw result.error();
}
```
This would require `Error` to derive from `std::exception` (or wrapping it in a custom exception). Alternatively, change the signature of `nearest()` to return `tl::expected<std::vector<uint64_t>, Error>` so callers can handle errors explicitly without exceptions.

LHT129 (commented):
## Review Summary

This PR introduces a standalone VSAG Lite library with a BruteForce baseline and a new GraphBackend. The overall architecture is clean: a public `Index` facade with PIMPL, an abstract `Backend` interface, and two concrete backends. The separation from Full VSAG is well-executed with its own CMake entry point and CI pipeline.

### What's good
- Clean separation of concerns between `Index` (public API) and `Backend` (internal storage/algorithm)
- `BruteForceBackend` correctly uses `SelectFP32Distance()` for SIMD dispatch
- Snapshot format with magic bytes and versioning for forward compatibility
- Comprehensive test coverage including randomized CRUD and recall benchmarks
- Good use of `tl::expected` for error propagation in the public API

### Suggestions (see inline comments)
1. **`GraphBackend::distance()` uses Generic SIMD only** — should use `SelectFP32Distance()` like `BruteForceBackend` does, so graph operations benefit from AVX512/AVX2/SSE4.1 acceleration.
2. **`GraphBackend::nearest()` throws `std::bad_alloc()` for non-memory errors** — if `Search` returns a validation error, the caller reports `NO_ENOUGH_MEMORY` which is misleading.

No critical issues found. The two suggestions above are performance and error-handling improvements that don't block merging.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,447 @@
...
+        }
+        try {
+            k = std::min(k, Size());
+            if (k == 0) {
+                return std::vector<Neighbor>{};
+            }
+            const uint64_t ef = std::min(Size(), std::max(k, ef_search_));
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> best(&closer);
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> accepted(
+                &closer);
```
LHT129, line 195: [critical] The `accepted` priority queue in `GraphBackend::SearchImpl` uses the wrong comparator, causing filtered search to return the *farthest* k candidates instead of the *closest* k.

The `accepted` queue is declared with `decltype(&closer)`. `closer(a, b)` returns `true` when `a` is closer than `b`. In a `std::priority_queue` (max-heap), the top element is the closest candidate. When `accepted.size() > k`, `accepted.pop()` removes the top (closest), so `accepted` retains the k *farthest* filtered candidates.

The fix is to use `farther` as the comparator for `accepted`, so that `pop()` removes the farthest candidate and the queue retains the k closest:
```cpp
std::priority_queue<Candidate, std::vector<Candidate>, decltype(&farther)> accepted(&farther);
```

Note: the `best` queue intentionally retains the farthest `ef` candidates for pruning, so its use of `closer` is correct. Only `accepted` needs to be changed.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,447 @@
...
+            for (uint64_t neighbor : extras_[found->second]) {
+                link(neighbor, found->second);
+            }
+            return {};
+        } catch (const std::bad_alloc&) {
+            return failure(ErrorType::NO_ENOUGH_MEMORY, "update allocation failed");
+        } catch (const std::length_error&) {
+            return failure(ErrorType::NO_ENOUGH_MEMORY, "update capacity exceeded");
+        }
+    }
```
LHT129, line 139: [suggestion] `GraphBackend::Update` replaces the updated slot's neighbor list but does not remove the slot from its *old* neighbors' adjacency lists. Only the new neighbors get bidirectional links via `link()`. This leaves stale references in the old neighbors that still point to the updated slot.

Consider either:
1. Removing the slot from all old neighbors before assigning the new neighbor list, or
2. Documenting that the graph may temporarily contain asymmetric edges after Update (which `Remove` does clean up globally).

For reference, `Remove` correctly cleans up all references to the deleted slot across the entire graph.

LHT129 (commented):
Code review comments.

LHT129 (commented):
Code review.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,462 @@
...
+            for (auto& neighbor : neighbors) {
+                if (neighbor == last) {
+                    neighbor = slot;
+                }
+            }
+        }
+        if (slot != last) {
+            std::copy_n(vectors_.data() + last * Dim(), Dim(), vectors_.data() + slot * Dim());
+            ids_[slot] = ids_[last];
+            extras_[slot] = std::move(extras_[last]);
```
LHT129, line 179: [suggestion] `GraphBackend::Remove` may introduce self-loops in the adjacency list when the removed slot is not the last element and the last element's adjacency list contains a reference to itself.

In `Remove`, the loop at lines 149-156 first erases all references to `slot` from every adjacency list, then rewrites references to `last` as `slot`. After this, `extras_[last]` is moved to `extras_[slot]` (line 160). If `extras_[last]` originally contained a self-reference to `last`, that reference becomes a self-reference to `slot` after the rewrite, producing a self-loop in `extras_[slot]`.

This self-loop will be serialized by `Save` and then rejected by `restore_graph_backend` (which checks `neighbor == slot` at line 441), making snapshots of such indices unloadable.

Consider removing the self-loop from `extras_[slot]` after the move, e.g.:
```cpp
extras_[slot] = std::move(extras_[last]);
extras_[slot].erase(std::remove(extras_[slot].begin(), extras_[slot].end(), slot), extras_[slot].end());
```

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,612 @@
...
+            }
+            if (fp16_) {
+                std::copy_n(encoded.data(), dim, fp16_vectors_.data() + slot * dim);
+            } else {
+                std::copy_n(vector, dim, vectors_.data() + slot * dim);
+            }
+            extras_[slot] = std::move(*neighbors);
+            for (uint64_t neighbor : extras_[slot]) {
+                link(neighbor, slot);
+            }
```
LHT129, line 141: [suggestion] `GraphBackend::Remove` may introduce self-loops in the adjacency list when the removed slot is not the last element and the last element's adjacency list contains a reference to itself.

In `Remove`, the loop at lines 149-156 first erases all references to `slot` from every adjacency list, then rewrites references to `last` as `slot`. After this, `extras_[last]` is moved to `extras_[slot]` (line 160). If `extras_[last]` originally contained a self-reference to `last`, that reference becomes a self-reference to `slot` after the rewrite, producing a self-loop in `extras_[slot]`.

This self-loop will be serialized by `Save` and then rejected by `restore_graph_backend` (which checks `neighbor == slot`), making snapshots of such indices unloadable.

Consider removing the self-loop from `extras_[slot]` after the move, e.g.:
```cpp
extras_[slot] = std::move(extras_[last]);
extras_[slot].erase(std::remove(extras_[slot].begin(), extras_[slot].end(), slot), extras_[slot].end());
```

LHT129 (commented):
After thorough review of the current code (commit 42f4354a), the Lite library implementation is well-structured with comprehensive test coverage, proper error handling, and clear API design.

The existing review comments have already covered the main concerns. I would like to highlight a few observations:

1. The `accepted` priority queue in `graph_backend.cpp` `SearchImpl` (using `closer` comparator) is actually correct — in a `std::priority_queue` max-heap, `top()` returns the element with the largest distance, and `pop()` removes it, so the queue correctly retains the k closest filtered candidates. The existing review flagging this as critical appears to be based on a misunderstanding of `std::priority_queue` semantics.

2. The `Remove` self-loop issue flagged in the existing review is a real concern that should be addressed before merging.

3. The `Update` function now correctly cleans up old reverse links (this was fixed since the earlier review).

Overall, the code quality is high. The main blocker is the `Remove` self-loop bug.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,627 @@
...
+        return ids_[slot];
+    }
+
+    [[nodiscard]] const float*
+    VectorAt(uint64_t slot) const override {
+        if (not fp16_) {
+            return vectors_.data() + slot * Dim();
+        }
+        for (uint64_t d = 0; d < Dim(); ++d) {
+            decoded_[d] = decode_fp16(fp16_vectors_[slot * Dim() + d]);
```
LHT129: [note] The `VectorAt` method for FP16 storage uses a `mutable std::vector<float> decoded_` member as a decode buffer and returns a pointer to its internal data. This has two implications:

1. **Not thread-safe for reads**: Even though `VectorAt` is `const`, it mutates `decoded_`. Concurrent calls to `VectorAt` from different threads will race on the decode buffer, producing corrupted results.

2. **Pointer lifetime**: The returned pointer is only valid until the next call to `VectorAt` (on the same instance), which overwrites the buffer. Callers that store the pointer across multiple `VectorAt` calls will read wrong data.

This is used in `make_graph_backend`/`make_fp16_graph_backend` where `source.VectorAt(slot)` is called in a loop — safe in that context since each call's result is consumed immediately. But the `const` interface implies read-only safety that doesn't hold.

Consider either:
- Making `decoded_` thread-local (`static thread_local`) if multi-threaded read access is expected, or
- Returning a `std::vector<float>` by value (at the cost of an allocation per call), or
- Documenting the single-reader constraint explicitly in the interface.

For the current single-threaded Lite usage this is not a bug, but it's a latent correctness hazard for any future concurrent use.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,627 @@
...
+                    accepted.push(next);
+                    if (accepted.size() > k) {
+                        accepted.pop();
+                    }
+                }
+            };
+            visit(0);
+            visit(Size() - 1);
+            constexpr uint64_t k_extra_entry_points = 6;
+            for (uint64_t i = 1; i <= k_extra_entry_points; ++i) {
```
LHT129, line 253: [note] The entry point selection in `SearchImpl` uses `visit(0)`, `visit(Size()-1)`, and 6 evenly-spaced slots. For small indices this produces duplicate entry points:

- When `Size()==1`: `visit(0)` and `visit(0)` (since `Size()-1 == 0`).
- When `Size()==2`: all 6 extra entry points compute to slot 0 (e.g., `1*1/7=0`, `2*1/7=0`, ...).

The `visit` lambda is idempotent (checks `visited[slot]`), so duplicates are harmless but wasteful — they still compute `distance()` and push to queues before the visited check returns early.

Consider guarding the extra entry points with a minimum size check:
```cpp
if (Size() > k_extra_entry_points + 1) {
    for (uint64_t i = 1; i <= k_extra_entry_points; ++i) {
        visit(i * (Size() - 1) / (k_extra_entry_points + 1));
    }
}
```
This avoids the wasted work on tiny indices without changing behavior.

src/lite/bruteforce_backend.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include <stdexcept>
+#include <unordered_map>
+
+#include "lite/backend.h"
+#include "lite/fp32_distance.h"
+
+namespace vsag::lite::detail {
+namespace {
+
+auto
```
LHT129: [note] `bruteforce_backend.cpp` and `graph_backend.cpp` each define identical `validate()` and `failure()` helper functions in their own anonymous namespaces. This duplication means any change to validation logic or error message format must be made in two places, with a risk of divergence.

Consider extracting these into a shared internal header (e.g., `src/lite/backend_utils.h`) so both backends share a single definition. This would also make it easier to add new backends in the future without copying the helpers again.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+    void
+    link(uint64_t source, uint64_t target) {
+        auto& neighbors = extras_[source];
+        if (std::find(neighbors.begin(), neighbors.end(), target) != neighbors.end()) {
+            return;
+        }
+        neighbors.push_back(target);
+        if (neighbors.size() > max_degree_) {
+            // max_degree_ is at most 64; cache each distance once before sorting.
+            std::array<Candidate, 65> ranked{};
```
LHT129: [suggestion] The `link()` function uses a hardcoded `std::array<Candidate, 65>` sized for `max_degree_ + 1` (max_degree ≤ 64). If the `max_degree` validation limit is ever raised above 64 without updating this array size, the `ranked[i] = ...` assignment at line 465 would silently write out of bounds.

Consider using `std::vector<Candidate> ranked(neighbors.size())` or adding a `static_assert` to keep the array size and the max_degree limit in sync:

```cpp
static_assert(65 >= 64 + 1, "ranked array must hold max_degree + 1 candidates");
```

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+    bool
+    Remove(int64_t id) override {
+        auto found = slots_.find(id);
+        if (found == slots_.end()) {
+            return false;
+        }
+        const uint64_t slot = found->second;
+        const uint64_t last = Size() - 1;
+        for (auto& neighbors : extras_) {
```
LHT129, line 162: [suggestion] `Remove()` iterates over `extras_` (all adjacency lists) to erase references to the removed `slot` and rename `last` to `slot`. This is O(total_edges) for every removal. For a graph with many nodes, repeated removals can become expensive.

Only the removed slot's neighbors need fixup: the neighbors of `slot` have a reverse edge pointing to `slot`, and the neighbors of `last` (which will move to `slot`) have edges pointing to `last`. All other adjacency lists are unaffected. Consider limiting the fixup to just those two sets:

```cpp
for (uint64_t neighbor : extras_[slot]) {
    auto& reverse = extras_[neighbor];
    reverse.erase(std::remove(reverse.begin(), reverse.end(), slot), reverse.end());
}
if (slot != last) {
    for (uint64_t neighbor : extras_[last]) {
        auto& reverse = extras_[neighbor];
        std::replace(reverse.begin(), reverse.end(), last, slot);
    }
}
```

This reduces the complexity from O(total_edges) to O(degree × 2), which is bounded by `max_degree_` (≤ 64).

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+                if (Size() > 1) {
+                    visit((current.slot + Size() - 1) % Size());
+                    visit((current.slot + 1) % Size());
+                }
+                for (const auto neighbor : extras_[current.slot]) {
+                    visit(neighbor);
+                }
+            }
+
+            auto& output = filter == nullptr ? best : accepted;
```
LHT129: [suggestion] When `filter == nullptr`, the result extraction pops all `ef` candidates from the `best` priority_queue into a vector and re-sorts them. Since `best` already uses the `closer` comparator (top = farthest), the elements are already in reverse-sorted order after popping. You can avoid the O(n log n) sort by building the result in reverse:

```cpp
auto& output = filter == nullptr ? best : accepted;
std::vector<Neighbor> result(output.size());
for (uint64_t i = result.size(); i > 0; --i) {
    const auto candidate = output.top();
    output.pop();
    result[i - 1] = {IdAt(candidate.slot), candidate.distance};
}
result.resize(std::min(k, static_cast<uint64_t>(result.size())));
```

This is consistent with how `BruteForceBackend::SearchImpl` already builds its result (lines 140-144 of bruteforce_backend.cpp) and avoids the extra sort for the unfiltered path.

src/lite/index.cpp
```diff
@@ -0,0 +1,387 @@
...
+        write(output, payload);
+        write(output, fp16 ? 3 : (graph ? 2 : 1));
+        if (graph) {
+            write(output, impl_->backend->MaxDegree());
+            write(output, impl_->backend->EfSearch());
+        }
+        for (uint64_t slot = 0; slot < Size(); ++slot) {
+            write(output, static_cast<uint64_t>(impl_->backend->IdAt(slot)));
+        }
+        std::vector<float> scratch;
```
LHT129, line 217: [note] The `scratch` vector is declared outside the serialization loop at line 218 and reused across iterations. For FP32 backends, `VectorAt` returns a direct pointer and `scratch` is never touched. For FP16 backends, `VectorAt` calls `scratch.resize(Dim())` each time, which is a no-op after the first iteration. The reuse is correct and avoids repeated allocations.

However, the `scratch` declaration could be moved inside the `if (fp16)` branch or scoped closer to its only use site (the `VectorAt` call at line 220) to make the ownership clearer to readers.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+    tl::expected<std::vector<uint64_t>, Error>
+    nearest(const float* vector,
+            uint64_t count,
+            uint64_t excluded = std::numeric_limits<uint64_t>::max()) const {
+        if (Size() == 0) {
+            return std::vector<uint64_t>{};
+        }
+        const auto found = Search(vector, Dim(), std::min(Size(), std::max(count * 4, ef_search_)));
+        if (not found) {
```
LHT129, line 465: [note] The `nearest()` helper at line 437 passes `std::max(count * 4, ef_search_)` as the search depth. During `BuildGraph`, `count` is `max_degree_` (at most 64), so the search depth is at least `max(256, ef_search_)`. This multiplier ensures the search explores enough candidates to find good neighbors for graph construction, but the `* 4` factor is a heuristic without an obvious derivation. Consider adding a brief comment explaining why 4× was chosen (e.g., "explore 4× the target count to improve graph quality during incremental construction").

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+        try {
+            k = std::min(k, Size());
+            if (k == 0) {
+                return std::vector<Neighbor>{};
+            }
+            const uint64_t ef = std::min(Size(), std::max(k, ef_search_));
+            // With closer as Compare, top() is the farthest candidate and pop() evicts it.
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> best(&closer);
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> accepted(
+                &closer);
```
LHT129, line 220: @/tmp/review_body_1.txt

src/lite/graph_backend.cpp
nw9663644-eng: After moving the last slot's adjacency list into the removed slot, Remove now explicitly erases any reference to the destination slot. I also added a regression check that verifies every remaining adjacency list contains no self-loop after removing a non-last element.

The Debug, ASan/UBSan, and coverage test suites all pass.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,618 @@
...
+            }
+
+            auto visit = [&](uint64_t slot) {
+                if (visited[slot] != 0) {
+                    return;
+                }
+                visited[slot] = 1;
+                Candidate next{slot, distance(query, encoded_query, slot)};
+                candidates.push(next);
+                best.push(next);
```
LHT129, line 236: @/tmp/review_body_2.txt

lite/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    include("${VSAG_ROOT}/cmake/VSAGThirdPartyOverride.cmake")
+    include("${VSAG_ROOT}/extern/catch2/catch2.cmake")
+    add_executable(lite_tests "${VSAG_ROOT}/src/lite/index_test.cpp")
+    add_executable(lite_graph_tests "${VSAG_ROOT}/src/lite/graph_backend_test.cpp")
+    target_link_libraries(lite_tests PRIVATE vsag::lite Catch2::Catch2WithMain)
```
LHT129, line 73: @/tmp/review_body_3.txt

src/lite/graph_backend.cpp
nw9663644-eng: The internal VectorAt interface now accepts caller-owned scratch storage. FP16 decoding writes into that scratch buffer instead of a shared mutable member, so independent readers no longer race on decoded_ and the returned pointer remains tied to the caller's buffer.

All internal call sites now own and reuse their scratch storage. A regression test uses two separate scratch buffers and verifies that a later VectorAt call does not overwrite the earlier result.

src/lite/graph_backend.cpp
LHT129, line 220: [critical] The `accepted` priority queue in `GraphBackend::SearchImpl` uses the wrong comparator (`&closer`), causing filtered search to return the *farthest* k candidates instead of the *closest* k.

With `closer` as the comparator, `std::priority_queue` (max-heap) places the *closest* candidate at `top()`. When `accepted.size() > k`, `accepted.pop()` removes the top (closest), so the queue retains the k *farthest* filtered candidates — the opposite of what is intended.

Fix: use `farther` as the comparator for `accepted`, so `pop()` removes the farthest and the queue retains the k closest:
```cpp
std::priority_queue<Candidate, std::vector<Candidate>, decltype(&farther)> accepted(&farther);
```

Note: the `best` queue intentionally uses `closer` to retain the farthest `ef` candidates for pruning — that is correct. Only `accepted` needs to be changed.

src/lite/graph_backend.cpp
LHT129, line 236: [suggestion] The `candidates` priority queue in `GraphBackend::SearchImpl` uses `&farther` as its comparator, which causes the search to explore candidates in *farthest-first* order (largest distance first). Standard graph-based ANN search (HNSW, NSG, etc.) explores *closest-first* because the closest candidate is most likely to lead to further improvements and trigger early termination sooner.

With `farther` as the max-heap comparator, `candidates.top()` returns the farthest candidate, and the search loop processes the worst candidates before the best ones. While the early-exit condition is correctly adapted for this order, the traversal wastes work exploring unpromising candidates.

Consider swapping to closest-first exploration:
```cpp
std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> candidates(&closer);
```
The early-exit condition would then need to use `farther`:
```cpp
if (best.size() == ef and farther(best.top(), current)) {
    break;
}
```

This is a performance concern, not a correctness bug. For a v0.1 release the current behavior is acceptable, but closest-first order will matter more as the index scales.

lite/CMakeLists.txt
LHT129, line 73: [suggestion] The `ENABLE_COVERAGE` option adds `--coverage` link flags only to the `vsag_lite` shared library target, but both test executables (`lite_tests` and `lite_graph_tests`) do not receive the coverage link flag. This means coverage data for the object files may be incomplete when running tests with coverage enabled.

To fix, add the coverage link flag to both test targets:

```cmake
if(ENABLE_COVERAGE)
    target_compile_options(vsag_lite_objects PRIVATE --coverage -O0 -g)
    target_link_options(vsag_lite PRIVATE --coverage)
    target_link_options(lite_tests PRIVATE --coverage)
    target_link_options(lite_graph_tests PRIVATE --coverage)
endif()
```

src/lite/graph_backend.cpp
nw9663644-eng: SearchImpl now visits slot 0 first, visits the last slot only when it is different from slot 0, and skips extra entry points that duplicate either endpoint.

I kept the remaining unique entry points for small indices instead of disabling all extra entry points below a size threshold, so the search behavior retains every distinct candidate while avoiding redundant endpoint visits.

src/lite/bruteforce_backend.cpp
nw9663644-eng: The shared failure() and validate() helpers have been moved to the private src/lite/backend_utils.h header, and both backends now use the same definitions. This keeps the validation and error behavior consistent without changing the public API.

src/lite/graph_backend.cpp
nw9663644-eng: I added a comment explaining that incremental graph construction explores four times the requested neighbor count before retaining the closest count candidates. This gives construction a broader candidate pool to improve graph quality while keeping the stored degree bounded by max_degree.

src/lite/index.cpp
```diff
@@ -0,0 +1,388 @@
...
+namespace vsag::lite {
+namespace {
+
+auto
+failure(ErrorType type, const char* message) {
+    return tl::unexpected(Error(type, message));
+}
+
+void
+write(std::ostream& out, uint64_t value, uint64_t bytes = 8) {
```
LHT129, line 23: [note] The `index.cpp` file (in `vsag::lite`) defines its own `failure()` helper in an anonymous namespace, while `backend_utils.h` already provides `vsag::lite::detail::failure()` with the same implementation. The two backends (`bruteforce_backend.cpp`, `graph_backend.cpp`) both include `backend_utils.h` and use `detail::failure()`. Consider having `index.cpp` also use `detail::failure()` to avoid the duplication, or alternatively move the helper to a shared internal header visible to both namespaces.

src/lite/index.cpp
```diff
@@ -0,0 +1,388 @@
...
+                }
+            }
+        }
+        if (not output) {
+            return failure(ErrorType::READ_ERROR, "snapshot write failed");
+        }
+        return {};
+    } catch (const std::ios_base::failure&) {
+        return failure(ErrorType::READ_ERROR, "snapshot write failed");
+    }
```
LHT129, line 247: [note] `Save()` returns `ErrorType::READ_ERROR` when the output stream fails (lines 3211, 3215). While `READ_ERROR` is the closest match in the existing `ErrorType` enum (which has no `WRITE_ERROR`), the name is misleading for write failures. This is a pre-existing limitation of the Full VSAG error type, not introduced by this PR, but worth noting for future API clarity.

LHT129 (commented):
## Review Summary

This is a large, well-structured PR (26 new files, ~3600 lines) adding a standalone VSAG Lite library with BruteForce and FP32/FP16 graph index capabilities. The implementation is clean, with proper separation between the public API (`include/vsag/lite/index.h`), the polymorphic backend interface (`src/lite/backend.h`), and the concrete backends.

### What works well

- **Clean API design**: The `Index` class with `tl::expected` error handling provides a consistent, hard-to-misuse interface. The `IdFilter` callback for filtered search is well-designed.
- **Snapshot format**: The versioned binary format (v1/v2/v3) with little-endian encoding, thorough validation (magic, dimensions, payload, adjacency bounds, self-links, duplicates, finite values), and big-endian host support is robust.
- **Test coverage**: Comprehensive tests cover input validation, CRUD operations, filtered search, snapshot roundtrips, corruption detection, randomized CRUD against reference, FP16 codec, ISA kernel consistency, and SIFT-scale probes. The `BuildGraph` failure leaving the index unchanged is tested.
- **SIMD dispatch**: Clean runtime CPU feature detection for FP32 (AVX512/AVX2/SSE4.1/Generic) and FP16 (AVX512/AVX2/AVX/F16C/Generic) with per-file compilation flags.
- **Error handling**: Consistent use of `tl::expected` throughout, with proper exception-to-error translation at boundaries.

### Existing review comments

There are 29 existing inline comments covering most substantive observations. Many have been addressed in the current code (geometric growth zero-capacity fix, unchecked `find()` dereference, comparator functor, ID copy elimination, two's-complement decoding, update reverse-edge removal, remove self-loop cleanup). Several [suggestion] items remain open for consideration (Remove O(total_edges) scan, link() array sizing, duplicate validate/failure helpers, entry point selection for small indices).

### New observations

Two minor [note]-level observations were added: a duplicated `failure()` helper in `index.cpp` vs `backend_utils.h`, and the use of `READ_ERROR` for write failures (pre-existing enum limitation).

### Verdict

The implementation is solid, well-tested, and the public API contract is clearly documented. The remaining open suggestions are optimizations or style preferences, not correctness issues. No blocking concerns.

src/lite/index.cpp
nw9663644-eng: The scratch vector intentionally remains immediately outside the serialization loop so FP16 backends can reuse the same caller-owned decode allocation across all slots. FP32 backends ignore it.

I added a comment at the declaration to make this ownership and reuse explicit. Moving it into a per-slot scope would lose the allocation reuse, while moving it into a separate fp16 branch would require duplicating the serialization loop.

src/lite/graph_backend.cpp
nw9663644-eng: The maximum graph degree is now defined once as K_MAX_DEGREE. Graph option validation, snapshot validation, and the ranked stack array all use this shared constant:

std::array<Candidate, K_MAX_DEGREE + 1>

This removes the possibility of raising the validation limit without updating the array capacity.

src/lite/graph_backend.cpp
nw9663644-eng: Thank you for the suggestion. I audited this path, but retained the full edge scan for correctness.

The current adjacency lists are not guaranteed to be symmetric. Degree pruning in link() can remove an outgoing edge without removing the opposite stored edge, and the snapshot loader intentionally accepts historical asymmetric adjacency lists. Therefore, extras_[slot] and extras_[last] do not necessarily enumerate every node that points to slot or last.

Limiting the repair to those two outgoing-neighbor sets could leave stale or out-of-range references after the last slot is moved.

In 3d76bfa0 I added an explanatory comment and a regression test using a restored asymmetric graph. The test verifies that incoming references to the removed slot are erased and references to the moved last slot are remapped correctly.

A future reverse-edge index could make removal local, but it would add memory and snapshot/runtime bookkeeping costs that should be evaluated separately for Lite.

src/lite/graph_backend.cpp
nw9663644-eng: The unfiltered path now pops the best heap directly into the result in reverse order, avoiding a full distance sort.

The existing public ordering also uses the external ID as the tie-break for equal distances, while the internal heap uses slots. To preserve that behavior exactly, only equal-distance groups are sorted by external ID before the result is truncated to k.

The filtered path retains its existing ordering logic. All search, filter, randomized CRUD, recall, and snapshot tests pass.

LHT129 (commented):
## Code Review Summary

Reviewed all 26 changed files (3715 additions) for PR #2904: standalone VSAG Lite library with BruteForce and FP32 graph index.

### Overall Assessment

The code quality is solid for a v0.1 release. The public API (`include/vsag/lite/index.h`) is clean and well-defined. The implementation handles edge cases properly (zero-capacity growth, empty indices, FP16 range errors, self-loop prevention in Remove). Test coverage is reasonable with ~20 test cases across index and graph backends, including randomized reference comparisons and a 100k-scale graph test.

### Issues Already Resolved

Most previously reported issues have been fixed in the current revision:
- Zero-capacity `grow()` handled explicitly
- `find()` dereference replaced with `slots_.at()` 
- Function-pointer comparator replaced with `NeighborWorseFirst` functor
- `int64_t` decoding uses `memcpy` for clarity
- Range-for loops use `const auto&` consistently
- `VectorAt` uses caller-owned scratch storage (no shared mutable member)
- `Update` cleans up old neighbor references before setting new ones
- `Remove` self-loop prevention added
- Duplicate entry points eliminated
- `failure()`/`validate()` deduplicated into `backend_utils.h`
- `index.cpp` reuses `detail::failure()` via using-declaration
- `GraphBackend::distance()` uses `select_fp32_distance()` for SIMD dispatch
- Coverage flags applied to all test targets

### On the `accepted` Queue Comparator

Two existing comments (IDs 4043902169, 4056983632) claim the `accepted` queue comparator is wrong. This assessment is **incorrect**. The current code is right:

- `closer(a, b)` returns `true` when `a` is closer than `b`
- In `std::priority_queue`, Compare returning `true` means the first argument has *lower* priority (goes after the second)
- So `closer` as Compare means closer candidates have lower priority → `top()` is the *farthest* candidate
- When `accepted.size() > k`, `pop()` removes the farthest, retaining the k closest
- This is exactly the desired behavior for filtered search

### Remaining Suggestions (Non-blocking)

Several performance/style suggestions remain open but are appropriate for a follow-up:
- **Search traversal order** (ID 4056984157): `candidates` queue uses farthest-first exploration; closest-first would be more efficient for larger graphs
- **`Remove` complexity** (ID 4056964020): O(total_edges) scan is conservative but correct given asymmetric graph edges from degree pruning; the code comment explains the rationale
- **Hardcoded array size** (ID 4056963584): `link()` uses `std::array<Candidate, 65>` tied to `K_MAX_DEGREE=64`; a `static_assert` would catch mismatches
- **`READ_ERROR` for write failures** (ID 4057027725): Pre-existing `ErrorType` limitation, not introduced by this PR

No new substantive issues found beyond what existing comments already cover.

src/lite/graph_backend.cpp
```diff
@@ -0,0 +1,647 @@
...
+        }
+        try {
+            k = std::min(k, Size());
+            if (k == 0) {
+                return std::vector<Neighbor>{};
+            }
+            const uint64_t ef = std::min(Size(), std::max(k, ef_search_));
+            // With closer as Compare, top() is the farthest candidate and pop() evicts it.
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> best(&closer);
+            std::priority_queue<Candidate, std::vector<Candidate>, decltype(&closer)> accepted(
```
LHT129, line 219: [critical] The `accepted` priority queue in `GraphBackend::SearchImpl` uses the wrong comparator, causing filtered search to return the *farthest* k candidates instead of the *closest* k.

The `accepted` queue is declared with `decltype(&closer)`. `closer(a, b)` returns `true` when `a` is closer than `b`. In a `std::priority_queue` (max-heap), the top element is the closest candidate. When `accepted.size() > k`, `accepted.pop()` removes the top (closest), so `accepted` retains the k *farthest* filtered candidates.

The final `std::sort` at line 305 sorts these k farthest candidates by distance, but they are still the wrong set — the farthest instead of the closest.

The fix is to use `farther` as the comparator for `accepted`, so that `pop()` removes the farthest candidate and the queue retains the k closest:

```cpp
std::priority_queue<Candidate, std::vector<Candidate>, decltype(&farther)> accepted(&farther);
```

Note: the `best` queue intentionally retains the farthest `ef` candidates for pruning, so its use of `closer` is correct. Only `accepted` needs to be changed.

The existing filter test at `graph_backend_test.cpp:136` may pass despite this bug because the graph topology limits which nodes are visited during search. Consider adding a test with a fully-connected small graph (e.g., 8 nodes with max_degree=7) where the search is guaranteed to visit all nodes, and verify that filtered results are truly the closest k candidates.

LHT129 (commented):
Reviewed the latest push (25f1d595, "refactor(lite): share snapshot error helper"). This is a clean refactoring that moves the `failure` helper from `index.cpp`'s anonymous namespace into `backend_utils.h` and adds clarifying comments about the `READ_ERROR` usage in save paths.

I also re-reviewed the full PR state across all core implementation files (`index.cpp`, `graph_backend.cpp`, `bruteforce_backend.cpp`, `backend.h`, `backend_utils.h`, `fp32_distance.*`, `fp16_codec.h`, `fp16_distance.*`, `lite/CMakeLists.txt`, `include/vsag/lite/index.h`). The code quality is high:

- Snapshot format validation is thorough (magic, version, dimension, count, payload, representation, overflow checks, duplicate ID detection, non-finite value rejection, trailing byte detection, endianness handling).
- Graph construction, CRUD, and search logic is correct with well-considered edge cases (empty graph, single node, filtered search, asymmetric adjacency from restored snapshots).
- The `Remove()` full-scan approach is explicitly justified by the comment about asymmetric adjacency from degree pruning and restored snapshots — a reasonable correctness-over-performance tradeoff for the stated small-data deployment goal.
- Error handling is consistent, using `tl::expected` throughout with appropriate error types.
- The CMake build is well-isolated from the Full VSAG build, with proper SIMD dispatch and install rules.

No new blocking issues found. The earlier review comments from LHT129 have been addressed in prior commits (geometric growth edge case, stable heap comparator, static archive installation).

PR #2906, opened 2026-09-09
fix(hgraph): prevent stale entry point after force-remove tail nodes (backport #2902)

LHT129:
## Summary

Backport of #2902 to the  branch.

fix(hgraph): prevent stale entry point after force-remove tail nodes

find_new_entry_point only searched route_graphs_ for a replacement entry point. When the deleted node was the sole upper-level node and also the swap-tail, entry_point_id_ remained stale and pointed past the live range. A subsequent Add used that stale ID in its bottom-graph probe, reading out-of-range codes, which crashed in SIMD load.

## Changes

- find_new_entry_point: fall back to bottom_graph_ when route_graphs_ are exhausted without a successor
- force_remove_one: retry find_new_entry_point after graph state is updated, guarding the swap_id == inner_id skip-move case
- shrink_to_fit: synchronise max_capacity_ so subsequent Add resize correctly grows flatten and graph storage
- flatten_datacell query: check codes validity before SIMD distance evaluation and throw READ_ERROR on null instead of crashing (adapted for 1.0 branch GetCodesById API)
- test_hgraph_remove: add regression tests for tail-force-remove followed by Add
- test_hgraph_stale_entry_point: add deterministic functional and concurrent stress coverage

## Files Changed

- src/algorithm/hgraph/hgraph_modify.cpp
- src/datacell/flatten_datacell.h (adapted for 1.0 API)
- tests/test_hgraph_remove.cpp
- tests/test_hgraph_stale_entry_point.cpp (new)

Fixes: #2901

### DIFF
--- a/src/algorithm/hgraph/hgraph_modify.cpp
+++ b/src/algorithm/hgraph/hgraph_modify.cpp
@@ -76,6 +76,22 @@ HGraph::find_new_entry_point() {
         }
         route_graphs_.pop_back();
     }
+
+    // No upper-level successor found -- fall back to the bottom graph.
+    // Pick any remaining node as the new entry point so the index stays
+    // usable for subsequent Add / search operations.
+    if (not find_new_ep and this->bottom_graph_ != nullptr and
+        this->bottom_graph_->TotalCount() > 0) {
+        auto total = this->total_count_.load();
+        for (InnerIdType candidate = 0; candidate < total; ++candidate) {
+            if (candidate != this->entry_point_id_ and
+                this->bottom_graph_->CheckIdExists(candidate)) {
+                this->entry_point_id_ = candidate;
+                find_new_ep = true;
+                break;
+            }
+        }
+    }
 }
 
 void
@@ -189,6 +205,13 @@ HGraph::force_remove_one(int64_t label) {
     bool was_mark_removed = false;
     {
         std::unique_lock lock(this->label_lookup_mutex_);
+        // After move_id (or when swap_id == inner_id so move_id is skipped),
+        // entry_point_id_ may still equal the deleted inner_id when
+        // find_new_entry_point failed to locate an upper-level successor.
+        // Retry with the bottom graph visible now that graph state is updated.
+        if (this->entry_point_id_ == inner_id and this->total_count_.load() > 1) {
+            this->find_new_entry_point();
+        }
         was_mark_removed = this->label_table_->IsRemoved(inner_id);
         this->label_table_->ForceRemove(label, inner_id);
         if (swap_id != inner_id) {
@@ -215,6 +238,8 @@ HGraph::shrink_to_fit() {
         route_graph->ShrinkToFit(total_count);
     }
     label_table_->ShrinkToFit(total_count);
+
+    this->max_capacity_.store(total_count);
 }
 
 void
--- a/src/datacell/flatten_datacell.h
+++ b/src/datacell/flatten_datacell.h
@@ -386,6 +386,10 @@ FlattenDataCell<QuantTmpl, IOTmpl>::query(float* result_dists,
             codes2 = this->GetCodesById(idx[i + 1], release2);
             codes3 = this->GetCodesById(idx[i + 2], release3);
             codes4 = this->GetCodesById(idx[i + 3], release4);
+            if (codes1 == nullptr or codes2 == nullptr or codes3 == nullptr or codes4 == nullptr) {
+                throw VsagException(ErrorType::READ_ERROR,
+                                    "failed to acquire codes for batch distance evaluation");
+            }
             computer->ComputeDistsBatch4(codes1,
                                          codes2,
                                          codes3,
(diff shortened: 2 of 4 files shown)


src/algorithm/hgraph/hgraph_modify.cpp
```diff
@@ -76,6 +76,22 @@ HGraph::find_new_entry_point() {
         }
         route_graphs_.pop_back();
     }
+
+    // No upper-level successor found -- fall back to the bottom graph.
+    // Pick any remaining node as the new entry point so the index stays
+    // usable for subsequent Add / search operations.
+    if (not find_new_ep and this->bottom_graph_ != nullptr and
+        this->bottom_graph_->TotalCount() > 0) {
+        auto total = this->total_count_.load();
```
LHT129, line 85: [suggestion] The `total_count_` read at line 85 is outside any lock, but `find_new_entry_point` is called from `force_remove_one` which holds `force_remove_mutex_` (via `Remove`) or `label_lookup_mutex_` (retry path). The atomic load is safe here, but the comment could clarify the locking context — specifically that this fallback is only reached after `graph_force_remove_one` has updated the bottom graph, so `CheckIdExists` reflects the post-removal state.

src/datacell/flatten_datacell.h
```diff
@@ -386,6 +386,10 @@ FlattenDataCell<QuantTmpl, IOTmpl>::query(float* result_dists,
             codes2 = this->GetCodesById(idx[i + 1], release2);
             codes3 = this->GetCodesById(idx[i + 2], release3);
             codes4 = this->GetCodesById(idx[i + 3], release4);
+            if (codes1 == nullptr or codes2 == nullptr or codes3 == nullptr or codes4 == nullptr) {
```
LHT129, line 389: [suggestion] The batch null-check error message does not identify which of the four IDs caused the failure, which will make debugging harder. Consider including the `idx[i]` range (e.g. `"failed to acquire codes for batch at ids [" + std::to_string(idx[i]) + "-" + std::to_string(idx[i+3]) + "]"`) so the failing range is traceable.

tests/test_hgraph_stale_entry_point.cpp
```diff
@@ -0,0 +1,424 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <atomic>
+#include <catch2/catch_test_macros.hpp>
+#include <future>
+#include <limits>
```
LHT129, line 18: [note] The `<future>` header is included but not used in this file. It can be removed to keep the include list minimal.

tests/test_hgraph_stale_entry_point.cpp
```diff
@@ -0,0 +1,424 @@
...
+    auto query = vsag::Dataset::Make();
+    query->Dim(DIM)->NumElements(1)->Float32Vectors(vectors.data())->Owner(false);
+    auto search_result = index->KnnSearch(query, 10, search_param);
+    REQUIRE(search_result.has_value());
+    REQUIRE(search_result.value()->GetDim() > 0);
+}
+
+TEST_CASE("HGraph stale entry point with concurrent Add (stress)",
+          "[ft][hgraph][force_remove][stale_ep][concurrent][stress]") {
+    fixtures::logger::LoggerReplacer _;
```
LHT129, line 276: [note] `kAddBatch` is defined but never used. It can be removed.

LHT129 (commented):
Overall this is a well-structured backport of #2902. The fix addresses a real crash bug where stale entry points after tail force-removes could cause out-of-bounds reads in SIMD operations.

**Core fix review:**

- `find_new_entry_point` fallback to `bottom_graph_` is correct and necessary when upper-level route graphs are exhausted.
- The retry of `find_new_entry_point` in `force_remove_one` under `label_lookup_mutex_` handles the case where `swap_id == inner_id` (skip-move path) and the first call failed to find an upper-level successor.
- `shrink_to_fit` synchronizing `max_capacity_` is essential — without it, a subsequent `Add` would not resize storage correctly, leading to out-of-bounds writes.
- The null-code checks in `flatten_datacell.h` are a good defensive layer that converts a segfault into a clean `READ_ERROR` exception.

**Test coverage:**

- Regression tests in `test_hgraph_remove.cpp` cover the tail-force-remove-then-Add and force-remove-all-then-Add scenarios.
- The new `test_hgraph_stale_entry_point.cpp` adds comprehensive coverage: small/medium/degenerate sizes, multiple seeds, sequential removes, and concurrent stress testing.
- The concurrent stress test is particularly valuable as it exercises the lock ordering between `force_remove_mutex_` and `label_lookup_mutex_`.

Left a few minor inline comments — none are blocking.

LHT129 (commented):
Reviewed the diff and the existing inline comments. The four inline comments already posted cover the key findings:

1. `total_count_` read outside lock in `find_new_entry_point` fallback — the atomic load is safe under `force_remove_mutex_` serialization, but the comment could clarify the locking context.
2. Batch null-check error message in `flatten_datacell.h` could include the failing ID range for easier debugging.
3. Unused `<future>` header in `test_hgraph_stale_entry_point.cpp`.
4. Unused `kAddBatch` constant in the concurrent stress test.

The core fix is correct: the `find_new_entry_point` fallback to `bottom_graph_`, the retry in `force_remove_one` under `label_lookup_mutex_`, and the `max_capacity_` synchronization in `shrink_to_fit` all address the root cause of the stale entry point crash. The test coverage is comprehensive across small/medium/degenerate sizes, sequential removes, and concurrent stress.

No additional blocking issues found.

wxyucs (approved):
lgtm

PR #2909, opened 2026-09-09
feat(attr): port ads attribute filtering capabilities to main

### LINKED ISSUE #2905: [feat](attr): Port ads attribute filtering capabilities to main
### Affected module
module/attr

### Affected area
not applicable

### Motivation
The ads branch contains attribute filtering capabilities that are absent from current main. Users need these capabilities upstream without carrying the older ads branch or losing fixes already present on main.

### Proposed solution
Port region-filter expressions and execution, reusable parsed expressions in search requests, attribute schema access, and support for introducing fields during attribute updates. Preserve the ads region-type semantics and bitmap intermediate reuse, adapt to current main APIs, and add regression tests and bilingual documentation.

### Alternatives considered
Continuing to maintain the ads branch, or copying its attribute implementation wholesale. Neither provides a narrowly scoped migration that preserves current main fixes.

Breaking change for precompiled C++ consumers: request layout and index virtual interfaces may change. Existing source-based string-filter usage should remain supported; binary consumers must rebuild.

### References
https://github.com/antgroup/vsag/commit/c15da2e48abe3f9a8d180d8cf48c2bc1a7149548
https://github.com/antgroup/vsag/commit/500accb121a8e1e9c5b7ff89099e7bdb37012b9f

### Additional context
This migration excludes all FastBitset memory-layout experiments.

## Related issues
_None found in the "region filter" duplicate search._

---
_Drafted with AI assistant: DeepSeekHarness:gpt-6-astra_

LHT129:
## Change Type
- [x] New feature

## Linked Issue
Closes: #2905

## What Changed
Port the attribute capabilities from ads commits c15da2e4 and 500accb1 onto current main, without the FastBitset memory-layout experiments or unrelated ads changes:
- Region filter grammar, AST and specialized FastBitset executor with reusable intermediate bitmaps.
- Public `AstParse`, borrowed `GetAttrTypeSchema()` and optional `SearchRequest::expression_` (prioritized over the string when attribute filtering is enabled) for IVF/HGraph/BruteForce.
- New-field support in the bucket attribute update overload receiving original attributes.
- Preserve FUNCTION grammar/AST compatibility only; no callback registry or function executor is introduced.
- Synchronize project-generated FC parser artifacts with ANTLR 4.13.2; no third-party version change. These checked-in artifacts are required: the current build does not regenerate the grammar.

Necessary migration corrections relative to ads:
- Honor expression-supplied field names; triggers still query the residence field.
- Check integer narrowing and negative-to-unsigned conversions; require signed type storage for -1.
- Reject unsupported non-FastBitset storage explicitly.
- Transactional executor Init, safe filter ownership and bucket reuse.
- Append the request member to preserve existing positional aggregate initialization.
- Enforce read-only schema access through const getter/parser signatures.
- Reject mismatched existing/new/original field types before erasing postings; set schema only for new fields.
- Make null attribute indexes fail fast, and classify FUNCTION execution as explicitly unsupported.
- Use unique_ptr scratch ownership and named region manager indices; preserve transactional Init and bitmap operations.

## Test Evidence
- [x] `make fmt` (clang-format 15 for maintained code; FC retains native ANTLR 4.13.2 generated layout, matching the existing main convention and avoiding unrelated reformatting)
- [ ] `make lint` (full target not run)
- [x] `make test` with targeted filter below
- [x] `make release`
- [x] Targeted clang-tidy 15: region executor, expression visitor, bucket attribute datacell

```text
make test COMPILE_JOBS=4 UT_FILTER='[AttributeBucketInvertedDataCell],[RegionFilterExecutor],[expression_visitor],[attribute_expression],[AttrTypeSchema]'
19 unit test cases: 467 assertions passed (including const schema and review regressions)
1 BruteForce functional test: 11 assertions passed
Other test binary: no matching cases (expected for these filters)
```
Local builds use `EXTRA_DEFINED` FetchContent source overrides for existing pinned CPUinfo/fmt caches because remote archive downloads were unreliable. No build workaround is committed.

Coverage includes region truth tables, all eight type-0 input combinations, bucket reuse, absent postings, malformed arguments, integer boundaries, unsupported storage, field selection, new-field update roundtrip, and pre-parsed expression precedence/result parity. Full-suite coverage threshold and separate IVF/HGraph expression integration coverage have not been measured here; CI remains required.

## Compatibility Impact
- Existing string-filter source usage remains supported. The new member is appended for aggregate compatibility.
- C++ ABI changes: SearchRequest layout and Index virtual interface; precompiled consumers must rebuild.
- Region type rules preserved: -1=all; 0=T OR (R AND H); 1=H; 2=R; 3=R OR H; 4=R AND H; 5=R AND NOT H.
- Schema pointer is borrowed. Expressions should be treated as immutable while shared by requests.

- Intermediate bitmap reuse is preserved. No full-IVF throughput/latency or concurrent-update guarantee is claimed by this PR.
- Existing index synchronization requirements continue to apply; no new concurrent schema mutation API is promised.

## Documentation Impact
- [x] Updated docs: `docs/docs/{en,zh}/src/advanced/attribute_filter.md`

## Risk and Rollback
- Medium: parser/API and specialized filter migration, with explicit ABI rebuild requirement.
- Revert this commit and rebuild consumers. No bitmap serialization-format change or memory PoC is included.

## Checklist
- [x] Linked issue
- [x] Added regression tests
- [x] Considered API/ABI compatibility
- [x] Updated bilingual documentation
- [x] Conventional commit with human DCO sign-off and AI assistance attribution


include/vsag/index.h
```diff
@@ -859,6 +860,15 @@ class Index {
             Error(ErrorType::UNSUPPORTED_INDEX_OPERATION, "Index does not support SetImmutable"));
```
LHT129, line 860: [suggestion] `GetAttrTypeSchema()` returns a non-const `AttrTypeSchema*` pointer, but the documentation explicitly states "do not mutate the schema through it". Returning `const AttrTypeSchema*` would enforce this constraint at compile time and prevent accidental mutation by callers.

Current:
```cpp
[[nodiscard]] virtual AttrTypeSchema*
GetAttrTypeSchema() const {
    return nullptr;
}
```

Suggested:
```cpp
[[nodiscard]] virtual const AttrTypeSchema*
GetAttrTypeSchema() const {
    return nullptr;
}
```

If `AttrTypeSchema`'s API isn't const-correct today, this could be deferred, but the intent should be communicated through the type system where possible.

src/attr/executor/region_filter_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+    if (this->bitset_ == nullptr) {
+        throw VsagException(ErrorType::INTERNAL_ERROR,
+                            "region_filter executor must be initialized before Run");
+    }
+    Clear();
+    Accumulate(this->region_bitset_, this->region_managers_, bucket_id);
+    Accumulate(this->residence_bitset_, this->residence_managers_, bucket_id);
+    Accumulate(this->trigger_bitset_, this->trigger_managers_, bucket_id);
+
+    this->intersection_bitset_->Or(this->region_bitset_);
```
LHT129: [note] The `intersection_bitset_` computation at lines 237-238 uses `Or(region_bitset_)` followed by `And(residence_bitset_)` to compute `R AND H`. Since `intersection_bitset_` was just cleared at the top of `Run()`, the `Or` is effectively a copy of `region_bitset_`. If `ComputableBitset` provides a `CopyFrom` or `Set` method, using it here would be more intention-revealing than relying on the `Or`-from-empty behavior.

Current:
```cpp
this->intersection_bitset_->Or(this->region_bitset_);
this->intersection_bitset_->And(this->residence_bitset_);
```

If available:
```cpp
this->intersection_bitset_->CopyFrom(this->region_bitset_);
this->intersection_bitset_->And(this->residence_bitset_);
```

src/attr/executor/executor.h
```diff
@@ -36,7 +36,8 @@ class Executor {
         : expr_(expression),
           attr_index_(attr_index),
           allocator_(allocator),
-          bitset_type_(attr_index->GetBitsetType()){};
+          bitset_type_(attr_index == nullptr ? ComputableBitsetType::FastBitset
```
LHT129: [suggestion] The base `Executor` constructor now silently falls back to `FastBitset` when `attr_index` is `nullptr`, whereas the original code would have crashed with a null-pointer dereference. This changes the failure mode for all executor subclasses: a bug that was previously a hard crash is now a silent misconfiguration.

If the intent is to allow `RegionFilterExecutor` (which validates `attr_index` itself) to pass through the base constructor safely, consider making `bitset_type_` a parameter that subclasses must supply explicitly, or add a separate protected constructor that skips the `attr_index` dereference.

Current:
```cpp
bitset_type_(attr_index == nullptr ? ComputableBitsetType::FastBitset
                                   : attr_index->GetBitsetType()){};
```

src/datacell/attribute_bucket_inverted_datacell.cpp
```diff
@@ -239,7 +239,13 @@ AttributeBucketInvertedDataCell::UpdateBitsetsByAttr(const AttributeSet& attribu
 
     for (const auto* attr : attributes.attrs_) {
         const auto& name = attr->name_;
+        auto iter = field_2_value_map_.find(name);
+        if (iter == field_2_value_map_.end()) {
+            field_2_value_map_[name] =
+                std::make_shared<AttrValueMap>(allocator_, this->bitset_type_);
+        }
         auto& value_map = this->field_2_value_map_[name];
+        this->field_type_map_.SetTypeOfField(name, attr->GetValueType());
```
LHT129: [note] `field_type_map_.SetTypeOfField(name, attr->GetValueType())` is called unconditionally on every `UpdateBitsetsByAttr` invocation, even for fields that already exist. If `SetTypeOfField` is not a trivial no-op for existing fields, this is wasteful on hot update paths. If the intent is to record the type only for newly introduced fields, the call should be moved inside the `if (iter == field_2_value_map_.end())` block.

Current:
```cpp
auto iter = field_2_value_map_.find(name);
if (iter == field_2_value_map_.end()) {
    field_2_value_map_[name] =
        std::make_shared<AttrValueMap>(allocator_, this->bitset_type_);
}
auto& value_map = this->field_2_value_map_[name];
this->field_type_map_.SetTypeOfField(name, attr->GetValueType());
```

Suggested:
```cpp
auto iter = field_2_value_map_.find(name);
if (iter == field_2_value_map_.end()) {
    field_2_value_map_[name] =
        std::make_shared<AttrValueMap>(allocator_, this->bitset_type_);
    this->field_type_map_.SetTypeOfField(name, attr->GetValueType());
}
auto& value_map = this->field_2_value_map_[name];
```

src/attr/expression.h
```diff
@@ -469,4 +472,72 @@ class NotExpression : public Expression {
...
+class FunctionExpression : public Expression {
+public:
+    FunctionExpression(std::string name,
+                       std::vector<std::string> args,
+                       std::vector<std::string> types)
+        : Expression(ExpressionType::kFunctionExpression, OpType::kNone),
+          function_name(std::move(name)),
+          arguments(std::move(args)),
+          argument_types(std::move(types)) {
+        if (arguments.size() != argument_types.size()) {
```
LHT129, line 487: [note] `FunctionExpression`'s constructor throws `std::invalid_argument` when argument/type counts mismatch, while the rest of the attribute expression subsystem consistently uses `VsagException(ErrorType::INVALID_ARGUMENT, ...)`. The `visitFunctionExpr` visitor catches `std::exception` and re-wraps it as `VsagException`, so callers see the right type, but for consistency with the rest of the codebase, consider throwing `VsagException` directly here.

Current:
```cpp
throw std::invalid_argument("function argument and type counts must match");
```

Suggested:
```cpp
                    "function argument and type counts must match");
```

src/attr/executor/region_filter_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+std::vector<const MultiBitsetManager*>
+get_managers(const AttrInvertedInterfacePtr& attr_index,
+             const std::string& field_name,
+             const std::vector<int64_t>& values) {
+    AttributeValue<T> attribute;
+    attribute.name_ = field_name;
+    auto& converted = attribute.GetValue();
+    converted.reserve(values.size());
+    for (const auto value : values) {
+        if constexpr (std::is_unsigned_v<T>) {
```
LHT129, line 36: [note] In the `get_managers` template, the `if constexpr (std::is_unsigned_v<T>)` branch contains a dead check `value < 0`. Since `T` is unsigned, `value` is first cast to `int64_t` (the parameter type), but the check `value < 0` can never be true for an unsigned target type — the `static_cast<uint64_t>(value) > std::numeric_limits<T>::max()` check already covers the lower bound (0 is the minimum for unsigned types). The `value < 0` sub-expression is dead code and should be removed to avoid misleading readers into thinking unsigned types can hold negative values.

Suggested fix: remove the `value < 0 ||` from the unsigned branch:
```cpp
if constexpr (std::is_unsigned_v<T>) {
    if (static_cast<uint64_t>(value) > std::numeric_limits<T>::max()) {
        throw VsagException(...);
    }
}
```

src/attr/executor/region_filter_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+        }
+        const auto* type_values = manager->GetOneBitset(bucket_id);
+        if (type_values == nullptr) {
+            return;
+        }
+        this->type_bitsets_[manager_index]->Or(matching_values);
+        this->type_bitsets_[manager_index]->And(type_values);
+        this->bitset_->Or(this->type_bitsets_[manager_index]);
+    };
+
```
LHT129, line 244: [suggestion] The `Run()` method uses hardcoded magic numbers (1, 2, 3, 4, 5, 6) as indices into `region_type_managers_` and `type_bitsets_`. These indices represent region filter types (e.g., residence, region, intersection, etc.) but their meaning is not documented. Consider defining named constants or an enum (e.g., `kRegionTypeResidence = 1`, `kRegionTypeRegion = 2`, `kRegionTypeAll = 6`) to make the code self-documenting and reduce the risk of off-by-one errors when the type layout changes.

For example, the check at line 263:
```cpp
if (this->region_type_managers_.size() > 6 && this->region_type_managers_[6] != nullptr)
```
would be clearer as:
```cpp
if (this->region_type_managers_.size() > kRegionTypeAll && this->region_type_managers_[kRegionTypeAll] != nullptr)
```

src/attr/executor/region_filter_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+    this->bitset_ = result.release();
+    this->region_bitset_ = region.release();
+    this->residence_bitset_ = residence.release();
+    this->trigger_bitset_ = trigger.release();
+    this->intersection_bitset_ = intersection.release();
+    for (uint64_t index = 0; index < this->type_bitsets_.size(); ++index) {
+        this->type_bitsets_[index] = type_bitsets[index].release();
+    }
+    this->own_bitset_ = true;
+}
```
LHT129, line 211: [suggestion] `Init()` releases `unique_ptr` ownership to raw member pointers before setting `own_bitset_ = true` at the very end. If any `NewBitset()` call throws (e.g. allocation failure), the already-released raw pointers will not be deleted by the destructor because `own_bitset_` is still `false`.

Consider setting `own_bitset_ = true` immediately after the first successful allocation, or wrapping the raw members in `std::unique_ptr` with a custom deleter. Alternatively, use a scope guard that sets `own_bitset_` in a deferred fashion so that partial initialization is still cleaned up.

Current flow in `Init()`:
```cpp
this->bitset_ = result.release();       // if next NewBitset() throws, bitset_ leaks
this->region_bitset_ = region.release(); // same
// ... more releases ...
this->own_bitset_ = true;               // too late
```

src/attr/executor/region_filter_executor.cpp
LHT129: This check is required: value comes from vector<int64_t>, independently of the unsigned target T. For T=uint64_t, converting -1 yields UINT64_MAX, so the upper-bound comparison alone accepts it. Keeping value < 0 prevents negative inputs from wrapping into valid unsigned values. Resolving without the suggested removal.

src/attr/executor/region_filter_executor.cpp
LHT129: The current Init allocates every bitmap into local unique_ptr owners before any release. All NewBitset calls finish before the member assignments; release and the following pointer/bool assignments do not allocate. If allocation throws, local owners clean up and no partial initialization is published. The described release-then-allocate path is absent, so resolving this specific leak report. Explicit member ownership is being improved separately.

include/vsag/index.h
LHT129: Fixed in 49acf131: GetAttrTypeSchema returns const AttrTypeSchema* throughout the public/index/inner chain; AstParse and visitor accept a const schema, and GetTypeOfField is const. Added a const-schema read regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/region_filter_executor.cpp
LHT129: ComputableBitset does not expose CopyFrom or a whole-bitmap Set operation. Kept OR from the explicitly cleared scratch bitmap to preserve the existing algorithm and reuse behavior, and added a comment explaining that it copies R before intersecting H. Addressed in 49acf131 without introducing a new bitmap API. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/executor.h
LHT129: Fixed in 49acf131: the base Executor constructor now throws INVALID_ARGUMENT immediately for a null attribute index instead of silently choosing FastBitset. Added a precise error-code/message regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/datacell/attribute_bucket_inverted_datacell.cpp
LHT129: Fixed in 49acf131: SetTypeOfField runs only for newly introduced fields. Existing field types and both incoming/original attribute sets are validated before any posting is erased, including conflicting duplicate new-field types. Tests verify rejected updates preserve schema and postings. Uses the robin-map mutable iterator value() accessor. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/expression.h
LHT129: Fixed in 49acf131: FunctionExpression now throws VsagException(INVALID_ARGUMENT) directly for argument/type count mismatch. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

include/vsag/index.h
```diff
@@ -859,6 +860,15 @@ class Index {
...
     }
 
+    /**
+     * @brief Return the attribute schema used by this index.
+     * @return nullptr when attribute filtering is not configured.
+     */
+    [[nodiscard]] virtual AttrTypeSchema*
```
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: GetAttrTypeSchema returns const AttrTypeSchema* throughout the public/index/inner chain; AstParse and visitor accept a const schema, and GetTypeOfField is const. Added a const-schema read regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/executor.cpp
```diff
@@ -37,6 +38,9 @@ Executor::MakeInstance(Allocator* allocator,
     if (std::dynamic_pointer_cast<LogicalExpression>(expression)) {
         return std::make_shared<LogicalExecutor>(allocator, expression, attr_index);
     }
+    if (std::dynamic_pointer_cast<RegionFilterExpression>(expression)) {
+        return std::make_shared<RegionFilterExecutor>(allocator, expression, attr_index);
+    }
```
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: MakeInstance explicitly rejects FunctionExpression with UNSUPPORTED_INDEX_OPERATION and a message that FUNCTION is parser compatibility only and not executable. Added an exact error-code/message regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/executor.h
```diff
@@ -36,7 +36,8 @@ class Executor {
...
+                                             : attr_index->GetBitsetType()){};
```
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: the base Executor constructor now throws INVALID_ARGUMENT immediately for a null attribute index instead of silently choosing FastBitset. Added a precise error-code/message regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/region_filter_executor.h
```diff
@@ -0,0 +1,67 @@
...
+    std::vector<const MultiBitsetManager*> region_type_managers_;
+    std::vector<const MultiBitsetManager*> region_managers_;
+    std::vector<const MultiBitsetManager*> residence_managers_;
+    std::vector<const MultiBitsetManager*> trigger_managers_;
+
+    ComputableBitset* region_bitset_{nullptr};
+    ComputableBitset* residence_bitset_{nullptr};
+    ComputableBitset* trigger_bitset_{nullptr};
+    ComputableBitset* intersection_bitset_{nullptr};
+    std::array<ComputableBitset*, 6> type_bitsets_{};
```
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: all derived scratch/type bitmaps now use unique_ptr, including the array; the destructor is defaulted. Init remains transactional; only the result bitmap transfers to the existing base-class ownership mechanism after all allocations succeed. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/region_filter_executor.cpp
LHT129: Fixed in 49acf131: added named manager indices for region semantics and replaced numeric indexing, explicitly separating MATCH_ALL slot from the stored type value -1. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

include/vsag/index.h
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: GetAttrTypeSchema returns const AttrTypeSchema* throughout the public/index/inner chain; AstParse and visitor accept a const schema, and GetTypeOfField is const. Added a const-schema read regression. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/region_filter_executor.h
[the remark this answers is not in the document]
LHT129: Fixed in 49acf131: all derived scratch/type bitmaps now use unique_ptr, including the array; the destructor is defaulted. Init remains transactional; only the result bitmap transfers to the existing base-class ownership mechanism after all allocations succeed. Release build and focused regressions passed (467 unit assertions and 11 functional assertions); core clang-tidy 15 checks also passed.

src/attr/executor/region_filter_executor.cpp
```diff
@@ -0,0 +1,275 @@
...
+get_managers(const AttrInvertedInterfacePtr& attr_index,
+             const std::string& field_name,
+             const std::vector<int64_t>& values) {
+    AttributeValue<T> attribute;
+    attribute.name_ = field_name;
+    auto& converted = attribute.GetValue();
+    converted.reserve(values.size());
+    for (const auto value : values) {
+        if constexpr (std::is_unsigned_v<T>) {
+            if (value < 0 || static_cast<uint64_t>(value) > std::numeric_limits<T>::max()) {
```
LHT129, line 37: [suggestion] Dead code: `value < 0` check in `if constexpr (std::is_unsigned_v<T>)` branch is always false for unsigned types. Consider removing the `value < 0` sub-expression from the unsigned branch.

src/attr/executor/region_filter_executor.h
```diff
@@ -0,0 +1,79 @@
...
+private:
+    enum ManagerIndex : uint64_t {
+        TRIGGER_OR_INTERSECTION = 0,
+        RESIDENCE = 1,
+        REGION = 2,
+        REGION_OR_RESIDENCE = 3,
+        REGION_AND_RESIDENCE = 4,
+        REGION_WITHOUT_RESIDENCE = 5,
+        MATCH_ALL = 6,
+        TYPE_COUNT = 6,
```
LHT129: [suggestion] `TYPE_COUNT = 6` and `MATCH_ALL = 6` share the same value. While this is technically correct (6 non-MATCH_ALL types, and MATCH_ALL is index 6 into `region_type_managers_`), it is confusing: if someone later adds a 7th type and increments TYPE_COUNT to 7, MATCH_ALL would silently become 7, breaking the `region_type_managers_` index. Consider making MATCH_ALL an explicit constant (e.g., `MATCH_ALL = 6`) and using a separate sentinel like `TYPE_COUNT = 6` with a static_assert or comment clarifying they happen to be equal but serve different purposes.

LHT129 (commented):
整体来看这个 PR 质量不错，将 ads 分支的属性过滤能力移植到 main，代码结构清晰，测试覆盖到位。

## 主要发现

### 建议项
1. **`region_filter_executor.cpp:42`** — `get_managers<T>()` 中 unsigned 分支的 `value < 0` 检查是死代码，对无符号类型永远为 false。
2. **`region_filter_executor.h:51`** — `TYPE_COUNT = 6` 和 `MATCH_ALL = 6` 共享同一数值，虽然当前逻辑正确（`type_bitsets_` 用 TYPE_COUNT 做数组大小，MATCH_ALL 用作 `region_type_managers_` 的索引），但未来若新增类型并递增 TYPE_COUNT，MATCH_ALL 会静默变为错误值。

## 正面评价
- `RegionFilterExecutor` 实现完整，包含完整的真值表测试覆盖
- `AstParse()` 公共 API 设计合理，通过 `SearchRequest::expression_` 支持预解析表达式复用
- 各算法模块（IVF/HGraph/BruteForce）的集成方式一致，优先使用预解析表达式
- 文档更新到位，attribute_filter.md 新增了 region_filter 语法说明

src/attr/executor/region_filter_executor.h
LHT129: [note] `MATCH_ALL = 6` and `TYPE_COUNT = 6` share the same numeric value but serve different purposes: `MATCH_ALL` indexes `region_type_managers_` (7 elements, index 6 → value -1), while `TYPE_COUNT` sizes `type_bitsets_` (6 elements, indices 0–5). The equality is coincidental — `MATCH_ALL` happens to equal the count of non-MATCH_ALL types. If the type list ever changes, these could diverge silently.

Consider expressing the relationship explicitly so future readers don't have to deduce it:
```cpp
MATCH_ALL = 6,
TYPE_COUNT = MATCH_ALL,
```
This makes it clear that `TYPE_COUNT` currently equals `MATCH_ALL` but is semantically a count for the `type_bitsets_` array.

src/datacell/attribute_bucket_inverted_datacell.cpp
```diff
@@ -239,7 +262,15 @@ AttributeBucketInvertedDataCell::UpdateBitsetsByAttr(const AttributeSet& attribu
...
-        auto& value_map = this->field_2_value_map_[name];
+        auto iter = field_2_value_map_.find(name);
+        if (iter == field_2_value_map_.end()) {
+            iter =
+                field_2_value_map_
+                    .emplace(name, std::make_shared<AttrValueMap>(allocator_, this->bitset_type_))
+                    .first;
+            this->field_type_map_.SetTypeOfField(name, attr->GetValueType());
+        }
+        auto& value_map = iter.value();
```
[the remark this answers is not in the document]
LHT129: `field_2_value_map_` is a `tsl::robin_map` (see `src/hash_types.h:31`), not `std::unordered_map`. `tsl::robin_map::iterator` does provide a `.value()` method that returns a reference to the mapped value, so this code is correct and compiles fine.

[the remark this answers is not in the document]
LHT129: Confirmed the clarification above: this is tsl::robin_map, whose iterator.value() provides mutable mapped-value access. This code passed Release compilation and targeted datacell tests. The std::unordered_map iterator assumption does not apply; resolving without changing the correct accessor.

src/attr/executor/region_filter_executor.cpp
LHT129: This repeats the earlier false positive: value has signed int64_t type from the input vector, not type T. For an unsigned destination, negative inputs are possible and must be rejected before conversion. In particular uint64_t(-1) equals UINT64_MAX, so an upper-bound check alone will not reject it. Keeping the necessary check and resolving.

src/attr/executor/region_filter_executor.h
LHT129: Addressed in 32e3c40b: MATCH_ALL remains explicitly 6 with a comment identifying the manager slot for stored type -1; TYPE_COUNT now equals MATCH_ALL, with a comment clarifying that only the preceding six types need scratch bitmaps. This makes the current layout relationship explicit without changing semantics. make fmt and targeted RegionFilterExecutor tests passed.

LHT129: Addressed in 32e3c40b: MATCH_ALL remains explicitly 6 with a comment identifying the manager slot for stored type -1; TYPE_COUNT now equals MATCH_ALL, with a comment clarifying that only the preceding six types need scratch bitmaps. This makes the current layout relationship explicit without changing semantics. make fmt and targeted RegionFilterExecutor tests passed.

LHT129 (commented):
## Code Review Summary

I have completed a thorough review of this PR (1914 additions, 404 deletions across 39 files) which ports ads attribute filtering capabilities to main.

### Overall Assessment

The code quality is very high. The region filter implementation is well-structured with clear separation of concerns, proper RAII memory management, and comprehensive test coverage for all region types (-1 through 5).

### Key Observations

1. **Region Filter Logic (region_filter_executor.cpp)**: The implementation correctly handles all region filter types (MATCH_ALL, TRIGGER_OR_INTERSECTION, RESIDENCE, REGION, REGION_OR_RESIDENCE, REGION_AND_RESIDENCE, REGION_WITHOUT_RESIDENCE). The `Not()` computation for REGION_WITHOUT_RESIDENCE is correct (R AND NOT H).

2. **Type Safety**: `get_managers<T>()` properly handles signed/unsigned integer narrowing with bounds checking. The `GetAttrTypeSchema()` returning `const AttrTypeSchema*` maintains const correctness.

3. **Memory Management**: Scratch bitmaps are created as `unique_ptr` in `Init()` and moved to members, preventing leaks during partial initialization.

4. **Previous Review Issues**: Most of the 30 existing review comments have been addressed in commit `49acf131`.

5. **Two Unresolved Comments Are False Positives**:
   - Comment about `iter.value()` not compiling: `tsl::robin_map` iterators do have a `.value()` method, so this is valid.
   - Comment about `value < 0` being dead code: `value` is `int64_t`, so the check is meaningful even in the `std::is_unsigned_v<T>` branch.

6. **Test Coverage**: `region_filter_executor_test.cpp` covers all region types -1 through 5, including edge cases.

### Conclusion

No blocking issues found. The PR is well-implemented and ready for merge.

src/datacell/attribute_bucket_inverted_datacell.cpp
LHT129, line 273: [critical] `iter` is a `std::unordered_map::iterator` from `field_2_value_map_.find(...)`. Iterators do not have a `.value()` method — this is a compilation error. Use `iter->second` instead.

```cpp
// Current (line 273):
auto& value_map = iter.value();

// Should be:
auto& value_map = iter->second;
```

LHT129: This is the same iterator-type false positive discussed in the earlier resolved threads. `field_2_value_map_` is declared as `vsag::UnorderedMap<std::string, ValueMapPtr>` in src/datacell/attribute_bucket_inverted_datacell.h:65. That alias is **tsl::robin_map**, not std::unordered_map (src/hash_types.h:30-35). Its iterator.value() is the supported mutable mapped-value accessor; iter->second exposes a const mapped value here and cannot bind to the existing insert_by_type(ValueMapPtr&) helper. The separate std::unordered_map used for incoming_types validation is not the container used by this iterator. Release compilation and the targeted attribute update regressions have passed with this accessor. No code change is needed; resolving this duplicate finding.

Pinned type definition: https://github.com/antgroup/vsag/blob/32e3c40b5aa740f7e7ec9fff3bef2e30124a1398/src/hash_types.h#L30-L35
Field declaration: https://github.com/antgroup/vsag/blob/32e3c40b5aa740f7e7ec9fff3bef2e30124a1398/src/datacell/attribute_bucket_inverted_datacell.h#L65

wxyucs (approved):
lgtm

PR #2927, opened 2026-09-10
refactor(attr): route region predicates through immutable executor factories

LHT129:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

Internal refactor (`kind/improvement`); a linked issue is not required. Related-issue search for region filter registry found no matches.

## What Changed

- Introduce an immutable named executor-factory registry and isolate compile-linked builtin registration from core executor dispatch.
- Route the existing region predicate through its registered factory while retaining the specialized bitmap executor.
- Keep existing query syntax and unsupported generic FUNCTION behavior. This is an internal extension seam, not a public plugin SDK or dynamic-library ABI.
- Exclude the experimental request toggle, benchmark target and POC artifacts from the production change.
- Add focused registry/dispatch/lifecycle coverage and explain scope in English and Chinese attribute-filter docs.

## Test Evidence

- Release build with tests enabled, isolated dependencies, GCC 11.4.0.
- `make fmt` with clang-format 15.0.7.
- Focused clang-tidy 15 for changed production and test translation units.
- Attribute executor/parser/manager suite: **31,013 assertions in 27 test cases**, seed 424242 (final `tests-verified.log`). Dedicated region/registry run: **1,683 assertions in 5 cases**.
- `git diff --check`.

No full-repository `make test`, `make lint`, sanitizer run or coverage measurement is claimed.

## Compatibility Impact

- No intended public API or query-semantic changes; final diff to verify removal of all POC-only public fields.
- Region's existing storage/type restrictions remain unchanged.
- Generic FUNCTION and top-level NOT execution are not added by this change.

The prior A/B POC compared real executor construction and actual IVF/BruteForce searches on a shared Xeon host. The revised POC observed approximately 20–56 ns additional construction/Init/destruction cost (1.88–3.02%); search paired median differences ranged from roughly -0.53% to +0.75% in that synthetic matrix. These are historical POC measurements, not a benchmark of this final refactored PR. They do not establish zero overhead, a universal sub-percent bound, or production-tail equivalence. Host load and independently randomized IVF training limit cross-process comparisons.

Factory resolution is in executor construction, not Run(bucket_id) or candidate checks. Registry definitions are immutable; query executors own independent mutable state. No runtime unload/hot replacement is introduced.

## Documentation Impact

- English and Chinese canonical attribute filtering documentation updated (exact paths to confirm in final diff).

## Risk and Rollback

- Risk: internal dispatch/registration correctness; focused tests exercise malformed registration, region behavior, composition and state isolation.
- Rollback: revert this refactor; no index-storage format migration intended.

## Checklist

- [x] Final diff, API compatibility and doc scope reviewed
- [x] Focused tests and exact formatting/lint evidence recorded
- One Conventional Commit with ordered human DCO and AI-assistance trailers.
- Fork-first submission; `kind/improvement` and `version/1.1` labels.

### DIFF
--- a/src/attr/executor/builtin_executor.cpp
+++ b/src/attr/executor/builtin_executor.cpp
@@ -1,0 +1,37 @@
+// Copyright 2024-present the vsag project
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy at http://www.apache.org/licenses/LICENSE-2.0
+#include "builtin_executor.h"
+
+#include "executor_factory_registry.h"
+#include "region_filter_executor.h"
+
+namespace vsag {
+namespace {
+ExecutorPtr
+make_region(Allocator* allocator,
+            const ExprPtr& expression,
+            const AttrInvertedInterfacePtr& attr_index) {
+    return std::make_shared<RegionFilterExecutor>(allocator, expression, attr_index);
+}
+}  // namespace
+
+const ExecutorFactoryRegistry&
+ExecutorFactoryRegistry::Builtins() {
+    static const ExecutorFactoryRegistry registry{{"region_filter", make_region}};
+    return registry;
+}
+
+ExecutorPtr
+CreateBuiltinExecutor(Allocator* allocator,
+                      const ExprPtr& expression,
+                      const AttrInvertedInterfacePtr& attr_index) {
+    // Keep the existing grammar and typed arguments; generic FUNCTION remains unsupported.
+    if (std::dynamic_pointer_cast<RegionFilterExpression>(expression)) {
+        return ExecutorFactoryRegistry::Builtins().Create(
+            "region_filter", allocator, expression, attr_index);
+    }
+    return nullptr;
+}
+}  // namespace vsag
--- a/src/attr/executor/executor_factory_registry.cpp
+++ b/src/attr/executor/executor_factory_registry.cpp
@@ -1,0 +1,35 @@
+// Copyright 2024-present the vsag project
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy at http://www.apache.org/licenses/LICENSE-2.0
+#include "executor_factory_registry.h"
+
+namespace vsag {
+
+ExecutorFactoryRegistry::ExecutorFactoryRegistry(std::initializer_list<Entry> entries)
+    : factories_([&entries] {
+          std::unordered_map<std::string, Factory> factories;
+          for (const auto& entry : entries) {
+              if (entry.first.empty() || entry.second == nullptr ||
+                  !factories.emplace(entry.first, entry.second).second) {
+                  throw VsagException(ErrorType::INVALID_ARGUMENT,
+                                      "invalid or duplicate executor factory registration");
+              }
+          }
+          return factories;
+      }()) {
+}
+
+ExecutorPtr
+ExecutorFactoryRegistry::Create(const std::string& name,
+                                Allocator* allocator,
+                                const ExprPtr& expression,
+                                const AttrInvertedInterfacePtr& attr_index) const {
+    auto factory = factories_.find(name);
+    if (factory == factories_.end()) {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
+                            "unknown executor factory: " + name);
+    }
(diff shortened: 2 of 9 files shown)


src/attr/executor/executor_factory_registry.cpp
```diff
@@ -0,0 +1,35 @@
...
+          return factories;
+      }()) {
+}
+
+ExecutorPtr
+ExecutorFactoryRegistry::Create(const std::string& name,
+                                Allocator* allocator,
+                                const ExprPtr& expression,
+                                const AttrInvertedInterfacePtr& attr_index) const {
+    auto factory = factories_.find(name);
```
LHT129, line 28: [suggestion] The variable name `factory` is misleading here — `find()` returns an iterator, not a factory. Consider renaming to `it` or `iter` for clarity.

Current:
```cpp
auto factory = factories_.find(name);
if (factory == factories_.end()) {
```

Suggested:
```cpp
auto it = factories_.find(name);
if (it == factories_.end()) {
                        "unknown executor factory: " + name);
}
return it->second(allocator, expression, attr_index);
```

src/attr/executor/executor_factory_registry_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+    auto expr = AstParse(R"(region_filter(kind,geo,home,"1","1","2"))");
+    // Instrument a factory only in this correctness test, never in timed production code.
+    static int factory_calls = 0;
+    factory_calls = 0;
+    const ExecutorFactoryRegistry counted{
+        {"region_filter",
+         +[](Allocator* allocator, const ExprPtr& expression, const AttrInvertedInterfacePtr& attrs)
+             -> ExecutorPtr {
```
LHT129, line 99: [suggestion] This test directly accesses private members of RegionFilterExecutor (filter_, bitset_, allocator_, expr_, attr_index_). These are implementation details that should not be exposed in tests. The test should verify behavior through the public API (Init, Run, CheckValid) rather than inspecting internal state.

If the goal is to verify that two executors have independent bitmaps, this can be tested by checking that mutations through one executor do not affect the other's CheckValid results — without accessing bitset_ directly.

If the goal is to verify that the executor retains shared ownership of the expression and index after reset(), a simpler approach is to verify that Run(0)->CheckValid(...) still returns the correct result after the shared pointers are released.

src/attr/executor/builtin_executor.cpp
```diff
@@ -0,0 +1,37 @@
...
+const ExecutorFactoryRegistry&
+ExecutorFactoryRegistry::Builtins() {
+    static const ExecutorFactoryRegistry registry{{"region_filter", make_region}};
+    return registry;
+}
+
+ExecutorPtr
+CreateBuiltinExecutor(Allocator* allocator,
+                      const ExprPtr& expression,
+                      const AttrInvertedInterfacePtr& attr_index) {
```
LHT129, line 29: [note] The `CreateBuiltinExecutor` function checks `std::dynamic_pointer_cast<RegionFilterExpression>(expression)` to decide whether to dispatch, then `RegionFilterExecutor`'s constructor internally performs the same `dynamic_cast` again. This results in a redundant RTTI check on the success path.

Since `CreateBuiltinExecutor` returns `nullptr` for non-matching types (which is a reasonable guard to avoid the exception path), the double-check is a minor overhead. If the registry is extended with more factories in the future, consider having `CreateBuiltinExecutor` pass a pre-validated typed pointer to avoid the redundant cast.

src/attr/executor/executor_factory_registry_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+         [](int64_t) { return false; }}};
+    for (const auto& item : cases) {
+        auto expr = AstParse(item.first, &index->field_type_map_);
+        auto registered = Executor::MakeInstance(alloc.get(), expr, index);
+        registered->Init();
+        registered->Init();
+        for (int repeat = 0; repeat < 4; ++repeat) {
+            registered->Clear();
+            const auto* result = registered->Run(0);
+            for (int64_t id = 0; id < 64; ++id) {
```
LHT129, line 83: [suggestion] `AstParse` is called with `&index->field_type_map_`, which accesses a private member of `AttributeInvertedInterface`. This is the same pattern as the `bitset_` / `filter_` accesses below — the test relies on internal implementation details.

Consider whether `AstParse` can obtain the field type map through a public accessor on the index, or whether the test can use a different approach that does not require peeking into private state.

Issue #2582, opened 2026-07-30
[improve](other): make streaming block deserialization forward-only


wxyucs:
### Kind of improvement

refactor

### Affected area

other

### Summary

`ReadSeekableBlockPayload()` currently materializes every known streaming TLV payload before invoking its component deserializer:

- payloads at or below `Options::block_size_limit()` are copied into a payload-sized `std::vector`;
- larger payloads are copied through a `block_size_limit` buffer into `std::tmpfile()` and read back through a seekable adapter.

This preserves random-access `StreamReader` behavior and validates the complete payload checksum before deserialization, but it defeats the intended forward-only, bounded-memory behavior for large streaming artifacts.

Refactor the streaming TLV deserialization path so all reachable component deserializers use only forward `Read` and `Skip` operations. Compute the payload checksum incrementally while bytes are consumed, drain any unconsumed suffix with a small fixed-size buffer, validate the checksum at the end, and remove the payload-sized memory and temporary-file materialization paths.

> [!IMPORTANT]
> Confirmed compatibility and failure-semantics decisions:
> - Existing `IVF_PARTITION_STRATEGY` V1 streaming artifacts may be rejected. Introduce V2 for the forward-only nested representation instead of retaining V1 seek/materialization compatibility.
> - If `DeserializeStreaming()` returns an error after component state has been modified, the target index object is considered unusable and must be discarded. Whole-block checksum validation is allowed to complete after deserialization.

Source ID is not a blocker for this work: HGraph's streaming-specific label reader already uses metadata and reads Source ID state sequentially without probing or rewinding.

### Expected benefit

- Eliminate the only `std::tmpfile()` use in the repository's streaming payload reader.
- Reduce extra payload-buffer memory from up to `block_size_limit` (128 MiB by default) to a small fixed drain buffer, proposed as 8 KiB.
- Avoid writing large TLV payloads to temporary storage and reading them back before component deserialization.
- Preserve true forward-only input support for `DeserializeStreaming()` and `Index::Load()`.
- Make hidden component dependencies on random access fail explicitly through a forward-only reader.
- Preserve supported ReaderIO behavior by modeling its requirement as forward `Skip`, rather than random `Seek`.

### Benchmark evidence

Not collected yet. This is primarily a serialization refactor. The implementation PR should compare at least:

- wall time for loading an artifact containing a TLV payload larger than the former `block_size_limit`;
- temporary-disk bytes written;
- peak additional memory attributable to the TLV payload adapter.

[...]

PR #2932, opened 2026-09-10
refactor(hgraph): deserialize streaming blocks forward-only

wxyucs:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

Part of #2582 — HGraph only. Other indexes retain the existing buffered/temp-file helper; this PR does not close the whole issue.

## What Changed

- Add bounded forward-only TLV payload reading with incremental CRC validation, avoiding payload-sized staging and temporary files for HGraph.
- Route HGraph block callbacks and external precise readers through this path.
- Add forward Skip support, preserve ReaderIO range binding, and deserialize ConjugateGraph sequentially without changing its valid wire format.
- Add regression coverage for optional data, fused/non-fused RaBitQ, malformed payloads, checksums, and streaming entry points.
- Include reproducible resource benchmarks under scripts/benchmarks/.

## Compatibility Impact — human review requested

Two legacy RaBitQ compatibility paths are disabled but deliberately preserved as commented code with explanations above them:
1. Missing supplement_io_type fallback that probes and rewinds older snapshots.
2. Loading legacy split payloads into fused storage via probing and rewinding.

Please review these comments before deciding whether to remove them permanently. Corresponding old compatibility assertions are also retained as review comments. These behavior changes affect ordinary deserialization as well as streaming. Current-format fused model-only and ordinary split payloads remain supported.

StreamReader::Skip becomes virtual: downstream binary consumers should rebuild.
CRC validation occurs after callback mutations; discard the destination index if deserialization fails.

## Test Evidence

- Debug and coverage builds succeeded.
- Original 944 unit cases passed across eight shards after updating the obsolete compatibility expectation and rerunning its entire 118-case shard; the subsequently added HGraph unit case also passed (166 assertions). Total tested: 945 cases.
- Targeted unit run: 58 cases / 48,835 assertions passed.
- Final HGraph functional run: 14 cases / 3,183 assertions passed.
- clang-format 15.0.7 checks and git diff --check passed.
- Unit-only coverage: 32,445 / 44,473 lines = **73.0%**.
- Added/changed executable production lines: **82/82 covered**, including **27/27** in the new forward reader/helper.

**Outstanding acceptance limitation:** repository-wide unit line coverage is below the required 90%. A main-branch coverage baseline was not measured, so this gap is not attributed to this PR. Functional/benchmark profiles were kept outside the unit coverage tree. This is not an all-gates-passed claim; the PR is draft for compatibility review and coverage follow-up. Full lint was not run.

A real HGraph artifact of 134,329,419 bytes (1,024 vectors; extra-info block above 128 MiB) was loaded in fresh processes using the same coverage-instrumented Debug library and component parsers. A benchmark-only LD_PRELOAD shim selects the old helper for comparison.

- Peak RSS reduced by approximately 128 MiB.
- Temporary output reduced from 134,221,824 bytes per load to zero.
- Three load samples: old helper 4.33437 / 4.33281 / 4.33939 s; forward helper 4.25345 / 4.26048 / 4.24500 s.

These are local resource measurements, not production latency guarantees or a historical full-checkout comparison. See scripts/benchmarks/README.md for methodology and reproduction. No new concurrency guarantee is introduced.

## Documentation Impact

- [x] Updated English and Chinese website serialization documentation.
- [x] Added benchmark methodology and reproduction instructions.

## Risk and Rollback

- Medium risk: explicitly disabled legacy compatibility, shared Skip/component code, and failure-after-mutation semantics.
- Rollback: revert this commit to restore the previous reader and compatibility paths.

## Checklist

- [x] Linked the relevant issue without closing the remaining index work.
- [x] Added/updated tests.
- [x] Considered API/ABI and behavioral compatibility.
- [x] Updated documentation.
- [x] Conventional commit with human sign-off and AI assistance attribution.
- [ ] Repository-wide >=90% unit coverage.
- [ ] Human approval of the two disabled legacy compatibility paths.

### DIFF
--- a/src/algorithm/hgraph/hgraph_serialize.cpp
+++ b/src/algorithm/hgraph/hgraph_serialize.cpp
@@ -15,7 +15,6 @@
 #include <fmt/format.h>
 
 #include <algorithm>
-#include <array>
 #include <limits>
 
 #include "common.h"
@@ -48,85 +47,6 @@ dump_basic_info_for_log(const JsonType& basic_info) {
     }
     return log_basic_info.Dump(4);
 }
-
-class PayloadChecksumStreamReader : public StreamReader {
-public:
-    PayloadChecksumStreamReader(StreamReader& reader, const StreamBlockHeader& header)
-        : StreamReader(header.value_len),
-          reader_(reader),
-          expected_checksum_(header.payload_checksum) {
-    }
-
-    void
-    Read(char* data, uint64_t size) override {
-        const auto cursor = reader_.GetCursor();
-        if (cursor > length_ || size > length_ - cursor) {
-            throw VsagException(ErrorType::READ_ERROR,
-                                "checksum stream reader exceeds payload boundary");
-        }
-        if (cursor > checksum_cursor_) {
-            checksum_range(checksum_cursor_, cursor - checksum_cursor_);
-        }
-
-        reader_.Read(data, size);
-        if (cursor + size > checksum_cursor_) {
-            const auto checksum_offset = checksum_cursor_ > cursor ? checksum_cursor_ - cursor : 0;
-            const auto checksum_size = size - checksum_offset;
-            crc_ = StreamHeader::UpdateChecksum(
-                crc_, std::string_view(data + checksum_offset, checksum_size));
-            checksum_cursor_ += checksum_size;
-        }
-    }
-
-    void
-    Seek(uint64_t cursor) override {
-        if (cursor > length_) {
-            throw VsagException(ErrorType::READ_ERROR,
-                                "checksum stream reader seek exceeds payload boundary");
-        }
-        reader_.Seek(cursor);
-    }
-
-    [[nodiscard]] uint64_t
-    GetCursor() const override {
-        return reader_.GetCursor();
-    }
-
-    void
-    Validate() {
-        if (checksum_cursor_ < length_) {
-            checksum_range(checksum_cursor_, length_ - checksum_cursor_);
-        }
-        if (StreamHeader::FinalizeChecksum(crc_) != expected_checksum_) {
-            throw VsagException(ErrorType::INVALID_BINARY,
-                                "streaming block payload checksum mismatch");
-        }
-    }
-
-private:
-    void
-    checksum_range(uint64_t offset, uint64_t size) {
-        constexpr uint64_t k_buffer_size = 8192;
-        std::array<char, k_buffer_size> buffer{};
-        const auto original_cursor = reader_.GetCursor();
-        reader_.Seek(offset);
-        uint64_t remaining = size;
-        while (remaining > 0) {
-            const auto read_size = std::min<uint64_t>(remaining, k_buffer_size);
-            reader_.Read(buffer.data(), read_size);
-            crc_ = StreamHeader::UpdateChecksum(crc_, std::string_view(buffer.data(), read_size));
-            remaining -= read_size;
-        }
-        checksum_cursor_ += size;
-        reader_.Seek(original_cursor);
-    }
-
-private:
-    StreamReader& reader_;
-    uint32_t expected_checksum_{0};
-    uint32_t crc_{StreamHeader::InitialChecksum()};
-    uint64_t checksum_cursor_{0};
-};
 
 }  // namespace
 
@@ -709,7 +629,7 @@ HGraph::read_streaming_body(StreamReader& reader,
 
         switch (static_cast<StreamSerializationTag>(block_header.tag)) {
             case StreamSerializationTag::LABEL_TABLE:
-                ReadSeekableBlockPayload(block_reader, block_header, [this](StreamReader& block) {
+                ReadForwardBlockPayload(block_reader, block_header, [this](StreamReader& block) {
                     this->deserialize_label_info_streaming(block);
                 });
                 loaded_label_table = true;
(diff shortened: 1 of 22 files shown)


src/storage/tlv_section.cpp
```diff
@@ -39,6 +39,53 @@ namespace vsag {
...
+    }
+
+    [[nodiscard]] uint64_t
+    GetCursor() const override {
+        return cursor_;
+    }
+
+    void
+    Finish(uint32_t expected_checksum) {
+        this->Skip(length_ - cursor_);
```
LHT129: [note] `this->Skip(length_ - cursor_)` drains remaining bytes through `StreamReader::Skip()` → `SkipForward()` → virtual `Read()`, so drain bytes are included in the incremental CRC before the final `FinalizeChecksum` comparison. This is correct but the indirection through `Skip` makes it non-obvious that drain bytes are checksummed — a one-line comment would help future readers verify this at a glance.

src/storage/tlv_section.cpp
```diff
@@ -329,6 +376,19 @@ ReadSeekableBlockPayload(StreamReader& reader,
     validate_seekable_block_cursor(block_reader, header);
 }
 
+void
+ReadForwardBlockPayload(StreamReader& reader,
+                        const StreamBlockHeader& header,
+                        const std::function<void(StreamReader&)>& deserialize) {
+    if (header.value_len == std::numeric_limits<uint64_t>::max()) {
+        throw VsagException(ErrorType::UNSUPPORTED_INDEX_OPERATION,
```
LHT129: [note] If `deserialize(block)` throws, `block.Finish()` is never called, leaving the source reader at an intermediate position within the payload. The docs correctly state that a failed `DeserializeStreaming` means the index must be discarded, but any caller that catches the exception and attempts to continue reading the stream (e.g. to skip the corrupted block and try recovering later blocks) would encounter corrupted stream state. Consider whether a best-effort drain in a catch block would be worthwhile for robustness.

src/storage/stream_reader.cpp
```diff
@@ -31,6 +31,11 @@
 
 namespace vsag {
 
+void
+StreamReader::Skip(uint64_t size) {
```
LHT129: [note] The default `StreamReader::Skip()` delegates to `SkipForward()` which reads through virtual `Read()`. This means any `StreamReader` subclass that validates consumed bytes in `Read()` (like `ForwardChecksumReader`) will correctly account for skipped bytes. The comment on the declaration says "Wrappers that validate consumed bytes must not bypass Read()" — this is the right design. However, subclasses like `BoundedForwardReader` (not modified in this PR) inherit this default and will use the 8 KiB buffered path for skip operations. If `Skip` becomes a hotter path in the future, those subclasses may want direct overrides for efficiency.

src/impl/conjugate_graph.cpp
```diff
@@ -213,39 +213,42 @@ tl::expected<void, Error>
...
-        in_stream.Seek(cur_pos + footer_offset);
-        footer_.Deserialize(in_stream);
-
-        offset = sizeof(memory_usage_);
-        in_stream.Seek(cur_pos + offset);
-        while (offset != memory_usage_ - FOOTER_SIZE) {
+        const uint32_t footer_offset = memory_usage_ - FOOTER_SIZE;
+        // The existing format stores the footer after all adjacency entries. Read it last so
+        // this component also works inside a forward-only streaming block.
+        while (offset < footer_offset) {
```
LHT129: [note] The forward-only refactoring reads adjacency entries sequentially and defers footer parsing to the end. The bounds checks (entry header overlap at line 231, neighbor count overflow at line 242) correctly guard against malformed payloads where `memory_usage_` is too small. The `memory_usage_ < sizeof(memory_usage_) + FOOTER_SIZE` check at line 223 was also tightened from `<=` to `<` which is more precise since `memory_usage_` includes its own size. These defensive checks are a good addition for the forward-only path.

src/algorithm/hgraph/hgraph_serialize.cpp
```diff
@@ -709,7 +629,7 @@ HGraph::read_streaming_body(StreamReader& reader,
 
         switch (static_cast<StreamSerializationTag>(block_header.tag)) {
             case StreamSerializationTag::LABEL_TABLE:
-                ReadSeekableBlockPayload(block_reader, block_header, [this](StreamReader& block) {
+                ReadForwardBlockPayload(block_reader, block_header, [this](StreamReader& block) {
```
LHT129: [note] `ReadForwardBlockPayload` is called inside a `while (true)` loop that reads block headers. If the checksum validation in `Finish()` throws `INVALID_BINARY`, the exception propagates out of `read_streaming_body`. The caller is expected to catch this and discard the index. Consider whether the loop should break on the first checksum failure or whether there is value in attempting to read subsequent blocks for diagnostic purposes. The current fail-fast behavior is correct for production but may complicate debugging of corrupted streams.

LHT129: [note] The `loaded_*` boolean flags (`loaded_label_table_`, `loaded_code_slot_map_`, etc.) are set inside the deserialization callbacks BEFORE `ReadForwardBlockPayload` returns. This means a flag can be `true` even though the checksum validation in `Finish()` has not yet run. If an exception occurs during `Finish()`, the index object will have inconsistent state (some flags set, some not). This is consistent with the documented "discard on failure" contract, but callers that inspect the index after a caught exception may see misleading state.

PR #2933, opened 2026-09-10
perf(hgraph): add opt-in adaptive NSW neighbor pruning

LightWant:
## What changed

Add opt-in adaptive L2 neighbor pruning for HGraph NSW bottom-layer insertion, including optimized RaBitQ split/fused construction. The selector relaxes or tightens alpha from first-pass accepted/rejected counts and can re-prune full reverse lists under the existing lock. Queries and the disabled build path retain their existing behavior.

Configuration is flat under `index_param`:

```json
{
  "adaptive_pruning": true,
  "adaptive_pruning_adjust_step": 0.06,
  "adaptive_pruning_apply_to_reverse": false,
  "adaptive_pruning_apply_to_upper": false
}
```

Unconditional rejected-neighbor filling and `fill_rejected` were removed. The Pyramid GIST 1M ablation of this shared selector showed no consistent single-factor benefit from filling (−3.4% to +4.7% QPS across two alphas and three recall targets; one parallel-built index per configuration). This motivates simplifying the selector, not a claim that filling never helps other datasets or quantizers. Reverse adaptation remains opt-in because its effect depended on alpha.

## Compatibility and scope

Defaults are disabled, step 0.06, reverse false, upper false. Enabled mode requires NSW, L2, finite thresholds, and bottom-layer scope; imported-cache construction is rejected. Non-full reverse lists still append directly. Serialized enabled indexes require matching alpha and policy settings. Missing adaptive fields retain legacy disabled behavior. The former experimental nested object and removed fill option are rejected; indexes written with that experimental schema must be rebuilt.

English/Chinese configuration and algorithm docs and regression tests are updated. Pyramid integration is tracked in #2934.

## Validation

- clang-format 15 and whitespace checks on changed C++ files.
- clang-tidy 15 on changed production translation units.
- Unit regression: 136 cases / 1768 assertions passed; after the final equivalent validation-expression cleanup, 41 focused cases / 593 assertions passed. Coverage includes selector boundaries, no-fill behavior including zero step, flat external mapping, obsolete-config rejection, serialization compatibility, FP32/RaBitQ build–reload–Add, forward/reverse scope and build-cache rejection.
- Instrumented selector coverage: 98.1% before the final equivalent boolean-expression cleanup; pruning strategy: 117/124 (94.4%). These are focused source coverage figures, not full-library coverage.
- This environment's aggregate test link repeats existing test archives before and inside whole-archive. The normal target was attempted; local validation uses the existing helper to remove duplicate archive arguments. No repository CMake workaround or symbol suppression was added.

## Performance and risk

Additional construction passes cost time; compare query QPS at matching recall, not only matching efSearch. The latest full ablation was on Pyramid RaBitQ 3-bit + SQ8 root search and the equivalent fill=false selector before this schema cleanup, not a fresh HGraph benchmark of this commit. Keep adaptive pruning disabled and rebuild to restore the original policy.

## Change type

- [x] Improvement/Refactor
- [x] Tests and English/Chinese docs updated

## Review follow-up (2b84fa45)

The tightened-pass reject list is explicitly named and used by the baseline fallback, while moving the first-pass buffer preserves allocation reuse. The reverse distance computer is constructed only after forward selection and the empty-result guard; it remains available to the reverse loop in both pruning modes.

Latest validation: 89 unit cases / 1200 assertions passed; clang-format 15 and clang-tidy 15 passed. Focused source coverage is 104/106 lines for the adaptive selector and 117/124 for pruning_strategy.cpp. The added regression verifies fallback recovery from tightened-pass rejects when the first-pass reject set is empty.


src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,178 @@
...
+            steps = 1;
+        }
+        result.second_alpha = alpha - static_cast<float>(steps) * parameter.adjust_step;
+        accepted.clear();
+        rejected.clear();
+        // The normalized original list is exactly sorted(A+B) followed by unscanned tail T.
+        scan(candidates, result.second_alpha, rejected);
+        if (accepted.size() < target_degree) {
+            Vector<PruningCandidate> remaining(allocator);
+            scan(rejected, alpha, remaining);
```
LHT129, line 168: [suggestion] The `rejected` vector is `clear()`ed and then immediately reused as the failure-output parameter for `scan()`, and later as the input for the fallback `scan()`. This dual role (first-pass rejects → second-pass failures → fallback input) is correct but subtle. A reader may not immediately realize that the second `scan(rejected, alpha, remaining)` operates on the *new* rejects from the tightened pass, not the original first-pass rejects.

Consider using a separate, descriptively-named container for the tightened-pass failures to make the data flow explicit:

```cpp
accepted.clear();
Vector<PruningCandidate> tightened_rejects(allocator);
scan(candidates, result.second_alpha, tightened_rejects);
if (accepted.size() < target_degree) {
    Vector<PruningCandidate> remaining(allocator);
    scan(tightened_rejects, alpha, remaining);
}
```

src/impl/pruning_strategy.cpp
```diff
@@ -133,10 +164,16 @@ mutually_connect_new_element(InnerIdType cur_c,
...
+                             const AdaptivePruningParameter* pruning) {
     PairwiseDistanceComputer pairwise_distance(distance_provider, allocator);
     const uint64_t max_size = graph->MaximumDegree();
-    select_edges_by_heuristic(top_candidates, max_size, distance_provider, allocator, alpha);
+    if (pruning != nullptr && pruning->enabled) {
+        select_adaptive_heap(
+            top_candidates, cur_c, max_size, distance_provider, allocator, alpha, *pruning);
+    } else {
+        select_edges_by_heuristic(top_candidates, max_size, distance_provider, allocator, alpha);
+    }
```
LHT129, line 176: [suggestion] The `PairwiseDistanceComputer` at this line is constructed unconditionally but is unused when `pruning != nullptr && pruning->enabled` because `select_adaptive_heap` creates its own instance internally. Consider moving the construction inside the `else` branch to avoid the unused allocation on the adaptive path.

```cpp
const uint64_t max_size = graph->MaximumDegree();
if (pruning != nullptr && pruning->enabled) {
    select_adaptive_heap(top_candidates, cur_c, max_size, distance_provider, allocator, alpha, *pruning);
} else {
    PairwiseDistanceComputer pairwise_distance(distance_provider, allocator);
    select_edges_by_heuristic(top_candidates, max_size, distance_provider, allocator, alpha);
}
```

src/impl/adaptive_pruning.cpp
LightWant: Addressed in 2b84fa45. The tightened-pass output is now named tightened_rejects and is the explicit input to the baseline fallback. It takes ownership of the first-pass vector allocation via std::move and clears its contents before the tightened scan, preserving buffer reuse without conflating the two sets. The regression asserts a case where the first pass rejects nothing but the tightened pass rejects a candidate that the fallback must recover. Validation: 89 unit cases / 1200 assertions; clang-format 15 and clang-tidy 15 passed.

src/impl/pruning_strategy.cpp
LightWant: Addressed in 2b84fa45 by deferring this computer until after forward selection, the empty-neighbor guard, and publication of the new node adjacency list, immediately before the reverse-neighbor loop. It is needed for center-to-candidate distances in full reverse lists even when forward pruning is adaptive, so moving it into the fixed-pruning else branch would make it unavailable where it is still needed. The adaptive selector retains its per-selection cache. This change avoids the outer computer on the empty-result path and avoids its lifetime overlapping the adaptive forward pass. Validation: forward/reverse and build–reload–Add regression passed within 89 cases / 1200 assertions.

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,41 @@ class PairwiseDistanceComputer {
 
 private:
     const DistanceProviderForGraph& distance_provider_;
-    // The helper lives for one pruning pass; its source IDs are bounded by the selected-edge limit.
+    // Per-call cache; adaptive passes may select different source IDs (at most twice the degree).
```
LHT129: [note] The `PairwiseDistanceComputer` is constructed after the forward selection guard (line ~190 in pruning_strategy.cpp), which means it is only constructed when `selected_neighbors` is non-empty. This is correct — the computer is only needed for reverse list pruning. However, the comment on line 55 of pruning_strategy.cpp says "Per-call cache; adaptive passes may select different source IDs (at most twice the degree)", but the computer is actually used across multiple reverse neighbor iterations within the same call. Consider clarifying the comment to reflect that the cache lifetime spans all reverse neighbors of a single `mutually_connect_new_element` call, not just one selection pass.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+                    float threshold,
+                    Vector<PruningCandidate>& failures) {
+        for (const auto& candidate : input) {
+            if (accepted.size() == target_degree) {
+                break;
+            }
+            bool good = true;
+            for (const auto& selected : accepted) {
+                float pair_distance = distance(selected.second, candidate.second);
+                ++result.distance_calls;
```
LHT129, line 118: [note] In `select_edges_adaptive`, the `scan` lambda captures `accepted` and `result` by reference. The lambda is called multiple times (first pass, relax scan, tighten scan, baseline fallback scan). Each call reads `accepted.size()` to check the `target_degree` guard. This is correct because `accepted` is cleared before the tighten rescan and the baseline fallback. However, the `failures` output parameter of `scan` is always a reference to a pre-allocated vector — the tighten path reuses `rejected` (moved to `tightened_rejects`) and the baseline fallback passes a fresh `remaining`. This is fine but subtle; a future maintainer might not realize that `scan` appends to `failures` rather than replacing it. The callers all pass freshly cleared vectors, so this is safe.

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,41 @@ class PairwiseDistanceComputer {
...
     UnorderedMap<InnerIdType, ComputerInterfacePtr> computers_;
 };
 
+void
+select_adaptive_heap(const DistHeapPtr& edges,
+                     InnerIdType center,
+                     uint64_t max_size,
```
LHT129, line 63: [suggestion] The `select_adaptive_heap` helper in pruning_strategy.cpp creates a local `PairwiseDistanceComputer` and passes it as a lambda to `select_edges_adaptive`. This means each forward selection constructs and destroys its own computer. In the reverse loop, `mutually_connect_new_element` also constructs a separate `PairwiseDistanceComputer` (line ~196). These two computers are independent — the forward one is destroyed before the reverse one is created. This is correct but means the forward computer's cache is discarded before the reverse loop. Since the forward and reverse selections operate on different centers (cur_c vs selected_neighbor), there is no cache reuse opportunity, so this is not a bug. Just noting for clarity.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+        result.second_alpha = alpha - static_cast<float>(steps) * parameter.adjust_step;
+        accepted.clear();
+        // Reuse the allocation while separating first-pass rejects from tightened-pass failures.
+        auto tightened_rejects = std::move(rejected);
+        tightened_rejects.clear();
+        // The normalized original list is exactly sorted(A+B) followed by unscanned tail T.
+        scan(candidates, result.second_alpha, tightened_rejects);
+        if (accepted.size() < target_degree) {
+            Vector<PruningCandidate> remaining(allocator);
```
LHT129, line 172: [suggestion] In `select_edges_adaptive`, the tighten branch at line ~170 does:\n```cpp\nauto tightened_rejects = std::move(rejected);\ntightened_rejects.clear();\nscan(candidates, result.second_alpha, tightened_rejects);\n```\nThe `std::move` + `clear()` pattern reuses the `rejected` vector's allocation for `tightened_rejects`. This is a good optimization. However, after the tighten scan, if `accepted.size() < target_degree`, the baseline fallback scan uses a new `Vector<PruningCandidate> remaining(allocator)` rather than reusing `rejected` (which is now in a moved-from state). Consider reusing `rejected` after the move for the fallback scan as well, to avoid an extra allocation. This is a minor optimization and not a correctness issue.

src/impl/pruning_strategy.cpp
```diff
@@ -150,10 +186,16 @@ mutually_connect_new_element(InnerIdType cur_c,
         top_candidates->Pop();
     }
 
+    if (selected_neighbors.empty()) {
```
LHT129, line 191: [suggestion] In `mutually_connect_new_element` (pruning_strategy.cpp), the early return at line ~189:\n```cpp\nif (selected_neighbors.empty()) {\n    graph->InsertNeighborsById(cur_c, selected_neighbors);\n    return cur_c;\n}\n```\nThis returns `cur_c` as the entry point when no neighbors are selected. This is a behavior change from the original code which would proceed to the reverse loop and potentially still update reverse neighbors. In the adaptive path, if forward selection produces zero neighbors (e.g., all candidates are self-edges that get removed), the function returns early without attempting reverse connections. The legacy path would also hit this case since `select_edges_by_heuristic` can also produce empty results. However, the `PairwiseDistanceComputer` construction after this guard is now correct — it avoids constructing the computer when there are no selected neighbors. Consider whether the early return is intentional for both paths or should be gated on the adaptive pruning path only.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <utility>
+
+#include "hash_types.h"
+
+namespace vsag {
+
+void
+AdaptivePruningParameter::FromJson(const JsonType& json) {
+    CHECK_ARGUMENT(json.IsObject(), "index parameters must be an object");
+    const bool has_removed_fill =
```
LHT129, line 29: [suggestion] The `AdaptivePruningParameter::FromJson` method resets `*this = AdaptivePruningParameter{}` before parsing. This means if `FromJson` is called on an already-configured parameter object, all fields (including `enabled`, `adjust_step`, etc.) are reset to defaults first. This is consistent with how other parameter classes in the codebase work (e.g., `HGraphParameter::FromJson` does not reset itself). However, `HGraphParameter::FromJson` calls `adaptive_pruning.FromJson(json)` without resetting the parent first. If a user calls `FromJson` twice on the same `HGraphParameter`, the adaptive pruning settings from the first call would be wiped by the `*this = AdaptivePruningParameter{}` in the second call. This is consistent behavior but worth noting — double-parsing is not an expected use case.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -296,7 +296,11 @@ HGraph::map_hgraph_param(const JsonType& hgraph_json) {
             inner_json[PARAMETER_USE_CONJUGATE_GRAPH].SetJson(value);
         } else if (key == HGRAPH_LABEL_REMAP_TYPE) {
             inner_json[LABEL_REMAP_TYPE_KEY].SetJson(value);
-        } else if (key == HGRAPH_USE_MCI || key == HGRAPH_MCI_MCS || key == HGRAPH_MCI_CLIQUE_MAX ||
+        } else if (key == HGRAPH_ADAPTIVE_PRUNING || key == "adaptive_pruning_adjust_step" ||
```
LHT129, line 299: [note] The `hgraph_param_mapping.cpp` change adds several adaptive pruning keys to the passthrough list in `map_hgraph_param`. Notably, it also adds `"fill_rejected"` and `"adaptive_pruning_fill_rejected"` to this list. These are obsolete keys that should trigger errors (as enforced in `AdaptivePruningParameter::FromJson`). By adding them to the passthrough list, they will be forwarded to the inner parameter parser which will then reject them. This is correct — the error message will be generated by `FromJson`. However, if someone removes the error check from `FromJson` in the future, these keys would silently pass through. Consider whether the error should be raised here in the mapping layer instead, or at least add a comment noting that these keys are intentionally passed through to be rejected downstream.

src/algorithm/hgraph/hgraph.cpp
```diff
@@ -76,6 +77,13 @@ HGraph::HGraph(const HGraphParameterPtr& hgraph_param, const vsag::IndexCommonPa
       mci_parameters_(hgraph_param->mci_parameters),
       use_old_serial_format_(common_param.use_old_serial_format_),
       use_conjugate_graph_(hgraph_param->use_conjugate_graph) {
+    adaptive_pruning_.Validate(alpha_);
```
LHT129, line 80: [note] In the HGraph constructor (hgraph.cpp), `adaptive_pruning_.Validate(alpha_)` is called before `alpha_` is validated elsewhere. The `Validate` method checks `std::isfinite(alpha)`, `alpha - 2*adjust_step > 0`, etc. This is the first validation that runs in the constructor. If a user passes a NaN or negative alpha, the error message will come from `Validate` rather than a dedicated alpha check. This is acceptable since the error message is clear ("adaptive_pruning requires finite alpha/step, step >= 0 and alpha - 2*step > 0"), but it may be surprising that enabling adaptive_pruning changes which code path catches an invalid alpha. Consider adding a separate alpha validation before the adaptive pruning check for consistency.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -163,6 +163,8 @@ HGraph::Build(const DatasetPtr& data) {
     std::vector<int64_t> ret;
     const bool using_build_cache = this->has_loaded_cache();
     if (using_build_cache) {
+        CHECK_ARGUMENT(not adaptive_pruning_.enabled,
```
LHT129, line 166: [note] The `hgraph_build.cpp` change adds:\n```cpp\nCHECK_ARGUMENT(not adaptive_pruning_.enabled,\n               "adaptive_pruning does not yet support building from imported cache");\n```\nThis is placed inside the `if (using_build_cache)` block. The check is correct — the adaptive pruning algorithm needs to run during Build to select edges, and imported caches bypass this. However, the check is only reached when `has_loaded_cache()` returns true. If a user enables adaptive pruning and imports a cache but the cache import fails silently, the check would not fire. This is unlikely in practice since `ImportCache` is a separate API call. The error message is clear and actionable.

src/impl/adaptive_pruning_test.cpp
```diff
@@ -0,0 +1,312 @@
+
```
LHT129, line 1: [note] The test coverage is comprehensive. The test matrix covers:\n- Unit tests for the core algorithm: relaxation/tightening boundaries, tail preservation, normalization, empty/singleton inputs, invalid inputs, parameter roundtrip, L2 geometry with permutation invariance, coincident vectors\n- Integration tests: forward pruning with short lists, disabled policy preserving legacy edges, reverse pruning with relaxation, reverse pruning with append vs full lists\n- HGraph-level tests: build-reload-add lifecycle (FP32), unsupported contexts (ODescent, IP, apply_to_upper, invalid thresholds), cache rejection, RaBitQ split/fused lifecycle\n- Parameter tests: legacy parameter acceptance, flat parameter mapping, obsolete option rejection, compatibility checks\n\nOne area that could be strengthened: there are no tests for the interaction between adaptive pruning and `use_elp_optimizer`. The ELP optimizer also modifies edge selection and the two features could theoretically interact. If they are mutually exclusive or compatible, a test documenting this would be helpful.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+            }
+            bool good = true;
+            for (const auto& selected : accepted) {
+                float pair_distance = distance(selected.second, candidate.second);
+                ++result.distance_calls;
+                const bool valid_distance = std::isfinite(pair_distance) && pair_distance >= 0;
+                CHECK_ARGUMENT(
+                    valid_distance,
+                    "adaptive_pruning requires finite non-negative pairwise L2 distances");
+                if (threshold * pair_distance < candidate.first) {
```
LHT129, line 124: [suggestion] In `select_edges_adaptive`, the `scan` lambda validates `std::isfinite(pair_distance) && pair_distance >= 0` on every pairwise distance comparison. This check runs in the hot inner loop — for a target_degree of 64 with 200 candidates, this could be thousands of `isfinite` calls per node. Since the distances come from `FlattenInterface::PairwiseDistance` (which computes squared L2 distances that are guaranteed non-negative and finite for finite inputs), these checks are purely defensive. Consider using `assert` (debug-only) or a `#ifndef NDEBUG` guard to eliminate this overhead in release builds while keeping the safety net during development and testing.

src/impl/pruning_strategy.cpp
LHT129, line 63: [suggestion] The `select_adaptive_heap` helper in `pruning_strategy.cpp` converts the `DistHeap` to a `Vector<PruningCandidate>`, calls `select_edges_adaptive` (which sorts the result), then pushes results back into the heap. This heap→vector→heap round-trip adds allocation and copy overhead on every edge selection. The heap already maintains sorted order; consider whether `select_edges_adaptive` could accept a sorted range (iterator pair) instead of requiring a vector, or whether the result could be directly consumed without re-heapifying. This is a minor optimization but the heap conversion happens on every node during Build/Add.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+        int steps = 2;
+        if (ratio >= 5) {
+            steps = 0;
+        } else if (ratio >= 2.5) {
+            steps = 1;
+        }
+        result.second_alpha = alpha - static_cast<float>(steps) * parameter.adjust_step;
+        accepted.clear();
+        // Reuse the allocation while separating first-pass rejects from tightened-pass failures.
+        auto tightened_rejects = std::move(rejected);
```
LHT129, line 167: [note] In the `tighten` branch of `select_edges_adaptive`, after `accepted.clear()`, the full `candidates` list is rescanned at the tightened alpha. The comment correctly notes "The normalized original list is exactly sorted(A+B) followed by unscanned tail T." This is correct because the first pass stopped early when `accepted.size() == target_degree`, leaving tail T unexamined. However, the tightened rescan re-examines all of A (already-accepted candidates from first pass) alongside B (first-pass rejects) and T (tail). Since A candidates were already accepted at the original alpha, they will also pass the tightened (lower) alpha check — the geometric condition `threshold * d(s,c) < d(u,c)` is strictly easier to satisfy with a smaller threshold. This means the tightened rescan effectively wastes distance computations on A candidates that are guaranteed to be re-accepted. Consider skipping the already-accepted candidates (A) in the tightened rescan and only scanning B+T, then prepending A to the result. This would save `|A| * target_degree` distance computations per tighten.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+                   "adaptive_pruning requires finite alpha/step, step >= 0 and alpha - 2*step > 0");
+    CHECK_ARGUMENT(not apply_to_upper,
+                   "adaptive_pruning currently supports only bottom-layer selection");
+}
+
+Vector<PruningCandidate>
+select_edges_adaptive(Vector<PruningCandidate>& candidates,
+                      InnerIdType center,
+                      uint64_t target_degree,
+                      float alpha,
```
LHT129, line 77: [note] The `CHECK_ARGUMENT(not apply_to_upper, ...)` check in `AdaptivePruningParameter::Validate` is called both at HGraph construction time (hgraph.cpp line 80) and inside `select_edges_adaptive` (adaptive_pruning.cpp line 82). The construction-time check is sufficient to prevent misconfiguration — if `apply_to_upper` is true, the index constructor throws before any edges are selected. The runtime check in `select_edges_adaptive` is therefore redundant. This is harmless (defense in depth) but the duplicate validation could be confusing to future readers who might wonder whether there is a code path that bypasses the constructor check. Consider removing the runtime check or adding a comment explaining it is a safety net for direct API callers.

src/impl/pruning_strategy.cpp
LHT129, line 191: [suggestion] In `mutually_connect_new_element` (pruning_strategy.cpp), when the forward selection produces zero neighbors, the function returns `cur_c` as the entry point without attempting any reverse connections. The original code (before this PR) did not have this early return — it would proceed to the reverse loop with an empty `selected_neighbors` vector, which would simply skip the loop body. The behavioral difference is that the original code would still construct a `PairwiseDistanceComputer` (now moved after the guard) and return `cur_c`. With this change, the `PairwiseDistanceComputer` construction is avoided when there are no selected neighbors, which is a minor optimization. However, the early return also skips the `next_closest_entry_point` computation. This is correct since `selected_neighbors.back()` would be UB on an empty vector. Consider adding a brief comment explaining why the early return is safe: "no forward neighbors means no reverse edges to update."

LHT129 (commented):
## Review Summary

This PR adds an opt-in adaptive NSW neighbor pruning mechanism to the HGraph index. The core idea is a two-pass alpha adjustment schedule that relaxes the geometric rejection threshold when too few candidates are accepted, and tightens it when the neighbor list fills up with many rejects — allowing the tail of the candidate list to compete for slots.

### What I reviewed

- **Core algorithm** (`adaptive_pruning.cpp/h`): The `select_edges_adaptive` function with normalization, two-pass schedule, and baseline fallback. The logic is sound and well-tested.
- **Integration** (`pruning_strategy.cpp/h`): The `select_adaptive_heap` bridge and `mutually_connect_new_element` modifications. The forward/reverse selection split is clean.
- **HGraph wiring** (`hgraph.cpp/h`, `hgraph_parameter.cpp/h`, `hgraph_build.cpp`, `hgraph_param_mapping.cpp`): Constructor validation, parameter plumbing, cache rejection, and compatibility checks are all correct.
- **Tests**: Comprehensive unit tests (89 cases / 1200+ assertions) covering relaxation/tightening boundaries, edge cases, L2 geometry, parameter roundtrip, and integration tests for build-reload-add lifecycle with both FP32 and RaBitQ.

### New findings (not covered by prior reviews)

1. **[suggestion] Hot-path distance validation overhead** — The `scan` lambda validates `std::isfinite` on every pairwise distance. These checks are defensive; distances from `FlattenInterface::PairwiseDistance` are guaranteed non-negative for L2. Consider `assert` or `#ifndef NDEBUG` for release builds.

2. **[suggestion] Heap→vector→heap round-trip** — `select_adaptive_heap` converts the DistHeap to a vector and back, adding allocation overhead on every node. Consider accepting a sorted range directly.

3. **[note] Tighten rescan wastes work on already-accepted candidates** — The tighten branch rescans the full candidate list including candidates already accepted in the first pass (which are guaranteed to pass the tighter check). Skipping them would save `|A| * target_degree` distance computations.

4. **[note] Duplicate `apply_to_upper` validation** — Checked both at construction time and inside `select_edges_adaptive`. Harmless but redundant.

5. **[suggestion] Early return on empty selected_neighbors** — The new guard avoids UB on `selected_neighbors.back()` but could benefit from a brief comment explaining the safety rationale.

### Overall assessment

The implementation is well-structured, the algorithm is clearly documented, and the test coverage is thorough. All of my comments are suggestions/notes — there are no correctness issues. The PR is in good shape.

src/impl/pruning_strategy.cpp
LightWant: Addressed in 9f76bb9c. The member comment now distinguishes a selection-local cache from the outer cache spanning the entire reverse-neighbor loop; the latter stores reverse centers. The helper also documents why retaining separate per-selection caches bounds scratch usage.

src/impl/pruning_strategy.cpp
LightWant: Confirmed, and clarified the cache lifetimes in 9f76bb9c. The forward selector owns a local cache; the outer reverse loop caches center-to-candidate computers, and each adaptive reverse selection has its own selected-source cache. Some source IDs could overlap, so reuse is theoretically possible, but retaining all selection caches across reverse lists could grow retained scratch to O(degree^2). The current lifetime keeps each selection bounded.

src/impl/adaptive_pruning.cpp
LightWant: Keeping the descriptively named fallback output. Moving rejected transfers its buffer to tightened_rejects, so reusing the moved-from rejected vector does not recover a spare allocation. Constructing an empty remaining vector does not allocate element storage; its subsequent growth would also be needed for the moved-from vector. This substitution therefore does not avoid the claimed allocation and makes the distinct pass outputs less clear. The existing fallback regression verifies recovery of tightened-pass failures.

src/impl/pruning_strategy.cpp
LightWant: The guard intentionally applies to both paths. With no selected neighbors, there are no reverse neighbors to iterate or update. Publishing the empty adjacency and returning cur_c also avoids accessing selected_neighbors.back() on an empty vector. Gating this on adaptive pruning would leave that invalid access in the legacy empty-input case. Added the explicit explanation in 9f76bb9c.

src/impl/adaptive_pruning.cpp
LightWant: Confirmed: FromJson parses a complete adaptive configuration rather than applying a patch. Omitted keys reset to defaults so stale enabled/step/reverse values cannot leak into the next parse. Added this contract in 9f76bb9c. The existing parameter-roundtrip test explicitly reparses an empty object after an enabled configuration and verifies that pruning is disabled.

src/impl/adaptive_pruning.cpp
LightWant: Keeping the release-mode validation. Finite floating-point inputs do not guarantee finite squared L2: subtraction/squaring/accumulation can overflow, and the callback/provider is not restricted to one exact FP32 implementation. Invalid pair distances would silently change pruning decisions if this became debug-only. Added the overflow rationale in 9f76bb9c. Existing tests require rejection of negative, infinite and NaN pair distances; a performance change here needs profiling and a stronger validated-distance contract first.

src/impl/pruning_strategy.cpp
LightWant: Keeping the current adapter in this PR. A heap guarantees its top element, not a globally sorted range; popping this max-heap yields descending distances, and normalization also removes self edges/duplicate IDs with deterministic distance/ID ordering. Reversing alone is not sufficient to preserve all those contracts for arbitrary DistHeap implementations. Both fixed and adaptive selectors currently return a heap to the shared connection path. Removing that round-trip is a worthwhile separately profiled interface refactor, but changing the selector/caller contract is beyond this correctness-focused review follow-up.

src/impl/pruning_strategy.cpp
LightWant: Addressed in 9f76bb9c: the guard now states that no forward neighbors means no reverse edges to update, and that back() would be invalid. This applies to both fixed and adaptive paths.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,183 @@
...
+                      const AdaptivePruningParameter& parameter,
+                      const std::function<float(InnerIdType, InnerIdType)>& distance,
+                      Allocator* allocator,
+                      AdaptivePruningStats* stats) {
+    CHECK_ARGUMENT(parameter.enabled, "select_edges_adaptive requires enabled adaptive_pruning");
+    parameter.Validate(alpha);
+    AdaptivePruningStats result;
+    result.second_alpha = alpha;
```
LHT129, line 85: [suggestion] `parameter.Validate(alpha)` is already called in `HGraph`'s constructor (`hgraph.cpp` line 657) before any selection happens. Calling it again here on the hot path (every `select_edges_adaptive` invocation) is redundant. The parameters cannot change after construction.

Consider removing this call and relying solely on the constructor-time validation, or guard it with a debug-only check:

```cpp
CHECK_ARGUMENT(parameter.enabled, "select_edges_adaptive requires enabled adaptive_pruning");
// Validation already performed at HGraph construction time.
```

src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -321,6 +323,13 @@ HGraphParameter::ToJson() const {
 bool
```
LHT129, line 323: [note] When `adaptive_pruning.enabled` is false on both sides, `alpha` is not checked for compatibility (line 900-905). The legacy `select_edges_by_heuristic` still uses `alpha`, so loading an index built with a different alpha could produce inconsistent graph structures even without adaptive pruning enabled.

This is pre-existing behavior (the original `CheckCompatibility` did not check `alpha` either), so this is just a note — not introduced by this PR. If you want to tighten compatibility checks, consider always comparing `alpha` regardless of the adaptive pruning state.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,183 @@
...
+select_edges_adaptive(Vector<PruningCandidate>& candidates,
+                      InnerIdType center,
+                      uint64_t target_degree,
+                      float alpha,
+                      const AdaptivePruningParameter& parameter,
+                      const std::function<float(InnerIdType, InnerIdType)>& distance,
+                      Allocator* allocator,
+                      AdaptivePruningStats* stats) {
+    CHECK_ARGUMENT(parameter.enabled, "select_edges_adaptive requires enabled adaptive_pruning");
+    parameter.Validate(alpha);
```
LHT129, line 83: [suggestion] `parameter.Validate(alpha)` at this line is redundant with the constructor-time validation in `HGraph::HGraph()` (hgraph.cpp line 80). Since `select_edges_adaptive` is called on the hot path for every node during Build/Add, this performs `std::isfinite` checks and `CHECK_ARGUMENT` evaluations on every invocation. The parameters cannot change after construction.

Consider removing this call and relying solely on the constructor-time validation, or guard it with `#ifndef NDEBUG`:

```cpp
CHECK_ARGUMENT(parameter.enabled, "select_edges_adaptive requires enabled adaptive_pruning");
#ifndef NDEBUG
parameter.Validate(alpha);  // already validated at HGraph construction time
#endif
```

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,183 @@
...
+        } else if (ratio >= 2.5) {
+            steps = 1;
+        }
+        result.second_alpha = alpha - static_cast<float>(steps) * parameter.adjust_step;
+        accepted.clear();
+        // Reuse the allocation while separating first-pass rejects from tightened-pass failures.
+        auto tightened_rejects = std::move(rejected);
+        tightened_rejects.clear();
+        // The normalized original list is exactly sorted(A+B) followed by unscanned tail T.
+        scan(candidates, result.second_alpha, tightened_rejects);
```
LHT129, line 170: [suggestion] In the tighten branch, `scan(candidates, result.second_alpha, tightened_rejects)` rescans the full `candidates` list including already-accepted candidates from the first pass (set A). Since A candidates were accepted at the original (higher) alpha, they will always pass the tightened (lower) alpha check — the geometric condition `threshold * d(s,c) < d(u,c)` is strictly easier to satisfy with a smaller threshold.

This means the tightened rescan wastes approximately `|A| * target_degree` distance computations per tighten event. Consider skipping the already-accepted candidates and only scanning B (first-pass rejects) + T (unscanned tail), then prepending A to the result:

```cpp
// A candidates are guaranteed to pass the tightened alpha; skip them.
Vector<PruningCandidate> b_and_t(allocator);
b_and_t.reserve(rejected.size() + (candidates.size() - accepted.size() - rejected.size()));
// Copy B (rejects) and T (tail) — the tail starts after accepted.size() + rejected.size().
// Then scan only b_and_t at the tightened alpha, and prepend accepted to the result.
```

This is a minor optimization but the tighten path fires on most nodes during construction when `target_degree` is reached in the first pass.

src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -292,6 +293,7 @@ HGraphParameter::ToJson() const {
     json[EF_CONSTRUCTION_KEY].SetUint64(this->ef_construction);
     json[RESIZE_INCREASE_COUNT_BIT].SetUint64(this->resize_increase_count_bit);
```
LHT129, line 294: [note] When `adaptive_pruning.enabled` is false, `alpha` is not checked for compatibility. The legacy `select_edges_by_heuristic` still uses `alpha`, so loading an index built with a different alpha could produce inconsistent graph structures even without adaptive pruning enabled.

This is pre-existing behavior (the original `CheckCompatibility` did not check `alpha`), so this is not introduced by this PR. However, now that this PR adds an explicit `alpha` compatibility check for the `enabled=true` case, the asymmetry is worth noting — a user who changes `alpha` with `adaptive_pruning=false` will not get a compatibility error on deserialization, which could lead to silently incorrect search results.

src/impl/adaptive_pruning.cpp
LightWant: Keeping validation at the selector boundary. This is a standalone callable helper accepting an AdaptivePruningParameter and alpha; it is also exercised directly without constructing HGraph. Constructor validation protects that caller, but does not establish an invariant for every caller of this function. Removing this check or making it debug-only changes its release-mode invalid-input contract. These are constant-time checks once per selection (not once per pairwise comparison). A validated-parameter type/private unchecked helper could support a future profiled optimization, but simply deleting the checks is not equivalent.

src/algorithm/hgraph/hgraph_parameter.cpp
LightWant: Confirmed that this is pre-existing compatibility behavior, and retaining it deliberately when adaptive pruning is disabled. Alpha controls graph construction/pruning, not the distance encoding or interpretation of a serialized adjacency list. Loading an existing graph with a different construction alpha does not by itself imply incorrect search results; subsequent Add operations may construct different edges, as was already possible. This PR requires matching alpha for the new adaptive policy without introducing a compatibility break for existing fixed-alpha indexes. A stricter legacy construction-policy compatibility contract would be a separate change.

src/impl/adaptive_pruning.cpp
LightWant: Keeping validation at the selector boundary. This is a standalone callable helper accepting an AdaptivePruningParameter and alpha; it is also exercised directly without constructing HGraph. Constructor validation protects that caller, but does not establish an invariant for every caller of this function. Removing this check or making it debug-only changes its release-mode invalid-input contract. These are constant-time checks once per selection (not once per pairwise comparison). A validated-parameter type/private unchecked helper could support a future profiled optimization, but simply deleting the checks is not equivalent.

src/impl/adaptive_pruning.cpp
LightWant: The proposed optimization reverses the meaning of the predicate: threshold * d(s,c) < d(u,c) is the REJECTION condition. Lowering threshold makes rejection easier, not acceptance. For example, with pair distance 2 and center distance 2, alpha=1.06 accepts (2.12 < 2 is false), whereas alpha=0.94 rejects (1.88 < 2 is true). Therefore an element of A can be removed by tightening, and retaining/prepending A would defeat tightening and change the result. The existing real-L2 regression explicitly requires first-pass {1,2} to become {1,3} after tightening, and the fallback test covers a newly rejected member of A. Keeping the full rescan is necessary for correctness.

src/algorithm/hgraph/hgraph_parameter.cpp
LightWant: Confirmed that this is pre-existing compatibility behavior, and retaining it deliberately when adaptive pruning is disabled. Alpha controls graph construction/pruning, not the distance encoding or interpretation of a serialized adjacency list. Loading an existing graph with a different construction alpha does not by itself imply incorrect search results; subsequent Add operations may construct different edges, as was already possible. This PR requires matching alpha for the new adaptive policy without introducing a compatibility break for existing fixed-alpha indexes. A stricter legacy construction-policy compatibility contract would be a separate change.

PR #2934, opened 2026-09-10
perf(pyramid): support adaptive NSW pruning with flat parameters

LightWant:
## What changed

Extend the shared adaptive neighbor selector from #2933 to Pyramid NSW bottom graphs. One global policy applies with each hierarchy's effective alpha; routing layers retain fixed-alpha selection. Flat-node promotion, root graphs and path-node insertion use the same existing distance provider and locking scheme.

Configuration is flat under `index_param`:

```json
{
  "graph_type": "nsw",
  "alpha": 1.06,
  "adaptive_pruning": true,
  "adaptive_pruning_adjust_step": 0.06,
  "adaptive_pruning_apply_to_reverse": false,
  "adaptive_pruning_apply_to_upper": false
}
```

Remove unconditional filling and its option from the shared selector. The complete GIST 1M ablation found no consistent filling benefit (−3.4% to +4.7% single-factor QPS change). Reverse adaptation remains optional: it hurt the measured recall targets at alpha=1.06 and helped at alpha=1.2.

## GIST 1M ablation evidence

Root-only RaBitQ 3-bit + SQ8 reorder, L2 960D, 1M vectors, 1000 queries, k=10, M=32, efConstruction=400, step=.06. Construction uses 16 threads pinned to CPUs 0–15; queries use CPU 0, five timed runs per matching-recall point.

| alpha | selector | QPS @ .85 / .90 / .92 recall | Build seconds |
|---|---|---|---:|
| 1.06 | fixed baseline | 1002.6 / 634.8 / 462.3 | 348.98 |
| 1.06 | forward adaptive, no fill | 1090.8 / 703.6 / 542.4 | 441.01 |
| 1.2 | fixed baseline | 672.8 / 353.2 / 211.4 | 416.29 |
| 1.2 | forward + reverse, no fill | 993.0 / 634.4 / 464.5 | 490.72 |

All four fill/reverse combinations at both alphas plus two baselines were tested (10 configurations; 6 new and 4 reused indexes; matching-recall queries remeasured this batch). 110 formal HDF5 results validated. Index memory stayed near 1600 MiB. Each configuration has one parallel-built index; small differences are not proof of stable gains. These measurements preceded the flat-schema cleanup and use the equivalent fill=false implementation; they are not a fresh benchmark of the final commit. Root-only results do not establish path-filtered performance.

## Compatibility and scope

Requires NSW and L2. Upper adaptation and nonempty imported-cache builds remain rejected. Full reverse lists adapt only when opted in; non-full lists append normally. Serialized enabled indexes require matching alpha and policy fields. The former experimental nested schema and removed fill option are rejected and require rebuilding those experimental indexes; absent adaptive fields still mean disabled mode.

## Validation

- clang-format 15, whitespace checks and clang-tidy 15 passed for the review changes.
- Joint HGraph/Pyramid unit regression: 140 cases / 2372 assertions passed.
- Pyramid functional regression: 72 cases / 320427 assertions passed.
- Instrumented shared selector: 103/105 executable lines (98.1%); pruning strategy: 117/124 (94.4%). These are focused source measurements, not whole-library coverage.
- The standard unit target was attempted. This environment still repeats test archives in its aggregate link; tests use the existing local archive de-duplication helper. Repository build rules and symbol resolution flags were not modified.
- English/Chinese algorithm documentation updated.

## Review follow-up

- Skip the redundant TIGHTEN rescan when the first-pass rejection ratio is at least 5 and the selected tightening step count is zero. Preserve the selected neighbors, branch metadata and first-pass counters. This can occur with a positive configured adjustment step and is distinct from the existing zero-delta path.
- Extend the boundary regression to assert selected IDs and exact distance-call counts at and above the threshold, while verifying nonzero-step rescans below it.
- Retain per-selection distance-computer caches and document why: sharing a cache across up to K reverse selections could retain O(K^2) distinct source computers until insertion completes; local caches keep live scratch O(K). The separate outer cache only retains reverse centers.
- Preserve the existing empty-neighbor guard and fixed-alpha cached-graph connection path; both match the documented scope.

This PR includes the shared implementation from #2933 and should follow that PR; review the Pyramid-specific delta against its branch where useful.

## Change type

- [x] Improvement/Refactor
- [x] Tests and English/Chinese docs updated


LHT129 (commented):
## Code Review Summary

This PR introduces adaptive NSW neighbor pruning for Pyramid (and HGraph) bottom graphs. The core idea is a multi-pass algorithm that adjusts the RNG pruning threshold (alpha) based on the accept/reject ratio from the first pass. Overall the implementation is solid with good test coverage across boundary conditions, serialization compatibility, and integration tests.

### Strengths
- **Well-tested**: The test suite covers ratio boundaries, empty/zero-step, self/duplicate normalization, invalid inputs, real L2 tightening across permutations, forward/reverse scope, cache rejection, parameter compatibility, and FP32/RaBitQ split/fused build-serialize-reload-Add lifecycle.
- **Clean API design**: The `AdaptivePruningParameter*` parameter is optional (defaults to `nullptr`), making it a non-breaking change for existing callers.
- **Defensive validation**: Both `Validate()` and `select_edges_adaptive()` check for finite/non-negative distances, valid alpha/step ranges, and reject unsupported configurations (non-NSW graphs, non-L2 metrics, apply_to_upper).
- **Serialization compatibility**: Deserialization rejects mismatched adaptive pruning configurations, preventing silent correctness issues.

### Suggestions (non-blocking)
1. `src/impl/adaptive_pruning.cpp:140` - The comment about "sorted(A+B) followed by unscanned tail T" is slightly misleading since the first pass scans all candidates; there is no separate tail. Consider rewording.
2. `src/impl/adaptive_pruning.cpp:126` - The hardcoded ratio thresholds (1.5/3 for relax, 2.5/5 for tighten) are not configurable. Consider documenting the rationale or making them tunable.
3. `src/impl/adaptive_pruning.cpp:148` - The tighten-branch fallback to original alpha may rarely produce additional accepted edges. Consider adding a comment explaining the expected benefit.
4. `src/impl/pruning_strategy.h:112` - The docstring should mention that `apply_to_reverse` also adaptively prunes existing neighbor lists, not just the new node's forward edges.
5. `src/algorithm/pyramid/pyramid.cpp:274` - Consider adding a comment explaining why `connect_cached_graph_point` intentionally skips adaptive pruning.
6. `src/algorithm/hgraph/hgraph_build.cpp:166` - The "does not yet support" phrasing implies a temporary limitation; clarify if this is permanent.

### Test Coverage
The test suite is comprehensive. Key scenarios:
- Relaxation boundaries (1-5 accepted edges, correct second_alpha)
- Tightening boundaries (0-10 rejected edges, correct second_alpha)
- Strict inequality and empty inputs
- Invalid numeric inputs (NaN, inf, negative)
- Parameter roundtrip (JSON serialize/deserialize)
- Real L2 tightening across all input permutations
- Singleton degree and coincident vectors
- Disabled policy preserves legacy behavior
- Forward pruning handles short lists and self edges
- Reverse pruning relaxes around existing neighbor
- Reverse pruning preserves append and prunes full lists
- Build+serialize+reload+Add lifecycle (FP32 and RaBitQ)
- Cache rejection for both HGraph and Pyramid
- Parameter compatibility with legacy JSON

### Verdict
The PR is well-engineered with thorough test coverage. The suggestions above are minor improvements for documentation clarity and configurability - none are blocking.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,178 @@
...
+            steps = 2;
+        }
+        result.second_alpha = alpha + static_cast<float>(steps) * parameter.adjust_step;
+        Vector<PruningCandidate> remaining(allocator);
+        scan(rejected, result.second_alpha, remaining);
+    } else if (accepted.size() == target_degree) {
+        result.branch = AdaptivePruningBranch::TIGHTEN;
+        const double ratio =
+            static_cast<double>(rejected.size()) / static_cast<double>(target_degree);
+        int steps = 2;
```
LHT129, line 155: [note] When `ratio >= 5` in the TIGHTEN branch, `steps = 0` and `result.second_alpha = alpha`. The tightened scan re-runs at the same alpha as the first pass, producing identical results — the same candidates are accepted and rejected. This is correct but redundant; consider skipping the rescan when `steps == 0` and keeping the first-pass result directly.

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,41 @@ class PairwiseDistanceComputer {
...
+select_adaptive_heap(const DistHeapPtr& edges,
+                     InnerIdType center,
+                     uint64_t max_size,
+                     const DistanceProviderForGraph& distance_provider,
+                     Allocator* allocator,
+                     float alpha,
+                     const AdaptivePruningParameter& parameter) {
+    // Keep distance-computer scratch bounded to one vertex selection.
+    PairwiseDistanceComputer pairwise_distance(distance_provider, allocator);
+    Vector<PruningCandidate> candidates(allocator);
```
LHT129, line 69: [suggestion] `select_adaptive_heap` creates its own `PairwiseDistanceComputer` internally (line 69), while the caller `mutually_connect_new_element` also creates one at line 99. In the adaptive forward path, the outer `PairwiseDistanceComputer` is only used later in the reverse neighbor loop (lines 153, 158). Consider passing the existing `PairwiseDistanceComputer` into `select_adaptive_heap` to avoid creating two instances, or document why a fresh cache is preferred for the adaptive selection pass.

src/impl/pruning_strategy.cpp
```diff
@@ -150,6 +187,10 @@ mutually_connect_new_element(InnerIdType cur_c,
         top_candidates->Pop();
     }
 
+    if (selected_neighbors.empty()) {
```
LHT129, line 190: [note] The early return when `selected_neighbors.empty()` correctly avoids the reverse-neighbor loop and prevents undefined behavior from `selected_neighbors.back()` on an empty vector. Good defensive check.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2115,8 +2123,14 @@ Pyramid::add_bottom_graph_point(const Hierarchy& hierarchy,
...
-            mutually_connect_new_element(
-                inner_id, results, node.graph_, codes, points_mutex_, allocator_, hierarchy.alpha);
+            mutually_connect_new_element(inner_id,
+                                         results,
+                                         node.graph_,
+                                         codes,
+                                         points_mutex_,
+                                         allocator_,
+                                         hierarchy.alpha,
+                                         &adaptive_pruning_);
```
LHT129, line 2133: [note] The `connect_cached_graph_point` path (line 2125) does not use adaptive pruning — only the `mutually_connect_new_element` path does. This is consistent with the documented scope (no imported-cache support), but worth noting that cached-graph connections use the original fixed-alpha selector.

src/impl/pruning_strategy.cpp
LHT129, line 69: [suggestion] `select_adaptive_heap` creates its own `PairwiseDistanceComputer` (line 69) while `mutually_connect_new_element` creates another one (line 174). When called from `mutually_connect_new_element`, this means two `PairwiseDistanceComputer` instances exist simultaneously, each with its own `UnorderedMap` cache. Consider passing the outer computer into `select_adaptive_heap` to reuse the cache, or documenting why the duplication is intentional (e.g., lifetime isolation for the reverse-neighbor path).

LHT129 (commented):
## Review Summary

This PR enables adaptive NSW pruning for Pyramid index, building on the shared infrastructure introduced in a prior PR. The implementation is clean and well-scoped: it adds `select_edges_adaptive()` in `adaptive_pruning.cpp`, integrates it via `select_adaptive_heap()` in `pruning_strategy.cpp`, and wires it into both HGraph and Pyramid bottom-layer construction paths.

### What's Good

- **Well-defined scope**: Adaptive pruning applies only to NSW/L2 bottom-layer forward selection (and optionally reverse), with clear guardrails rejecting unsupported combinations (ODescent, non-L2, imported cache, upper layers).
- **Thorough test coverage**: `adaptive_pruning_test.cpp` covers ratio boundaries, strict inequality, normalization, zero step, invalid inputs, and real L2 tightening examples. `pruning_strategy_test.cpp` covers forward/reverse scope and disabled policy.
- **Clean serialization**: Flat parameter fields with backward-compatibility rejection of the removed experimental nested schema.
- **Documentation**: Comprehensive `adaptive_pruning.md` with algorithm details, configuration, and scope limitations.

### Items Noted

1. **[suggestion] `PairwiseDistanceComputer` duplication**: `select_adaptive_heap` creates its own `PairwiseDistanceComputer` while `mutually_connect_new_element` creates another. When called from `mutually_connect_new_element`, two instances coexist with separate caches. Consider passing the outer computer in or documenting the rationale.

2. **[note] TIGHTEN `steps==0` no-op**: When `delta==0` in `select_edges_adaptive`, the TIGHTEN branch computes `alpha_tightened == alpha0` and re-scans with the same alpha, producing identical results. This is correct but redundant — the `delta==0` early return already handles this case.

3. **[note] Cached-graph path**: `connect_cached_graph_point` in `pyramid.cpp` does not use adaptive pruning, consistent with the documented scope. Worth noting for future cache-build support.

4. **[note] Empty neighbors guard**: The early return when `selected_neighbors.empty()` correctly avoids undefined behavior from `selected_neighbors.back()` on an empty vector.

### Verdict

The PR is well-structured with good test coverage and clear documentation. The noted items are minor and non-blocking. No correctness, security, or performance concerns identified.

LHT129 (commented):
## Code Review Summary

This PR adds adaptive NSW pruning with flat parameters to both Pyramid and HGraph indexes. Overall the implementation is well-structured with thorough validation, comprehensive test coverage, and clear documentation.

### What this PR does well

- **Clean parameter design**: Flat JSON keys (`adaptive_pruning`, `adaptive_pruning_adjust_step`, etc.) avoid nested-object parsing issues and integrate naturally with the existing parameter mapping infrastructure.
- **Defense in depth**: Validation happens at multiple layers — `AdaptivePruningParameter::Validate`, index constructors (both Pyramid and HGraph), `CheckCompatibility` for reload safety, and explicit rejection of imported caches when adaptive pruning is enabled.
- **Comprehensive test coverage**: 22+ new test cases covering unit-level algorithm behavior (relaxation/tightening boundaries, edge cases, invalid inputs), integration-level build-reload-add lifecycles, parameter roundtrips, compatibility checks, and unsupported context rejection.
- **All prior review comments addressed**: The `steps == 0` skip is implemented, `PairwiseDistanceComputer` duplication is documented with explanatory comments, `connect_cached_graph_point` correctly excludes adaptive pruning, and the `selected_neighbors.empty()` early return is present.
- **Backward compatibility**: Legacy parameters without `adaptive_pruning` default to disabled. Obsolete `fill_rejected` keys are explicitly rejected with clear error messages. Disabled policy preserves identical graph topology to the legacy path (verified by test).

### Minor observations

A few non-blocking notes for consideration:

1. **`adaptive_pruning_fill_rejected` / `fill_rejected` in passthrough lists**: These keys appear in both `CheckAndMappingExternalParam` (Pyramid) and `map_hgraph_param` (HGraph) passthrough lists, then are explicitly rejected in `AdaptivePruningParameter::FromJson`. This is the correct pattern for producing a clear error message, but consider whether these passthrough entries should carry a comment noting they exist only for error-reporting purposes.

2. **HGraph `CheckCompatibility` checks `alpha`**: When adaptive pruning is enabled, `HGraphParameter::CheckCompatibility` checks `alpha` equality. This is correct since alpha directly affects pruning decisions. The Pyramid equivalent also checks this. Both are consistent.

3. **Test tag `[adaptive_pruning]` on existing tests**: The `"Pyramid promotes flat node at index minimum size"` and `"Pyramid multi-layer root builds routes and survives serialization"` tests now also run with `adaptive=true` via `GENERATE`. This is good coverage — existing functionality is verified to work correctly with adaptive pruning enabled.

No blocking issues found. The implementation is solid and ready to merge.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+    result.initial_rejected = rejected.size();
+    if (parameter.adjust_step == 0) {
+        // A zero step keeps the single-pass result, including any unused capacity.
+    } else if (not accepted.empty() && accepted.size() < target_degree) {
+        result.branch = AdaptivePruningBranch::RELAX;
+        const double ratio =
+            static_cast<double>(target_degree) / static_cast<double>(accepted.size());
+        int steps = 1;
+        if (ratio > 3) {
+            steps = 3;
```
LHT129, line 144: [note] **RELAX branch guard may skip relaxation when all candidates are rejected**

The RELAX branch condition `not accepted.empty() && accepted.size() < target_degree` prevents relaxation when `accepted` is empty (i.e., the first pass with the initial alpha rejects ALL candidates).

If `adjust_step > 0` and the initial alpha is too tight (e.g., due to a user-set alpha value that is lower than optimal), the algorithm returns an empty accepted list without attempting relaxation. The `adjust_step` parameter is specifically designed to handle this case by widening the threshold, but the `not accepted.empty()` guard prevents it from being used.

Consider whether the intent is:
1. **Intentional**: Empty accepted means no candidates are close enough to be useful neighbors, and relaxing would add low-quality edges. If so, this is fine as-is.
2. **Unintentional**: The algorithm should attempt relaxation even when `accepted` is empty, since `adjust_step` exists precisely to handle overly tight initial alphas. If so, consider removing the `not accepted.empty()` guard or adding a separate empty-accepted handling path.

src/impl/adaptive_pruning.cpp
LightWant: Implemented in b1af5d98: when the rejection ratio selects zero tightening steps, the selector keeps the first-pass result and skips the redundant scan. Branch metadata and first-pass counters are preserved. The regression now checks selected IDs and exact distance-call counts at ratio=5 and above, and retains rescan checks below the threshold. This is distinct from a configured adjust_step=0. Joint unit regression passed: 140 cases / 2372 assertions; clang-format 15 and clang-tidy 15 passed.

src/impl/pruning_strategy.cpp
LightWant: Documented in b1af5d98 why the caches have separate lifetimes. Each adaptive selection retains O(K) source computers. Sharing one cache across up to K full reverse-neighbor selections could retain O(K^2) distinct source computers until insertion finishes. The outer cache only retains the reverse centers for center-to-candidate distances. Keeping the inner cache local bounds live scratch to O(K), so I retained the separate instances and expanded the code comment.

src/impl/pruning_strategy.cpp
LightWant: Confirmed; the empty-neighbor early return is retained. It publishes the empty adjacency list and returns before selected_neighbors.back() or reverse-neighbor processing. No behavior change was needed for this note. The updated joint unit regression and Pyramid functional regression passed.

src/algorithm/pyramid/pyramid.cpp
LightWant: Confirmed; cached-graph connections retain the fixed-alpha selector. Adaptive Pyramid Build rejects a nonempty imported cache, while an empty cache is allowed. This remains the documented scope, and cache-rejection regression coverage passed. No cached-graph adaptive behavior was introduced.

src/impl/pruning_strategy.cpp
LightWant: Documented in b1af5d98 why the caches have separate lifetimes. Each adaptive selection retains O(K) source computers. Sharing one cache across up to K full reverse-neighbor selections could retain O(K^2) distinct source computers until insertion finishes. The outer cache only retains the reverse centers for center-to-candidate distances. Keeping the inner cache local bounds live scratch to O(K), so I retained the separate instances and expanded the code comment.

src/impl/adaptive_pruning.cpp
LightWant: This state cannot occur with positive target_degree and a nonempty normalized candidate list. In scan(), good starts as true and rejection only occurs while iterating over already accepted neighbors. For the first valid candidate, accepted is empty, so there are no comparisons and that candidate is unconditionally accepted, regardless of alpha. An empty result therefore means zero capacity or no valid candidates after self/duplicate normalization; there are no rejected candidates to recover by relaxation. Invalid distances are rejected as invalid input. The guard is intentional and also avoids dividing by zero in K / accepted.size(); it does not skip any viable relaxation case. Existing empty/self-only, normalization and relaxation-boundary tests cover these paths, so no code change is needed.

LHT129 (commented):
## Code Review

This PR adds adaptive NSW neighbor pruning with flat parameters to both Pyramid and HGraph indexes. The implementation is well-structured with thorough test coverage and clear documentation. All prior review comments have been addressed.

### Items noted

1. **`FromJson` full reset semantics** (`src/impl/adaptive_pruning.cpp:30`): The `*this = AdaptivePruningParameter{}` line resets all fields before parsing. This is correct for fresh construction, but means any partial JSON update (e.g., toggling only `enabled`) silently resets other fields to defaults. Since this is only called during index construction from a complete JSON parameter block, this is not a practical issue — just worth noting that the struct is not designed for incremental updates.

2. **`Validate()` called twice in `select_edges_adaptive`** (`src/impl/adaptive_pruning.cpp:81-82`): `select_edges_adaptive` calls both `CHECK_ARGUMENT(parameter.enabled, ...)` and `parameter.Validate(alpha)`. The `Validate` call is redundant since the index constructors (HGraph, Pyramid) already call `Validate` during initialization. The duplicate validation is harmless but adds a minor runtime cost on every edge selection. Consider removing the `parameter.Validate(alpha)` call inside `select_edges_adaptive` since the enabled check already guards the hot path.

3. **`scan` lambda captures `accepted` by reference across multiple calls** (`src/impl/adaptive_pruning.cpp:104-128`): The `scan` lambda captures `accepted` by reference and modifies it in place. This works correctly because each branch (RELAX, TIGHTEN) calls `scan` with different semantics: RELAX keeps the first-pass accepted list and scans only rejects; TIGHTEN clears accepted before rescanning the full candidate list. The lambda's dual-use (append-only in first pass and RELAX, clear-then-append in TIGHTEN) is correct but subtle. A comment explaining the accepted/rejected state machine across passes would help future readers.

4. **`CheckAndMappingExternalParam` passthrough for removed keys** (`src/algorithm/pyramid/pyramid.cpp:1878-1882` and `src/algorithm/hgraph/hgraph_param_mapping.cpp:299-302`): The `fill_rejected` and `adaptive_pruning_fill_rejected` keys are passed through to the inner JSON only to be rejected in `AdaptivePruningParameter::FromJson`. This is the correct pattern for producing clear error messages, but consider adding a brief comment noting these keys exist in the passthrough list solely for error-reporting purposes.

### Summary

No blocking issues. The implementation is solid with comprehensive test coverage across unit-level algorithm behavior, integration-level build-reload-add lifecycles, parameter roundtrips, and compatibility checks. All prior review comments have been addressed.

LHT129 (changes requested):
Thanks for this PR. The implementation is solid — well-structured, well-tested, and the algorithm documentation in `adaptive_pruning.md` is thorough. I have a few minor suggestions below.

**src/impl/adaptive_pruning.cpp — `Validate()` error message could include the alpha value**

When `Validate()` rejects an invalid alpha/step combination, the error message says "adaptive_pruning requires finite alpha/step, step >= 0 and alpha - 2*step > 0" but doesn't include the actual values. In Pyramid, where `Validate()` is called once for the global alpha and once per hierarchy, a user with multiple hierarchies won't know which alpha triggered the failure. Consider including the alpha value in the message, e.g.:



Similarly, the `apply_to_upper` rejection message could mention that the field is reserved for future use rather than just saying it's not supported.

**src/algorithm/pyramid/pyramid.h — Hierarchy alpha validation loop**

The constructor calls `adaptive_pruning_.Validate(hierarchy.alpha)` in a loop over hierarchies. If one fails, the error message won't distinguish which hierarchy caused the problem. A small improvement would be to catch and re-throw with hierarchy context, or include the hierarchy name in the message.

**src/impl/pruning_strategy.cpp — `select_adaptive_heap` local cache comment**

The comment explains why a local `PairwiseDistanceComputer` is created (to avoid O(K^2) cache pollution across reverse lists). This is correct, but when adaptive pruning is enabled for forward selection only (`apply_to_reverse=false`), the outer `PairwiseDistanceComputer` in `mutually_connect_new_element` is still constructed and only used by the reverse path's `select_edges_by_heuristic`. This is harmless but worth noting — the outer cache allocation could be avoided when adaptive forward selection is active but reverse is not, though the savings are minimal.

Overall this is a clean PR. The test coverage is excellent — the unit tests cover edge cases (empty inputs, coincident vectors, invalid parameters, serialization compatibility), and the integration tests cover both HGraph and Pyramid with various build configurations. The documentation clearly explains the algorithm, scope, and limitations.

LHT129 (changes requested):
Thanks for this PR. The implementation is solid — well-structured, well-tested, and the algorithm documentation in `adaptive_pruning.md` is thorough. I have a few minor suggestions below.

**src/impl/adaptive_pruning.cpp — `Validate()` error message could include the alpha value**

When `Validate()` rejects an invalid alpha/step combination, the error message says "adaptive_pruning requires finite alpha/step, step >= 0 and alpha - 2*step > 0" but doesn't include the actual values. In Pyramid, where `Validate()` is called once for the global alpha and once per hierarchy, a user with multiple hierarchies won't know which alpha triggered the failure. Consider including the alpha value in the message.

Similarly, the `apply_to_upper` rejection message could mention that the field is reserved for future use rather than just saying it's not supported.

**src/algorithm/pyramid/pyramid.h — Hierarchy alpha validation loop**

The constructor calls `adaptive_pruning_.Validate(hierarchy.alpha)` in a loop over hierarchies. If one fails, the error message won't distinguish which hierarchy is problematic. A small improvement would be to include the hierarchy name in the validation context.

**src/impl/pruning_strategy.cpp — `select_adaptive_heap` local cache**

The comment explains why a local `PairwiseDistanceComputer` is created (to avoid O(K^2) cache pollution across reverse lists). This is correct, but when adaptive pruning is enabled for forward selection only (`apply_to_reverse=false`), the outer `PairwiseDistanceComputer` in `mutually_connect_new_element` is still constructed and only used by the reverse path's `select_edges_by_heuristic`. This is harmless but worth noting — the outer cache allocation could be avoided when adaptive forward selection is active but reverse is not, though the savings are minimal.

Overall this is a clean PR. The test coverage is excellent — the unit tests cover edge cases (empty inputs, coincident vectors, invalid parameters, serialization compatibility), and the integration tests cover both HGraph and Pyramid with various build configurations. The documentation clearly explains the algorithm, scope, and limitations.

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,44 @@ class PairwiseDistanceComputer {
...
+        max_size,
+        alpha,
+        parameter,
+        [&](InnerIdType left, InnerIdType right) {
+            return pairwise_distance.PairwiseDistance(left, right);
+        },
+        allocator);
+    for (const auto& candidate : selected) {
+        edges->Push(candidate.first, candidate.second);
+    }
```
LHT129, line 90: [note] `select_adaptive_heap` pushes `(distance, id)` with positive distances into the max-heap, while `select_edges_by_heuristic` pushes `(-distance, id)` with negated distances. This reverses the heap order between the two paths.

In `mutually_connect_new_element`, after draining the heap, `selected_neighbors` is in opposite order: adaptive path produces [farthest, ..., closest] while the non-adaptive path produces [closest, ..., farthest]. The return value `selected_neighbors.back()` therefore differs (closest vs farthest).

No production caller currently uses this return value (all call sites in hgraph_build.cpp and pyramid.cpp discard it), so this is functionally harmless. However, it is a latent inconsistency — if a future caller relies on the documented "farthest selected neighbor" semantics, the adaptive path would silently return the wrong value.

src/impl/adaptive_pruning.h
```diff
@@ -0,0 +1,62 @@
...
+    uint64_t initial_rejected{0};
+    uint64_t distance_calls{0};
+    float second_alpha{0};
+    AdaptivePruningBranch branch{AdaptivePruningBranch::SINGLE};
+};
+
+using PruningCandidate = std::pair<float, InnerIdType>;
+
+// Pure L2 selector. Candidates are normalized in place; the returned list is nearest first.
+// The caller decides whether this policy applies to its graph layer and selection role.
```
LHT129, line 53: [note] The header comment describes `select_edges_adaptive` as a "Pure L2 selector", but the function is actually distance-agnostic — it accepts an arbitrary `std::function<float(InnerIdType, InnerIdType)>` and applies the PRUNE condition `threshold * d(s,c) < d(u,c)` regardless of the metric. The L2 constraint is enforced at a higher level (HGraph/Pyramid constructors reject non-L2SQR metrics). Consider rewording to "Applies the PRUNE selection rule with the supplied distance function" to avoid implying the function itself is coupled to L2.

src/impl/pruning_strategy.cpp
LightWant: Checked both heap paths: current_pair.first is already NEGATIVE when stored in return_list, so edges->Push(-current_pair.first, ...) restores a POSITIVE distance. Both paths therefore drain the max-heap farthest-to-nearest and return the nearest neighbor. The implementation is consistent, but the header incorrectly documented farthest. Fixed the header in 43acf6a7 to document nearest, or cur_c for an empty selection. Header format/diff checks passed; the unchanged implementation passed 24 selector/connection cases with 1109 assertions.

src/impl/adaptive_pruning.h
LightWant: Addressed in 43acf6a7. The comment now says the selector applies adaptive pruning using the supplied distance function, and separately documents its finite, non-negative squared-L2 distance contract and caller-side metric enforcement. Accepting a generic callback does not validate which metric it computes, so keeping that contract explicit avoids suggesting support for unvalidated metrics. No algorithm behavior changed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2115,8 +2123,14 @@ Pyramid::add_bottom_graph_point(const Hierarchy& hierarchy,
             connect_cached_graph_point(
```
LHT129, line 2123: **[suggestion]** `connect_cached_graph_point` does not pass adaptive pruning parameters, always using fixed-alpha `select_edges_by_heuristic`. This means the cached-graph build path (triggered when `use_self_as_entry=true` and the node already has cached neighbors) will not apply adaptive pruning even when it is enabled for the bottom graph.

In `add_bottom_graph_point` and `add_routed_point`, the non-cached path correctly passes `&adaptive_pruning_` to `mutually_connect_new_element`, but the cached path calls `connect_cached_graph_point` which internally uses `select_edges_by_heuristic` with only `alpha`.

If this is intentional (adaptive pruning is not yet supported for cached builds, similar to how imported caches are rejected), consider adding a comment or validation. Otherwise, `connect_cached_graph_point` should accept and forward the `AdaptivePruningParameter`.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+        }
+        result.second_alpha = alpha + static_cast<float>(steps) * parameter.adjust_step;
+        Vector<PruningCandidate> remaining(allocator);
+        scan(rejected, result.second_alpha, remaining);
+    } else if (accepted.size() == target_degree) {
+        result.branch = AdaptivePruningBranch::TIGHTEN;
+        const double ratio =
+            static_cast<double>(rejected.size()) / static_cast<double>(target_degree);
+        int steps = 2;
+        if (ratio >= 5) {
```
LHT129, line 156: **[suggestion]** In the TIGHTEN fallback path, when `accepted.size() < target_degree` after tightening, the code rescans `rejected` with the original `alpha`. However, `rejected` at this point only contains candidates rejected during the tightening scan — it does not include the unscanned tail `T` from the original candidate list that was never examined in the first pass (because the first pass stopped at `target_degree` accepted candidates).

This means the fallback cannot discover tail candidates that were never seen in the first pass. If the tightening scan rejects some previously-accepted candidates and the fallback needs to fill the remaining slots, it can only choose from candidates that were already rejected during tightening, potentially missing better candidates from the unscanned tail.

Consider whether the fallback should scan the original tail `T` (candidates from index `target_degree` onward in the sorted list) rather than only the tightening-rejected subset.

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,44 @@ class PairwiseDistanceComputer {
...
+                     uint64_t max_size,
+                     const DistanceProviderForGraph& distance_provider,
+                     Allocator* allocator,
+                     float alpha,
+                     const AdaptivePruningParameter& parameter) {
+    // Keep this cache local to one selection: adaptive passes need at most O(K) source
+    // computers. Sharing the caller's cache across up to K full reverse lists could retain
+    // O(K^2) distinct source computers until insertion finishes. The caller's separate cache
+    // is only for center-to-candidate distances and retains at most K reverse centers.
+    PairwiseDistanceComputer pairwise_distance(distance_provider, allocator);
```
LHT129, line 71: **[note]** The `select_adaptive_heap` function creates its own local `PairwiseDistanceComputer` (line 71) rather than reusing the caller's cache. The comment explains this is to avoid retaining O(K²) source computers across K full reverse lists. This is a reasonable design choice, but it means each adaptive selection call pays the cost of creating fresh distance computers for the source IDs it encounters.

The caller's `PairwiseDistanceComputer` (created at line 172 in `mutually_connect_new_element`) is only used for the center-to-neighbor distances in the reverse edge selection loop (line 223), not for the adaptive selection itself. This separation is clean but worth documenting as a deliberate trade-off between memory and CPU.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+            steps = 3;
+        } else if (ratio > 1.5) {
+            steps = 2;
+        }
+        result.second_alpha = alpha + static_cast<float>(steps) * parameter.adjust_step;
+        Vector<PruningCandidate> remaining(allocator);
+        scan(rejected, result.second_alpha, remaining);
+    } else if (accepted.size() == target_degree) {
+        result.branch = AdaptivePruningBranch::TIGHTEN;
+        const double ratio =
```
LHT129, line 153: **[note]** In the TIGHTEN branch (line ~147-155), `accepted.clear()` and `rejected.clear()` are called before rescanning the full `candidates` list at the tightened alpha. The comment on line 153 says "The normalized original list is exactly sorted(A+B) followed by unscanned tail T" — this is correct because the first pass only scanned up to `target_degree` accepted candidates, so `candidates` still contains the full sorted list.

However, the fallback on line 156 rescans `rejected` (which now contains only tightening-rejected candidates) rather than the original tail. See the separate comment on line 156 for the potential gap.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,181 @@
...
+namespace vsag {
+
+void
+AdaptivePruningParameter::FromJson(const JsonType& json) {
+    CHECK_ARGUMENT(json.IsObject(), "index parameters must be an object");
+    const bool has_removed_fill =
+        json.Contains("fill_rejected") || json.Contains("adaptive_pruning_fill_rejected");
+    CHECK_ARGUMENT(not has_removed_fill,
+                   "fill_rejected has been removed; rejected neighbors are never filled");
+    *this = AdaptivePruningParameter{};
```
LHT129, line 32: **[suggestion]** `FromJson` resets `*this` to default at line 32 (`*this = AdaptivePruningParameter{}`) before parsing fields. This means if `FromJson` is called on an already-configured object, any fields not present in the new JSON will revert to defaults rather than retaining their previous values.

This is consistent with how other parameter structs in the codebase work (full replacement semantics), but it means callers cannot incrementally update individual fields. If partial updates are ever needed, consider parsing into a temporary and merging.

LHT129 (commented):
## Code Review Summary

This PR adds adaptive NSW neighbor pruning to Pyramid and HGraph with flat configuration parameters. Overall the implementation is well-structured with clear separation between the core algorithm (`adaptive_pruning.cpp/.h`), integration layer (`pruning_strategy.cpp/.h`), and index-specific wiring (pyramid, hgraph). The test coverage is thorough across unit, integration, parameter, and serialization dimensions.

### Key observations:

1. **`connect_cached_graph_point` does not apply adaptive pruning** — The cached-graph build path in pyramid uses fixed-alpha `select_edges_by_heuristic` while the non-cached path correctly passes `&adaptive_pruning_`. This may be intentional (similar to how imported caches are rejected), but should be explicitly documented or validated.

2. **TIGHTEN fallback may miss tail candidates** — When tightening rejects some previously-accepted candidates and the fallback needs to fill remaining slots, it rescans only the tightening-rejected subset, not the original unscanned tail `T`. This could miss better candidates that were never examined in the first pass.

3. **Clean separation of distance computer caches** — `select_adaptive_heap` creates its own local `PairwiseDistanceComputer` to avoid O(K²) memory retention across reverse lists. The design is deliberate and well-commented.

4. **Serialization compatibility is properly enforced** — Both Pyramid and HGraph check `adaptive_pruning.enabled`, `alpha`, `adjust_step`, `apply_to_reverse`, and `apply_to_upper` in `CheckCompatibility`, preventing mismatched deserialization.

5. **Good rejection of unsupported configurations** — Non-L2 metrics, ODescent graphs, upper-layer adaptation, imported caches, and the removed `fill_rejected` option are all properly rejected with clear error messages.

6. **Comprehensive test coverage** — Tests cover: parameter parsing/roundtrip, build/reload/add lifecycle, reverse pruning behavior, tightening/relaxation boundaries, normalization edge cases, invalid inputs, serialization compatibility, and unsupported contexts.

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include "adaptive_pruning.h"
+
+#include <algorithm>
+#include <cmath>
+
+#include "hash_types.h"
+
+namespace vsag {
+
```
LHT129, line 24: [suggestion]  performs an overly strict  check for a sub-struct parser. Since  is called from  with a sub-object (), this check is fine in practice. However, the error message "index parameters must be an object" is misleading in the context of a nested parameter block — it would be clearer as something like "adaptive_pruning must be an object".

src/impl/pruning_strategy.cpp
```diff
@@ -52,10 +52,41 @@ class PairwiseDistanceComputer {
...
-    // The helper lives for one pruning pass; its source IDs are bounded by the selected-edge limit.
+    // Per-call cache; adaptive passes may select different source IDs (at most twice the degree).
     UnorderedMap<InnerIdType, ComputerInterfacePtr> computers_;
 };
 
+void
+select_adaptive_heap(const DistHeapPtr& edges,
+                     InnerIdType center,
+                     uint64_t max_size,
+                     const DistanceProviderForGraph& distance_provider,
```
LHT129, line 63: [note] The `select_adaptive_heap` helper creates a local `PairwiseDistanceComputer` per call. This is intentional — the comment explains it keeps distance-computer scratch bounded to one vertex selection, avoiding O(K^2) memory. The design is sound, but consider whether the per-call construction cost is measurable in hot paths (e.g., during incremental insertion of many elements).

src/impl/adaptive_pruning.cpp
```diff
@@ -0,0 +1,191 @@
...
+    result.initial_rejected = rejected.size();
+    if (parameter.adjust_step == 0) {
+        fill(rejected);
+    } else if (not accepted.empty() && accepted.size() < target_degree) {
+        result.branch = AdaptivePruningBranch::RELAX;
+        const double ratio =
+            static_cast<double>(target_degree) / static_cast<double>(accepted.size());
+        int steps = 1;
+        if (ratio > 3) {
+            steps = 3;
```
LHT129, line 144: [note] In the TIGHTEN branch, when `steps != 0`, `accepted` and `rejected` are cleared and the full `candidates` list is rescanned at the tighter alpha. This means the first-pass scan work is discarded entirely. If the expected rejection ratio is typically high (making tightening common), consider whether the first pass could be skipped when tightening is anticipated, or whether the first-pass results could be partially reused.

src/impl/adaptive_pruning.h
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include <functional>
+
+#include "parameter.h"
+
+namespace vsag {
+
+struct AdaptivePruningParameter {
+    bool enabled{false};
```
LHT129, line 25: [note] The `fill_rejected` field exists in `AdaptivePruningParameter` and `AdaptivePruningStats` but the current `select_edges_adaptive` implementation does not appear to use it (the `fill` lambda is not called in the main flow). If `fill_rejected` is planned for a future change, consider removing it from this PR to keep the diff minimal, or add a brief comment explaining it is reserved for future use.

src/algorithm/hgraph/hgraph.h
```diff
@@ -931,7 +931,9 @@ class HGraph : public InnerIndexInterface {
     HGraphMCIParameters mci_parameters_{};
 
     uint64_t ef_construct_{400};  // expansion factor during graph construction
-    float alpha_{1.0};            // Relative Neighborhood Graph pruning coefficient
+    AdaptivePruningParameter adaptive_pruning_;
```
LHT129, line 934: [suggestion] The HGraph constructor calls adaptive_pruning_.Validate(alpha_) unconditionally. When adaptive_pruning_.enabled is false, Validate is a no-op, which is fine. However, if alpha_ is invalid (e.g., negative or NaN) but adaptive pruning is disabled, this will silently pass. Consider whether alpha_ should be validated independently of the adaptive pruning feature, since it is also used by the legacy select_edges_by_heuristic path.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -163,6 +163,8 @@ HGraph::Build(const DatasetPtr& data) {
     std::vector<int64_t> ret;
     const bool using_build_cache = this->has_loaded_cache();
```
LHT129, line 164: [note] The `CHECK_ARGUMENT` guard at the top of the `using_build_cache` branch correctly blocks adaptive pruning with imported caches. However, the check happens *after* `this->has_loaded_cache()` but *before* `this->check_fused_mutation_supported(...)`. If a future refactor moves `check_fused_mutation_supported` before this guard, the fused-mutation check could throw before the more specific adaptive-pruning rejection message is reached. Consider grouping all build-cache preconditions together at the top of the branch.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -163,6 +163,8 @@ HGraph::Build(const DatasetPtr& data) {
...
     if (using_build_cache) {
+        CHECK_ARGUMENT(not adaptive_pruning_.enabled,
+                       "adaptive_pruning does not yet support building from imported cache");
         this->check_fused_mutation_supported("Build with imported cache");
```
LHT129, line 168: [note] `publish_unique_to_bottom_graph` now passes `&adaptive_pruning_` to `mutually_connect_new_element`. This means the adaptive selector runs on the optimized build path (including RaBitQ split/fused). The `connect_cached_graph_point` path at line 2125 does not pass adaptive pruning, which is consistent with the documented scope. However, the asymmetry means the same `Build()` call uses two different selectors depending on whether the insertion goes through the optimized path or the cached-graph path — worth keeping in mind when comparing recall between build strategies.

src/algorithm/hgraph/hgraph_param_mapping.cpp
```diff
@@ -296,7 +296,11 @@ HGraph::map_hgraph_param(const JsonType& hgraph_json) {
             inner_json[PARAMETER_USE_CONJUGATE_GRAPH].SetJson(value);
         } else if (key == HGRAPH_LABEL_REMAP_TYPE) {
             inner_json[LABEL_REMAP_TYPE_KEY].SetJson(value);
-        } else if (key == HGRAPH_USE_MCI || key == HGRAPH_MCI_MCS || key == HGRAPH_MCI_CLIQUE_MAX ||
```
LHT129, line 299: [note] The passthrough list includes `adaptive_pruning_fill_rejected` and `fill_rejected` keys. These are forwarded to the inner parameter parser where `AdaptivePruningParameter::FromJson` rejects them with an error. This is correct behavior — the user gets a clear rejection message rather than silent ignoring — but the passthrough entries are technically dead code since they always result in a thrown exception. If the rejection logic in `FromJson` is ever relaxed, these keys could silently pass through.

src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -129,6 +129,7 @@ HGraphParameter::FromJson(const JsonType& json) {
         CHECK_ARGUMENT(this->ef_construction > 0, "ef_construction must be positive");
     }
```
LHT129, line 130: [note] `adaptive_pruning.FromJson(json)` is called at line 132, *before* `alpha` is parsed from JSON at line 134. This means `FromJson` cannot validate the adaptive pruning parameters against the user-provided alpha — it only validates the internal field constraints (non-negative step, finite bounds). The actual alpha-aware validation happens later in the HGraph constructor via `adaptive_pruning_.Validate(alpha_)`. This two-phase validation is correct but means a user who provides `alpha=0.5` with `adjust_step=0.3` will pass `FromJson` and only fail in the constructor. The error message will come from `Validate` rather than `FromJson`, which may be less discoverable.

src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -292,6 +293,7 @@ HGraphParameter::ToJson() const {
```
LHT129, line 291: [note] `CheckCompatibility` compares `alpha` only when `adaptive_pruning.enabled` is true on both sides. This is correct — when disabled, alpha changes are harmless for the adaptive selector. However, if one side has `enabled=true` and the other has `enabled=false`, the `CHECK_FIELD_EQ(*this, *p, adaptive_pruning.enabled)` at line 327 will fail first, so the alpha comparison is never reached. The logic is sound.

src/algorithm/pyramid/pyramid_zparameters.cpp
```diff
@@ -357,6 +358,7 @@ JsonType
 PyramidParameters::ToJson() const {
     JsonType json = InnerIndexParameter::ToJson();
```
LHT129, line 359: [note] `PyramidParameters::CheckCompatibility` has the same pattern as HGraph: it compares `alpha` only when `adaptive_pruning.enabled` is true. However, Pyramid has multiple hierarchy alphas — this check only compares the top-level `alpha` field, not the per-hierarchy alphas. The constructor validates all hierarchy alphas against the adaptive pruning parameters, so a mismatch would be caught at construction time rather than at `CheckCompatibility` time. This is consistent with the design where `CheckCompatibility` validates serialized parameter objects.

include/vsag/constants.h
```diff
@@ -182,6 +182,8 @@ extern const char* const HGRAPH_BASE_QUANTIZATION_TYPE;
 extern const char* const HGRAPH_GRAPH_MAX_DEGREE;
 extern const char* const HGRAPH_BUILD_EF_CONSTRUCTION;
```
LHT129, line 183: [note] Both `HGRAPH_ADAPTIVE_PRUNING` and `PYRAMID_ADAPTIVE_PRUNING` are defined as the same string literal `"adaptive_pruning"`. This is intentional — the flat parameter model uses a single key name across index types. The two constants exist for code clarity when referencing the key in HGraph vs Pyramid context. No risk of collision since they have the same value.

src/impl/pruning_strategy.h
```diff
@@ -38,8 +39,8 @@ DEFINE_POINTER(MutexArray);
  * @param max_size Maximum number of edges to select.
  * @param flatten Flatten interface for computing pairwise vector distances.
  * @param allocator Allocator for memory management.
- * @param alpha Diversity parameter controlling the trade-off between proximity
- *              and diversity. Higher values allow more diverse neighbors.
+ * @param alpha Pruning relaxation parameter. Larger values make the geometric
```
LHT129, line 42: [note] The doc comment for `mutually_connect_new_element`'s return value changed from "The ID of the farthest selected neighbor" to "The ID of the nearest selected neighbor, or cur_c if no neighbor is selected." This reflects the actual behavior: `selected_neighbors.back()` returns the last element after heap drain, which is the nearest (smallest distance) since the max-heap pops largest first. The adaptive path also returns `selected_neighbors.back()`. However, as noted in another comment, the adaptive path's heap pushes `(distance, id)` while the non-adaptive path pushes `(-distance, id)`, so `back()` gives opposite semantics between the two paths. The doc update correctly describes the non-adaptive path but may mislead readers about the adaptive path's return value.

src/impl/adaptive_pruning_test.cpp
```diff
@@ -0,0 +1,313 @@
+
```
LHT129, line 1: [note] The test suite for `adaptive_pruning` is thorough: it covers relaxation boundaries (1-5 initial accepted), tightening boundaries (0-11 rejected), zero-step single-pass, invalid inputs (empty, self-only, K=0, negative distance), parameter roundtrip with legacy JSON, and a real L2 tightening example across input permutations. The `GENERATE`-based combinatorial testing for boundary ratios is particularly effective at catching off-by-one errors in the ratio-to-step mapping.

src/algorithm/hgraph/hgraph_add_test.cpp
```diff
@@ -1539,3 +1539,148 @@ TEST_CASE("HGraph optimized build accepts MRLE RaBitQ split with raw-vector stor
     vsag::HGraphOptimizedBuildSession session(*hgraph);
```
LHT129, line 1539: [note] The HGraph integration tests cover the full lifecycle: Build -> Serialize -> Deserialize -> KnnSearch -> Add -> KnnSearch, with both 1 and 4 threads. The compatibility checks verify that mismatched `adjust_step`, `apply_to_reverse`, and `alpha` all cause `Deserialize` to fail. The unsupported contexts test covers ODescent, inner product, `apply_to_upper`, and invalid threshold. The imported cache rejection test and RaBitQ fused lifecycle test provide additional coverage. This is solid integration-level validation.

src/algorithm/pyramid/pyramid_test.cpp
```diff
@@ -65,7 +65,8 @@ MakePyramidIndex(uint32_t index_min_size,
                  bool use_mrle_split = false,
```
LHT129, line 65: [note] The Pyramid integration tests cover flat-node promotion, multi-layer root builds with both `flat` and `compressed` graph storage, and root/path query validation — all with `GENERATE(adaptive, reverse)` combinatorial coverage. The `MakePyramidIndex` and `MakeRootPyramidIndex` helpers gained `adaptive` and `reverse` parameters with sensible defaults (`false`), preserving backward compatibility for existing callers. The test coverage for Pyramid is consistent with the HGraph integration tests.

src/algorithm/hgraph/hgraph_parameter_test.cpp
```diff
@@ -17,6 +17,7 @@
```
LHT129, line 17: [note] The parameter tests verify that legacy JSON (without `adaptive_pruning` key) produces `enabled=false`, and that `CheckCompatibility` ignores `alpha` and `adjust_step` differences when disabled. The flat parameter mapping test confirms that `fill_rejected` and `adaptive_pruning_fill_rejected` are rejected, and that nested `{"adaptive_pruning": {"enabled": true}}` (the old experimental format) is also rejected. Good coverage of the migration path from the experimental schema.

src/algorithm/pyramid/pyramid_zparameters_test.cpp
```diff
@@ -861,3 +861,92 @@ TEST_CASE("Pyramid validates explicit factor", "[ut][PyramidParameters]") {
     REQUIRE_THROWS(
```
LHT129, line 861: [note] The Pyramid parameter tests cover legacy compatibility (disabled by default), roundtrip serialization, and per-field compatibility checks (`alpha`, `adjust_step`, `apply_to_reverse`, `apply_to_upper`). The hierarchy alpha validation test confirms that the constructor rejects configurations where any hierarchy's alpha violates the adaptive pruning constraints, and covers the same unsupported contexts as HGraph (ODescent, non-L2, nested policy, fill_rejected).

src/impl/pruning_strategy_test.cpp
```diff
@@ -172,6 +185,60 @@ TEST_CASE("Pruning Strategy Select Edges With Heuristic", "[ut][pruning_strategy
...
+        auto connect = [&](const GraphInterfacePtr& graph, const AdaptivePruningParameter* p) {
+            for (InnerIdType node = 0; node < 5; ++node) {
+                auto heap = std::make_shared<StandardHeap<true, false>>(allocator.get(), -1);
+                for (InnerIdType other = 0; other < node; ++other) {
+                    heap->Push(FlattenDistanceProvider(flatten, nullptr)
+                                   .PairwiseDistance(node, other, nullptr),
+                               other);
+                }
+                mutually_connect_new_element(
+                    node, heap, graph, flatten, mutexes, allocator.get(), 1.5F, p);
```
LHT129, line 206: [note] The reverse pruning tests are well-designed: the "relaxes around the existing neighbor" test uses `GENERATE(incoming_is_nearer, policy.enabled, policy.apply_to_reverse)` to verify that relaxation admits the nearer distant candidate while the baseline rejects both. The "preserves append and prunes full lists" test confirms that non-full reverse lists append directly (bypassing the selector) while full lists run the adaptive selector. The disabled-policy test verifies bit-exact equivalence with the legacy path even when `adjust_step=99`.

docs/docs/en/src/indexes/adaptive_pruning.md
```diff
@@ -0,0 +1,91 @@
+# Adaptive neighbor pruning in HGraph and Pyramid
```
LHT129, line 1: [note] The `adaptive_pruning.md` documentation is thorough: it covers scope and compatibility, the distance rule with the mathematical inequality, the adaptive schedule with ratio-to-step tables, a worked example with concrete numbers, cost and performance interpretation, related work with citations, and validation coverage. The explicit disclaimer that "this is not a claim of inventing adaptive neighbor selection" and the detailed provenance section with links to Descartes, HNSW, Vamana, SymphonyQG, and Alpha-CNG are particularly well-done.

docs/docs/en/src/indexes/hgraph.md
```diff
@@ -236,6 +236,66 @@ The chosen `dtype` only constrains the **input** representation. The on-disk / i
 still controlled by `base_quantization_type` (and optionally `precise_quantization_type` when
```
LHT129, line 236: [note] The HGraph documentation's adaptive pruning section is clear and complete: it lists all flat parameters with defaults, explains the scope (forward edges, optional reverse, no upper layers), describes the alpha schedule with ratio-to-step tables, and explicitly states that this is "a GIST reference experiment, not a dimension-dependent default or a claim of improved performance." The cross-reference to `adaptive_pruning.md` for algorithm details avoids duplication.

docs/docs/en/src/indexes/pyramid.md
```diff
@@ -96,6 +96,7 @@ Build-time parameters live under `index_param`.
 | `graph_storage_type` | string | `"flat"` | Bottom-graph storage for a `multi_layer` root: `flat` favors construction and search speed, while `compressed` reduces graph memory. Compressed storage requires `max_degree <= 255`. Single-layer roots, routing graphs, and child graphs remain sparse. |
```
LHT129, line 96: [note] The Pyramid documentation correctly notes that the adaptive pruning policy is global (not per-hierarchy), that routing layers are not adapted, and that each hierarchy retains its own effective alpha. The parameter table and JSON example are consistent with the HGraph documentation. The note that "there is no hierarchy-specific `adaptive_pruning` override" is important for users who might expect per-level tuning.

LHT129 (commented):
## Code Review Summary for PR #2934

This PR adds an experimental adaptive NSW neighbor pruning selector for HGraph and Pyramid bottom-layer construction, with flat parameter configuration.

### Overall Assessment

The implementation is solid, well-tested, and thoroughly documented. The code follows VSAG conventions (4-space indent, `uint64_t`, `vsag` namespace, `.cpp` suffix). The test coverage is excellent — unit tests cover ratio boundaries, normalization, invalid inputs, and parameter roundtrip; integration tests cover the full build-serialize-reload-add lifecycle with 1 and 4 threads for both HGraph and Pyramid.

### Key Observations

1. **Architecture**: The adaptive selector is cleanly separated into `adaptive_pruning.cpp/h` with a well-defined `AdaptivePruningParameter` struct. Integration into `mutually_connect_new_element` is minimal and gated behind `pruning != nullptr && pruning->enabled`, preserving the legacy path.

2. **Validation**: Both HGraph and Pyramid constructors validate the adaptive pruning parameters against their respective alphas, rejecting unsupported configurations (ODescent, non-L2, imported cache, `apply_to_upper`). `CheckCompatibility` ensures serialized indexes cannot be reloaded with mismatched parameters.

3. **Documentation**: The `adaptive_pruning.md` document is exemplary — it covers the distance rule, adaptive schedule with ratio-to-step tables, a worked example, cost interpretation, and related work with proper citations. The explicit provenance section is particularly well-done.

4. **Backward Compatibility**: Legacy JSON without `adaptive_pruning` defaults to disabled. The experimental nested-object format and removed `fill_rejected` option are explicitly rejected with clear error messages.

### Items Flagged in Inline Comments

Most comments are `[note]` level — observations about design choices that are correct but worth documenting for future readers. A few `[suggestion]` items cover potential improvements to the `FromJson` strictness and the `PairwiseDistanceComputer` instance management.

No `[critical]` issues were found.

Issue #2939, opened 2026-09-11
[improve](eval): add reproducible fixed-ef calibration benchmarks


Utopia-V:
### Kind of improvement

test

### Affected module

not applicable

### Affected area

area/tools

### Summary

Add a runnable example that calibrates fixed `ef_search` with the existing evaluation and AutoTune tools. It will provide the reference for the early-stop comparison in #2941.

The workflow should:

- Read the existing evaluation data format and record the dataset, metric and index settings.
- Use separate query content for calibration and evaluation, with an explicit weighting for repeated queries.
- Find a fixed `ef_search` that meets the declared quality target, then save that configuration. Record when no tested budget meets the target.
- Use the query-coverage metric from #2841 / #2864 alongside average recall, and report latency, QPS and search work on the evaluation queries.

Include a runnable calibration example and configurations covering at least two datasets, three recall targets, `k=10/100`, and single-/multi-threaded runs.

### Expected benefit

A contributor should be able to reproduce the fixed baseline before comparing a new stopping policy with it.

### Benchmark evidence

The DBpedia, Cohere and LAION comparisons in #2941 provide examples for this workflow.

### References

At upstream `fd077c3f`:

- `tools/eval/README.md:13` — evaluation entry point.
- `tools/eval/eval_template.yaml:15` — dataset, index and search configuration.
- `docs/docs/en/src/resources/autotune.md:153` — fixed-ef candidates.

### Additional context

Part of #2840. Relevant entry points are the existing `tools/eval` examples and AutoTune documentation.

#2562 covers runtime recall-target parameter mapping; this Issue covers offline calibration and reproducible experiments.

## Related issues

#2840, #2841, #2562, #2940, #2941; evaluation support in #2864.

---
_Drafted with AI assistant: Codex:GPT-6_

PR #2943, opened 2026-09-13
test(eval): prepare calibration and validation query subsets

Utopia-V:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

Part of #2939.

## What Changed

Add `tools/eval/prepare_query_split.py` to generate calibration and validation HDF5 inputs from explicit query-row lists. It rejects cross-split duplicate query values, keeps query/ground-truth rows aligned in the requested order, and shares the original base vectors through relative HDF5 links.

The helper supports unfiltered dense float32/int8 datasets, preserves row weighting within each split, and writes the source/row mapping alongside the outputs. Python callers can supply Python or NumPy integer row indices; accepted indices are normalized for JSON output. Existing output directories are not replaced.

## Test Evidence

- [x] Other: 11 Python tests passed, including a native `eval_performance` test that builds one HGraph index and searches both generated subsets.
- [x] Local documentation checks passed after generating `src/version.h` with the repository generator. The clean-checkout Docs CI run failed on existing testing-guide references to that generated file.
- [x] CI YAML and the added shell step validated; `git diff --check` passed.

The native check reused a Release build with the same dataset-loader source as this PR's base. The existing X86 CI job runs these tests with the PR's ASan-built evaluation binary. Its previous run failed at binary startup with an `is_inited` ODR violation; that failure remains unresolved.

## Compatibility Impact

No C++ API/ABI changes. The new standalone helper requires NumPy and h5py. Sparse, multi-vector and filtering layouts are rejected.

Query data is copied in bounded blocks. Base vectors are linked rather than duplicated; the source HDF5 file must remain unchanged and available at the recorded relative path. No changes to search execution or measurement code.

## Documentation Impact

- [x] Updated English and Chinese evaluation guides and the tool README links.

## Risk and Rollback

The generated subsets depend on the source dataset through external links; this is documented with the usage example. Revert the helper, tests and documentation to roll back this change.

## Checklist

- [x] Linked the relevant Issue.
- [x] Added behavior and integration tests.
- [x] Considered API compatibility and input preservation.
- [x] Updated the user documentation.
- [x] Commit follows the repository's Conventional Commit and attribution rules.


tools/eval/prepare_query_split.py
```diff
@@ -0,0 +1,160 @@
...
+    batch = max(1, BLOCK_BYTES // (dataset.shape[1] * dataset.dtype.itemsize))
+    for start in range(0, len(indices), batch):
+        selected = np.asarray(indices[start:start + batch], dtype=np.int64)
+        order = np.argsort(selected)
+        yield start, dataset[selected[order]][np.argsort(order)]
+
+
+def validate_content(source, rows):
+    seen = {}
+    for role in ROLES:
```
LHT129, line 83: [suggestion] The row index type check `type(i) is not int` on line 83 rejects `numpy.integer` subclasses (e.g. `np.int64`, `np.uint32`) that are valid Python indices and work correctly with h5py slicing. While the CLI path always produces plain `int` from `json.loads`, the `prepare_split` function is importable and callable from other Python code (as the test file does via `importlib`). Consider using `not isinstance(i, (int, np.integer))` to accept numpy integer types as well.

Current:
```python
if any(type(i) is not int or not 0 <= i < query_count for i in selected):
```

Suggested:
```python
if any(not isinstance(i, (int, np.integer)) or not 0 <= i < query_count for i in selected):
```

tools/eval/prepare_query_split.py
```diff
@@ -0,0 +1,160 @@
...
+def copy_attributes(source, destination):
+    for name, value in source.attrs.items():
+        destination.attrs.create(name, value, dtype=source.attrs.get_id(name).dtype)
+
+
+def prepare_split(source_path, rows, output_path):
+    source_path = Path(source_path).resolve(strict=True)
+    output_path = Path(output_path)
+    if output_path.exists() or output_path.is_symlink():
+        raise FileExistsError(f"output directory already exists: {output_path}")
```
LHT129, line 112: [suggestion] There is a TOCTOU (time-of-check time-of-use) race condition between the existence check and directory creation. If two concurrent invocations target the same output directory, both could pass the `exists()` / `is_symlink()` check on lines 112-113 before either reaches `mkdir()` on line 135. The second `mkdir()` would then raise `FileExistsError`, but the first invocation's staging directory would have already been cleaned up by `TemporaryDirectory`, leaving a potentially incomplete output.

Consider removing the explicit pre-check and relying solely on `output_path.mkdir()` (which raises `FileExistsError` if the directory exists) as the atomic gate. Move the check closer to the `mkdir()` call or use it as the exclusive guard:

```python
def prepare_split(source_path, rows, output_path):
    source_path = Path(source_path).resolve(strict=True)
    output_path = Path(output_path).resolve()
    with h5py.File(source_path, "r") as source:
        validate_dataset(source)
        validate_rows(rows, source["test"].shape[0])
        validate_content(source, rows)
        output_path.parent.mkdir(parents=True, exist_ok=True)
        relative_source = os.path.relpath(source_path, output_path)
        with tempfile.TemporaryDirectory(prefix=".query-split-", dir=output_path.parent) as temp:
            staging = Path(temp)
            for role in ROLES:
                with h5py.File(staging / f"{role}.hdf5", "x") as target:
                    copy_attributes(source, target)
                    target["train"] = h5py.ExternalLink(relative_source, "/train")
                    for name in ("test", "neighbors", "distances"):
                        original = source[name]
                        selected = target.create_dataset(
                            name, shape=(len(rows[role]), original.shape[1]), dtype=original.dtype
                        )
                        copy_attributes(original, selected)
                        for offset, values in read_rows(original, rows[role]):
                            selected[offset:offset + len(values)] = values
            (staging / "query_rows.json").write_text(
                json.dumps({"source": relative_source, "query_rows": rows}, indent=2) + "\n",
                encoding="utf-8",
            )
            output_path.mkdir()  # atomic gate — raises FileExistsError if concurrent
            for path in staging.iterdir():
                path.rename(output_path / path.name)
    return output_path
```

This also avoids the confusing behavior where `output_path` is resolved before the check but the resolved path is never re-checked for existence.

Utopia-V: The existing mkdir() already guards against concurrent writes. Each invocation has its own temporary directory, so a failed invocation only cleans up its own files.
The initial check also rejects dangling symlinks before resolve() follows them, so I’d prefer to keep it.

LHT129 (commented):
Reviewed commit ddc2f7b7.

The changes in this commit address the earlier review feedback on `validate_rows`:
- Replaced `type(i) is not int` with `isinstance(i, (int, np.integer))` (with explicit `bool` exclusion) to accept numpy integer types from programmatic callers.
- `validate_rows` now returns normalized `int` lists, making the output deterministic regardless of input integer subtype.
- Added `test_numpy_integer_rows_preserve_data_order_and_json_manifest` and expanded rejection test cases for `np.bool_`, `np.int64(-1)`, and `np.uint64(2**64-1)`.

The fix is correct and the test coverage is adequate. No new issues found.

LHT129 (commented):
Thanks for this PR. The changes look solid:

- **`prepare_query_split.py`**: The validation logic is thorough — dataset format checks, row index validation (now correctly accepting `numpy.integer` types per the earlier review), cross-split duplicate detection with signed-zero handling, and bounded-block I/O. The atomic `mkdir()` gate and staging-directory approach provide clean failure semantics.
- **CMakeLists.txt**: The link order fix (`vsag` before `simd`) correctly resolves the ASan ODR violation for the shared `is_inited` state.
- **Tests**: 11 test cases covering normal paths, edge cases (signed zero, duplicate content within/across splits, numpy integer types, large subsets), error paths (invalid inputs, unsupported layouts), and a native integration test with `eval_performance`. The test structure is clean and each case is self-contained.
- **CI**: The new workflow step properly sets up the environment and runs the test suite against the ASan-built binary.
- **Documentation**: Both English and Chinese docs are updated with clear usage examples and explanations.

No blocking issues found. The two earlier review comments (numpy integer type check and ODR link order) have been addressed in this revision.

LHT129 (commented):
Reviewed the latest commit (b8ed98c9). All previously raised issues have been addressed:

- The `validate_rows` function now accepts `numpy.integer` subclasses via `isinstance(i, (int, np.integer))` (resolving LHT129's suggestion).
- The CMakeLists.txt link order fix (`simd` after `vsag`) resolves the ASan ODR `is_inited` startup failure (resolving Copilot's concern).
- The TOCTOU concern is mitigated by `output_path.mkdir()` as the atomic gate, with the early `exists()`/`is_symlink()` check serving as a fast-path rejection for dangling symlinks.

The new code is well-structured with comprehensive test coverage (11 test cases covering ordering, type normalization, content validation, error paths, and native integration). No new issues found.

Issue #2614, opened 2026-08-04
[bug](hgraph): harden duplicate handling


jac0626:
### Affected area

hgraph

### Interface

cpp

### VSAG version

`v0.18.10` (`39afdfa27ed91b3b892c014a6cd172bdb25cdd6e`) and the current `0.18` branch.

### Summary

The v0.18 branch still uses `LabelTable` for duplicate tracking. It needs the v0.18-native equivalents of the duplicate deserialization hardening and KNN query fixes developed in #2258 and #2264.

Two problems remain:

1. Native `LabelTable` deserialization trusts duplicate group counts and IDs, so malformed data can index `duplicate_records_` out of bounds or publish overlapping groups.
2. Iterator duplicate expansion, entry-point handling, and final drain need to be consistent with standard and parallel KNN. Unreturned duplicate members need pending state independent of the ef-limited discard heap, with the same per-group limit and filter semantics.

### Expected behavior

- Keep the existing v0.18 native serialization format unchanged.
- Reject out-of-range, repeated, overlapping, or cross-role duplicate IDs before publishing records.
- Support `max_duplicates_per_group` for standard, parallel, and iterator KNN search: `-1` is unlimited, `0` suppresses members, and positive values cap accepted members per group.
- Expand duplicate members during search, before reorder, without enqueuing duplicate-only IDs as graph nodes. Preserve unreturned members across Iterator pages and final drain; do not redesign persistence of unconsumed graph candidates.
- Apply filters before counting the per-group cap.
- Preserve default `support_duplicate` compatibility and existing RangeSearch behavior.

### Scope

Implement directly against the v0.18 `LabelTable` architecture. Do not backport the graph-level tracker refactor (#1797) or duplicate vector storage (#2244).

### References

- #2258
- #2264
- #2574
- #1822

### Multi-version tracking

The query/iterator work and the SQ construction fix are separate changes; a small SQ-fix PR alone is not an equivalent backport of #2615.

**Query / Iterator parity**

- [x] `0.18`: #2615 — merged; implemented against LabelTable.
- [ ] `1.0`: #2944 — query/iterator adaptation using the existing graph-level tracker, plus the Analyzer adjacency fix.
- [ ] `main`: #2264 — query/iterator implementation.

**Other duplicate fixes included in the 0.18 delivery**

- [x] `0.18`: SQ duplicate-candidate fix included in #2615.
- [ ] `1.0`: SQ duplicate-candidate fix — #2892, independent of #2944.
[...]

PR #2944, opened 2026-09-14
fix(hgraph): backport duplicate query handling to 1.0

jac0626:
## Summary

Backport the HGraph duplicate-query behavior from #2615 / #2264 to `1.0`, using its existing graph-level duplicate tracker.

- Add `hgraph.max_duplicates_per_group` to standard, parallel, and iterator KNN graph search: `-1` preserves unlimited expansion, `0` suppresses extra members, and positive values count members accepted by the filter, excluding the representative.
- Expand the entry-point group even when its representative is filtered out, and trim the result heap after expanding a group.
- Keep unreturned iterator duplicate members in lazily allocated pending state across pages and final drain. Expansion still happens during graph search, before reorder; this does not introduce delayed expansion or redesign graph-candidate persistence.
- Skip duplicate-only IDs when the Analyzer reads graph adjacency for degree statistics.
- Add regression coverage and document the parameter in English and Chinese.

## Scope

This is the `1.0` query/iterator equivalent, not a wholesale cherry-pick of the `0.18` implementation. The graph-level deserialization validation from #2258 is already in `1.0`. The `0.18`-specific LabelTable/v0.14 streaming-format changes and `min_distance` parameter are not copied. Duplicate vector storage (#2244), construction locking, and RangeSearch/brute-force behavior are unchanged.

The separate SQ duplicate-candidate fix remains in #2892; it is not bundled into this PR. Both PRs are tracked separately in the issue's multi-version checklist.

## Validation

- Red: on unmodified `1.0` commit `87a8160ca3f7933eb59f62f7ff7307cda12fab78`, 8 of the 11 new query regression cases failed.
- Analyzer red: initialized unused memory-IO rows expose incorrect degree statistics before the fix (incoming degree sum 5 vs. outgoing sum 2).
- Green: 35 targeted query, iterator context, parameter, basic/parallel searcher, and Analyzer tests pass (266,060 assertions). All 12 query/Analyzer regressions also pass three additional runs (377 assertions each).
- Built with `-j $(nproc)`; changed C++ files checked with clang-format 15 and clang-tidy 15; `git diff --check` passes.

This is targeted validation, not a claim that the full test suite or coverage run passed. Two existing non-finite bridge tests were excluded from the combined local Release run; the branch's existing non-finite distance handling is unchanged.

Fixes: #2614

Related: #2264, #2615, #2892.


src/impl/searcher/basic_searcher.cpp
```diff
@@ -166,13 +166,72 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
...
+        const auto group_id = graph->GetGroupId(group_member_id);
+        if (group_id != group_member_id) {
+            return;
+        }
+        int64_t duplicate_count = 0;
+        for (const auto duplicate_id : graph->GetDuplicateIds(group_id)) {
+            if (inner_search_param.max_duplicates_per_group >= 0 and
+                duplicate_count >= inner_search_param.max_duplicates_per_group) {
+                break;
+            }
```
LHT129, line 191: [suggestion] The `add_pending_duplicates` lambda checks `group_id != group_member_id` and returns early, while `add_duplicate_results` in the non-iterator path passes `group_member_id` directly to `GetDuplicateIds` without this check. This asymmetry means the iterator path only expands duplicates when visiting the representative node, while the non-iterator path expands duplicates for any node. If `GetDuplicateIds` returns the same group regardless of which member is queried (i.e., it internally resolves to the representative), then the `group_id` check in `add_pending_duplicates` is redundant. If `GetDuplicateIds` requires the representative ID, then `add_duplicate_results` should also resolve the group representative first.

Please clarify the intended semantics and align both paths.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -181,6 +240,7 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
```
LHT129, line 180: [note] In the entry-point initialization path, `add_pending_duplicates(dist, ep)` is called unconditionally (position 74), but the original `lower_bound` update was also moved inside the `if (not top_candidates->Empty())` guard. If `ep` is filtered out and its duplicates are added via `add_pending_duplicates` but then all evicted by `shrink_top_candidates`, `top_candidates` may become empty while `lower_bound` retains a stale value. Consider whether `lower_bound` should be reset to a sentinel (e.g. `std::numeric_limitsts<float>::max()`) when `top_candidates` is empty after this block, to prevent the subsequent search loop from using an incorrect lower bound.

src/impl/searcher/parallel_searcher.cpp
```diff
@@ -171,6 +171,28 @@ ParallelSearcher::search_impl(const GraphInterfacePtr& graph,
...
+            inner_search_param.max_duplicates_per_group == 0) {
+            return;
+        }
+
+        if (not is_result_distance_eligible(duplicate_dist)) {
+            return;
+        }
+
+        int64_t duplicate_count = 0;
+        for (const auto duplicate_id : graph->GetDuplicateIds(group_member_id)) {
```
LHT129, line 185: [suggestion] The `add_duplicate_results` lambda in `parallel_searcher.cpp` (and the equivalent in `basic_searcher.cpp` non-iterator path) calls `graph->GetDuplicateIds(group_member_id)` directly without resolving the group representative first. This differs from `add_pending_duplicates` which checks `graph->GetGroupId(group_member_id) != group_member_id` and returns early if the node is not the representative.

If `GetDuplicateIds` is only meaningful when called with the representative ID, both `add_duplicate_results` implementations should resolve the group representative first (e.g., via `graph->GetGroupId`). Otherwise, if `GetDuplicateIds` handles non-representative IDs gracefully, the check in `add_pending_duplicates` is unnecessary and can be removed for consistency.

See the related comment on `basic_searcher.cpp` for the same issue.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -118,10 +118,20 @@ HGraph::KnnSearch(const DatasetPtr& query,
...
+            for (const auto& [pending_id, pending_dist] : *pending_duplicates) {
+                if (iter_filter_ctx->CheckPoint(pending_id)) {
+                    search_result->Push(pending_dist, pending_id);
+                }
+            }
+        }
         while (!iter_filter_ctx->Empty()) {
             uint32_t cur_inner_id = iter_filter_ctx->GetTopID();
             float cur_dist = iter_filter_ctx->GetTopDist();
-            search_result->Push(cur_dist, cur_inner_id);
```
LHT129, line 124: [note] In the `is_last_filter` drain path, pending duplicates are pushed to `search_result` before the discard heap is drained (lines 121-128), and pending-duplicate entries in the discard heap are skipped (lines 133-135). After both loops complete, the `pending_duplicates_` map still contains the drained entries. While `SetPoint` clears individual entries when they are later marked as visited, the map itself is not cleared after the final drain. This is harmless for correctness since the iterator context is typically discarded after the last page, but if the context were reused (e.g., for a new query with the same iterator), stale pending entries could interfere. Consider calling a `ClearPendingDuplicates()` or resetting `pending_duplicates_` after the final drain for robustness.

src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -233,6 +233,21 @@ HGraphSearchParameters::FromJson(const std::string& json_string) {
             fmt::format("brute_force_threshold({}) must in range[0.0, 1.0]",
                         obj.brute_force_threshold));
     }
+    if (params[INDEX_TYPE_HGRAPH].Contains(HGRAPH_PARAMETER_MAX_DUPLICATES_PER_GROUP)) {
+        const auto& max_duplicates_json =
+            params[INDEX_TYPE_HGRAPH][HGRAPH_PARAMETER_MAX_DUPLICATES_PER_GROUP];
+        CHECK_ARGUMENT(max_duplicates_json.IsNumberInteger(),
+                       "max_duplicates_per_group must be an integer");
+        if (max_duplicates_json.IsNumberUnsigned()) {
+            CHECK_ARGUMENT(max_duplicates_json.GetUint64() <=
```
LHT129, line 242: [suggestion] The validation for `max_duplicates_per_group` checks `IsNumberUnsigned()` and validates the uint64 value against `int64_t::max()`, then calls `GetInt()` which returns `int64_t`. However, if the JSON value is a negative integer (e.g., `-1`), `IsNumberUnsigned()` returns false and the range check is skipped, which is correct. But if the JSON value is a large unsigned integer just below `int64_t::max()`, `GetInt()` should handle it correctly since the range check ensures it fits in `int64_t`. This logic is sound, but consider adding a comment explaining why the unsigned path is needed (to catch large unsigned values that would overflow `int64_t`), as the flow is not immediately obvious to a reader.

LHT129 (commented):
I noticed one additional issue beyond the existing review comments:

**Range search path missing `max_duplicates_per_group` propagation**

In `SearchWithRequest()`, the range-search branch (line 489) sets `consider_duplicate = true` but does **not** propagate `max_duplicates_per_group` from `params`. The KNN branch (line 504) correctly sets both:

```cpp
// Range search (line 488-494)
search_param.consider_duplicate = true;
// NOTE: max_duplicates_per_group is NOT set, stays at default -1 (unlimited)

// KNN search (line 503-504)  
search_param.consider_duplicate = true;
search_param.max_duplicates_per_group = params.max_duplicates_per_group;
```

This means:
1. Range search now expands duplicates (a behavioral change from before this PR, since `consider_duplicate` was already `true` pre-PR), but always uses unlimited expansion regardless of the user-specified limit.
2. The test `HGraph duplicate limit does not change range-search expansion` passes `max_duplicates_per_group=0` and `-1` and observes identical results — because the parameter is never read by the range-search code path.

If this is intentional (range search should always expand all duplicates regardless of the limit), consider adding a comment to make it explicit. If the limit should apply to range search as well, add `search_param.max_duplicates_per_group = params.max_duplicates_per_group;` in the range-search block.

PR #2955, opened 2026-09-14
ci: separate PR ASan builds from performance benchmarks

### LINKED ISSUE #2922: [improve](ci): move ASan build benchmarks out of the PR critical path
### Kind of improvement
perf

### Affected module
not applicable

### Affected area
area/ci

### Summary
Move the X86 ASan clean/warm/no-op build benchmark into a dedicated manual/scheduled workflow, and retain a single normal ASan build in PR CI.

The current PR workflow invokes the build metrics collector with `--clear-ccache`, making every applicable PR pay for a cold build and additional benchmark rebuilds.

**Prerequisite:** #2899, which addresses trustworthy dependency metrics and no-op measurement. This issue should be implemented after that PR merges, without duplicating its collector changes.

Proposed scope:

- Replace the PR collector invocation with one `make asan COMPILE_JOBS=3`.
- Preserve existing sanitizer/compiler settings, system OpenBLAS, tests, examples, tools, changed-source lint, required check identities, and build artifact contracts.
- Move benchmark timers, collector metadata, summaries, and diagnostic artifacts into a dedicated `workflow_dispatch`/scheduled workflow.
- Use isolated benchmark compiler caches, read-only permissions, credential-free checkout, restore-only dependency archives, and safe manual ref handling.
- Add workflow regression tests and English/Chinese documentation.

### Expected benefit
Remove repeated benchmark builds from the PR critical path while preserving functional validation and independently available build-performance measurements.

Historical samples spent approximately **29.7–35.6 minutes** in the two additional phases. These are observed costs, not a guaranteed future speedup; the no-op correction in #2899 will independently change that baseline. Benefits must not be double-counted.

### Benchmark evidence
Two successful PR CI runs used GCC 11.4.0 and three parallel build jobs:

| Phase | Run 34179906667 | Run 34115637924 |
|---|---:|---:|
| Configure | 4.413 s | 4.318 s |
| Cold build | 4612.629 s | 3647.514 s |
| Warm ccache rebuild | 1066.005 s | 886.221 s |
| Reported “no-op” build | 1070.132 s | 894.449 s |

These measurements predate #2899. The reported no-op phase was not a genuine no-op: the collector deleted `.ninja_log` before each phase and both samples repeated the full build scheduling.

No post-change CI speedup has been measured yet.

### References
- Historical configuration: [`pr-ci.yml:128–169` at 30c217bc](https://github.com/antgroup/vsag/blob/30c217bc/.github/workflows/pr-ci.yml#L128-L169)
- Historical collector: [`collect_build_metrics.py:235–279` at 30c217bc](https://github.com/antgroup/vsag/blob/30c217bc/scripts/collect_build_metrics.py#L235-L279)
- https://github.com/antgroup/vsag/actions/runs/34179906667
- https://github.com/antgroup/vsag/actions/runs/34115637924
- Prerequisite: https://github.com/antgroup/vsag/pull/2899

### Additional context
Acceptance criteria:

- PR X86 performs one normal ASan build without invoking the benchmark loop or clearing ccache.
- Existing applicable lint, test, example, and compatibility checks remain enabled and pass.
- The independent benchmark records the actual tested commit, toolchain, cache semantics, phase timings, and failure diagnostics.
- Workflow tests verify trigger boundaries, permissions, cache policies, and unchanged PR artifact consumers.
- Persistent compiler-cache prewarming, PCH/sloppiness tuning, larger runners, compiler flag changes, and dependency binary caching are out of scope.

## Related issues
- #2893 — Establish trustworthy dependency build metrics; prerequisite scope.
- #2771 — Cache prebuilt third-party dependencies for cold builds.
- #2555 — Move PR ASan builds to host and prefetch third-party sources.
- #2798 — Reduce flatten_interface.cpp build and lint time.
- #2818 — Composable storage access architecture; duplicate-search match, different scope.
- #2518 — Disabled-reasoning overhead threshold; duplicate-search match, different scope.

---
_Drafted with AI assistant: DeepSeek-Harness:gpt-6-astra_

LHT129:
## Change Type

- [x] Improvement/Refactor
- [x] CI/Build/Infra

## Summary

Separate normal PR X86 ASan validation from the clean/warm/no-op build-performance benchmark.

- Keep one normal ASan build in PR CI, preserving build options, lint, artifact contracts and downstream tests.
- Intentionally change PR X86 dependency archives from restore-and-save to restore-only, matching #2899's trust policy: PR jobs do not publish dependency caches. This is separate from the benchmark's restore-only policy and does not add persistent compiler caching.
- Add a separate manual/scheduled benchmark with isolated compiler cache, restore-only dependency archives, exact revision reporting and read-only permissions.
- Preserve the collector ownership boundary with #2899; do not cherry-pick or duplicate its implementation.
- Document supported benchmark revisions and add workflow/guard regression tests.

Closes: #2922
Related prerequisite: #2899 / #2893

## Draft / dependency boundary

This PR is being developed from current main before #2899 merges, per maintainer direction. Current main's collector is legacy and does not provide verified no-op metrics. The independent benchmark must fail clearly on unsupported collector revisions rather than run the old loop and claim valid no-op measurements. Rebase and compatibility verification after #2899 lands remain required before marking this ready.

Persistent compiler-cache prewarming, PCH/sloppiness, compiler flags, parallelism changes, dependency binary caching and production C++ changes are out of scope.

## Validation

- `python3 -m unittest tests/scripts/test_build_benchmark_workflow.py tests/scripts/test_collect_build_metrics.py`: 10 tests passed, including actual inline preflight/postflight execution and real temporary Git history/ref rejection.
- `python3 -m unittest discover -s tests/scripts -p '*test.py'`: 21 existing tests passed on unchanged main baseline.
- `python3 -m unittest discover -s scripts/linters -p 'test*.py'`: 4 passed.
- `make test-cmake COMPILE_JOBS=3`: 7 focused CMake contract suites passed.
- Pinned actionlint 1.7.7 (release SHA256 checked), `-shellcheck=''` on both workflows: passed. ShellCheck is unavailable; no claim of ShellCheck validation.
- `node .github/scripts/check-docs.mjs` and `git diff --check`: passed.
- `make fmt` with clang-format 15: passed; no unrelated formatting changes.
- Actual #2899 `9073806` collector snapshot passes the new preflight without importing/executing it. Its 28 prerequisite tests passed in an isolated snapshot (with existing ResourceWarnings); these are not a full VSAG benchmark.
- Confirmed byte-identical workflow suffix from changed-source lint through downstream jobs versus main, and unchanged collector/Makefile/CMake/daily/push workflows.

Full ASan builds, upstream hosted dispatch, performance comparison and downstream CI are still pending. The benchmark intentionally rejects current legacy main until schema 3 collector support lands. No full C++ test/coverage run is claimed for this workflow-only change.

## Compatibility and risk

- No production API/ABI changes. No modifications to the collector, CMake/Makefile compiler settings, push/daily build workflows or downstream test targets.
- CI behavior intentionally changes: PR builds once; benchmark observations move to manual/weekly runs. Unmerged PR refs are not supported by the conservative benchmark target selector.
- Main-history code is trusted as repository code, not sandboxed merely by `contents: read`. No compiler-cache persistence is added.
- Medium workflow integration risk; keep this draft until the collector prerequisite lands and hosted CI is verified. Revert the workflow migration independently if needed; do not restore misleading no-op measurement claims.
- English and Chinese building guides updated.

## Performance evidence

Historical runs 34179906667 and 34115637924 spent approximately 29.7–35.6 minutes on the two extra warm/nominal-no-op phases combined. These are historical observations, not promised post-change improvements; #2899's no-op correction changes that baseline independently. No additive/double-counted speedup claim.


LHT129:
Integration rehearsal against prerequisite #2899 head fde6bfd passed 34 focused tests after applying the expected post-merge wiring in a disposable snapshot: retain both dependency/collector helper test invocations in PR CI, and move the dependency-cache-write-policy assertion to the new benchmark workflow. No prerequisite code was imported into this branch. The rehearsal also caught an overly broad assertion in our own test that rejected the collector unit-test filename; fixed in f3ee36dc to forbid benchmark execution rather than unit tests. The task branch still passes its 10 focused tests. Real hosted CI and final post-merge integration remain pending.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,258 @@
...
+          PY
+      - name: Prepare trusted Ubuntu host
+        # Fixed workflow-owned package list matches PR OpenBLAS setup, not target-ref sudo scripts.
+        run: |
+          sudo rm -rf /opt/hostedtoolcache
+          sudo apt-get update
+          sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y \
+            ccache cmake g++ gcc gfortran lcov libaio-dev libcurl4-openssl-dev \
+            liblapacke-dev libomp-15-dev libopenblas-dev liburing-dev make \
+            ninja-build python3-dev clang-tidy-15 time
```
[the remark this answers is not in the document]
LHT129: Checked the Ubuntu 22.04 package inventory: both clang-tidy-15 and libomp-15-dev are available as version 15.0.7 in the official jammy security packages (https://packages.ubuntu.com/jammy/clang-tidy-15 and https://packages.ubuntu.com/jammy/libomp-15-dev). The hosted ubuntu-22.04 image also lists clang-tidy 15.0.7: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md . Keeping the explicit -15 packages preserves the repository requirement and parity with the existing PR host setup; downgrading to unsuffixed packages or adding an unnecessary external LLVM repository would be less appropriate. Actual benchmark hosted execution remains pending and is not claimed by these inventory checks.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,258 @@
...
+        uses: actions/checkout@v4
+        with:
+          repository: antgroup/vsag
+          ref: main
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Resolve approved benchmark commit
+        id: target
+        env:
+          REQUESTED_REF: ${{ inputs.ref || 'main' }}
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Changed the fallback to github.event.inputs.ref || 'main', so scheduled runs read the optional event payload explicitly. Kept the input in an environment variable rather than interpolating it into shell source; workflow contract checks cover the expression. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,258 @@
...
+        run: |
+          set -euo pipefail
+          base_sha=$(git rev-parse HEAD)
+          if [[ "$REQUESTED_REF" == main ]]; then
+            sha="$base_sha"
+          elif [[ "$REQUESTED_REF" =~ ^[0-9a-fA-F]{40}$ ]]; then
+            if [[ "$(git cat-file -t "$REQUESTED_REF")" != commit ]]; then
+              echo '::error::Benchmark SHA must identify a commit object, not a tag, tree or blob.'
+              exit 1
+            fi
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Added an explicit object lookup with captured exit status and suppressed raw Git stderr, followed by a clear workflow error for an unknown SHA. Existing commit-type and ancestry checks still reject tag/tree objects and unapproved history; regression cases cover an absent 40-hex SHA. Validation: 31 script tests, actionlint, docs check and diff check passed.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,162 @@
...
+                bad = copy.deepcopy(report)
+                bad['phases'][0][key] = value
+                variants.append(bad)
+            variants.extend([dict(report, schema_version=1), dict(report, status='failed'),
+                             dict(report, phases=[]), dict(report, commit_sha='b' * 40)])
+            for candidate in [report, *variants]:
+                (root / 'build-metrics/build-metrics.json').write_text(json.dumps(candidate), encoding='utf-8')
+                result = self.run_python('Verify benchmark report', root)
+                self.assertEqual(result.returncode == 0, candidate is report, result.stderr)
+            path = root / 'build-metrics/build-metrics.json'
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Replaced the object-identity expectation with explicit candidate/expected-result pairs, so equivalent valid reports do not depend on Python identity. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/pr-ci.yml
```diff
@@ -117,83 +117,42 @@ jobs:
...
           g++ --version
-      - name: Start Dependency Preparation Timer
-        id: dependency-preparation-start
-        run: echo "started_ns=$(date +%s%N)" >> "$GITHUB_OUTPUT"
       - name: Prepare FetchContent Sources
         uses: ./.github/actions/prepare-fetchcontent
         with:
           include-tests: "true"
           include-tools: "true"
       - name: Cache ExternalProject Archives
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Renamed the PR step to Restore ExternalProject Archives to make restore-only behavior explicit in CI logs. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,258 @@
...
+                         'Revalidate this gate if the collector API changes.')
+              Path('build-metrics/preflight.txt').write_text(message + '\n', encoding='utf-8')
+              with open(os.environ['GITHUB_STEP_SUMMARY'], 'a', encoding='utf-8') as summary:
+                  summary.write('**' + message + '**\n')
+              raise SystemExit(message)
+          PY
+      - name: Prepare trusted Ubuntu host
+        # Fixed workflow-owned package list matches PR OpenBLAS setup, not target-ref sudo scripts.
+        run: |
+          sudo rm -rf /opt/hostedtoolcache
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Removed broad hostedtoolcache deletion from the new benchmark workflow. Host preparation no longer removes runner-managed tool installations. Existing PR cleanup behavior is outside this targeted change and remains unchanged. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/build_performance.yml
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Added an explicit handled object lookup and friendly error for unknown SHAs, with a missing-object regression. Commit-type and ancestry checks remain fail-closed. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,258 @@
...
+          source = Path('scripts/collect_build_metrics.py')
+          try:
+              tree = ast.parse(source.read_text(encoding='utf-8'))
+              schemas = [value for node in ast.walk(tree) if isinstance(node, ast.Dict)
+                         for key, value in zip(node.keys, node.values)
+                         if isinstance(key, ast.Constant) and key.value == 'schema_version']
+              schema3 = bool(schemas) and all(
+                  isinstance(value, ast.Constant) and type(value.value) is int and value.value == 3
+                  for value in schemas
+              )
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. Narrowed capability inspection to the report assignment in Collector.write_reports and the true_noop assignment in capture_build. Unrelated dictionaries no longer invalidate an otherwise supported collector. Added acceptance/rejection fixtures; this remains only a static compatibility preflight, with actual output validated separately. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/pr-ci.yml
```diff
@@ -117,83 +117,42 @@ jobs:
...
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: .ci-downloads
           key: external-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('extern/**/*.cmake') }}
           restore-keys: |
             external-${{ runner.os }}-${{ runner.arch }}-
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. The PR dependency archive restore-only policy is intentional and aligns with #2899: untrusted PR jobs do not save dependency caches. Clarified this separately from benchmark cache behavior in the bilingual docs and PR summary. No persistent compiler-cache mechanism is added. Validation: 31 script tests, actionlint, docs check and diff check passed.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+ROOT = Path(__file__).resolve().parents[2]
+PR = (ROOT / '.github/workflows/pr-ci.yml').read_text(encoding='utf-8')
+BENCH = (ROOT / '.github/workflows/build_performance.yml').read_text(encoding='utf-8')
+
+
+def step(workflow, name):
+    marker = '      - name: ' + name + '\n'
+    return workflow.split(marker, 1)[1].split('\n      - ', 1)[0]
```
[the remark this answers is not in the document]
LHT129: Addressed in 2c03672d. The step-extraction helper now raises an explicit assertion identifying the missing step and workflow rather than leaking IndexError. Regression coverage checks this diagnostic. Validation: 31 script tests, actionlint, docs check and diff check passed.

.github/workflows/pr-ci.yml
```diff
@@ -117,83 +117,42 @@ jobs:
...
       - name: Prepare FetchContent Sources
         uses: ./.github/actions/prepare-fetchcontent
         with:
           include-tests: "true"
           include-tools: "true"
-      - name: Cache ExternalProject Archives
-        uses: actions/cache@v4
+      - name: Restore ExternalProject Archives
+        uses: actions/cache/restore@v4
         with:
```
LHT129, line 139: [note] The comment says "No persistent compiler cache is used", but the Makefile's `asan` target enables `-DENABLE_CCACHE=ON`, and this step no longer sets `CCACHE_DIR`. Without an explicit `CCACHE_DIR`, ccache defaults to `~/.ccache` on the runner, which may retain objects from previous workflow runs. This contradicts the stated intent.

If the goal is truly no persistent compiler cache, consider either:
- Setting `CCACHE_DIR` to a temporary location that is cleaned up after the build, or
- Passing `CCACHE_DISABLE=1` or `-DENABLE_CCACHE=OFF` to override the Makefile default.

Otherwise, update the comment to reflect that ccache is active but using the default cache location.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,195 @@
...
+
+class WorkflowContractTest(unittest.TestCase):
+    def test_single_pr_build_and_preserved_options(self):
+        with self.assertRaisesRegex(AssertionError, "Missing step 'Missing step' in pr-ci.yml"):
+            step(PR, 'Missing step')
+        self.assertIn('uses: actions/cache/restore@v4', step(PR, 'Restore ExternalProject Archives'))
+        x86 = PR.split('  build-asan-x86:\n', 1)[1].split('  build-asan-macos-arm64:', 1)[0]
+        self.assertEqual(x86.count('make asan COMPILE_JOBS=3'), 1)
+        # Collector unit tests remain valid PR checks; only executing the benchmark is forbidden.
+        for forbidden in ('python3 scripts/collect_build_metrics.py', '--clear', 'build-performance-metrics', 'COMPILER_CACHE_KEY'):
```
LHT129: [suggestion] The forbidden substring `'--clear'` is too broad and may cause false-positive test failures. It matches any flag containing `--clear`, such as `--clear-logs`, `--clear-temps`, or `--clear-state`. Consider matching the exact flag that was removed from the PR CI workflow, e.g. `'--clear-ccache'`, so the test only catches the specific benchmark-collector flag it intends to guard against.

Current:
```python
for forbidden in ('python3 scripts/collect_build_metrics.py', '--clear', 'build-performance-metrics', 'COMPILER_CACHE_KEY'):
```

Suggested:
```python
for forbidden in ('python3 scripts/collect_build_metrics.py', '--clear-ccache', 'build-performance-metrics', 'COMPILER_CACHE_KEY'):
```

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured and the separation of PR ASan builds from performance benchmarks is cleanly implemented.

1. **Clean separation**: The new `build_performance.yml` workflow is isolated from PR CI with its own concurrency group, permissions, and cache directories. PR CI is simplified to a single `make asan` with ccache diagnostics.

2. **Strong security posture**: Read-only permissions, no persisted credentials, SHA/ancestry validation rejecting PR refs, and a fixed workflow-owned package list instead of running target-ref scripts.

3. **Defense in depth**: The preflight AST check gates on schema 3 capability without executing the collector, and the post-collection verification enforces `true_noop: true` and zero build edges even if the collector reports success with a false no-op.

4. **Comprehensive testing**: The test suite covers workflow contracts, security/isolation properties, preflight rejection of legacy collectors, report gate validation with many edge cases, and real Git history ref gate testing including injection attempts.

5. **Good documentation**: Both English and Chinese building docs are updated with clear explanations of the new workflow behavior, security boundaries, and the "cold" definition.

The existing review comments have already addressed the substantive issues I would have raised. I have no additional blocking or suggestion-level comments to add.

LHT129 (commented):
## Review Summary

After thorough review of all 5 changed files (520 additions, 51 deletions), this PR is well-constructed with comprehensive test coverage. The separation of PR ASan builds from performance benchmarks is cleanly implemented.

### Issues already flagged by other reviewers (still relevant):

1. **`tests/scripts/test_build_benchmark_workflow.py:28`** — `python_step()` raises an opaque traceback when the heredoc format changes. Consider adding an explicit `AssertionError` similar to `step()`.

2. **`.github/workflows/pr-ci.yml:135`** — The comment says "No persistent compiler cache is used" but `CCACHE_DIR` is no longer set, so ccache defaults to `~/.ccache` on the runner. Either set `CCACHE_DIR` to a temp location or update the comment.

3. **`tests/scripts/test_build_benchmark_workflow.py:39`** — The forbidden substring `'--clear'` is too broad and may cause false positives. Consider using `'--clear-ccache'` instead.

### Already resolved:

- `inputs.ref` → `github.event.inputs.ref` (fixed)
- `git cat-file -t` error handling with `2>/dev/null` (fixed)
- `sudo rm -rf /opt/hostedtoolcache` removed (fixed)

### Note on Copilot comment about `clang-tidy-15`:

The comment suggesting `clang-tidy-15` and `libomp-15-dev` are unavailable on Ubuntu 22.04 default repos is incorrect — these packages are available in jammy and are already used successfully by the existing PR CI.

### Overall assessment:

The workflow design is sound: read-only permissions, no persisted credentials, fixed package lists, SHA ancestry validation, conservative preflight gating, and thorough contract tests. No new substantive issues found.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,195 @@
...
+    marker = '      - name: ' + name + '\n'
+    workflow_name = 'pr-ci.yml' if workflow == PR else 'build_performance.yml' if workflow == BENCH else 'provided workflow'
+    if marker not in workflow:
+        raise AssertionError(f'Missing step {name!r} in {workflow_name}; check its name and indentation')
+    return workflow.split(marker, 1)[1].split('\n      - ', 1)[0]
+
+
+def python_step(name):
+    body = step(BENCH, name).split("python3 - <<'PY'\n", 1)[1]
+    return textwrap.dedent(body.rsplit('\n          PY', 1)[0])
```
LHT129: [suggestion] The `python_step()` helper at module level will raise `IndexError` if a benchmark step changes its inline Python delimiter (e.g., from `python3 - <<'PY'` to `python - <<'EOF'` or different indentation). This produces a confusing test failure unrelated to the contract being checked.

Consider adding an explicit guard similar to the `step()` helper:

```python
def python_step(name):
    body = step(BENCH, name)
    marker = "python3 - <<'PY'\n"
    if marker not in body:
        raise AssertionError(f"Step {name!r} missing expected inline Python marker; check delimiter and indentation")
    return textwrap.dedent(body.split(marker, 1)[1].rsplit('\n          PY', 1)[0])
```

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+      ref:
+        description: 'main or a full 40-character commit SHA reachable from main (no PR refs)'
+        required: false
+        default: main
+        type: string
+  schedule:
+    - cron: '23 3 * * 1'
+
+permissions:
+  contents: read
```
LHT129, line 15: [suggestion] The workflow declares only `contents: read` but uses `actions/cache/restore@v4` (requires `actions: read`) and `actions/upload-artifact@v5` (requires `actions: write`). If the repository ever moves to restrictive default permissions, cache restore and artifact upload will fail with 403.

Consider adding the minimum required permissions explicitly:

```yaml
permissions:
  contents: read
  actions: read
```

(Note: `actions/upload-artifact@v5` needs `actions: write`, but since `pr-ci.yml` already uses artifact upload successfully with the repo's current defaults, the `actions: write` scope is implicitly available. Adding `actions: read` explicitly covers the cache restore requirement and documents the intent.)

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+              raise SystemExit(message)
+          PY
+      - name: Prepare trusted Ubuntu host
+        # Fixed workflow-owned package list matches PR OpenBLAS setup, not target-ref sudo scripts.
+        run: |
+          sudo apt-get update
+          sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y \
+            ccache cmake g++ gcc gfortran lcov libaio-dev libcurl4-openssl-dev \
+            liblapacke-dev libomp-15-dev libopenblas-dev liburing-dev make \
+            ninja-build python3-dev clang-tidy-15 time
```
LHT129, line 133: [note] The `Prepare trusted Ubuntu host` step installs `clang-tidy-15` and `libomp-15-dev`. On ubuntu-22.04, these version-suffixed LLVM packages are not in the default repositories — they require an additional apt source (e.g., `apt.llvm.org`). If the runner image doesn't pre-configure this source, `apt-get install` will fail.

The `pr-ci.yml` workflow avoids this by using `scripts/deps/install_deps_ubuntu.sh` which handles LLVM repository setup. Consider either:
1. Using the same `install_deps_ubuntu.sh` script (with appropriate flags to limit to the needed packages), or
2. Adding the LLVM apt repository explicitly before the `apt-get install` command.

This is currently masked if the ubuntu-22.04 runner image pre-installs these packages or pre-configures the LLVM repo.

LHT129 (commented):
Overall this is a well-structured PR that cleanly separates PR ASan validation from build-performance benchmarking. The workflow design is thorough: conservative security posture (read-only contents, no persisted credentials), careful SHA validation with ancestry checks, schema 3 preflight gating, and comprehensive contract tests with real git history exercises.

**Summary of findings:**

1. **`python_step()` error handling** (`tests/scripts/test_build_benchmark_workflow.py:28`): The helper will raise `IndexError` with a confusing message if the inline Python delimiter changes. An explicit guard (like the `step()` helper already has) would improve diagnosability.

2. **Missing `actions` permission** (`.github/workflows/build_performance.yml:15`): The workflow uses `actions/cache/restore@v4` and `actions/upload-artifact@v5` but only declares `contents: read`. Adding `actions: read` explicitly documents the cache restore requirement.

3. **Versioned LLVM packages** (`.github/workflows/build_performance.yml:132`): `clang-tidy-15` and `libomp-15-dev` may not be available from default ubuntu-22.04 repos without an LLVM apt source.

The three Copilot comments that were addressed in this revision (SHA validation error handling, `inputs.ref` → `github.event.inputs.ref`) are correctly resolved. The test coverage is excellent — 10 workflow contract tests plus real git history validation for the ref gate.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+
+          now = time.time_ns()
+          elapsed = (now - int(os.environ['STARTED_NS'])) / 1_000_000_000
+          archive = (now - int(os.environ['ARCHIVE_STARTED_NS'])) / 1_000_000_000
+          print(f'seconds={elapsed:.3f}')
+          print(f'archive_seconds={archive:.3f}')
+          PY
+      - name: Collect isolated cold warm and no-op metrics
+        env:
+          BUILD_BASE_SHA: ${{ steps.target.outputs.base_sha }}
```
LHT129, line 191: [note] `g++ -dumpfullversion -dumpversion` passes both flags to GCC, which outputs the full version and then the major version on separate lines. `COMPILER=$(...)` captures both lines, embedding a newline in the compiler identifier string used for `--compiler "gcc-$COMPILER"`. This produces a multi-line argument value that may cause formatting issues in the collector report.

Consider using only one flag:

```bash
COMPILER=$(g++ -dumpversion)
```

or if the full version is preferred:

```bash
COMPILER=$(g++ -dumpfullversion)
```

.github/workflows/build_performance.yml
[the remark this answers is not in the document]
LHT129: Verified the action implementation: normal same-run artifact upload and cache restore authenticate using ACTIONS_RUNTIME_TOKEN, not GITHUB_TOKEN actions scopes. See https://github.com/actions/toolkit/blob/main/packages/artifact/src/internal/shared/config.ts and https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheHttpClient.ts . Keeping contents: read avoids unnecessary privilege; actions:write is not needed for this workflow. REST access to other workflow runs is a different operation and is not used here.

LHT129: Verified both cache and artifact toolkit authentication use ACTIONS_RUNTIME_TOKEN for these operations, not GITHUB_TOKEN actions permissions: https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheHttpClient.ts and https://github.com/actions/toolkit/blob/main/packages/artifact/src/internal/shared/config.ts . Also, specifying contents:read sets unspecified GITHUB_TOKEN permissions to none, so success is not due to implicitly inherited actions:write. Keeping the least-privilege permissions unchanged.

.github/workflows/build_performance.yml
LHT129: Checked official Ubuntu jammy packages: https://packages.ubuntu.com/jammy/clang-tidy-15 and https://packages.ubuntu.com/jammy/libomp-15-dev both provide 15.0.7. The hosted ubuntu-22.04 inventory also lists LLVM15. The current install_deps_ubuntu.sh simply apt-installs these packages and does not add an LLVM repository. Therefore retaining LLVM15 and the trusted inline apt bootstrap preserves parity without executing selected-revision scripts through sudo; no extra apt repository is needed.

.github/workflows/build_performance.yml
LHT129: Verified on the task Ubuntu host: g++ -dumpfullversion -dumpversion outputs one line, 11.4.0, not full and major versions on separate lines. The combined flags are an established full-version/fallback invocation and do not introduce a newline here. Keeping this unchanged avoids reducing compiler metadata precision.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,271 @@
...
+          sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y \
+            ccache cmake g++ gcc gfortran lcov libaio-dev libcurl4-openssl-dev \
+            liblapacke-dev libomp-15-dev libopenblas-dev liburing-dev make \
+            ninja-build python3-dev clang-tidy-15 time
+          {
+            g++ --version
+            cmake --version
+            ninja --version
+            ccache --version
+            printf '%s\n' 'jobs=3; Ninja; compile_commands=ON; system OpenBLAS=ON; examples/tools/tests=ON; Sanitize ASan/UBSan'
```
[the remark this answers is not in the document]
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Metadata now records actual workflow environment values and the exact shell-escaped collector argv executed through the same Bash array. BENCHMARK_JOBS is shared with --jobs. Regressions change env values and replay the recorded command to verify argument identity. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,195 @@
...
+PR = (ROOT / '.github/workflows/pr-ci.yml').read_text(encoding='utf-8')
+BENCH = (ROOT / '.github/workflows/build_performance.yml').read_text(encoding='utf-8')
+
+
+def step(workflow, name):
+    marker = '      - name: ' + name + '\n'
+    workflow_name = 'pr-ci.yml' if workflow == PR else 'build_performance.yml' if workflow == BENCH else 'provided workflow'
+    if marker not in workflow:
+        raise AssertionError(f'Missing step {name!r} in {workflow_name}; check its name and indentation')
+    return workflow.split(marker, 1)[1].split('\n      - ', 1)[0]
```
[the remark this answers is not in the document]
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Replaced exact whitespace splitting with an indentation-aware stdlib block extractor. Regression fixtures cover alternate indentation, comments, and unnamed uses steps without requiring PyYAML. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Inline Python extraction now explicitly validates both opening and closing heredoc markers with targeted AssertionError diagnostics. Missing-marker regressions cover both cases. Task-focused 14 tests, actionlint and docs checks passed after rebase.

.github/workflows/pr-ci.yml
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Clarified the comment and bilingual docs: ccache remains enabled using its default directory on a disposable GitHub-hosted VM; no compiler cache is restored or saved across jobs. No cache disabling or new persistence mechanism is introduced. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Narrowed the forbidden flag to --clear-ccache rather than the overly broad --clear substring. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Added targeted AssertionError diagnostics for missing inline Python opener/closer, with negative regression fixtures instead of leaking IndexError. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
```diff
@@ -0,0 +1,195 @@
...
+    body = step(BENCH, name).split("python3 - <<'PY'\n", 1)[1]
+    return textwrap.dedent(body.rsplit('\n          PY', 1)[0])
+
+
+class WorkflowContractTest(unittest.TestCase):
+    def test_single_pr_build_and_preserved_options(self):
+        with self.assertRaisesRegex(AssertionError, "Missing step 'Missing step' in pr-ci.yml"):
+            step(PR, 'Missing step')
+        self.assertIn('uses: actions/cache/restore@v4', step(PR, 'Restore ExternalProject Archives'))
+        x86 = PR.split('  build-asan-x86:\n', 1)[1].split('  build-asan-macos-arm64:', 1)[0]
```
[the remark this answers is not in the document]
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Replaced brittle job-header splits with a defensive indentation-aware job block extractor. Missing-job diagnostics and reordered/alternate-indentation fixtures are covered. Task-focused 14 tests, actionlint and docs checks passed after rebase.

tests/scripts/test_build_benchmark_workflow.py
LHT129: Fixed in d953ac8c, rebased onto latest main c350ac85. Both heredoc delimiters are now explicitly checked and reported with the step name via AssertionError; tests cover missing opener and closer. Task-focused 14 tests, actionlint and docs checks passed after rebase.

LHT129:
Review follow-up pushed as single commit d953ac8c on freshly fetched main c350ac85; all current inline threads replied to and resolved. Task-focused 14 tests, actionlint (ShellCheck unavailable), and documentation checks pass. Broad discovery after rebase: 51/52 pass; the newly merged SIMD ownership fixture fails on this Git 2.34.1 host and reproduces identically in a pristine main archive at simd_coverage_test.py:278, so it is not introduced by this patch. CircleCI 18190 on the prior head failed with SIGSEGV in HGraph companion MCI serializes concurrent initial Add (hgraph_mci_test.cpp:222), not artifact download. No stack trace establishes a production fix yet; test and C++ implementation were untouched by this PR. New head CircleCI 18230 is pending; not claiming the crash fixed, and no test coverage has been removed.

LHT129:
CI follow-up: CircleCI 18230 on d953ac8c is now SUCCESS. Independently built an isolated archive of the original failing base 13dd2cbf and ran the exact MCI concurrent-initial-Add case 30 times; all passed (5 assertions/1 test case per run, not an empty filter). Thus no runtime reproduction or backtrace was obtained, and no production fix is claimed. Original full-suite CircleCI18190 SIGSEGV remains unexplained; isolated runs do not exclude order/concurrency/environment effects. No C++ test was removed or weakened. Latest GitHub Actions checks remain queued at the time of this update.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,277 @@
...
+          BUILD_COMMIT_SHA: ${{ steps.target.outputs.sha }}
+          DEPENDENCY_CACHE_KEY: ${{ steps.archives.outputs.cache-primary-key }}
+          DEPENDENCY_CACHE_MATCHED_KEY: ${{ steps.archives.outputs.cache-matched-key }}
+          DEPENDENCY_CACHE_STATE: ${{ steps.archives.outputs.cache-hit == 'true' && 'exact-hit' || (steps.archives.outputs.cache-matched-key != '' && 'fallback-hit' || 'miss') }}
+          DEPENDENCY_PREPARATION_SECONDS: ${{ steps.dependency-preparation.outputs.seconds }}
+          DEPENDENCY_SOURCE_PREPARATION_SECONDS: ${{ steps.sources.outputs.seconds }}
+          DEPENDENCY_CACHE_RESTORE_SECONDS: ${{ steps.dependency-preparation.outputs.archive_seconds }}
+        run: |
+          COMPILER=$(g++ -dumpfullversion -dumpversion)
+          # build/ belongs only to this fresh hosted job; keep the existing collector CLI.
```
LHT129, line 201: [suggestion] The `DEPENDENCY_CACHE_STATE` expression uses GitHub Actions `&&`/`||` short-circuit evaluation which can produce unexpected results when intermediate values are empty strings. Specifically, `steps.archives.outputs.cache-matched-key != ''` evaluates to `true` (a non-empty string), but if `cache-hit` is `'true'`, the expression short-circuits correctly. However, if `cache-hit` is `'false'` and `cache-matched-key` is `''`, the expression `'' || 'miss'` evaluates to `'miss'` only because `''` is falsy in JavaScript — this is correct but non-obvious.

Consider extracting this into a separate step with a shell one-liner for clarity:

```yaml
- name: Resolve dependency cache state
  id: cache-state
  env:
    CACHE_HIT: ${{ steps.archives.outputs.cache-hit }}
    CACHE_MATCHED_KEY: ${{ steps.archives.outputs.cache-matched-key }}
  run: |
    if [[ "$CACHE_HIT" == true ]]; then
      echo 'state=exact-hit' >> "$GITHUB_OUTPUT"
    elif [[ -n "$CACHE_MATCHED_KEY" ]]; then
      echo 'state=fallback-hit' >> "$GITHUB_OUTPUT"
    else
      echo 'state=miss' >> "$GITHUB_OUTPUT"
    fi
```

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,277 @@
...
+                      for key, value in zip(report.keys, report.values)
+                  ) for report in reports
+              )
+              noop_field = any(
+                  isinstance(node, ast.Subscript) and isinstance(node.ctx, ast.Store)
+                  and isinstance(node.value, ast.Name) and node.value.id == 'phase'
+                  and isinstance(node.slice, ast.Constant) and node.slice.value == 'true_noop'
+                  for method in methods if method.name == 'capture_build'
+                  for node in ast.walk(method)
+              )
```
LHT129, line 113: [note] The preflight AST check for `phase['true_noop'] = True` is tightly coupled to the variable name `phase` and the subscript-assignment pattern. If #2899 refactors `capture_build` to use a different local variable name (e.g. `stats['true_noop'] = True`) or a different assignment style (e.g. `setattr`), this gate will reject a valid collector. The test suite covers this well (`test_preflight_rejects_legacy_without_executing_it`), and the PR description acknowledges the trade-off. Just noting that after #2899 lands, the preflight and its tests should be re-validated against the final collector API to avoid a false-negative gate on main.

.github/workflows/build_performance.yml
```diff
@@ -0,0 +1,277 @@
...
+      VSAG_FETCHCONTENT_BASE_DIR: ${{ github.workspace }}/.ci-fetchcontent
+      VSAG_THIRDPARTY_DOWNLOAD_DIR: ${{ github.workspace }}/.ci-downloads
+      CCACHE_DIR: ${{ github.workspace }}/.benchmark-ccache
+      BENCHMARK_JOBS: '3'
+      CMAKE_GENERATOR: Ninja
+      CMAKE_EXPORT_COMPILE_COMMANDS: 'ON'
+      EXTRA_DEFINED: '-DVSAG_USE_SYSTEM_OPENBLAS=ON'
+      VSAG_ENABLE_EXAMPLES: 'ON'
+      VSAG_ENABLE_TOOLS: 'ON'
+    steps:
```
LHT129, line 37: [suggestion] `BENCHMARK_JOBS: '3'` is declared as a YAML string (quoted), while `CMAKE_GENERATOR: Ninja` and other env vars that hold non-numeric values are unquoted. This is inconsistent and could confuse readers about whether the value is intended as a string or number. In bash the distinction doesn't matter since `"$BENCHMARK_JOBS"` expands the same way, but for consistency with the rest of the file, consider removing the quotes:

```yaml
BENCHMARK_JOBS: 3
```

(If the quoting is intentional to force YAML string type for safety, the same treatment should apply to `CMAKE_GENERATOR: Ninja` and other non-numeric values.)

.github/workflows/pr-ci.yml
```diff
@@ -121,83 +121,43 @@ jobs:
         run: |
           python3 -m unittest discover -s scripts/linters -p 'test_changed_source_filter.py'
           python3 -m unittest tests/scripts/configure_git_safe_directory_test.py
+          python3 -m unittest tests/scripts/test_build_benchmark_workflow.py
       - name: Prepare Host
         run: |
           sudo env VSAG_INSTALL_INTEL_MKL=OFF bash ./scripts/deps/install_deps_ubuntu.sh
           sudo apt-get install -y clang-tidy-15 time
           ninja --version
           cmake --version
```
LHT129, line 130: [suggestion] After this PR, no workflow will save to the `external-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('extern/**/*.cmake') }}` cache key. The PR CI switches from `actions/cache@v4` (save+restore) to `actions/cache/restore@v4` (restore-only), and the new benchmark workflow also uses restore-only. Once the existing cache entries expire (7 days of no access on GitHub), every PR and benchmark build will have a permanent cache miss, re-downloading all ExternalProject archives on every run.

Consider either:
1. Adding a cache save step to the benchmark workflow (or another main-branch workflow like `daily_test.yml`) to keep the cache warm, or
2. Documenting explicitly that cache misses are expected and acceptable for PR builds.

LHT129 (commented):
I have reviewed this PR and found no blocking issues. The changes are well-structured and focused:

**Workflow separation** (`pr-ci.yml` + new `build_performance.yml`):
- PR CI now runs a single `make asan` build with `ccache --zero-stats` + `/usr/bin/time -p`, removing the `collect_build_metrics.py` invocation. Downstream test/artifact contracts are preserved.
- Dependency archives changed from `actions/cache@v4` (save+restore) to `actions/cache/restore@v4` (restore-only), consistent with the stated trust policy.
- The new benchmark workflow is manual/scheduled only, with conservative security (read-only permissions, no persisted credentials, fixed package list, SHA ancestry validation).

**Preflight gate**: The AST-based schema check correctly rejects legacy collectors without executing them. The post-collection verification enforces `true_noop: true`, `build_edges: 0`, and available Ninja telemetry — a reasonable defense-in-depth approach given that #2899 can report false no-ops without failing.

**Tests** (`test_build_benchmark_workflow.py`):
- The `block()`/`job()`/`step()`/`python_step()` helpers now have proper error messages for missing steps/delimiters, addressing the earlier `IndexError` concern.
- Contract tests cover: single PR build, preserved build options, security/isolation properties, toolchain metadata, collector command recording, preflight rejection cases, report gate validation, and real-git-history ref gating.
- The ref gate test correctly rejects tags, trees, unapproved SHAs, PR refs, and injection attempts.

**Documentation**: Both English and Chinese building guides now document the PR ASan build and standalone benchmark workflow, including the legacy no-op caveat and dispatch instructions.

No critical issues found. The previously flagged Copilot comments (inputs.ref, git cat-file error handling, hostedtoolcache cleanup, python_step IndexError, permissions) have all been addressed in the latest commit.

docs/docs/zh/src/development/building.md
```diff
@@ -24,8 +24,7 @@ Usage: make <target>
 Targets:
 help:                    ## Show the help.
 ##
-## ================ development ================
-debug:                   ## Build vsag with debug options.
+## ================ development =========debug:                   ## Build vsag with debug options.
```
LHT129: [critical] The Makefile help-target section header comments and their following target lines have been accidentally merged, losing the newline separator.

The diff shows:
```
-## ================ development ================
-debug:                   ## Build vsag with debug options.
+## ================ development =========debug:                   ## Build vsag with debug options.
```

This pattern repeats for all three section headers (development, integration, distribution). The section comment line and the target line are concatenated into a single line, which breaks the `make help` output formatting and makes the Makefile targets unparseable.

This appears to be an unintentional edit — the section headers should remain on separate lines from the targets they precede. Please restore the original formatting for these three lines.

LHT129: [critical] Same issue as above: the `integration` section header and the `fmt:` target line are merged into one line.

```
-## ================ integration ================
-fmt:                     ## Format codes.
+## ================ integration =========fmt:                     ## Format codes.
```

Please restore the newline between the section comment and the target.

docs/docs/zh/src/development/building.md
```diff
@@ -34,17 +33,15 @@ tsan:                    ## Build with ThreadSanitizer option.
 test_tsan: tsan          ## Run unit tests with ThreadSanitizer option.
 clean:                   ## Clear build/ directory.
 ##
```
LHT129, line 36: [critical] Same issue as above: the `integration` section header and the `fmt:` target line are merged into one line.

```
-## ================ integration ================
-fmt:                     ## Format codes.
+## ================ integration =========fmt:                     ## Format codes.
```

Please restore the newline between the section comment and the target.

docs/docs/zh/src/development/building.md
```diff
@@ -34,17 +33,15 @@ tsan:                    ## Build with ThreadSanitizer option.
...
-## ================ integration ================
-fmt:                     ## Format codes.
+## ================ integration =========fmt:                     ## Format codes.
 cov:                     ## Build unit tests with code coverage enabled.
 lint:                    ## Check coding styles defined in `.clang-tidy`.
 fix-lint:                ## Fix coding style issues in-place via clang-apply-replacements, use it be careful!!!
 test_parallel:           ## Run all tests parallel (used in CI).
 test_asan_parallel: asan ## Run unit tests parallel with AddressSanitizer option.
 test_tsan_parallel: tsan ## Run unit tests parallel with ThreadSanitizer option.
 ##
```
LHT129, line 45: [critical] Same issue as above: the `distribution` section header and the `release:` target line are merged into one line.

```
-## ================ distribution ================
-release:                 ## Build vsag with release options.
+## ================ distribution =========release:                 ## Build vsag with release options.
```

Please restore the newline between the section comment and the target.

docs/docs/zh/src/development/building.md
```diff
@@ -24,8 +24,7 @@ Usage: make <target>
 Targets:
 help:                    ## Show the help.
```
LHT129, line 25: [critical] The section separator lines in the Chinese docs (`docs/docs/zh/src/development/building.md`) are being merged with the following target line, breaking the Makefile help output formatting.

Three occurrences in the diff:
- `## ================ development =========debug:` (was `## ================ development ================` + newline + `debug:`)
- `## ================ integration =========fmt:` (was `## ================ integration ================` + newline + `fmt:`)
- `## ================ distribution =========release:` (was `## ================ distribution ================` + newline + `release:`)

The section separator `## ================ ... ================` must remain on its own line, with the target name (`debug:`, `fmt:`, `release:`) on the next line. The current merged form will produce garbled `make help` output and is inconsistent with the English docs and the rest of the file.

.github/workflows/build_performance.yml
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Already addressed: toolchain.txt now records the actual workflow environment values (not a fixed string), and collector-command.txt records the exact shell-escaped collector invocation. These two files make drift visible in the artifact without hard-coding. If metadata is out of date, the toolchain file directly shows the discrepancy. No additional hard-coded summary string is needed.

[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Already addressed: toolchain.txt now records the actual workflow environment values (not a fixed string), and collector-command.txt records the exact shell-escaped collector invocation. These two files make drift visible in the artifact without hard-coding. If metadata is out of date, the toolchain file directly shows the discrepancy. No additional hard-coded summary string is needed.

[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Already addressed: toolchain.txt now records the actual workflow environment values (not a fixed string), and collector-command.txt records the exact shell-escaped collector invocation. These two files make drift visible in the artifact without hard-coding. If metadata is out of date, the toolchain file directly shows the discrepancy. No additional hard-coded summary string is needed.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. This was fixed by replacing whitespace-fragile string splits with an explicit indentation-aware block extractor that handles comments, alternate indentation, and unnamed steps robustly. Regression fixtures cover these cases.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Fixed by adding explicit delimiter checks with targeted AssertionError diagnostics for both the opener and closer markers, plus regression coverage for missing or changed markers.

.github/workflows/pr-ci.yml
LHT129: Addressed in 3f3f086d, rebased onto latest main. The comment and bilingual docs have been updated to clarify that ccache uses its default directory on the disposable hosted VM with no restore or save across jobs. ccache is kept enabled (matching the Makefile default) but has no persistent cache. This is accurate and matches the behavior.

tests/scripts/test_build_benchmark_workflow.py
LHT129: Addressed in 3f3f086d, rebased onto latest main. The forbidden flag has been narrowed to --clear-ccache instead of the overly broad --clear substring. The contract test now catches only the exact flag that matters.

.github/workflows/build_performance.yml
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Official source in actions/toolkit (both artifacts and cache packages) confirms that ACTIONS_RUNTIME_TOKEN handles auth for these operations, not GITHUB_TOKEN permissions. contracts: read is the intended least-privilege design. No permission change is necessary.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Fixed alongside the other Python extraction comments (4003553597): explicit opener/closer delimiter checks with clear diagnostic messages.

tests/scripts/test_build_benchmark_workflow.py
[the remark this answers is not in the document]
LHT129: Addressed in 3f3f086d, rebased onto latest main. Fixed in the same extraction robustness update: job headers are now extracted via an indentation-aware block helper that provides meaningful assertion failures for missing or renamed jobs.

tests/scripts/test_build_benchmark_workflow.py
LHT129: Addressed in 3f3f086d, rebased onto latest main. This duplicates 4003553597 and 4012892801. Both sides were fixed with targeted opener/closer diagnostics and negative tests.

.github/workflows/build_performance.yml
LHT129: Addressed in 3f3f086d, rebased onto latest main. Addressed together with 4012892731: cache/artifact auth uses ACTIONS_RUNTIME_TOKEN, not GITHUB_TOKEN actions scopes. contracts: read remains correct.

docs/docs/zh/src/development/building.md
LHT129: [critical] The Chinese documentation Makefile targets section has corrupted formatting: three section separator lines lost their trailing `====` and newline, causing them to merge with the following target line.

- Line 27: `## ================ development =========debug:` (was `## ================ development ================\ndebug:`)
- Line 36: `## ================ integration =========fmt:` (was `## ================ integration ================\nfmt:`)
- Line 44: `## ================ distribution =========release:` (was `## ================ distribution ================\nrelease:`)

Each separator line should end with `====` followed by a newline, and the target line should start on its own line. This appears to be an accidental truncation during editing.

.github/workflows/build_performance.yml
LHT129: Fixed in d29795ed. Attempted rebase had accidentally truncated the three separator lines in the Chinese building docs. Restored the full separator lines with trailing newlines so each section header is independent.

LHT129 (commented):
I have reviewed the latest commit (d29795ed) for this PR. Here is my assessment:

**Summary**: This PR cleanly separates PR ASan builds from performance benchmarks. The workflow design is well-thought-out with strong security boundaries (read-only permissions, no persisted credentials, fixed package lists, SHA/ancestry validation). The test coverage is comprehensive, including contract tests for workflow structure, preflight rejection of legacy collectors, report gate validation, and real Git history ref gate testing.

**Previously flagged issues that have been addressed**:
- `inputs.ref` → `github.event.inputs.ref` (schedule safety): Fixed.
- `git cat-file -t` error handling for unknown SHA: Fixed with explicit `2>/dev/null` check.
- Test helper fragility (`IndexError` on job/step extraction): Fixed with `block()`/`job()` helpers and explicit guards.
- `python_step()` missing delimiter guard: Fixed with explicit marker validation.

**Remaining open items from prior reviews**:
1. `clang-tidy-15` / `libomp-15-dev` availability on ubuntu-22.04 without an LLVM apt source (flagged by Copilot). The PR CI workflow gets these through `install_deps_ubuntu.sh` which likely adds the repo; the benchmark workflow does not run that script. This could cause the benchmark to fail on a fresh runner.
2. `permissions: contents: read` is insufficient for `actions/cache/restore@v4` (needs `actions: read`) and `actions/upload-artifact@v5` (needs `actions: write`). Currently works due to repo defaults but would break under restrictive default permissions (flagged by LHT129).

**New findings**: No new critical or suggestion-level issues identified. The PR is in good shape.

**Overall**: The changes are well-structured, the test coverage is thorough, and the security posture is conservative and appropriate for a CI workflow. The two remaining open items from prior reviews are worth addressing before merge but are not blocking given current repo permission defaults and the existing runner image.

LHT129 (commented):
## Review Summary

This PR cleanly separates PR ASan validation from build performance benchmarks. The changes are well-structured with comprehensive contract tests and thorough documentation. Previous review comments have been largely addressed:

- `inputs.ref` → now uses `github.event.inputs.ref` ✓
- `git cat-file -t` error handling → now uses `2>/dev/null` with explicit existence check ✓
- Test extraction fragility → replaced with regex-based `block()`/`job()`/`step()` helpers with proper error messages ✓
- `python_step()` missing guard → now has explicit delimiter checking with clear AssertionError messages ✓

### Observations

**[note]** The benchmark workflow declares only `contents: read` but uses `actions/cache/restore@v4` and `actions/upload-artifact@v5`. This is consistent with `pr-ci.yml` (which also uses these actions without explicit `actions` permissions), so it matches existing project practice. If the repository ever moves to restrictive default permissions, both workflows would need `actions: read` (and `actions: write` for artifact upload) added explicitly.

**[note]** The `clang-tidy-15` and `libomp-15-dev` packages in the benchmark workflow's "Prepare trusted Ubuntu host" step match the PR CI's host setup. If these versioned packages work in PR CI on `ubuntu-22.04`, they should work here too. The `sudo rm -rf /opt/hostedtoolcache` was correctly removed from the earlier revision.

### Test coverage

The new `test_build_benchmark_workflow.py` provides strong contract coverage:
- Block/job/step extraction with proper error messages
- Python inline script extraction with delimiter validation
- Toolchain metadata uses runtime environment variables (not hardcoded values)
- Collector command recording matches executed arguments (including replay verification)
- PR build preserves required options and forbids benchmark-only artifacts
- Security and isolation assertions (permissions, no secrets, no PR triggers, etc.)
- Preflight rejects legacy collectors without executing them (11 test cases)
- Report gate requires verified no-op (15+ variant test cases)
- Ref gate with real git history (injection resistance, ancestry validation)

No blocking issues found.

LHT129 (commented):
Reviewed the latest commit d29795ed. The PR has been iterated well:

- The `inputs.ref` → `github.event.inputs.ref` fix is in place.
- The `git cat-file -t` error handling with `2>/dev/null` fallback is in place.
- The `step()` / `block()` extraction helpers now use regex-based parsing with explicit `AssertionError` on missing markers.
- The `python_step()` helper now validates the inline Python delimiter before extraction.
- The `pr-ci.yml` ASan build correctly runs a single `make asan COMPILE_JOBS=3` with `ccache --zero-stats` and `/usr/bin/time -p`, preserving lint, downstream tests, and artifact contracts.
- The benchmark workflow has conservative security posture: `contents: read`, `persist-credentials: false`, fixed package list, SHA ancestry validation, and no compiler-cache persistence.
- Test coverage is thorough: workflow contract tests, preflight AST gate tests, report verification gate tests, and real-Git ref gate tests all pass.

No new blocking issues found in this revision.

LHT129 (commented):
## Code Review Summary

This PR cleanly separates PR ASan builds from performance benchmarks. The changes are well-structured and the test coverage is thorough.

### What was reviewed
- `.github/workflows/build_performance.yml` (new): Standalone benchmark workflow with conservative security posture, SHA validation, schema-3 preflight gate, and verified-noop enforcement.
- `.github/workflows/pr-ci.yml`: Removed benchmark collection from PR CI, kept single `make asan` build with restore-only dependency archives.
- `tests/scripts/test_build_benchmark_workflow.py` (new): Comprehensive contract tests covering block extraction, preflight rejection, report gate validation, and real git history ref gate testing.
- `tests/scripts/test_dependency_workflows.py`: Updated restore-only assertion to reference the benchmark workflow.
- `docs/docs/en/src/development/building.md` and `docs/docs/zh/src/development/building.md`: Documented the new PR ASan build behavior and standalone benchmark workflow.

### Issues previously raised and now resolved
- `inputs.ref` → `github.event.inputs.ref` for schedule safety (fixed)
- `git cat-file -t` error handling for unknown SHAs (fixed)
- `IndexError` from fragile string splitting in tests (fixed, replaced with `block()`/`job()` helpers)
- `python_step()` missing guard for delimiter changes (fixed)

### Remaining open item
The permissions discussion (adding explicit `actions: read`) is already tracked as a suggestion on the workflow file and does not block this PR.

LHT129 (commented):
Reviewed commit 21a4552a. This iteration cleanly addresses most prior feedback: the `inputs.ref` → `github.event.inputs.ref` fix, the unknown-SHA error handling in the ref resolution step, the narrowed AST preflight gate, the `block()`/`job()` extraction helpers with explicit assertions, and the `python_step()` delimiter guard are all correctly implemented. The test coverage (preflight, report gate, ref gate with real git history, toolchain metadata, collector command recording) is thorough.

Remaining open items from earlier reviews (permissions scope for `actions`, ccache default-directory persistence on the PR build step, clang-tidy-15 package availability) are already tracked in existing inline comments and do not need duplication.

No new substantive issues found in this commit.

LHT129 (commented):
This PR has been thoroughly reviewed with 30 inline comments already covering correctness, security, permissions, workflow contracts, and test robustness. Most issues raised in earlier review rounds have been addressed in subsequent commits.

The separation of PR ASan builds from performance benchmarks is well-structured: `pr-ci.yml` is simplified to a single `make asan` build with ccache diagnostics, while the new `build_performance.yml` provides an isolated, security-conscious benchmark with strict SHA validation, schema gating, and no-op verification.

No new blocking issues identified beyond what has already been discussed in the existing review threads. The remaining open discussion threads (permissions, test fragility, toolchain metadata drift) are non-blocking suggestions that can be addressed iteratively.

LHT129 (commented):
## Review Summary

This PR cleanly separates PR ASan builds from performance benchmarks. The changes are well-structured with comprehensive test coverage (290-line test suite for workflow contracts, plus updates to existing tests).

### What was reviewed

- **`.github/workflows/build_performance.yml`** (new, 277 lines): Isolated benchmark workflow with conservative security posture (read-only contents, no persisted credentials, fixed package list, SHA ancestry validation, AST-based preflight gate for schema 3 collector support).
- **`.github/workflows/pr-ci.yml`** (simplified): Removed the multi-phase collector invocation from PR CI, replaced with a single `make asan COMPILE_JOBS=3` + ccache diagnostics. Dependency archives changed to restore-only (`actions/cache/restore@v4`).
- **Documentation**: Both English and Chinese building guides updated with clear explanations of the separation and security boundaries.
- **`tests/scripts/test_build_benchmark_workflow.py`** (new, 290 lines): Thorough contract tests covering block extraction, Python delimiter validation, toolchain metadata, collector command recording, preflight rejection of legacy collectors, report gate validation, and real git history ref gate testing.
- **`tests/scripts/test_dependency_workflows.py`**: Updated to validate the benchmark workflow's restore-only policy.

### Assessment

The implementation is thorough and defensively coded. Key security properties (no secrets, no persisted credentials, SHA validation, fixed package list, no privileged target-ref scripts) are all verified by tests. The preflight gate correctly blocks legacy collectors without executing them, and the post-collection verification enforces `true_noop: true` with zero build edges.

Previous review comments from Copilot and self-review have been addressed in this revision (proper `github.event.inputs.ref` usage, explicit git cat-file error handling, robust YAML block extraction with AssertionError messages, and proper Python delimiter validation).

No blocking issues found.

LHT129 (commented):
Reviewed commit d1cd9a79.

**Summary**: This is a well-structured CI refactor that cleanly separates PR ASan validation from build-performance benchmarking. The new `build_performance.yml` workflow has strong security boundaries (read-only permissions, no persisted credentials, SHA/ancestry validation, fixed package list), comprehensive preflight/postflight gates, and thorough test coverage.

**Previously raised issues resolved in this commit**:
- `inputs.ref` → `github.event.inputs.ref` for schedule safety
- `git cat-file -t` now handles missing SHAs gracefully with `2>/dev/null`
- `python_step()` and `step()` helpers now have proper guard assertions
- `block()`/`job()` extraction helpers added for robust workflow parsing

**Remaining open items** (already noted in prior comments):
- `clang-tidy-15` / `libomp-15-dev` may not be available from default ubuntu-22.04 repos without an LLVM apt source (the benchmark workflow does not call `install_deps_ubuntu.sh`)
- `permissions` block declares only `contents: read`; `actions/cache/restore` and `actions/upload-artifact` may need explicit `actions` scope if the repo ever tightens default permissions

No new issues found. The test suite is thorough, covering preflight rejection, report gate validation, ref gate with real git history, toolchain metadata, and collector command fidelity.

wxyucs (approved):
lgtm

PR #2957, opened 2026-09-14
feat: support fp16 reranking in sindi indexes

### LINKED ISSUE #2956: [feat](sindi): support FP16 rerank storage
### Affected module

module/index

### Affected area

not applicable

### Motivation

SINDI and SINDI V2 currently support FP32 sparse forward storage for candidate reranking, while DMQ8 is the only compressed rerank alternative. Workloads that need lower forward-store memory without adopting 8-bit quantization need an intermediate FP16 option.

Because sparse term IDs remain 32-bit, storing values in FP16 reduces the core per-term rerank payload from 8 bytes to 6 bytes. This can reduce memory and bandwidth pressure for large candidate sets while preserving FP32 query values and accumulation.

### Proposed solution

Add `rerank_type: "fp16"` to both SINDI and SINDI V2.

- Require `use_reorder: true`.
- Store document sparse values as IEEE FP16 while retaining `uint32_t` term IDs.
- Keep query values and distance accumulation in FP32.
- Support build, search, incremental insertion, serialization, and SINDI V2 rerank layouts and I/O backends.
- Preserve the existing FP32 serialization format and use a versioned payload for FP16 to prevent mismatched decoding.
- Report FP16 distance evaluations through the `sparse_fp16` backend.
- Keep the default `rerank_type` as `fp32`.

### Alternatives considered

- Continue using FP32 rerank storage, which preserves exact values but consumes more memory and bandwidth.
- Use `rerank_type: "dmq8"`, which provides stronger compression but has larger quantization error and additional codebook constraints.
- Apply FP16 to query values and accumulation as well; this was rejected to avoid unnecessary precision loss.

additive only (no break)

### References

- `src/algorithm/sindi/sindi_parameter.h:37`
- `src/algorithm/sindi/sindi.cpp:153`
- `src/algorithm/sindi_v2/sindi_v2.cpp:216`
- `src/quantization/sparse_quantization/sparse_quantizer.h:255`
- `src/datacell/sparse_vector_datacell.inl:167`
- `docs/docs/en/src/indexes/sindi.md`
- `docs/docs/zh/src/indexes/sindi.md`

### Additional context

FP16 conversion adds computation and does not guarantee lower latency for every workload. Cache-resident workloads with high term overlap may regress, while large memory-bound candidate scans can benefit from the smaller payload. Representative workload benchmarking is recommended before enabling FP16.

## Related issues

- #2544 — [feat](other): expose per-search distance evaluation statistics
  https://github.com/antgroup/vsag/issues/2544

---
_Drafted with AI assistant: Codex:GPT-5_

Roxanne0321:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2956

## What Changed

- Add `rerank_type: "fp16"` to SINDI and SINDI V2 while keeping FP32 as the default.
- Store rerank document values as FP16, with `uint32_t` term IDs and FP32 query values and accumulation.
- Support build, search, incremental insertion, serialization, statistics, memory estimation, and SINDI V2 rerank layouts.
- Preserve the existing FP32 sparse payload and add a versioned FP16 payload with value-type validation.
- Add parameter, sparse datacell, SINDI, and SINDI V2 coverage, including non-zero SINDI V2 rerank layouts.
- Document configuration, compatibility, storage, precision, and performance tradeoffs in English and Chinese.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build --target unittests_algorithm unittests_datacell -j6
cmake --build build --target vsag -j6
cmake --build build-release --target vsag -j6

./build/tests/unittests_algorithm '[ut][SINDI]'
All tests passed (184065 assertions in 55 test cases)

./build/tests/unittests_algorithm '[ut][SINDIV2]'
All tests passed (1544 assertions in 25 test cases)

./build/tests/unittests_algorithm 'SINDIV2 Top Terms Rerank Layout End-To-End'
All tests passed (28 assertions in 1 test case; generated FP32 and FP16 cases)

./build/tests/unittests_datacell '[ut][SparseDataCell]'
All tests passed (14126 assertions in 8 test cases)

/root/.local/bin/clang-format --dry-run --Werror <changed C++ files>
/root/.local/bin/clang-tidy -p build --quiet <changed production translation units>
git diff --check
All passed.

The documentation checker could not be run because Node.js is unavailable in the local environment.
```

## Compatibility Impact

- API/ABI compatibility: Additive configuration only; no public C++ ABI change.
- Behavior changes: `rerank_type: "fp16"` is newly accepted and requires `use_reorder: true`. Existing FP32 and DMQ8 configurations retain their behavior and FP32 keeps its existing serialized representation. Older VSAG readers cannot read the new FP16 rerank payload.

- Performance impact: The core per-term rerank payload decreases from 8 bytes to 6 bytes. On an Intel Xeon Platinum 8269CY Release build, a cache-resident full-overlap microbenchmark was about 2.66x slower because every matching value requires conversion, while a 500,000-candidate sparse-overlap scan was about 16%-18% faster and used about 16.3% less total datacell memory. Results depend on candidate size, term overlap, and cache pressure.
- Concurrency/thread-safety impact: None; existing sparse datacell synchronization is unchanged.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: English and Chinese SINDI/SINDI V2 website documentation and the C++ SINDI example.

## Risk and Rollback

- Risk level: Medium. FP16 introduces expected rounding and a new sparse forward-store layout, guarded by an explicit serialization version and value-type check.
- Rollback plan: Revert this commit; existing FP32 and DMQ8 payloads are unchanged.

## Checklist


src/quantization/sparse_quantization/sparse_quantizer.h
```diff
@@ -32,16 +36,22 @@ struct BufferEntry {
...
  * @brief Sparse Quantizer stores sparse vectors with only non-zero elements.
  *
- * code layout:
+ * FP32 code layout (also used for queries):
  * +----------+------------------+------------------+-----+------------------+
+ *
+ * FP16 base code layout:
+ * +----------+------------------+------------------+
+ * | len      | ids              | values           |
+ * | [4B]     | uint32[len]      | uint16[len]      |
```
LHT129, line 48: [note] The FP16 code layout diagram (lines 44-48) is visually adjacent to the FP32 detail lines (lines 50-52, "| len | entry[0]...") without a clear separator. The FP32 detail lines appear to continue the FP16 diagram, which could confuse readers. Consider adding a blank ` *` comment line or a small separator comment between the two layout diagrams to make the boundary explicit.

LHT129 (commented):
## Review Summary

This PR adds FP16 reranking support to SINDI indexes. Overall the implementation is well-structured with good test coverage. The core design — storing base vectors as FP16 while keeping queries and accumulation in FP32 — is sound and provides a 25% memory reduction for rerank payloads (6 bytes/entry vs 8 bytes/entry).

### What was checked
- **Correctness**: FP16 encode/decode roundtrip, distance computation (FP32 query × FP16 base, FP16 base × FP16 base), V3 serialization with value_type validation, backward compatibility with V2/legacy formats.
- **Code style**: Consistent with existing codebase patterns, proper use of `const`, `static`, `[[nodiscard]]`.
- **Test coverage**: New test cases for FP16 encode/decode, serialization roundtrip, distance computation correctness, parameter parsing. Covers both SINDI and SINDI V2.
- **Performance**: The FP16 encode path uses a temporary `Vector<BufferEntry>` allocation (flagged by Copilot). The compute paths use direct pointer access without intermediate allocations.
- **Security**: No user-controlled input paths identified that could be exploited.

### Issues found
- 1 inline [note] about FP16 layout comment clarity in `sparse_quantizer.h`.
- 5 existing Copilot comments cover the main substantive issues (float equality in tests, hard-coded offsets, Vector allocation overhead, nested cast readability).

No blocking issues found. The Copilot suggestions are worth addressing before merge.

LHT129 (commented):
Overall this is a well-structured PR that adds FP16 reranking support to SINDI and SINDI V2 indexes. The implementation is clean and follows existing patterns.

**What this PR does:**
- Adds `SparseQuantizerValueType` enum (FP32/FP16) and extends `SparseQuantizerParameter` to parse `"value_type"` from JSON
- Extends `SparseQuantizer` with FP16 encode/decode/compute paths: encoding sorts entries and stores values as `uint16_t` via `FloatToFP16`, query is always FP32, distance computation dispatches to `ComputeFP32QueryToFP16Codes` when base is FP16
- Adds serialization format V3 to `SparseVectorDataCell` with a `value_type` byte for validation; V2 legacy format rejects FP16 payloads
- Plumbs `rerank_type: "fp16"` through SINDI and SINDI V2 parameter parsing and `create_rerank_flat`
- Comprehensive tests covering encode/decode round-trip, serialization, distance computation, type mismatch rejection, and backward compatibility
- Documentation updates in both English and Chinese

**Key design decisions:**
- Query always encoded as FP32, base codes as FP16 — avoids precision loss in query and keeps accumulation in FP32
- FP16 encoding sorts entries by id (O(n log n)) and stores ids+values in separate arrays (struct-of-arrays layout)
- Serialization versioning with sentinel (0xFFFFFFFF) for backward compatibility with legacy 32-bit offset format
- `rerank_type` is independent of posting-list `use_quantization` — they control separate stores

The existing inline comments from Copilot cover the main areas for improvement (test float comparison tolerance, hardcoded byte offset, heap allocation in encode path, nested cast readability).

src/datacell/sparse_vector_datacell.inl
```diff
@@ -161,7 +161,19 @@ SparseVectorDataCell<QuantTmpl, IOTmpl>::Deserialize(LvalueOrRvalue<StreamReader
...
+            CHECK_ARGUMENT(not quantizer_->IsFP16(),
+                           "SparseVectorDataCell v2 payload requires fp32 sparse values");
+        } else if (version == SERIALIZE_FORMAT_VERSION_V3) {
+            uint8_t value_type = 0;
+            StreamReader::ReadObj(reader, value_type);
+            CHECK_ARGUMENT(value_type <= static_cast<uint8_t>(SparseQuantizerValueType::FP16),
+                           "invalid SparseVectorDataCell sparse value type");
+            CHECK_ARGUMENT(value_type == static_cast<uint8_t>(quantizer_->IsFP16()
+                                                                  ? SparseQuantizerValueType::FP16
+                                                                  : SparseQuantizerValueType::FP32),
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 4cb0c458: the expected value type is computed in a named local before comparison, removing the nested casts. Serialization v3 also writes the FP16 value-type constant directly, since v3 is emitted only for FP16 data.

src/quantization/sparse_quantization/sparse_quantizer.h
Roxanne0321: Fixed in 4cb0c458: the FP32 and FP16 layout diagrams are now grouped and separated explicitly, so each description appears immediately before its own diagram.

src/datacell/sparse_vector_datacell_test.cpp
```diff
@@ -128,6 +131,117 @@ TEST_CASE("SparseDataCell Basic Test", "[ut][SparseDataCell] ") {
...
+    std::stringstream fp32_stream;
+    IOStreamWriter fp32_writer(fp32_stream);
+    fp32->Serialize(fp32_writer);
+    const auto fp16_bytes = fp16_stream.str();
+    const auto fp32_bytes = fp32_stream.str();
+    REQUIRE(fp16_bytes.size() < fp32_bytes.size());
+    uint32_t version = 0;
+    std::memcpy(&version, fp16_bytes.data() + 16, sizeof(version));
+    REQUIRE(version == 3);
+
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 4cb0c458: the test now advances through the base header via FlattenInterface::Deserialize, then reads and verifies the sentinel and v3 marker. It no longer depends on the base header being exactly 16 bytes. The focused FP16 SparseDataCell test passes.

src/quantization/sparse_quantization/sparse_quantizer.h
```diff
@@ -225,18 +331,82 @@ template <MetricType metric>
...
+    std::sort(entries.begin(), entries.end(), [](const BufferEntry& a, const BufferEntry& b) {
         return a.id < b.id;
     });
+    auto* ids = reinterpret_cast<uint32_t*>(codes + sizeof(uint32_t));
+    auto* values = reinterpret_cast<uint16_t*>(codes + sizeof(uint32_t) +
+                                               static_cast<uint64_t>(sv.len_) * sizeof(uint32_t));
+    for (uint32_t i = 0; i < sv.len_; ++i) {
+        ids[i] = entries[i].id;
+        values[i] = generic::FloatToFP16(entries[i].val);
+    }
```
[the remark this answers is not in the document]
Roxanne0321: Addressed in 4cb0c458: sorted IDs, which are the normal SINDI path, now encode directly into the compact output with no scratch allocation or BufferEntry copy. Unsorted input keeps the allocator-backed fallback because the 6-byte-per-entry destination cannot safely stage 8-byte BufferEntry values while preserving ID/value pairing. The focused SINDI FP16 and SINDI V2 layout tests pass (8 and 28 assertions).

src/datacell/sparse_vector_datacell_test.cpp
```diff
@@ -128,6 +131,117 @@ TEST_CASE("SparseDataCell Basic Test", "[ut][SparseDataCell] ") {
...
+    };
+    REQUIRE(distances[0] ==
+            1.0F - query_values[0] * rounded(values0[0]) - query_values[1] * rounded(values0[2]));
+    REQUIRE(distances[1] ==
+            1.0F - query_values[0] * rounded(values1[1]) - query_values[1] * rounded(values1[0]));
+    REQUIRE(JsonType::Parse(statistics.Dump())["distance_evaluations_by_backend"]["sparse_fp16"]
+                .GetUint64() == 2);
+
+    REQUIRE(fp16->ComputePairVectors(0, 1) == 1.0F - rounded(values0[0]) * rounded(values1[1]) -
+                                                  rounded(values0[2]) * rounded(values1[0]));
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 4cb0c458 together with the related assertion: the restored distance comparison now uses an absolute tolerance of 1e-6. The focused FP16 SparseDataCell test passes with 19 assertions.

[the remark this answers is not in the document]
Roxanne0321: Fixed in 4cb0c458: the distance and pair-distance assertions now use an absolute tolerance of 1e-6 instead of exact float equality. Validated with the FP16 SparseDataCell test (19 assertions).

src/datacell/sparse_vector_datacell.h
```diff
@@ -155,7 +155,7 @@ class SparseVectorDataCell : public FlattenInterface {
     Deserialize(LvalueOrRvalue<StreamReader> reader) override;
 
     inline void
-    SetQuantizer(std::shared_ptr<Quantizer<QuantTmpl>> quantizer) {
+    SetQuantizer(std::shared_ptr<QuantTmpl> quantizer) {
```
LHT129: [note] The `SetQuantizer` parameter type and `quantizer_` member type were narrowed from `std::shared_ptr<Quantizer<QuantTmpl>>` to `std::shared_ptr<QuantTmpl>`. This is a breaking change for any external code that passes the base type. Since `SparseVectorDataCell` is an internal implementation detail and the factory (`FlattenInterface::MakeInstance`) always creates `shared_ptr<QuantTmpl>`, this is safe in practice. The change eliminates the need for `static_cast<const QuantTmpl&>(*this->quantizer_)` in callers that need the derived type, which is a net improvement. Just flagging for awareness in case there are downstream consumers of this header.

LHT129 (commented):
## Code Review Summary

This PR adds FP16 reranking support to SINDI and SINDI V2 indexes. Overall the implementation is clean and well-structured.

### What changed
- New `SparseQuantizerValueType` enum (FP32=0, FP16=1) with JSON serialization
- FP16 code layout in `SparseQuantizer`: `[len:4B][ids:uint32[len]][values:uint16[len]]`
- New distance computation methods: `ComputeFP32QueryToFP16Codes`, `ComputeFP16Codes`
- `DecodeSparseVector` for FP16→FP32 conversion during `GetSparseVectorByInnerId`
- V3 serialization format in `SparseVectorDataCell` with `value_type` byte
- `create_rerank_flat` sets `value_type = FP16` when `rerank_type == "fp16"`
- `EstimateMemory` uses `sizeof(uint16_t)` for FP16 rerank values
- Comprehensive tests covering FP16 encode/decode, serialization, distance computation, and parameter validation

### Positive observations
- **Serialization versioning**: V3 format with `value_type` byte is a clean approach that maintains backward compatibility (V2 remains FP32-only)
- **Sorted fast-path**: The `EncodeOneImpl` FP16 path avoids allocating a `Vector<BufferEntry>` when input is already sorted — good performance optimization
- **Test coverage**: Tests cover the key paths: encode/decode round-trip, serialization compatibility checks, distance computation correctness, and parameter validation
- **Memory estimation**: Correctly uses `sizeof(uint16_t)` for FP16 values in both `sindi.cpp` and `sindi_v2.cpp`
- **API cleanup**: `DecodeSparseVector` centralizes the decode logic, replacing manual BufferEntry loops in `GetSparseVectorByInnerId`

### One note
- The `SetQuantizer`/`quantizer_` type narrowing from `shared_ptr<Quantizer<QuantTmpl>>` to `shared_ptr<QuantTmpl>` is a breaking API change for the header, though safe in practice since the factory always creates the derived type.

No blocking issues found. The PR is well-implemented.

LHT129 (commented):
Thank you for the update in 4cb0c458. The changes address all the issues raised in the previous review round:

- Float equality assertions now use an absolute tolerance of `1e-6F` instead of exact comparison, resolving the potential test flakiness.
- The serialization version byte offset is no longer hardcoded; the test now uses `StreamReader` to parse the header properly.
- The nested `static_cast` in the v3 deserialization check has been replaced with a named local variable for clarity.
- The `EncodeOneImpl` FP16 path now has a fast-path for already-sorted IDs, avoiding unnecessary heap allocation in the common case.

The test coverage is comprehensive across SINDI, SINDI V2, and SparseDataCell, including serialization round-trips, parameter validation, and memory estimation. No blocking issues found.

src/datacell/sparse_vector_datacell.h
Roxanne0321: Fixed in d479182e: SetQuantizer and quantizer_ now retain the original shared_ptr<Quantizer<QuantTmpl>> type. A private typed_quantizer accessor is used only for the SparseQuantizer-specific FP16 operations, so existing internal callers keep source compatibility. The datacell target builds and the focused FP16 SparseDataCell test passes with 22 assertions.

Roxanne0321:
Addressed the suppressed FP16 correctness findings from the latest Copilot review in d479182e:

- FloatToFP16/FP16ToFloat now handle IEEE zero, subnormal, normal, infinity, and NaN values with round-to-nearest-even conversion, preventing zero and tiny sparse weights from becoming a nonzero minimum-normal value.
- SINDI UpdateVector normalizes the candidate values through the same FP16 round trip before comparing them with the stored FP16 rerank vector. An unchanged vector containing 100.1 is accepted after storage rounds it to 100.125, while a genuinely changed value is still rejected.
- Added focused regression coverage: 12 FP16 boundary assertions, 22 SparseDataCell FP16 assertions, and 10 SINDI FP16 assertions. The existing 2025 FP16 conversion assertions and 28 SINDI V2 layout assertions also pass. clang-tidy 15 passes for the affected production sources.

LHT129 (commented):
I have completed a thorough review of this PR. Here is my assessment:

**Summary**: The PR adds FP16 reranking support to SINDI and SINDI V2 sparse indexes. The implementation is well-structured and covers all necessary layers: quantization (SparseQuantizer), serialization (SparseVectorDataCell V3 format), index-level integration (sindi.cpp, sindi_v2.cpp), and comprehensive tests.

**Key observations**:

1. **FloatToFP16 / FP16ToFloat (generic.cpp)**: The new IEEE 754 half-precision conversion functions correctly handle subnormals, infinities, NaN, and round-to-nearest-even. The subnormal rounding path that overflows into the exponent field (e.g., max subnormal rounding up to min normal) is correct IEEE 754 behavior.

2. **Mixed-precision scoring (ComputeFP32QueryToFP16Codes)**: Query is always encoded as FP32 and base values are decoded from FP16 on the fly during distance computation. This is the right design — it preserves query precision while saving storage.

3. **Serialization compatibility (V2→V3)**: The V3 format adds a single `value_type` byte after the version field. V2 deserialization explicitly rejects FP16 data, and legacy (pre-sentinel) format also rejects FP16. Backward compatibility is properly handled.

4. **UpdateVector FP16 rounding (sindi.cpp)**: The subset check correctly rounds new values through FP16 before comparing with stored (already FP16-rounded) values. SINDI V2 does not support UpdateVector, so no change is needed there.

5. **Existing review comments**: The issues flagged by Copilot and LHT129 (test float equality, byte offset hardcoding, nested cast readability, FP16 code layout diagram, SetQuantizer type narrowing) have been acknowledged by the author with fixes referenced as commit 4cb0c458. These fixes are not yet pushed to this PR branch.

**No new substantive issues found.** The implementation is correct and complete.

LHT129 (commented):
## Code Review Summary

After a thorough review of all 23 changed files (794 additions, 109 deletions), I find the FP16 reranking implementation to be well-structured and correct. The core changes are:

1. **FP16 conversion functions** (`FloatToFP16`/`FP16ToFloat`) with proper IEEE 754 round-to-nearest-even semantics
2. **Sparse quantizer** extended with FP16 code layout `[len:4B][ids:uint32[len]][values:uint16[len]]`
3. **Distance computation** with `ComputeFP32QueryToFP16Codes` (query in FP32, base in FP16) and `ComputeFP16Codes` (both FP16)
4. **Serialization** with V3 format including value_type byte
5. **SINDI/SINDIV2** parameter validation, `UpdateVector` FP16 rounding, and memory estimation

### Issues already flagged by Copilot (still open):

- **FP16 code alignment** (`GetCodeSizeByLength`): FP16 codes are not 4-byte aligned for odd-length vectors, which may cause issues with `VariableRecordLayout`
- **Hard-coded value type in serialization**: V3 always writes `FP16` instead of deriving from the quantizer
- **NaN quiet bit**: `FloatToFP16` does not guarantee the FP16 quiet-NaN bit (bit 9) is set

### Issues already addressed by the author:

- Float comparison precision in tests (now uses `std::abs() < 1e-6F`)
- Hard-coded byte offset in tests (now uses proper deserialization)
- Nested static_cast in serialization check (simplified)
- FP16 layout diagram grouping (separated from FP32)
- `EncodeOneImpl` extra allocation (added sorted fast path)
- `SetQuantizer` type narrowing (reverted to original type)

### Additional observations:

The `UpdateVector` FP16 path correctly rounds new vector values to FP16 precision before comparing with stored (already FP16-rounded) values, ensuring consistent subset detection. The `EncodeOneImpl` sorted fast path with `std::is_sorted` is safe since the default comparator requires strict ordering.

No additional substantive issues found beyond what Copilot has already covered.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -228,9 +243,14 @@ void
 SparseVectorDataCell<QuantTmpl, IOTmpl>::Serialize(StreamWriter& writer) {
     FlattenInterface::Serialize(writer);
     const uint32_t sentinel = SERIALIZE_FORMAT_SENTINEL;
-    const uint32_t version = SERIALIZE_FORMAT_VERSION_V2;
+    const uint32_t version =
+        typed_quantizer().IsFP16() ? SERIALIZE_FORMAT_VERSION_V3 : SERIALIZE_FORMAT_VERSION_V2;
     StreamWriter::WriteObj(writer, sentinel);
     StreamWriter::WriteObj(writer, version);
+    if (version == SERIALIZE_FORMAT_VERSION_V3) {
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 308cf21a: V3 serialization now writes typed_quantizer().GetValueType() rather than a hard-coded FP16 constant. The focused SparseDataCell serialization test reads and verifies the emitted FP16 value type (27 assertions).

src/quantization/sparse_quantization/sparse_quantizer.h
```diff
@@ -225,18 +331,90 @@ template <MetricType metric>
...
+}
+
+template <MetricType metric>
+uint64_t
+SparseQuantizer<metric>::GetCodeSizeByLength(uint32_t length) const {
+    if (not IsFP16()) {
+        return GetFP32CodeSize(length);
+    }
+    return sizeof(uint32_t) + static_cast<uint64_t>(length) * (sizeof(uint32_t) + sizeof(uint16_t));
+}
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 308cf21a: FP16 sparse code sizes are now rounded up to alignof(uint32_t), so every record starts on a 4-byte boundary. EncodeOneImpl explicitly zeroes the trailing padding bytes, making serialized payloads deterministic. The focused SparseDataCell FP16 test verifies two odd-length records are aligned and padding is zero (27 assertions).

src/simd/generic.cpp
```diff
@@ -203,28 +203,75 @@ FloatToBF16(const float fp32_value) {
...
+    const uint16_t sign = static_cast<uint16_t>((fp32.int_value >> 16) & 0x8000U);
+    const uint32_t exponent = (fp32.int_value >> 23) & 0xFFU;
+    const uint32_t mantissa = fp32.int_value & 0x007FFFFFU;
+
+    if (exponent == 0xFFU) {
+        if (mantissa == 0) {
+            return static_cast<uint16_t>(sign | 0x7C00U);
+        }
+        const uint16_t nan_mantissa =
+            static_cast<uint16_t>((mantissa >> 13) == 0 ? 1 : mantissa >> 13);
```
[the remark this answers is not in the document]
Roxanne0321: Fixed in 308cf21a: the FP16 NaN path now sets mantissa bit 9, guaranteeing a quiet NaN while retaining the representable payload bits. The IEEE conversion boundary test verifies the quiet bit and NaN round trip (13 assertions).

LHT129:
test

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
```diff
@@ -15,12 +15,24 @@
...
 #pragma once
 
+#include <fmt/format.h>
+
+#include <cstdint>
+
 #include "quantization/quantizer_parameter.h"
 #include "typing.h"
+#include "utils/param_compat_macros.h"
 #include "utils/pointer_define.h"
```
LHT129, line 25: [note] `SparseQuantizerParameter::FromJson()` uses `std::string` on line 50 (`const std::string value_type_name = ...`), but this header does not include `<string>`. While it may compile today through transitive includes from `typing.h` or `quantizer_parameter.h`, this is fragile and can break with header refactoring. Consider adding `#include <string>` to make the dependency explicit.

src/algorithm/sindi/sindi.cpp
```diff
@@ -688,7 +692,17 @@ SINDI::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) {
     auto check_and_cleanup = [this, inner_id, &new_sv](auto&& get_sparse_vector) -> bool {
         SparseVector old_sv;
         get_sparse_vector(inner_id, &old_sv, this->allocator_);
-        bool ret = is_subset_of_sparse_vector(old_sv, new_sv);
+        bool ret = false;
+        if (rerank_type_ == SPARSE_RERANK_TYPE_FP16) {
+            Vector<float> rounded_values(new_sv.len_, allocator_);
```
LHT129, line 697: [suggestion] Consider caching the `rounded_values` buffer to avoid per-call allocation in `UpdateVector`.

The temporary `Vector<float> rounded_values(new_sv.len_, allocator_)` is allocated on every `UpdateVector` call for FP16 rerank. For workloads with frequent incremental updates, this could add measurable allocation overhead. Consider reusing a member buffer or using a stack-allocated small buffer with a fallback to heap for large vectors.

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
```diff
@@ -30,13 +42,39 @@ class SparseQuantizerParameter : public QuantizerParameter {
 
     void
     FromJson(const JsonType& json) override {
+        value_type = SparseQuantizerValueType::FP32;
+        if (json.Contains(SPARSE_QUANTIZER_VALUE_TYPE_KEY)) {
+            const auto value_type_json = json[SPARSE_QUANTIZER_VALUE_TYPE_KEY];
+            CHECK_ARGUMENT(value_type_json.IsString(), "sparse value_type must be a string");
+            const std::string value_type_name = value_type_json.GetString();
+            CHECK_ARGUMENT(
```
LHT129, line 50: [suggestion] Missing `#include <string>` for `std::string` usage.

`FromJson()` uses `std::string value_type_name = value_type_json.GetString();` but `<string>` is not explicitly included. While it may be transitively included through other headers, an explicit include makes the dependency clear and avoids potential breakage if transitive includes change.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -161,7 +161,20 @@ SparseVectorDataCell<QuantTmpl, IOTmpl>::Deserialize(LvalueOrRvalue<StreamReader
...
         uint32_t version = 0;
         StreamReader::ReadObj(reader, version);
-        if (version != SERIALIZE_FORMAT_VERSION_V2) {
+        if (version == SERIALIZE_FORMAT_VERSION_V2) {
+            CHECK_ARGUMENT(not typed_quantizer().IsFP16(),
+                           "SparseVectorDataCell v2 payload requires fp32 sparse values");
+        } else if (version == SERIALIZE_FORMAT_VERSION_V3) {
+            uint8_t value_type = 0;
+            StreamReader::ReadObj(reader, value_type);
+            CHECK_ARGUMENT(value_type <= static_cast<uint8_t>(SparseQuantizerValueType::FP16),
```
LHT129, line 170: [note] The `CHECK_ARGUMENT(value_type <= static_cast<uint8_t>(SparseQuantizerValueType::FP16), ...)` check in the V3 deserialization path correctly validates the upper bound, but future value types added beyond FP16 will need this check updated. Consider using a named constant (e.g., `MAX_SPARSE_QUANTIZER_VALUE_TYPE`) to make this boundary explicit and easier to maintain.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -267,7 +287,7 @@ template <typename QuantTmpl, typename IOTmpl>
 void
 SparseVectorDataCell<QuantTmpl, IOTmpl>::InsertVector(const void* vector, InnerIdType idx) {
     auto sparse_vector = (const SparseVector*)vector;
-    uint64_t code_size = (static_cast<uint64_t>(sparse_vector->len_) * 2 + 1) * sizeof(uint32_t);
+    uint64_t code_size = typed_quantizer().GetCodeSizeByLength(sparse_vector->len_);
```
LHT129, line 290: [suggestion] The `InsertVector` change from manual `(len * 2 + 1) * sizeof(uint32_t)` to `typed_quantizer().GetCodeSizeByLength(len)` is a good refactor that centralizes code size logic. However, verify that `GetCodeSizeByLength` produces identical results for the FP32 path (the old formula) to avoid subtle serialization incompatibilities.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -344,10 +363,7 @@ SparseVectorDataCell<QuantTmpl, IOTmpl>::GetSparseVectorByInnerId(
         data->ids_ = nullptr;
         throw;
     }
-    for (uint32_t i = 0; i < data->len_; ++i) {
-        data->ids_[i] = entries[i].id;
-        data->vals_[i] = entries[i].val;
-    }
+    typed_quantizer().DecodeSparseVector(codes, data->ids_, data->vals_);
```
LHT129, line 366: [note] The refactoring of `GetSparseVectorByInnerId` to use `typed_quantizer().DecodeSparseVector()` instead of manual `BufferEntry` iteration is a clean improvement. This centralizes the decoding logic and ensures FP16 values are correctly converted to FP32 during deserialization.

src/algorithm/sindi/sindi_parameter.cpp
```diff
@@ -134,10 +134,12 @@ SINDIParameter::FromJson(const JsonType& json) {
     } else {
         rerank_type = SPARSE_RERANK_TYPE_FP32;
     }
-    CHECK_ARGUMENT(rerank_type == SPARSE_RERANK_TYPE_FP32 || rerank_type == SPARSE_RERANK_TYPE_DMQ8,
-                   fmt::format("rerank_type must be fp32 or dmq8, got {}", rerank_type));
+    CHECK_ARGUMENT(rerank_type == SPARSE_RERANK_TYPE_FP32 ||
+                       rerank_type == SPARSE_RERANK_TYPE_FP16 ||
+                       rerank_type == SPARSE_RERANK_TYPE_DMQ8,
+                   fmt::format("rerank_type must be fp32, fp16, or dmq8, got {}", rerank_type));
```
LHT129, line 140: [note] The updated `CHECK_ARGUMENT` error message that dynamically includes the invalid `rerank_type` value (via `fmt::format`) is a helpful improvement over the previous static message. This will make debugging configuration errors easier.

src/algorithm/sindi/sindi.cpp
```diff
@@ -688,7 +692,17 @@ SINDI::UpdateVector(int64_t id, const DatasetPtr& new_base, bool force_update) {
...
+            for (uint32_t i = 0; i < new_sv.len_; ++i) {
+                rounded_values[i] = generic::FP16ToFloat(generic::FloatToFP16(new_sv.vals_[i]));
+            }
+            SparseVector rounded_new_sv{new_sv.len_, new_sv.ids_, rounded_values.data()};
+            ret = is_subset_of_sparse_vector(old_sv, rounded_new_sv);
+        } else {
+            ret = is_subset_of_sparse_vector(old_sv, new_sv);
+        }
```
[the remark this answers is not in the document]
Roxanne0321: Thanks. This is a valid optimization opportunity, but not a correctness issue. The buffer is scoped to the UpdateVector check; reusing mutable scratch state would broaden this PR and introduce additional lifetime/concurrency considerations without a demonstrated regression. No code change in e559564e.

src/datacell/sparse_vector_datacell.inl
Roxanne0321: Agreed. The decoder centralization keeps FP32 and FP16 conversion in one implementation and the existing SparseDataCell FP16 and SINDI tests cover the behavior. No additional code change is needed.

src/datacell/sparse_vector_datacell.inl
Roxanne0321: The current upper bound intentionally tracks the currently supported enum values. A named maximum would add another source of truth without changing behavior; adding a future value still requires explicit deserialization and compatibility handling. No code change in e559564e.

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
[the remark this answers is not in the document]
Roxanne0321: Fixed in e559564e by adding an explicit #include <string> for std::string used by FromJson(). The affected datacell and algorithm targets build successfully, and the focused parameter test passes with 4 assertions.

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
Roxanne0321: This duplicate suggestion is covered by e559564e: sparse_quantizer_parameter.h now explicitly includes <string>. The focused parameter test passes with 4 assertions.

src/algorithm/sindi/sindi_parameter.cpp
Roxanne0321: Agreed. Including the invalid rerank_type in the error makes configuration failures easier to diagnose, and no further implementation change is needed.

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
Roxanne0321: Also fixed in e559564e by adding the direct #include <string> dependency. The focused SparseVectorDataCellParameter FP16 value type test passes with 4 assertions.

src/datacell/sparse_vector_datacell.inl
Roxanne0321: Verified: the FP32 path returns sizeof(uint32_t) + length * sizeof(BufferEntry), which is 4 + 8*length and exactly matches the previous (length * 2 + 1) * sizeof(uint32_t) formula. Existing datacell and algorithm builds/tests pass, so no code change is needed.

src/quantization/sparse_quantization/sparse_quantizer_parameter.h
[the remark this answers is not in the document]
Roxanne0321: Fixed in e559564e by adding an explicit #include <string> for std::string used by FromJson(). The affected datacell and algorithm targets build successfully, and the focused parameter test passes with 4 assertions.

src/algorithm/sindi/sindi.cpp
Roxanne0321: This is the same optimization opportunity as the Copilot comment above. It is not a correctness issue; the buffer is scoped to the UpdateVector check, while reusable mutable scratch state would broaden this PR and introduce additional lifetime/concurrency considerations. No code change in e559564e.

src/quantization/sparse_quantization/sparse_quantizer.h
```diff
@@ -199,7 +252,65 @@ SparseQuantizer<metric>::ComputeImpl(const uint8_t* codes1, const uint8_t* codes
...
         }
     }
-    return 1 - inner_product;
+    return 1.0F - inner_product;
+}
+
+template <MetricType metric>
+float
+SparseQuantizer<metric>::ComputeFP32QueryToFP16Codes(const uint8_t* query_codes,
+                                                     const uint8_t* base_codes) {
```
LHT129, line 261: [note] `ComputeFP32QueryToFP16Codes` performs FP16 to FP32 conversion on every base value for every intersecting term during the merge loop. For queries with many matching terms, this could become a hot path. Consider whether the FP16 values could be pre-converted to FP32 and cached in a small buffer before the merge loop, or whether SIMD-accelerated FP16 dot-product (e.g., `fp16_simd.h` already provides `FP16ComputeIP`) could be used here for additional throughput.

src/quantization/sparse_quantization/sparse_quantizer.h
```diff
@@ -199,7 +252,65 @@ SparseQuantizer<metric>::ComputeImpl(const uint8_t* codes1, const uint8_t* codes
...
+            ++base_index;
+        }
+    }
+    return 1.0F - inner_product;
+}
+
+template <MetricType metric>
+float
+SparseQuantizer<metric>::ComputeFP16Codes(const uint8_t* codes1, const uint8_t* codes2) {
+    const uint32_t length1 = *reinterpret_cast<const uint32_t*>(codes1);
```
LHT129, line 290: [note] `ComputeFP16Codes` converts both sides from FP16 to FP32. For pair-wise distance computation this is fine since it is not on the critical search path, but the same observation about SIMD acceleration applies if this function ever becomes a bottleneck.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -161,7 +161,20 @@ SparseVectorDataCell<QuantTmpl, IOTmpl>::Deserialize(LvalueOrRvalue<StreamReader
         // New 64-bit format. Layout written by Serialize().
         uint32_t version = 0;
         StreamReader::ReadObj(reader, version);
-        if (version != SERIALIZE_FORMAT_VERSION_V2) {
+        if (version == SERIALIZE_FORMAT_VERSION_V2) {
+            CHECK_ARGUMENT(not typed_quantizer().IsFP16(),
+                           "SparseVectorDataCell v2 payload requires fp32 sparse values");
+        } else if (version == SERIALIZE_FORMAT_VERSION_V3) {
```
LHT129, line 167: [suggestion] The V2 format rejection for FP16 is correct, but the error message could be more actionable. Consider adding a hint like "deserialize with an fp32-configured datacell or upgrade the payload to V3" so operators who encounter this during a rollback understand their options.

src/datacell/sparse_vector_datacell.inl
```diff
@@ -161,7 +161,20 @@ SparseVectorDataCell<QuantTmpl, IOTmpl>::Deserialize(LvalueOrRvalue<StreamReader
...
+            uint8_t value_type = 0;
+            StreamReader::ReadObj(reader, value_type);
+            CHECK_ARGUMENT(value_type <= static_cast<uint8_t>(SparseQuantizerValueType::FP16),
+                           "invalid SparseVectorDataCell sparse value type");
```
LHT129, line 171: [note] The validation `value_type <= static_cast<uint8_t>(SparseQuantizerValueType::FP16)` is future-proof: it will automatically reject any new enum value larger than FP16 until the deserialization code is updated. Good defensive pattern.

src/algorithm/sindi/sindi_parameter.cpp
```diff
@@ -134,10 +134,12 @@ SINDIParameter::FromJson(const JsonType& json) {
...
     CHECK_ARGUMENT(use_reorder || rerank_type == SPARSE_RERANK_TYPE_FP32,
-                   "rerank_type=dmq8 requires use_reorder=true");
+                   fmt::format("rerank_type={} requires use_reorder=true", rerank_type));
```
LHT129, line 143: [suggestion] The validation `CHECK_ARGUMENT(use_reorder || rerank_type == SPARSE_RERANK_TYPE_FP32, ...)` correctly enforces that FP16 and DMQ8 both require reorder. However, the error message says "rerank_type={} requires use_reorder=true" which is generic. Consider adding a more specific message explaining *why* reorder is required (e.g., the forward store is the only place FP16 values are stored; without reorder there is no forward store to read from).

src/algorithm/sindi_v2/sindi_v2_parameter.cpp
```diff
@@ -153,10 +153,12 @@ SINDIV2Parameter::FromJson(const JsonType& json) {
...
     }
-    CHECK_ARGUMENT(rerank_type == SPARSE_RERANK_TYPE_FP32 || rerank_type == SPARSE_RERANK_TYPE_DMQ8,
-                   fmt::format("rerank_type must be fp32 or dmq8, got {}", rerank_type));
+    CHECK_ARGUMENT(rerank_type == SPARSE_RERANK_TYPE_FP32 ||
+                       rerank_type == SPARSE_RERANK_TYPE_FP16 ||
+                       rerank_type == SPARSE_RERANK_TYPE_DMQ8,
+                   fmt::format("rerank_type must be fp32, fp16, or dmq8, got {}", rerank_type));
     CHECK_ARGUMENT(use_reorder || rerank_type == SPARSE_RERANK_TYPE_FP32,
-                   "rerank_type=dmq8 requires use_reorder=true");
+                   fmt::format("rerank_type={} requires use_reorder=true", rerank_type));
```
LHT129, line 162: [suggestion] Same suggestion as SINDI: the error message for `rerank_type={} requires use_reorder=true` could explain the rationale to help users self-diagnose.

src/simd/generic.cpp
```diff
@@ -203,28 +203,75 @@ FloatToBF16(const float fp32_value) {
 
 float
 FP16ToFloat(const uint16_t fp16_value) {
-    uint32_t sign = (fp16_value >> 15) & 0x1;
```
LHT129, line 206: [note] The `FP16ToFloat` and `FloatToFP16` implementations handle subnormals, NaN, and infinity correctly with round-to-nearest-even. The subnormal handling in `FloatToFP16` (lines checking `0x1p-24F` and `0x1p-25F`) is particularly important for sparse vectors where near-zero values are common. The test in `sparse_vector_datacell_test.cpp` explicitly verifies this boundary behavior — nice work.

src/algorithm/sindi/sindi_test.cpp
```diff
@@ -946,6 +947,95 @@ TEST_CASE("SINDI Basic Test", "[ut][SINDI]") {
     }
 }
 
+TEST_CASE("SINDI FP16 rerank build search add and serialize", "[ut][SINDI][fp16]") {
+    auto allocator = SafeAllocator::FactoryDefaultAllocator();
```
LHT129, line 951: [note] The FP16 rerank test covers build, search, add, serialize/deserialize, CalcDistanceById, EstimateMemory, and UpdateVector — excellent coverage. The memory estimation assertion (`fp32_memory - fp16_memory == estimate_count * 4`) correctly validates the 2-byte-per-value savings (4 bytes per float to 2 bytes per half, so 4 bytes saved per entry when counting both id and value overhead is 8 to 6, but the per-vector savings is 4 bytes per value).

LHT129 (commented):
Thank you for this well-structured PR adding FP16 reranking support to SINDI indexes. The implementation is clean and the test coverage is thorough. I have a few observations and suggestions below.

**Summary of findings:**

The PR adds FP16 half-precision storage for the SINDI/SINDIV2 rerank forward store, complementing the existing FP32 and DMQ8 options. Key changes include:
- `SparseQuantizerValueType::FP16` enum and parameter plumbing
- `ComputeFP32QueryToFP16Codes` / `ComputeFP16Codes` with FP16 to FP32 conversion during scoring
- V3 serialization format for `SparseVectorDataCell` with a `value_type` byte
- Parameter validation requiring `use_reorder=true` for FP16 rerank
- Comprehensive tests across unit, parameter, serialization, and integration levels

The implementation correctly keeps query values and accumulation in FP32 while only storing document values in FP16, which is the right trade-off for precision. The backward-compatibility handling (V2 format rejects FP16, legacy format rejects FP16) is properly enforced with clear error messages.

I found a few areas that deserve attention — see the inline comments below.

src/datacell/sparse_vector_datacell_test.cpp
```diff
@@ -128,6 +133,152 @@ TEST_CASE("SparseDataCell Basic Test", "[ut][SparseDataCell] ") {
     }
 }
 
+TEST_CASE("SparseDataCell FP16 values", "[ut][SparseDataCell][fp16]") {
+    auto allocator = SafeAllocator::FactoryDefaultAllocator();
```
LHT129, line 137: [note] The FP16 datacell test is thorough: it validates code layout (sorted ids, FP16 values, alignment), roundtrip through GetSparseVectorByInnerId, Query accuracy with FP16 rounding, ComputePairVectors, serialization header (sentinel + version + value_type byte), deserialization mismatch detection, and boundary value handling (zero, subnormal threshold). This is a model test for new serialization formats.

src/algorithm/sindi/sindi_parameter_test.cpp
```diff
@@ -292,6 +296,16 @@ TEST_CASE("SINDI Index Parameters Compatibility Test", "[ut][SINDIParameter]") {
         auto sindi_param1 = std::make_shared<vsag::SINDIParameter>();
         REQUIRE_THROWS(sindi_param1->FromString(param_str1));
     }
+
+    SECTION("fp16 requires reorder") {
+        SINDIDefaultParam param1;
```
LHT129, line 301: [suggestion] The test verifies that `fp16 requires reorder` throws, but does not test the case where `rerank_type="fp16"` is combined with `use_reorder=false` and `use_quantization="fp16"` simultaneously. While the current validation catches this, adding a test for the combined invalid case would guard against future refactoring that might accidentally allow it.

src/algorithm/sindi/sindi_test.cpp
```diff
@@ -946,6 +947,95 @@ TEST_CASE("SINDI Basic Test", "[ut][SINDI]") {
...
+    auto fp32_parameter = std::make_shared<SINDIParameter>();
+    auto fp32_json = parameter->ToJson();
+    fp32_json[SPARSE_RERANK_TYPE].SetString(SPARSE_RERANK_TYPE_FP32);
+    fp32_parameter->FromJson(fp32_json);
+    SINDI fp32(fp32_parameter, common_param);
+    constexpr uint64_t estimate_count = 10'000;
+    const uint64_t fp16_memory = loaded.EstimateMemory(estimate_count);
+    const uint64_t fp32_memory = fp32.EstimateMemory(estimate_count);
+    REQUIRE(fp16_memory < fp32_memory);
+    REQUIRE(fp32_memory - fp16_memory == estimate_count * 4);
```
LHT129, line 1036: [suggestion] The exact memory difference assertion depends on avg_doc_term_length=3 producing FP16 aligned code size exactly 4 bytes smaller than FP32. If avg_doc_term_length changes, the assertion would break. The REQUIRE(fp16_memory < fp32_memory) above is sufficient. Consider removing the exact equality assertion.

src/quantization/sparse_quantization/sparse_quantizer.h
Roxanne0321: Thanks for the note. `FP16ComputeIP` from `fp16_simd.h` is not applicable on this path: it is a dense, dim-wise half x half inner product (`float(const uint8_t*, const uint8_t*, uint64_t dim)`), whereas `ComputeFP32QueryToFP16Codes` merge-joins FP32 query `BufferEntry` records against the sparse `(uint32 ids, uint16 values)` layout, so there is no dense dimension to feed that kernel.

Pre-converting the base values would mean allocating and filling an FP32 scratch buffer inside per-candidate scoring (`ComputeDistImpl` -> `SearchBatchDistImpl`), trading a few integer ops in `generic::FP16ToFloat` (branch-light bit manipulation, with a short normalization loop only for subnormals) for an allocation plus a full copy per candidate. The FP32 sparse path is likewise a scalar merge join, so the FP16 path is not a regression against the existing backend. No code change in e559564e; if FP16 rerank ever shows a measurable scoring cost, an id/value-aware SIMD kernel would be the right follow-up.

src/quantization/sparse_quantization/sparse_quantizer.h
Roxanne0321: Agreed, and this call site is even further off the hot path: `ComputeFP16Codes` is reached from `ComputeImpl` (quantizer-level pair distance), not from `ComputeDistImpl`. The same limitation as the note above applies: `FP16ComputeIP` is a dense dim-wise kernel and cannot consume the sparse merge-join layout. No code change in e559564e.

src/datacell/sparse_vector_datacell.inl
Roxanne0321: Thanks. The message already states the requirement the operator has to satisfy (`v2 payload requires fp32 sparse values`) and the failure is deterministic, so there is no ambiguity about which setting is wrong. Wording would not change the rollback options either: a v2 payload cannot be read by an FP16-configured datacell at all, and producing a v3 payload requires an explicit re-serialization step rather than anything the deserializer can offer. Since this is user-facing polish rather than a behavior or compatibility fix, I kept the diff unchanged in e559564e.

src/datacell/sparse_vector_datacell.inl
Roxanne0321: Thanks for checking the bound. Keeping the check at the current enum maximum means any future value type is rejected until the deserialization path explicitly handles it. No action needed in e559564e.

src/algorithm/sindi/sindi_parameter.cpp
Roxanne0321: Thanks. The message already carries both the offending value and the exact fix (`rerank_type=fp16 requires use_reorder=true`), which is what a user needs to self-diagnose, and the rationale (FP16 base values only exist in the reorder forward store while the query stays FP32) is documented in the English and Chinese sindi docs updated by this PR. This is wording-only, so no code change in e559564e.

src/algorithm/sindi_v2/sindi_v2_parameter.cpp
Roxanne0321: Same as the SINDI case: the message names the offending value and the required setting, and the rationale is covered by the docs updated in this PR. No code change in e559564e.

src/simd/generic.cpp
Roxanne0321: Thanks for verifying the subnormal, NaN, and infinity handling and the round-to-nearest-even paths. No action needed in e559564e.

src/algorithm/sindi/sindi_test.cpp
Roxanne0321: Thanks. One clarification on the arithmetic for the memory assertion discussed below: the per-entry saving is 2 bytes (`BufferEntry` 8 B -> id 4 B + value 2 B = 6 B), and with `avg_doc_term_length=3` the alignment applied by `GetCodeSizeByLength` makes the FP16 rerank code 24 B versus 28 B for FP32, so the per-element saving in `EstimateMemory` is 4 B and the assertion is `estimate_count * 4`. I answered the assertion itself in the thread on line 1036.

src/datacell/sparse_vector_datacell_test.cpp
Roxanne0321: Thanks for the detailed review of the coverage. No action needed in e559564e.

src/algorithm/sindi/sindi_parameter_test.cpp
Roxanne0321: In SINDI there is no separate `use_quantization` -> `use_reorder` validation (that pairing only exists in SINDIV2 through `rerank_layout`), so `use_reorder=false` + `rerank_type="fp16"` hits the single check at `sindi_parameter.cpp:141` regardless of what `use_quantization` is set to. The proposed combination therefore exercises the same branch and the same message as the existing `fp16 requires reorder` section, so it would duplicate coverage rather than guard a distinct path. No code change in e559564e.

src/algorithm/sindi/sindi_test.cpp
Roxanne0321: The difference is exactly `num_elements * (fp32_code_size - aligned_fp16_code_size)`: with `avg_doc_term_length=3` that is `10000 * (28 - 24) = 40000`. The `block_size_limit()` term (`sindi.cpp:1936-1937`) is identical for both indexes and cancels in the subtraction, so the assertion is deterministic rather than environment-dependent. It intentionally pins the alignment-aware FP16 sizing (`sindi.cpp:1929-1932`) that mirrors `GetCodeSizeByLength` rounding up to `alignof(uint32_t)`; `REQUIRE(fp16_memory < fp32_memory)` alone would not catch a regression that drops that rounding. No code change in e559564e.

src/simd/generic.cpp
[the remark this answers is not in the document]
Roxanne0321: Updated in 340864fe per maintainer direction: the shared generic FloatToFP16/FP16ToFloat implementation change, including the quiet-NaN behavior and its dedicated SIMD assertions, has been reverted. The low-level converter remains unchanged in this PR; any correction to that shared utility should be proposed and reviewed separately. Focused SIMD, SparseDataCell FP16, and SINDI FP16 tests pass.

src/simd/generic.cpp
Roxanne0321: Scope update in 340864fe: the generic FloatToFP16/FP16ToFloat rewrite and the IEEE boundary test were removed at maintainer request so this feature PR does not alter the shared low-level converter. SINDI FP16 continues to use the repository existing conversion behavior. Focused SIMD, SparseDataCell FP16, and SINDI FP16 tests pass.

src/datacell/sparse_vector_datacell_test.cpp
Roxanne0321: Scope update in 340864fe: the SparseDataCell zero/subnormal assertions tied to the reverted generic FP16 conversion rewrite were removed. The remaining FP16 datacell test still covers layout/alignment, encode/decode consistency under the existing converter, query/pair distance, serialization, and value-type validation; it passes with 24 assertions.

Roxanne0321:
Scope update in 340864fe: per maintainer direction, this PR no longer changes the shared generic FloatToFP16/FP16ToFloat implementation. The dedicated IEEE boundary test and the SparseDataCell zero/subnormal assertions tied to that rewrite were also removed. The SINDI FP16 feature continues to use the repository existing converter, while its feature-specific layout, serialization, scoring, and UpdateVector behavior remain covered. Local validation passed: clang-format 15 dry-run, git diff --check, SIMD (64 cases), SparseDataCell FP16 (24 assertions), and SINDI FP16 (11 assertions).

LHT129 (commented):
## Review Summary

I have completed a thorough review of the FP16 reranking support for SINDI indexes. The implementation is well-structured and the code quality is high.

### What was reviewed
- Core FP16 quantizer implementation (`sparse_quantizer.h`): encoding, decoding, distance computation, code size calculation with alignment
- Data cell integration (`sparse_vector_datacell.h/.inl`): V3 serialization with value_type validation, InsertVector/GetSparseVectorByInnerId refactoring
- SINDI integration (`sindi.cpp`): create_rerank_flat, UpdateVector FP16 rounding, EstimateMemory alignment
- SINDI V2 integration (`sindi_v2.cpp`): create_rerank_flat, EstimateMemory alignment
- Parameter handling (`sparse_quantizer_parameter.h`, `sindi_parameter.cpp`, `sindi_v2_parameter.cpp`)
- Test coverage across all layers (sindi_test, sindi_v2_test, sparse_vector_datacell_test, parameter tests)
- Documentation updates

### Key observations
1. **FP16 code layout alignment**: `GetCodeSizeByLength` correctly rounds up to `alignof(uint32_t)` (4 bytes) and zeroes padding bytes in `EncodeOneImpl`. Both `EstimateMemory` implementations in sindi.cpp and sindi_v2.cpp use the same alignment logic — consistent.

2. **Serialization compatibility**: V3 format properly validates value_type on deserialization, checking both upper bound and exact match with the configured quantizer. Legacy and V2 formats correctly reject FP16 payloads.

3. **UpdateVector FP16 handling**: Rounds new values through FP16 before subset comparison, ensuring the update check is consistent with stored precision. The per-call allocation is scoped and acceptable for correctness.

4. **Test coverage**: Comprehensive — covers FP16 encode/decode round-trip, serialization/deserialization, distance computation accuracy (with tolerance), statistics backend reporting, memory estimation comparison, parameter validation, and compatibility checks. The SINDI V2 "Top Terms Rerank Layout" test now parameterizes over fp32/fp16.

5. **All previously identified issues have been addressed** in follow-up commits (alignment, NaN handling, test precision, hardcoded constants, etc.).

### Verdict
No new substantive issues found. The implementation is correct, well-tested, and ready to merge.

wxyucs (approved):
lgtm

PR #2959, opened 2026-09-15
perf(pyramid): accelerate RaBitQ split construction with scalar SIMD

LightWant:
## Change Type

- [x] Improvement/Refactor

## Linked Issue

Related: #2861. This separate PR covers ordinary Pyramid Build, including NSW and non-MRLE RaBitQ split construction.

## What Changed

Pyramid RaBitQ split construction previously searched packed codes even when scalar SIMD construction was available. Reuse HGraph's existing optimized-build interface, build computers and distance provider for ordinary NSW/ODescent builds, then finalize the packed split codes before returning.

Reuse Pyramid's batch encoding for ODescent and avoid its extra SQ8 build copy during the scalar session. Drain all submitted ODescent workers before propagating errors or releasing temporary storage. Preserve sequential encoding when logical datacell capacity has no physical backing yet.

## Test Evidence

- [x] `make fmt` (clang-format 15)
- [ ] Full `make lint` (targeted clang-tidy 15 results below)
- [x] `make test` with focused cases below
- [ ] Full instrumented coverage report (not run locally)
- [x] Documentation structure/link check

Tests cover NSW/ODescent, L2/IP/cosine, fast encoding on/off, raw storage on/off, hierarchy layouts, optimized-session activation, serialization/reload, subsequent Add, and cleanup after worker/build exceptions. A small-capacity raw-storage regression also passed 100 standalone repetitions.

```sh
make test CASE='"[pyramid],[PyramidParameters],[pyramid_build_cache],[ODescent],[optimized_build]"' COMPILE_JOBS=16
make fmt
node .github/scripts/check-docs.mjs
```

Results: 59 unit test cases / 9,397 assertions and 74 functional test cases / 324,053 assertions passed. The eval-monitor binary had no matching cases. Targeted clang-tidy 15 passed for both changed production translation units; the struct naming checker passed for 1,009 files. Full repository lint and instrumented coverage were not run locally.

Local dependency overrides reused cached Catch2/CRoaring and installed OpenBLAS.

## Compatibility Impact

- Public API and serialized format: unchanged.
- Build uses temporary scalar codes for supported configurations, then restores the existing packed query representation. Approximate graph edges may differ.

50,000 × 1,536 vectors from `openai-1536-50k-angular.hdf5`; cosine, NSW, max_degree=32, ef_construction=100, build_thread_count=16. Single path `all`, no_build_levels=[0], split 3+5, query bits=32, reorder enabled, fast encoding enabled with 6 rounds, FHT/raw storage disabled.

AMD EPYC 9T24, Release (-O3/NDEBUG), affinity to 16 physical cores; BLAS/OMP/MKL threads=1. Three fresh builds per version, interleaved ABBAAB. Timed the complete Build(), including training/encoding/graph construction/final packing, excluding HDF5 loading and queries. Baseline: 2674b8c49b38a0621767bd387938b8f243df8d23; optimized: that commit plus this patch. The intervening upstream commit only changes testing documentation.

| Measurement | Before | After |
|---|---:|---:|
| Build times, seconds | 333.592 / 333.520 / 332.964 | 11.987 / 12.213 / 12.006 |
| Median build time | 333.520 s | 12.006 s |
| Median Recall@10 (1,000 queries, ef_search=100) | 0.9754 | 0.9742 |
| Median peak process RSS | 646.4 MiB | 720.3 MiB |
| Final index memory | 265.5 MiB | 265.5 MiB |

**27.78× faster construction (96.4% less time)** for this configuration. Recall varies across parallel builds; observed ranges were 0.9748–0.9760 before and 0.9742–0.9744 after. RSS includes loaded data and libraries. No performance claim is made for ODescent or multiple hierarchy graphs.

Worker futures are drained on both submission and execution failures before temporary codes are released.

## Documentation Impact

- [x] Updated English and Chinese RaBitQ quantization documentation.

## Risk and Rollback

- Medium: construction uses a different distance representation and temporary memory; graph topology and recall can vary.
- Revert this commit to restore the previous construction path. No serialized-index migration is required.

## Checklist

- [x] Added tests for new behavior and failure cleanup.
- [x] Considered API compatibility.
- [x] Updated relevant documentation.
- [x] Conventional commit with human DCO and AI attribution trailers.


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -281,14 +282,22 @@ Pyramid::search_graph_for_add(const GraphInterfacePtr& graph,
...
-    auto results =
-        searcher_->Search(graph, distance_provider, vl_guard.get(), search_param, nullptr, nullptr);
+    DistHeapPtr results;
+    if (optimized_build_codes_ != nullptr) {
+        FlattenDistanceProvider distance_provider(
+            codes, optimized_build_codes_->FactoryComputerForBuild(vector, inner_id));
+        results = searcher_->Search(
+            graph, distance_provider, vl_guard.get(), search_param, nullptr, nullptr);
+    } else {
+        FlattenIdDistanceProvider distance_provider(codes, inner_id);
```
LHT129, line 299: [suggestion] When `optimized_build_codes_ != nullptr` but `vector == nullptr`, `FactoryComputerForBuild` receives a null query pointer. The original code handled `vector == nullptr` by falling through to `FlattenIdDistanceProvider`, but the new optimized-build branch does not check for null before calling `FactoryComputerForBuild(vector, inner_id)`.

While current callers always pass non-null vectors during build, the function signature explicitly accepts nullptr. Consider either:
- Adding a `vector != nullptr` guard before entering the optimized branch, or
- Asserting/documenting that `vector` must be non-null when `optimized_build_codes_` is active.

```cpp
if (optimized_build_codes_ != nullptr) {
    assert(vector != nullptr);
    FlattenDistanceProvider distance_provider(
        codes, optimized_build_codes_->FactoryComputerForBuild(vector, inner_id));
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -22,6 +22,7 @@
 #include <exception>
 #include <limits>
 #include <nlohmann/json.hpp>
+#include <numeric>
 
 #include "algorithm/inner_index_interface.h"
 #include "analyzer/analyzer.h"
```
LHT129, line 29: [suggestion] The variable name `storage_preallocated` is misleading — it is `true` when `data_num` exceeds `max_capacity_`, meaning `resize()` will actually grow the stores (allocating physical backing). The name suggests the storage was already preallocated before this point, but the logic means the opposite: it is `true` precisely when the storage was *not* yet preallocated and needs to be grown by the following `resize()` call.

Consider renaming to something like `storage_will_grow` or `needs_resize` to make the intent clearer. The comment on line 93–94 already explains the rationale well, so a matching variable name would help readers connect the comment to the condition.

Current:
```cpp
// A datacell's initial logical capacity may not have physical backing yet. As in
// prepare_add_batch(), parallel writes are safe only after Resize actually grows each store.
const bool storage_preallocated =
    data_num > static_cast<int64_t>(base_codes_->max_capacity_) and ...;
```

Suggested:
```cpp
// A datacell's initial logical capacity may not have physical backing yet. As in
// prepare_add_batch(), parallel writes are safe only after Resize actually grows each store.
const bool storage_will_grow =
    data_num > static_cast<int64_t>(base_codes_->max_capacity_) and ...;
```

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -602,6 +611,13 @@ Pyramid::build_by_odescent(const DatasetPtr& base) {
     const auto* data_ids = base->GetIds();
     const auto* source_ids = base->GetSourceID();
 
+    // A datacell's initial logical capacity may not have physical backing yet. As in
```
LHT129, line 616: [note] The `storage_preallocated` name is carried over from `prepare_add_batch()` and is consistent with the existing codebase convention. The comment on lines 93–94 explains the semantics clearly — no change needed, just noting for awareness.

src/algorithm/pyramid/pyramid.cpp
LightWant: Checked the optimized-build implementation; a null query is safe here. optimized_build_codes_ is assigned only after BeginOptimizedBuild() succeeds and is reset when the session finishes or aborts. In the active RaBitQ split session, FactoryComputerForBuild(query, inner_id) reads the scalar code and code sum by inner_id and does not dereference query (src/datacell/rabitq_split_datacell.h). The query-based fallback is used only outside that session. An assertion would unnecessarily disallow a supported ID-based query, so I am leaving this branch unchanged.

src/algorithm/pyramid/pyramid.cpp
LightWant: Keeping the existing name, consistent with prepare_add_batch() and AddBatch::storage_preallocated, as also noted in your follow-up discussion_r4011890833. The condition is evaluated before resize(), but the value is consumed after resize() has allocated the backing storage. The existing comment documents that ordering. No rename is needed.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2086,11 +2117,33 @@ Pyramid::Build(const DatasetPtr& base) {
...
+        if (graph_type_ == GRAPH_TYPE_VALUE_NSW) {
+            result = this->Add(base);
+        } else {
+            this->Train(base);
+            result = this->build_by_odescent(base);
+        }
+        if (optimized_build_codes_ != nullptr) {
+            optimized_build_codes_->FinalizeOptimizedBuild();
+            optimized_build_codes_.reset();
+        }
```
LHT129, line 2137: [note] If `FinalizeOptimizedBuild()` throws after partially packing scalar codes, the catch block calls `AbortOptimizedBuild()` on the same session. Depending on the implementation, `AbortOptimizedBuild` may not be safe to call after a failed `FinalizeOptimizedBuild` — the session may already be in an indeterminate state (e.g. some records packed, others still scalar). Consider whether `AbortOptimizedBuild` is designed to be idempotent after a failed finalization, or whether the catch block should simply reset the pointer without calling abort.

Current code (`pyramid.cpp:2137-2144`):
```cpp
if (optimized_build_codes_ != nullptr) {
    optimized_build_codes_->FinalizeOptimizedBuild();
    optimized_build_codes_.reset();
}
return result;
} catch (...) {
    if (optimized_build_codes_ != nullptr) {
        optimized_build_codes_->AbortOptimizedBuild();
        optimized_build_codes_.reset();
    }
    throw;
}
```

wxyucs (approved):
lgtm

Issue #2835, opened 2026-09-02
[improve](hgraph): support efficient graph index merging


ykindred:
### Kind of improvement

perf

### Affected area

hgraph

### Summary

HGraph currently supports index merging, but the existing merge implementation does not implement the graph-reuse strategy described in FGIM.

For workloads where multiple independently built HGraph indexes need to be consolidated, rebuilding or extensively reconstructing the merged graph may introduce significant computation and memory overhead, while the input indexes already contain useful local graph structures.

We propose adding an efficient graph index merging strategy for HGraph that reuses existing intra-index graph structures, establishes cross-index connections, and performs necessary graph refinement to maintain search quality.

The initial scope is dense-vector HGraph indexes over disjoint datasets. FGIM can be used as the main reference for the implementation.

### Expected benefit

The proposed merge strategy is expected to reduce the cost of consolidating multiple existing HGraph indexes by reusing graph structures that have already been constructed.

Compared with rebuilding an HGraph index over the complete dataset, the implementation should aim to:

- reduce merge/build time and distance computations;
- reuse valid intra-index graph edges instead of reconstructing them;
- preserve comparable Recall-QPS performance of the merged index;
- avoid significant query-performance regression;
- keep peak memory usage within a practical range.

The implementation will be evaluated against full rebuild on multiple datasets and different input-index configurations.

### Benchmark evidence

No VSAG before/after benchmark is available yet because this issue tracks the proposed implementation.

The implementation will be evaluated against full HGraph rebuild using reproducible experiments covering at least two dataset/configuration settings and different input-index sizes or shard counts.

Planned metrics include:

- merge/build time;
- speedup over full rebuild;
- number of distance computations, when available;
- Recall@K;
- QPS / Recall-QPS trade-off;
- peak memory usage.

The benchmark results and exact parameters will be attached to the implementation PR.

### References

- FGIM: *A Fast Graph-based Indexes Merging Framework for Approximate Nearest Neighbor Search*, SIGMOD 2026.
- #2128 - discussion about the difference between the current VSAG Merge implementation and FGIM.
- `src/algorithm/hgraph/hgraph.cpp`
- `src/algorithm/hgraph/hgraph_build.cpp`

### Additional context

Initial scope:

- Dense-vector HGraph indexes.
- Two or more independently built indexes over disjoint datasets.
- Reuse the existing HGraph merge API where possible.
- Add correctness tests and reproducible performance evaluation.

[...]

PR #2961, opened 2026-09-15
feat(hgraph): add FGIM initial k-NNG construction

ykindred:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

Related to #2835

## What Changed
- Add the initial k-NNG construction stage for FGIM-based HGraph merging.
- Support multiple homogeneous HGraph source indexes in a merged inner-ID space.
- Reuse original bottom-graph neighbors and perform minimum cross-querying across the other source graphs.
- Add an internal HGraph CrossQuery path that reuses the existing HGraph search implementation while returning local inner IDs and distances directly.
- Add focused FGIM tests in the consolidated hgraph_fgim_test.cpp.

This PR only introduces the initial PGs-to-k-NNG stage. FGIM refinement, graph reconstruction, and public HGraph::Merge integration are intentionally left to follow-up PRs.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

FGIM-focused unit tests pass locally.

The tests cover:

- multiple HGraph sources
- merged inner-ID mapping
- original-neighbor candidates
- internal/public CrossQuery equivalence
- minimum-querying and top-k behavior
- external-label / inner-ID separation
- invalid and unsupported source configurations
- ID-range, duplicate, self-loop, distance and ordering invariants

The repository-wide non-daily functional suite is still running locally. Its final result will be reported when available.

## Compatibility Impact

- API/ABI compatibility: none
- Behavior changes: none; this PR does not modify the public HGraph::Merge behavior

- Performance impact: the internal CrossQuery path avoids public search result packaging and unnecessary external-label round trips for FGIM.
- Concurrency/thread-safety impact: none; the current FGIM implementation is single-threaded.

## Documentation Impact

- [x] No docs update needed
- [ ] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Other: 

## Risk and Rollback

- Risk level: low
- Rollback plan: revert this PR; no existing public HGraph behavior depends on the new internal FGIM stage.

## Checklist


ykindred:
This PR is intended as `kind/improvement` for #2835.
Could a maintainer please add `kind/improvement` and `version/1.2`?

src/algorithm/hgraph/hgraph.h
```diff
@@ -80,6 +81,8 @@ class HGraph : public InnerIndexInterface {
                                  const IndexCommonParam& common_param);
 
     friend class HGraphAnalyzer;
+    friend class HGraphFGIM;
+    friend class HGraphFGIMTest;
```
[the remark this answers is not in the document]
ykindred: The fixture friendship supports independent inspection of labels, vectors, and bottom neighbors without adding public getters. We agree that it couples the header to the fixture name. However, ENABLE_TESTS is not uniformly propagated to all relevant targets, so conditionally declaring the friend is not a safe drop-in change.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+                                                   param,
+                                                   VisitedListPtr{},
+                                                   &ctx);
+        if (!route->Empty()) {
+            param.ep = route->Top().second;
+        }
+    }
+    const int64_t topk = std::min(l, target.GetNumElements());
+    param.ef = l;
+    param.topk = l;
```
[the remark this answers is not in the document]
ykindred: This intentionally mirrors HGraph’s public search path: the output count is clamped, while the internal search width remains L. The FP32 BasicSearcher uses dynamically growing heaps rather than allocating an L-sized buffer. Existing equivalence tests cover L greater than the target size. We therefore prefer to retain the current search semantics unless a measurable issue is demonstrated.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+    while (result->Size() > static_cast<uint64_t>(topk)) {
+        result->Pop();
+    }
+    FGIMNeighborList neighbors;
+    neighbors.reserve(result->Size());
+    while (!result->Empty()) {
+        const auto record = result->Top();
+        result->Pop();
+        if (!std::isnan(record.first)) {
```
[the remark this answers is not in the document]
ykindred: CrossQuery follows public HGraph search behavior, which preserves infinity when no threshold is provided. BuildInitialKnnGraph then explicitly rejects non-finite FGIM candidates. Replacing this with silent filtering would change an error into a potentially successful but incomplete candidate list. We prefer to retain explicit rejection rather than silently discard these results.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+            source.GetVectorByInnerId(local_u, vector.data());
+            for (uint64_t j = 0; j < source_graphs.size(); ++j) {
+                if (j != i) {
+                    const auto cross = CrossQuery(*source_graphs[j], vector.data(), query_count);
+                    for (const auto& neighbor : cross) {
+                        // CrossQuery IDs are target-local; apply the target prefix exactly once.
+                        candidates.push_back({offsets[j] + neighbor.id, neighbor.distance});
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
ykindred: Reserving may reduce reallocations, but these vectors are the persistent output rows, and resize(k) does not release capacity. Reserving a loose candidate upper bound for every node may increase retained memory. We suggest evaluating allocation count and peak memory before choosing per-row reservation or a reusable scratch buffer.

src/algorithm/hgraph/hgraph_fgim.cpp
LHT129, line 152: [suggestion] `std::isnan` only filters NaN but not infinity. `BuildInitialKnnGraph` later checks `std::isfinite` on all candidates, so if `CrossQuery` returns an `inf` distance (e.g. from a degenerate distance computation), the `CHECK_ARGUMENT(std::isfinite(...))` in `BuildInitialKnnGraph` will fire an assertion failure. Consider using `std::isfinite` here for consistency:

```cpp
if (std::isfinite(record.first)) {
    neighbors.push_back({record.second, record.first});
}
```

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+    const auto query_count = static_cast<int64_t>(k / other_count + (k % other_count != 0));
+    FGIMKnnGraph graph(offsets.back());
+    std::vector<float> vector(source_graphs.front()->dim_);
+
+    for (uint64_t i = 0; i < source_graphs.size(); ++i) {
+        const auto& source = *source_graphs[i];
+        for (InnerIdType local_u = 0; local_u < source.GetNumElements(); ++local_u) {
+            auto& candidates = graph[offsets[i] + local_u];
+            AppendOriginalCandidates(source, local_u, offsets[i], candidates);
+            source.GetVectorByInnerId(local_u, vector.data());
```
LHT129, line 177: [suggestion] The `static_cast<int64_t>(k / other_count + (k % other_count != 0))` can overflow when `k` is large and `other_count` is small. For example, with `k = std::numeric_limits<int64_t>::max()` (the validated upper bound) and `other_count = 1`, `k / other_count` exceeds `int64_t::max()`, making the `static_cast` implementation-defined (and typically wrapping to a negative value). Consider clamping or using a saturating cast:

```cpp
const uint64_t raw = k / other_count + (k % other_count != 0);
const auto query_count = static_cast<int64_t>(std::min(raw, static_cast<uint64_t>(std::numeric_limits<int64_t>::max())));
```

src/algorithm/hgraph/hgraph_fgim.h
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
```
LHT129, line 18: [suggestion] `#include <string>` is unused in this header. No `std::string` appears in the class declaration, forward declarations, or type aliases. Removing it reduces unnecessary transitive includes for every translation unit that includes this header.

```cpp
#include <cstddef>
#include <vector>

#include "basic_types.h"
#include "container_types.h"
```

src/algorithm/hgraph/hgraph_fgim.h
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+#include "basic_types.h"
+#include "container_types.h"
+#include "vsag/dataset.h"
```
LHT129, line 23: [suggestion] `#include "vsag/dataset.h"` is unused in this header. The header only forward-declares `HGraph` and `SearchStatistics`; `Dataset` is only needed in the `.cpp` and test files. Including it here unnecessarily widens the compile-time dependency graph.  Move this include to `hgraph_fgim.cpp` if needed there (it already includes `hgraph.h` which likely pulls in `Dataset` transitively).

src/algorithm/hgraph/hgraph_fgim.h
```diff
@@ -0,0 +1,67 @@
...
+    static FGIMKnnGraph
+    BuildInitialKnnGraph(const Vector<const HGraph*>& source_graphs, std::size_t k);
+
+    // Internal-only search, not a public index API. Same source preconditions as Build.
+    // Optional statistics are for diagnostics; normal FGIM runs do not collect them.
+    static FGIMNeighborList
+    CrossQuery(const HGraph& target,
+               const float* query,
+               int64_t l,
+               SearchStatistics* stats = nullptr);
```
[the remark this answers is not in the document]
ykindred: Agreed. 
Clarified the internal CrossQuery contract: the caller must validate the target as FGIM-compatible and provide a query buffer containing at least target.dim_ floats. The comment now explicitly states that CrossQuery does not repeat full source validation. No per-query validation or search behavior was changed.

test/hgraph_fgim_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+            graph.data_type_ = DataTypes::DATA_TYPE_INT8;
+        } else if (condition == "L2") {
+            graph.metric_ = MetricType::METRIC_TYPE_IP;
+        } else if (condition == "dimensions") {
+            ++graph.dim_;
+        } else if (condition == "deduplicate") {
+            graph.deduplicate_storage_ = true;
+        } else if (condition == "base storage") {
+            graph.basic_flatten_codes_ = nullptr;
+        } else if (condition == "fully built") {
```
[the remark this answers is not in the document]
ykindred: Agreed.
Renamed the mutation selectors to “non-float32” and “non-L2” and separated them from the expected diagnostic substring. The tests still verify the existing “float32” and “L2” diagnostic text. All four focused PR1 FGIM tests passed (26,485 assertions).

src/algorithm/hgraph/hgraph.h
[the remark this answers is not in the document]
ykindred: Discussed Earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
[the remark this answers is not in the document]
ykindred: Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thanks for pointing this out. CrossQuery currently mirrors public HGraph search, which preserves infinity when no threshold is supplied. BuildInitialKnnGraph then rejects non-finite candidates through an INVALID_ARGUMENT exception, rather than an assertion abort. Filtering infinity here would change rejection into potentially successful construction with dropped candidates. Would you prefer retaining explicit rejection, possibly earlier, or intentionally allowing these candidates to be discarded?

src/algorithm/hgraph/hgraph_fgim.h
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
```
LHT129: [suggestion] The header includes `<string>` and `"vsag/dataset.h"` which are not referenced anywhere in this header file. These add unnecessary compile-time dependencies for every translation unit that includes `hgraph_fgim.h`. Consider removing them.

`<string>` is not used by any declaration in this header. `Dataset` is only used in the `.cpp` implementation file — the forward declaration of `HGraph` and `SearchStatistics` is sufficient for the header.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+                                     InnerIdType source_offset,
+                                     FGIMNeighborList& candidates) {
+    Vector<InnerIdType> neighbors(source.allocator_);
+    source.bottom_graph_->GetNeighbors(local_u, neighbors);
+    for (InnerIdType local_v : neighbors) {
+        CHECK_ARGUMENT(local_v < source.GetNumElements() && local_v != local_u,
+                       "FGIM encountered an invalid original neighbor");
+        // Both distance operands remain in the source-local internal ID space.
+        const float distance = source.basic_flatten_codes_->ComputePairVectors(local_u, local_v);
+        candidates.push_back({source_offset + local_v, distance});
```
LHT129, line 98: [note] `AppendOriginalCandidates` computes distances via `basic_flatten_codes_->ComputePairVectors()`, while `CrossQuery` may return reranked distances from `high_precise_codes_` or `raw_vector_` when `use_reorder_` is enabled. Although `ValidateSourcesAndGetOffsets` requires all three codec layers to be fp32, the distance values could theoretically differ if the code paths use different SIMD kernels or accumulation orders. In practice with fp32 this is unlikely to affect top-k selection, but it is worth documenting that FGIM assumes all three storage layers produce numerically identical L2 distances for the same vector pairs.

If this invariant is expected to hold permanently, consider adding a brief comment in `AppendOriginalCandidates` noting that original-neighbor distances intentionally use `basic_flatten_codes_` and are comparable to CrossQuery results because all storage is fp32.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thank you for pointing this out. This value is bounded by the validation before this expression: 0 < k <= INT64_MAX and other_count >= 1. Therefore ceil(k / other_count) <= k <= INT64_MAX.

src/algorithm/hgraph/hgraph_fgim.h
ykindred: Thank you. Fixed. Removed <string> from the FGIM production header and included it directly in hgraph_fgim_test.cpp.

src/algorithm/hgraph/hgraph_fgim.h
ykindred: Thank you. Fixed. Removed "vsag/dataset.h" from the FGIM production header and included it directly in hgraph_fgim_test.cpp.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+                (source->raw_vector_ != nullptr &&
+                 source->raw_vector_->GetQuantizerName() == QUANTIZATION_TYPE_VALUE_FP32),
+            "FGIM requires fp32 raw vector storage");
+        CHECK_ARGUMENT(
+            static_cast<uint64_t>(count) <= std::numeric_limits<InnerIdType>::max() - total,
+            "FGIM total node count exceeds the merged internal ID capacity");
+        CHECK_ARGUMENT(source->bottom_graph_ != nullptr &&
+                           source->bottom_graph_->TotalCount() == count &&
+                           source->basic_flatten_codes_->TotalCount() == count &&
+                           source->label_table_->GetTotalCount() == count,
```
LHT129, line 76: [suggestion] `label_table_->GetTotalCount()` is called without a prior null check on `label_table_`, while `bottom_graph_` and `basic_flatten_codes_` are explicitly checked for null on the same line. Although a properly built HGraph always has a valid `label_table_`, the inconsistency with the other null checks on this line is misleading. Consider adding `source->label_table_ != nullptr &&` for consistency and defense-in-depth.

```cpp
CHECK_ARGUMENT(source->bottom_graph_ != nullptr &&
                   source->label_table_ != nullptr &&
                   source->bottom_graph_->TotalCount() == count &&
                   source->basic_flatten_codes_->TotalCount() == count &&
                   source->label_table_->GetTotalCount() == count,
               "FGIM requires dense, fully built source HGraphs");
```

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+    param.rerank_topk = topk;
+    param.consider_duplicate = target.support_duplicate_;
+    param.is_inner_id_allowed = target.create_search_filter(nullptr, false);
+    ctx.distance_phase = DistanceEvaluationPhase::APPROXIMATE;
+    auto result = target.search_one_graph(
+        query, target.bottom_graph_, target.basic_flatten_codes_, param, VisitedListPtr{}, &ctx);
+    if (target.use_reorder_) {
+        target.reorder(query, target.get_reorder_codes(), result, topk, nullptr, ctx);
+    }
+    while (result->Size() > static_cast<uint64_t>(topk)) {
```
LHT129, line 146: [suggestion] After `target.reorder(...)` on line 142, the code proceeds directly to `result->Size()` and `result->Top()` without verifying that `result` is non-empty. While `search_one_graph` should always return at least the entry point, `reorder` may modify the result heap in ways that could theoretically leave it empty (e.g., if all candidates are filtered out by a distance threshold in future reorder implementations). Consider adding a defensive empty check after reorder:

```cpp
if (target.use_reorder_) {
    target.reorder(query, target.get_reorder_codes(), result, topk, nullptr, ctx);
    if (result->Empty()) {
        return {};
    }
}
```

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+    return neighbors;
+}
+
+FGIMKnnGraph
+HGraphFGIM::BuildInitialKnnGraph(const Vector<const HGraph*>& source_graphs, std::size_t k) {
+    const auto offsets = ValidateSourcesAndGetOffsets(source_graphs, k);
+    const uint64_t other_count = source_graphs.size() - 1;
+    // ceil(k / (m - 1)), without overflowing k + m - 2.
+    const auto query_count = static_cast<int64_t>(k / other_count + (k % other_count != 0));
+    FGIMKnnGraph graph(offsets.back());
```
LHT129, line 171: [note] `FGIMKnnGraph graph(offsets.back())` pre-allocates `offsets.back()` default-constructed `FGIMNeighborList` vectors upfront, then fills them one by one in the nested loop. For large merges (e.g., millions of nodes), this eagerly constructs many empty vectors that are immediately overwritten. Since `FGIMKnnGraph` is `std::vector<FGIMNeighborList>`, you could `reserve(offsets.back())` and `emplace_back()` each row in order to avoid the upfront default construction cost. This is a minor optimization and not blocking.

```cpp
FGIMKnnGraph graph;
graph.reserve(offsets.back());
// ... in the loop:
graph.emplace_back();
auto& candidates = graph.back();
```

src/algorithm/hgraph/hgraph_fgim.h
ykindred: Thank you. Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                }
+            }
+            for (const auto& candidate : candidates) {
+                CHECK_ARGUMENT(std::isfinite(candidate.distance), "FGIM requires finite distances");
+            }
+            std::sort(candidates.begin(),
+                      candidates.end(),
+                      [](const FGIMNeighbor& a, const FGIMNeighbor& b) {
+                          return a.distance < b.distance ||
```
LHT129, line 195: [note] In CrossQuery, the reorder call passes nullptr for stats while ctx.stats is available. If use_reorder_ is enabled and caller passes non-null SearchStatistics*, distance evaluations during reorder will not be counted. This is consistent with "normal FGIM runs do not collect them" but could surprise debugging with stats on a reorder-capable graph. Consider adding a brief comment, or passing ctx.stats for consistency.

src/algorithm/hgraph/hgraph.h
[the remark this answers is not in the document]
ykindred: Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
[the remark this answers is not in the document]
ykindred: Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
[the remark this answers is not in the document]
ykindred: Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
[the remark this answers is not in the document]
ykindred: FGIM currently requires L2 and fp32 storage, including the precise/raw layers when applicable. The normal HGraph insertion path writes the same input vector and internal ID to these layers, so the candidate distances represent the same squared-L2 metric rather than mixed quantization accuracies. Also, get_reorder_codes() currently selects base or high-precision storage. We can document this invariant without promising bitwise identity across all distance kernels.

src/algorithm/hgraph/hgraph_fgim.cpp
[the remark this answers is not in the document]
ykindred: Discussed earlier.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thank you. The nullptr argument is IteratorFilterContext*, not SearchStatistics*. Statistics are passed through QueryContext& ctx. FlattenReorder updates reorder_distance_count and invokes flatten_->Query with the same context under DistanceEvaluationPhase::RERANK, so reorder distance evaluations are already counted when tracking is enabled.

LHT129 (commented):
## Review Summary

I have completed a thorough review of this PR. The code quality is high overall.

### What was reviewed
- `src/algorithm/hgraph/hgraph_fgim.h` (67 lines) — FGIM class declaration
- `src/algorithm/hgraph/hgraph_fgim.cpp` (204 lines) — FGIM implementation
- `test/hgraph_fgim_test.cpp` (385 lines) — comprehensive test suite
- `src/algorithm/hgraph/hgraph.h` (+3 lines) — friend declarations
- CMakeLists.txt changes — build integration

### Observations

1. **Validation is thorough.** `ValidateSourcesAndGetOffsets` checks all preconditions (null, distinct, non-empty, data type, metric, dimension, storage format, deleted nodes, deduplicate, conjugate graph, capacity overflow, fully built). This is exactly the right level of defensive checking for an internal merge stage.

2. **Test coverage is excellent.** The test suite covers: multi-source k-NNG correctness, singleton sources with ceiling semantics, deterministic tie-breaking, all unsupported input rejection paths, CrossQuery vs. public search equivalence, and source immutability after FGIM operations.

3. **The `friend class HGraphFGIMTest`** declaration in `hgraph.h` was noted by Copilot. The existing codebase already uses `friend class HGraphAnalyzer` in the same location, so this follows established patterns. Consider gating behind `#if defined(ENABLE_TESTS)` in a follow-up if desired.

4. **The `param.ef` / `param.topk` clamping** in `CrossQuery` was noted by Copilot. When `l > target.GetNumElements()`, clamping `ef` to `topk` would reduce unnecessary search work. This is a minor optimization opportunity for a follow-up.

No blocking issues found. The implementation is clean, well-tested, and follows VSAG conventions.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thank you. The rows are filled in place through references, not overwritten. reserve/emplace_back would still construct one neighbor-list vector per node, but at a different point in the loop. Since the current indexed layout directly expresses the merged-ID mapping and no benefit has been measured, I suggest retaining it in this correctness-focused PR.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: An empty heap is already handled safely. Size() is valid on an empty heap, and Top() is only called inside while (!result->Empty()). The truncation loop is also skipped for an empty result. The function therefore already returns an empty list without requiring an additional post-reorder check.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thank you. Fixed. I added an explicit label_table_ != nullptr check before accessing GetTotalCount(), together with a regression test for a missing label table.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: Thank you. Fixed. I documented the invariant here: FGIM validates the relevant storage layers as FP32/L2 over the same vectors, so the original-neighbor distances are comparable with CrossQuery distances. The comment also explicitly avoids assuming bitwise-equal floating-point results.

src/algorithm/hgraph/hgraph_fgim.cpp
ykindred: We checked this against the existing HGraph search semantics and decided to keep the current behavior. We prefer explicit rejection over silently discarding non-finite candidates, as filtering them here would change the existing search semantics.

src/algorithm/hgraph/hgraph_fgim.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+// Own the result independently of the source indexes and their allocators.
+using FGIMNeighborList = std::vector<FGIMNeighbor>;
+using FGIMKnnGraph = std::vector<FGIMNeighborList>;
+
+class HGraphFGIM {
+public:
+    // Sources are non-owning, non-empty, disjoint HGraphs with successful Builds.
+    // No concurrent mutation is allowed. Only float32/L2/fp32 without deletion
```
LHT129: [note] The `k` parameter uses `std::size_t`, but the VSAG coding standards prefer `uint64_t` over `size_t` to avoid macOS compilation issues (see `docs/agents/coding-standards.md`). The internal logic already casts to `int64_t` for `query_count`, and the validation bounds `k` to `INT64_MAX`. Using `uint64_t` consistently would be more idiomatic for this codebase.

```cpp
// Current:
static FGIMKnnGraph
BuildInitialKnnGraph(const Vector<const HGraph*>& source_graphs, std::size_t k);

// Suggested:
static FGIMKnnGraph
BuildInitialKnnGraph(const Vector<const HGraph*>& source_graphs, uint64_t k);
```

This would also make the `#include <cstddef>` in the header unnecessary.

ykindred: Thank you. Updated k to uint64_t consistently in both declarations and definitions, including the validation helper. Replaced <cstddef> with <cstdint>. The existing bounds validation and algorithm behavior are unchanged.

LHT129 (commented):
Overall this is a well-structured PR with thorough input validation, clean separation of concerns, and comprehensive test coverage.

**Summary of review:**

The FGIM initial k-NNG construction is implemented as a stateless utility class (`HGraphFGIM`) with clear static methods. The validation in `ValidateSourcesAndGetOffsets` is exhaustive, covering data types, metrics, storage quantizers, deletion state, deduplication, conjugate graphs, and structural integrity. The `CrossQuery` method cleanly reuses the existing HGraph search internals while avoiding public API overhead.

The test suite (`hgraph_fgim_test.cpp`) is thorough, covering:
- Multi-source graphs with varying sizes and k values
- Deterministic tie-breaking and singleton sources
- Comprehensive unsupported input rejection
- CrossQuery equivalence with public KnnSearch (including distance evaluation statistics)
- Source graph immutability verification (snapshot before/after)

**Existing comments from prior review rounds cover the main discussion points** (e.g., `friend class HGraphFGIMTest` coupling, `ef`/`topk` clamping in CrossQuery). I have no additional blocking concerns to add.

The `uint64_t` type alignment and formatting cleanup in the latest commit (78f2eca) is a nice polish.

LHT129 (commented):
Overall this is a well-structured PR with thorough input validation, clean separation of concerns, and comprehensive test coverage.

The FGIM initial k-NNG construction is implemented as a stateless utility class with clear static methods. Validation in ValidateSourcesAndGetOffsets is exhaustive. CrossQuery cleanly reuses existing HGraph search internals while avoiding public API overhead. The test suite is thorough, covering multi-source graphs, deterministic tie-breaking, unsupported input rejection, CrossQuery equivalence with public KnnSearch, and source graph immutability verification.

Existing comments from prior review rounds cover the main discussion points (friend class coupling, ef/topk clamping). I have no additional blocking concerns. The uint64_t type alignment and formatting cleanup in the latest commit is a nice polish.

LHT129 (commented):
Overall this is a well-structured PR with thorough input validation, clean separation of concerns, and comprehensive test coverage.

The FGIM initial k-NNG construction is implemented as a stateless utility class with clear static methods. Validation in ValidateSourcesAndGetOffsets is exhaustive. CrossQuery cleanly reuses existing HGraph search internals while avoiding public API overhead. The test suite is thorough, covering multi-source graphs, deterministic tie-breaking, unsupported input rejection, CrossQuery equivalence with public KnnSearch, and source graph immutability verification.

Existing comments from prior review rounds cover the main discussion points (friend class coupling, ef/topk clamping). I have no additional blocking concerns. The uint64_t type alignment and formatting cleanup in the latest commit is a nice polish.

LHT129 (commented):
Thanks for the update. The changes in this push (78f2eca7) are clean improvements:

- Changed `k` parameter type from `std::size_t` to `uint64_t` for consistency with VSAG coding standards.
- Extracted boolean validation conditions into named variables (`valid_k`, `valid_precise_storage`, `valid_raw_storage`, `fully_built`, `valid_neighbor`, `valid_query`), which improves readability and makes the validation logic self-documenting.

The two suggestions from the earlier Copilot review remain unaddressed:

1. **`param.ef` and `param.topk` in `CrossQuery`**: These are still set to `l` even when `l > target.GetNumElements()`. Consider clamping `ef` (and `topk` if the search implementation requires it) to `topk` to avoid unnecessary work in `search_one_graph`.

2. **`friend class HGraphFGIMTest` in `hgraph.h`**: This couples the production header to a specific test fixture name. Consider gating behind `#if defined(ENABLE_TESTS)` or using a test accessor pattern.

Otherwise the code looks solid — thorough input validation, good test coverage, and clear comments throughout.

PR #2966, opened 2026-09-17
fix(pyramid): exclude reserved slots from label lookup

### LINKED ISSUE #2964: [bug](pyramid): unused label slots cause incorrect label 0 lookup and possible crash
### Affected module
module/index

### Affected area
not applicable

### Interface
cpp

### VSAG version
bad2f6f69a536213274a8f6193cf8f06a9dee81d (prebuilt binary used by the original reproducer).

The additional core-dump case below comes from a deployed build whose exact VSAG commit has not been established. Do not assume that this binary is identical to the reproducer binary.

### Summary
Unused slots in a Pyramid label table can be treated as valid labels. This issue tracks two manifestations: (1) a real label 0 loses its correct reverse mapping after serialization/deserialization; (2) a nonexistent label 0 is falsely found by linear lookup in an unused zero-filled slot, leading to an out-of-range vector read and, in an observed deployed build, a process crash. The original public C++ API reproducer below reproduces the first manifestation 10/10 times; the second is supported by two independent core dumps.

### Minimal reproducer: existing label 0
```cpp
#include <vsag/vsag.h>

#include <cassert>
#include <iostream>
#include <sstream>

int main() {
    const char* params = R"({"dim":8,"dtype":"float32","metric_type":"ip",
        "index_param":{"base_quantization_type":"fp32","use_reorder":false}})";
    const auto create = [&]() {
        auto result = vsag::Factory::CreateIndex("pyramid", params);
        assert(result.has_value());
        return result.value();
    };
    int64_t id = 0;
    float vector[8] = {0.25F};
    float query[8] = {1.0F};
    std::string path;
    auto base = vsag::Dataset::Make()->NumElements(1)->Dim(8)->Ids(&id)
                    ->Float32Vectors(vector)->Paths(&path)->Owner(false);
    auto index = create();
    auto built = index->Build(base);
    assert(built.has_value() && built.value().empty());
    auto before = index->CalcDistanceById(query, id);
    assert(before.has_value() && before.value() == 0.75F);
    std::cout << "before: id=0 distance=" << before.value() << '\n';

    std::stringstream stream(std::ios::in | std::ios::out | std::ios::binary);
    auto saved = index->Serialize(stream);
    assert(saved.has_value());
    stream.seekg(0);
    auto reloaded = create();
    auto loaded = reloaded->Deserialize(stream);
    assert(loaded.has_value());
    assert(reloaded->GetNumElements() == 1);
    assert(reloaded->CheckIdExist(0));
    auto after = reloaded->CalcDistanceById(query, id);
    if (!after.has_value()) {
        std::cout << "after: id=0 exists=true ERROR=" << after.error().message << '\n';
        return 1;
    }
    std::cout << "after: id=0 distance=" << after.value() << '\n';
    return after.value() == 0.75F ? 0 : 1;
}
```

### Environment
`bash scripts/check_environment.sh` output:
```text
- vsag version: v0.18.0-317-g9e098afd
- compiler version: c++ (GCC) 10.2.1 20200825 (Alibaba 10.2.1-3 2.17)
```
The script reports the local source checkout; the reproducer links the binary at the commit listed above. Compiled with C++17 and `_GLIBCXX_USE_CXX11_ABI=0`.

### Expected behavior
Label 0 returns distance `0.75` both before and after the round-trip.

### Actual behavior
```text
before: id=0 distance=0.75
after: id=0 exists=true ERROR=failed to acquire codes for id 9
```

### References
[LabelTable::Deserialize](https://github.com/antgroup/vsag/blob/bad2f6f69a536213274a8f6193cf8f06a9dee81d/src/impl/label_table/label_table.cpp#L152-L159), [Serialize/Resize](https://github.com/antgroup/vsag/blob/bad2f6f69a536213274a8f6193cf8f06a9dee81d/src/impl/label_table/label_table.h#L180-L208).

### Additional context
Pyramid initially reserves 10 label slots. Unused slots contain 0 and are serialized too. Deserialization rebuilds the reverse map over all slots, overwriting `label 0 -> slot 0` with `label 0 -> slot 9`. Zero is a valid label, so simply skipping it is not a fix.

### Additional case: nonexistent label 0 found by linear lookup
Two independent core dumps show the same state in a deserialized FP32 Pyramid index used for recall evaluation. The small index is one evaluation partition, not the entire serving index.

| Core observation | Value |
| --- | --- |
| Requested external label | `0` |
| Resulting internal ID | `9` |
| `Pyramid::cur_element_count_` | `9` |
| `Pyramid::max_capacity_` | `10` |
| Label-table length and `total_count_` | `10` |
| `LabelTable::use_reverse_map_` | `false` (linear lookup) |
| Vector-data `total_count_` | `9` |
| Dimension / bytes per vector | `1024` / `4096` |
| `MemoryBlockIO::size_` | `36864` (`9 * 4096`) |
| `MemoryBlockIO::has_deserialized_` | `true` |
| Codes pointer passed to distance computation | `nullptr` |

The label table contains:

```text
valid slots 0..8: 118, 136, 123, 125, 114, 151, 116, 117, 161
unused slot   9: 0
```

There is no real label 0. Linear lookup nevertheless matches the unused slot and returns internal ID 9. The requested vector begins at offset `9 * 4096 = 36864`, exactly at the end of the stored vector data. The null codes pointer then reaches the SIMD inner-product implementation.

Relevant native call chain (outermost to innermost):

```text
Pyramid::CalcDistanceById(query=<non-null>, id=0)
  -> InnerIndexInterface::calc_distance_by_id
  -> InnerIndexInterface::cal_distance_by_id (*ids=0, count=1)
  -> FlattenDataCell<FP32Quantizer<IP>, MemoryBlockIO>::query (*idx=9, codes=nullptr)
  -> Computer::ComputeDist
  -> FP32Quantizer::ComputeDistImpl
  -> avx2::FP32ComputeIP
  -> SIGSEGV
```

The original query pointer is non-null. `ComputeDistImpl` passes the stored codes as the first argument to `FP32ComputeIP`; although that callee names its first parameter `query`, the null value in this frame is the stored-vector pointer. The process's JVM signal handler subsequently converts the SIGSEGV to an abort/SIGABRT core.

Unlike the original reproducer, this case does not require overwriting an existing label 0 in the reverse map. It exposes a false-positive label lookup with the reverse map disabled. A fix limited to reverse-map reconstruction would not cover this case. The different final outcomes (returned error versus crash) were observed in different binaries; identical behavior on the reproducer commit has not been established for this second case.

### Fix scope and acceptance criteria
- Distinguish valid label slots from reserved capacity consistently in serialization/deserialization, reverse-map construction, and linear lookup.
- Preserve legitimate label 0; skipping every zero-valued slot is not a correct fix.
- Reject invalid internal IDs or failed codes reads before SIMD distance computation, returning an error instead of dereferencing a null pointer.
- Add regression coverage with spare capacity for both real label 0 and absent label 0, with reverse mapping enabled and disabled, before and after serialization/deserialization.
- For a real label 0, existence and distance must remain correct. For an absent label 0, `CheckIdExist(0)` must be false and `CalcDistanceById(..., 0)` must fail safely.

## Related issues
No matching issue found.

---

wxyucs:
## Change Type

- [x] Bug fix

## Linked Issue

Fixes: #2964

## What Changed

Pyramid now reserves label storage without creating labels. A real label `0` retains its mapping and distance across serialization; absent `0` is not found, and single-ID distance lookup returns an error. Older capacity-sized payloads are trimmed using their stored vector count before the restored index is used.

Dense datacells reject invalid internal IDs before prefetch/SIMD, check failed batch reads, and validate pair-distance leases. Pyramid protects label-vector growth with its existing storage lock. Shared explicitly sized LabelTable callers retain their existing behavior; HNSW and DiskANN are unchanged.

The shared distance helper from current main compacts resolved labels before querying and skips code reads for all-missing batches. This preserves `-1` results on empty indexes and for missing labels in mixed batches while retaining the datacell bounds checks. The earlier empty-index guard is now subsumed by upstream; this PR leaves the current shared helper and distance APIs unchanged.

## Test Evidence

Rebased onto main at `7dc90ad5e61f972b9bab8aff448d3c80d31cf51a`. On the rebased head, aggregate `make test` rebuilt successfully and `HGraph Search Empty Index` passed 3,408 assertions; replay with the original CircleCI seed `2759684390` also passed. Affected Pyramid, LabelTable, InnerIndexInterface, and invalid-read unit tests passed 49 cases / 2,542 assertions. Current Pyramid/HGraph distance contract tests passed 3 cases / 84 assertions. Recompiled batch and legacy-payload probes passed 1,152 and 192 checks across both lookup modes and reader formats. Documentation validation and `git diff --check` passed. Exact clang-format/tidy 15 and broader sanitizer/coverage checks remain with normal CI.

Earlier validation history:

- Original regression baseline: main failed 32 assertions in the new Pyramid cases. The isolated datacell regression reproduced a SIGSEGV with a reserved internal ID and separately demonstrated an ignored failed batch read. These are local reproductions, not a claim that the issue's original and deployed binaries behave identically.
- Focused aggregate `make test`: 15 cases, 1,460 assertions passed.
- Broader unit regressions: 56 cases, 8,663,351 assertions passed; additional shared-caller regressions: 157 cases, 190,393 assertions passed.
- Focused ASan/UBSan datacell regression: 9 assertions passed. The test and instantiated header code were instrumented; the linked library was not fully instrumented.
- Functional regressions: 122 cases, 562,761 assertions passed; additional HGraph serialization and IVF distance regressions: 14 cases, 41,202 assertions passed.
- Documentation checks and `git diff --check` passed.
- The earlier head passed CI's exact clang-format 15 check. Normal CI validates the updated head; no checks are skipped by the commit.

CircleCI [job 18352](https://circleci.com/gh/antgroup/vsag/18352) exposed an empty-index batch regression: `HGraph Search Empty Index` failed `multi_result.has_value()` at `tests/test_index/test_index_calc.cpp:181` because the shared helper queried placeholder internal ID zero with no stored codes. The existing targeted regression failed before the follow-up fix (3 assertions), then passed all 3,408 assertions with CircleCI seed `2759684390`. The focused unit suite passed 18 cases / 1,509 assertions; batch and legacy-payload probes passed 1,152 and 192 distance checks against the updated library. Earlier broader results above predate this helper fix. No transient-failure rerun was requested.

## Compatibility Impact

The length-prefixed label-vector format is unchanged; no fields or version markers are added. Seekable and streaming readers recover old unused suffixes using Pyramid's dense stored-vector count, preserving real zero labels. Short label payloads are rejected. Older readers can parse new payloads but retain their lookup bug and should also be upgraded. Existing tombstone persistence is unchanged; this patch adds no deletion metadata.

Single-ID Pyramid distance lookup now returns an error for missing or mark-removed labels. Batch distance conventions and other indexes' missing-ID behavior are unchanged.

Distance evaluation adds a count snapshot and ID bounds checks. Label insertion briefly holds the existing exclusive storage lock to protect vector size changes from concurrent search; graph insertion retains its existing locking. No benchmark claim is made.

## Documentation Impact

- [x] Updated English and Chinese Pyramid documentation.

## Risk and Rollback

Risk is concentrated in shared dense datacell validation and Pyramid persistence. Reverting the commit restores the previous behavior without a format migration.

## Checklist

- [x] Linked issue and regression tests
- [x] Considered compatibility and concurrency
- [x] Updated behavior documentation
- [x] Conventional Commit with human DCO and AI-assistance trailers


src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2627,7 +2635,15 @@ Pyramid::CalcDistanceById(const float* query, int64_t id, bool calculate_precise
     if (raw_vector_ != nullptr && calculate_precise_distance) {
         flat = this->raw_vector_;
     }
-    return InnerIndexInterface::calc_distance_by_id(query, id, flat);
+    InnerIdType inner_id;
+    {
```
LHT129, line 2639: [suggestion] The new `CalcDistanceById` implementation uses `GetIdByLabel()` which throws `VsagException` for missing/mark-removed labels, while the batch path `CalcDistancesById` → `CalDistanceById` still delegates to `InnerIndexInterface::cal_distance_by_id`, which uses `TryGetIdByLabel` and returns `-1` for invalid IDs without throwing.

This creates an inconsistency: single-ID callers get an exception, batch callers get a `-1` sentinel. If this difference is intentional (as the PR description suggests — "Batch distance conventions … are unchanged"), consider adding a brief comment explaining why the two paths diverge, so future readers don't "fix" the inconsistency.

```cpp
// Single-ID: throw for missing labels (consistent with CheckIdExist contract).
// Batch path retains the legacy -1 sentinel for compatibility.
```

src/datacell/flatten_datacell.h
```diff
@@ -389,6 +389,14 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::query(float* result_dists,
...
 
+    const auto total_count = this->TotalCount();
+    for (InnerIdType i = 0; i < id_count; ++i) {
+        if (idx[i] >= total_count) {
+            throw VsagException(ErrorType::READ_ERROR,
+                                "invalid internal id " + std::to_string(idx[i]));
+        }
+    }
+
     for (uint32_t i = 0; i < this->prefetch_stride_code_ and i < id_count; i++) {
```
LHT129, line 400: [suggestion] The ID bounds check is added before the prefetch loop, which is correct. However, for `InMemory` layouts the prefetch at this line still executes before the individual `Acquire` calls later in the function perform their own lease validation. The prefetch itself is a hint and won't crash on an invalid ID, so this is not a correctness issue — just noting that the prefetch stride loop could be moved after the bounds check but before the `if constexpr (not LayoutTmpl::InMemory)` branch for clarity, since it's now guarded.

src/datacell/flatten_datacell.h
```diff
@@ -389,6 +389,14 @@ FlattenDataCell<QuantTmpl, LayoutTmpl>::query(float* result_dists,
                                               QueryContext* ctx) {
     Allocator* search_alloc = select_query_allocator(ctx, allocator_);
 
+    const auto total_count = this->TotalCount();
+    for (InnerIdType i = 0; i < id_count; ++i) {
+        if (idx[i] >= total_count) {
```
LHT129, line 394: [note] The ID bounds validation loop iterates over all `id_count` IDs on every `query()` call. In hot search paths where `id_count` can be large (e.g. top-K candidates), this adds O(N) overhead per query call.

Consider whether the validation could be done once at a higher level (e.g. when the ID list is first constructed) rather than on every `query()` invocation. Alternatively, a debug-only assertion (`assert`) could be used in release builds if the callers are trusted to provide valid IDs.

The `MultiRead` failure check and `ComputePairVectors` lease checks are less concerning since they guard against I/O failures that can occur in production.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -2627,7 +2635,15 @@ Pyramid::CalcDistanceById(const float* query, int64_t id, bool calculate_precise
     if (raw_vector_ != nullptr && calculate_precise_distance) {
         flat = this->raw_vector_;
     }
-    return InnerIndexInterface::calc_distance_by_id(query, id, flat);
+    InnerIdType inner_id;
```
LHT129, line 2638: [suggestion] `CalcDistanceById` (single-ID) now correctly resolves the label to an inner_id via `label_table_->GetIdByLabel(id)`. However, the batch variant `CalDistanceById` (called at line ~2643 in the original file) still delegates to `InnerIndexInterface::cal_distance_by_id(query, ids, count, flat, &validity)`, which passes the external label IDs directly as inner IDs to the data cell. This means the batch path still suffers from the same label-vs-inner-id confusion that this PR fixes for the single-ID path.

If `CalDistanceById` is expected to receive external labels (matching `CalcDistanceById`'s contract), it should also resolve each label through `label_table_` before querying the data cell. If it is only called with inner IDs internally, consider adding a comment to clarify the contract difference.

src/algorithm/pyramid/pyramid.cpp
wxyucs: The single-ID error and batch -1 sentinel distinction is intentional for compatibility, as documented in the PR. This optional comment suggestion does not require a code change.

src/datacell/flatten_datacell.h
wxyucs: As the review notes, this is a clarity suggestion, not a correctness defect. Keeping the safety checks and avoiding unrelated prefetch restructuring.

src/datacell/flatten_datacell.h
wxyucs: Runtime validation is required for the reported production out-of-range/null-code failure and must not become debug-only. No measured regression is supplied here; retain the safety guard and defer speculative optimization. Concrete performance evidence can be considered separately.

src/algorithm/pyramid/pyramid.cpp
wxyucs: I traced the full batch path on b9c46e9de450f70b678e2e452e102a489eca4d88; it already translates external labels.

- The public [`Index::CalcDistancesById` bridge](https://github.com/antgroup/vsag/blob/b9c46e9de450f70b678e2e452e102a489eca4d88/include/vsag/index.h#L1187-L1220) calls the legacy virtual, then [`IndexImpl::CalDistanceById`](https://github.com/antgroup/vsag/blob/b9c46e9de450f70b678e2e452e102a489eca4d88/src/index/index_impl.h#L131-L147) dispatches to Pyramid. Dataset-backed dense queries also dispatch each row through that float-pointer overload.
- [`Pyramid::CalDistanceById`](https://github.com/antgroup/vsag/blob/b9c46e9de450f70b678e2e452e102a489eca4d88/src/algorithm/pyramid/pyramid.cpp#L2656-L2675) calls the shared helper. In [`InnerIndexInterface::cal_distance_by_id`](https://github.com/antgroup/vsag/blob/b9c46e9de450f70b678e2e452e102a489eca4d88/src/algorithm/inner_index_interface.cpp#L1080-L1103), each `ids[i]` goes through `TryGetIdByLabel`, the mapped value is stored in `inner_ids[i]`, and `data->Query` receives `inner_ids.data()`. Missing/removed positions are overwritten with the legacy `-1` sentinel; the validity mask preserves top-k ordering.
- [`TryGetIdByLabel`](https://github.com/antgroup/vsag/blob/b9c46e9de450f70b678e2e452e102a489eca4d88/src/impl/label_table/label_table.cpp#L98-L115) handles both reverse-map and linear lookup and rejects mark-removed rows.

Reproduced on this head with labels `[118, 0, 9999]` (and `136` replacing zero for the absent-zero case), deliberately distinct vectors, missing labels `1` and `2147483647`, and mark removal. The public float-pointer and Dataset APIs, default and top-k outputs, both precision flags, mutable/immutable lookup, and seekable/streaming round trips passed 1,152 distance checks. A separate synthetic legacy capacity-sized payload probe passed 192 batch distance checks across both lookup modes and reader formats, including post-load removal filtering. The existing focused suite passed 15 cases / 1,460 assertions.

Removal checks concern live tombstones; this does not claim new tombstone persistence support. The batch contract is external labels with `-1` for invalid entries, as already documented. No additional mapping or source/test change is needed for this comment.

LHT129 (commented):
I have completed an independent review of this PR in addition to the existing inline comments from LHT129.

**Summary of findings:**

The changes are well-structured and address the core issue (#2964) comprehensively:

1. **LabelTable changes** (`Reserve` + `TrimUnusedSlots`): Clean separation between capacity reservation and slot creation. `TrimUnusedSlots` correctly rebuilds both the padding-label set and the reverse map after trimming, and the test covers the legacy payload recovery path with both reverse-map modes and label-0 scenarios.

2. **Pyramid changes**: Switching `Resize` → `Reserve` in `resize()` is the right fix — reserved capacity should not create spurious label slots. The `prepare_add_batch` lock scope expansion correctly protects label-vector growth from concurrent search readers. `CalcDistanceById` now properly resolves external labels through `label_table_` before querying the datacell, and the `label_lookup_mutex_` usage is appropriate.

3. **FlattenDataCell defenses**: The ID bounds check, `MultiRead` failure check, and `ComputePairVectors` lease validation are well-placed safety nets. The test covers reserved IDs, sentinel values, failed leases, and failed batch reads.

4. **`inner_index_interface.cpp`**: The all-missing-batch guard correctly avoids querying the datacell when no valid IDs remain, preventing the empty-index crash reported in CircleCI job 18352.

5. **Tests**: Good coverage across the new code paths — reserved-slot label semantics, legacy payload recovery, datacell rejection of invalid reads, and LabelTable reservation/trimming. The use of `GENERATE` for combinatorial testing (real_zero × immutable × streaming × reverse_map) is thorough.

The existing review comments from LHT129 have been addressed by the author with detailed explanations. I did not identify additional blocking issues.

LHT129 (commented):
This is a well-structured fix for the reserved-slot label lookup issue (#2964). The changes are focused and well-tested.

**Summary of changes:**

1. **LabelTable**: Added `Reserve()` to allocate capacity without creating label slots, and `TrimUnusedSlots()` to recover from legacy capacity-sized payloads. The `Resize` → `Reserve` change in `Pyramid::resize()` correctly prevents spurious label-0 creation.

2. **Pyramid**: `CalcDistanceById` now resolves external labels through `label_table_->GetIdByLabel()` before querying the data cell. `prepare_add_batch` holds `resize_mutex_` during label insertion to protect concurrent search readers. Deserialization paths call `TrimUnusedSlots` to discard unused suffix slots from legacy payloads.

3. **FlattenDataCell**: Added bounds checks before prefetch/SIMD, `MultiRead` failure detection, and `ComputePairVectors` lease validation — all throwing descriptive `VsagException` on failure.

4. **InnerIndexInterface**: `FactoryComputer` and `Query` are now skipped when all IDs in a batch are missing, avoiding a code read on an empty index.

5. **Tests**: Comprehensive coverage across 4 new test cases covering real-zero/absent-zero, mutable/immutable, seekable/streaming, legacy payload recovery, and datacell error paths.

The existing inline comments from the earlier review round have been addressed by the author with detailed explanations. No blocking issues remain.

LHT129 (approved):
LGTM

Issue #2967, opened 2026-09-17
[improve](build): tolerate floating-point rounding in int8 tests for 1.0


wxyucs:
### Kind of improvement

test

### Affected area

build

### Summary

Target release line: **1.0**. Improve the floating-point assertions in the INT8 quantizer tests so that expected rounding differences do not block distribution packaging. This is a test robustness improvement, not a request to change runtime index behavior.

A `v1.0.0` distribution build using `dockcross/manylinux2014-x64`, GCC 10.2.1, x86_64, Release mode and the pre-C++11 ABI completed compilation, then failed these two unit tests:

- `INT8 Compute`
- `INT8 Serialize and Deserialize`

Both stop in the shared `TestComputerINT8` helper at `src/quantization/int8_quantizer_test.cpp:199`:

```text
REQUIRE( gt == dists1[j] )
with expansion:
  0.823804736f == 0.823804796f
```

The difference is approximately `5.96e-8` (one float ULP at this magnitude). The helper performs exact equality before its numeric-error checks. Both callers already supply an error tolerance of `2e-5`.

Use a justified floating-point tolerance for the cosine distance comparison while retaining strict checks where exact integer-derived results are expected. Keep the two tests enabled in the normal distribution workflow.

### Expected benefit

- Expected floating-point rounding differences no longer interrupt 1.0 distribution builds.
- Both distance computation and serialization round-trip coverage remain enabled.
- Genuine distance errors outside the chosen tolerance still fail.
- Runtime algorithms, public APIs and release version detection remain unchanged.

### Benchmark evidence

Not applicable: this is a test-only improvement.

### References

- [src/quantization/int8_quantizer_test.cpp:161](https://github.com/antgroup/vsag/blob/v1.0.0/src/quantization/int8_quantizer_test.cpp#L161): reference distance calculation.
- [src/quantization/int8_quantizer_test.cpp:199](https://github.com/antgroup/vsag/blob/v1.0.0/src/quantization/int8_quantizer_test.cpp#L199): exact equality assertion.
- [src/quantization/int8_quantizer_test.cpp:230](https://github.com/antgroup/vsag/blob/v1.0.0/src/quantization/int8_quantizer_test.cpp#L230): compute test.
- [src/quantization/int8_quantizer_test.cpp:282](https://github.com/antgroup/vsag/blob/v1.0.0/src/quantization/int8_quantizer_test.cpp#L282): serialization test.
[...]

Issue #2945, opened 2026-09-14
[bug](quantization): fix exact-equality failures in int8 cosine tests on release


LightWant:
### Affected module

module/datacell

### Affected area

area/testing

### Interface

cpp

### VSAG version

Clean upstream main: `fd077c3f3d88dbd55b0936a6b3fb03ebb27cdd31` (2026-09-14).

### Summary

The existing `INT8 Compute` and `INT8 Serialize and Deserialize` unit tests fail in a fresh GCC 15 Release build of upstream main. Both stop at the exact floating-point equality assertion in `src/quantization/int8_quantizer_test.cpp:199`, with an approximately one-ULP difference in cosine distance. This also reproduces on PR #2908 at `c71a8d701e96736238c48b85b3241aa6402f7b56`; it is not specific to the fused cleanup.

### Minimal reproducer

From a fresh checkout of the main commit above, with the documented build dependencies installed:

```bash
make release CMAKE_GENERATOR='Unix Makefiles' VSAG_ENABLE_TESTS=ON COMPILE_JOBS=24 \
  EXTRA_DEFINED='"-DCMAKE_CXX_FLAGS=-include cstdint -Wno-error=ignored-attributes" "-DCMAKE_C_FLAGS=-Wno-error=stringop-overflow"'
./build-release/tests/unittests 'INT8 Compute,INT8 Serialize and Deserialize'
```

The existing reproducer generates deterministic INT8 data (dimensions 64/128; counts 10/101; query seed 165). Its critical assertion is:

```cpp
auto gt = gt_func(j, i);
quant.EncodeOne(reinterpret_cast<float*>(vecs.data() + j * dim), code);
quant.ComputeDist(*computer, code, dists1.data() + j);
REQUIRE(gt == dists1[j]);
```

### Environment

Output of `bash scripts/check_environment.sh` in the clean main worktree:

```text
- OS: Ubuntu 26.04
- vsag version: v1.0.0-170-gfd077c3f
- compiler version: c++ (Ubuntu 15.2.0-16ubuntu1) 15.2.0
```

x86_64, AMD EPYC 9T24; CMake 4.2.3; Catch2 3.7.1. Release compile commands contain `-O3 -DNDEBUG ... -Ofast` for both the library and these tests. The main comparison reused the same cached third-party dependency sources/OpenBLAS library as the PR build, but compiled VSAG and its tests from scratch with ccache disabled. No main source files were changed.

### Expected behavior

[...]

PR #2968, opened 2026-09-17
fix(test): tolerate INT8 cosine rounding on 1.0

wxyucs:
The INT8 compute and serialization tests require bitwise equality between independently evaluated cosine distances, so an approximately one-ULP rounding difference can fail release packaging. Use Catch2 `WithinAbs` with the callers' existing `2e-5` absolute tolerance for every cosine result. Keep exact L2/IP assertions, aggregate error checks, and single/batch consistency checks.

The absolute bound handles zero distances without division by the expected value. L2/IP remain exact at the tested dimensions because their integer sums fit in float's exact integer range. The matcher also rejects injected NaNs under the tested Release flags; a direct fast-math `abs(...) <= error` assertion did not.

Fixes: #2967 (release branch `1.0` only). Related main issue: #2945; main is delivered separately in #2969.

Validation:

- Fresh GCC 15.2 Release (`-O3 -DNDEBUG -Ofast`) library and focused INT8 test executable; cached dependency sources, build-only GCC compatibility flags (`-include cstdint`, warning exceptions for ignored attributes/stringop overflow).
- All three INT8 cases pass: 30,708 assertions. The two reported cases pass with seed `938459996`: 29,376 assertions.
- Rebuilding the original assertion against the same library reproduces both failures with `0.823804736f == 0.823804796f` on GCC 15.
- Local test-source mutation checks reject out-of-tolerance cosine errors, NaNs, and one-ULP L2/IP errors in both compute and serialization. Fourteen assertion boundary controls cover the reported ULP difference, zero, both signs at/beyond tolerance, negative expected values and non-finite inputs, including with Catch2's matcher compiled under `-Ofast`.
- No GCC 10.2.1/manylinux2014 or pre-C++11 ABI reproduction claimed. Full-suite coverage and exact clang-format/tidy 15 checks are left to CI.

Test-only change; no runtime, API, serialization-format or documentation changes. Revert this commit to roll back.


LHT129 (commented):
Reviewed the change in `src/quantization/int8_quantizer_test.cpp`.

The fix correctly replaces the exact `REQUIRE(gt == dists1[j])` assertion with `REQUIRE_THAT(dists1[j], Catch::Matchers::WithinAbs(gt, error))` for the COSINE metric path in `TestComputerINT8`. This is the right approach:

- COSINE distance involves square roots and division of stored norms, where the quantized norm and the reference norm can round differently at the last ULP. `WithinAbs` with the caller-provided `error` tolerance (2e-5) correctly handles this.
- L2 and IP remain exact because their integer sums fit in float's exact integer range at the tested dimensions, so the `REQUIRE(gt == dists1[j])` assertion is kept for those metrics.
- The `count_unbounded_numeric_error` tracking below the assertion is preserved and remains correct: it still counts cases where the absolute error exceeds the tolerance.

No issues found. The change is minimal, well-scoped, and the comments explain the rationale clearly.

LHT129 (approved):
LGTM

PR #2969, opened 2026-09-17
fix(test): tolerate INT8 cosine rounding on main

### LINKED ISSUE #2945: [bug](quantization): fix exact-equality failures in int8 cosine tests on release

wxyucs:
The INT8 compute and serialization tests require bitwise equality between independently evaluated cosine distances, so an approximately one-ULP rounding difference can fail Release tests. Use Catch2 `WithinAbs` with the callers' existing `2e-5` absolute tolerance for every cosine result. Keep exact L2/IP assertions, aggregate error checks, and single/batch consistency checks.

The absolute bound handles zero distances without division by the expected value. L2/IP remain exact at the tested dimensions because their integer sums fit in float's exact integer range. The matcher also rejects injected NaNs under the tested Release flags; a direct fast-math `abs(...) <= error` assertion did not.

Fixes: #2945 (`main` only). Related release issue: #2967; the `1.0` fix is delivered separately in #2968.

Validation:

- Fresh GCC 15.2 Release (`-O3 -DNDEBUG -Ofast`) library and focused INT8 test executable; cached dependency sources, build-only GCC compatibility flags (`-include cstdint`, warning exceptions for ignored attributes/stringop overflow).
- All three INT8 cases pass: 30,708 assertions. The two reported cases pass with seed `938459996`: 29,376 assertions.
- Rebuilding the original assertion against the same library reproduces both failures with `0.823804736f == 0.823804796f` on GCC 15.
- Local test-source mutation checks reject out-of-tolerance cosine errors, NaNs, and one-ULP L2/IP errors in both compute and serialization. Fourteen assertion boundary controls cover the reported ULP difference, zero, both signs at/beyond tolerance, negative expected values and non-finite inputs, including with Catch2's matcher compiled under `-Ofast`.
- No GCC 10.2.1/manylinux2014 or pre-C++11 ABI reproduction claimed. Full-suite coverage and exact clang-format/tidy 15 checks are left to CI.

Test-only change; no runtime, API, serialization-format or documentation changes. Revert this commit to roll back.


LHT129 (approved):
LGTM

Issue #2970, opened 2026-09-17
[bug](build): fix 1.0 manylinux cxx11 dependency build regressions


wxyucs:
### Affected area

build

### Interface

cpp

### VSAG version

Observed: `v1.0.0` (`efdaf17a10e96cdb5222baf558d50dfacbdc672e`).
Source comparison: `v0.18.0` and `v0.18.15` versus `v1.0.0`.
Target release line: **1.0**. Current branch snapshot inspected: `87a8160ca3f7933eb59f62f7ff7307cda12fab78`.

### Summary

The 1.0 manylinux C++11 ABI distribution build fails in CRoaring and bundled OpenBLAS with GCC 14.2.1. Source history identifies two dependency build changes since 0.18 that explain these failures without requiring a newly upgraded container image:

1. Release `-Ofast` now reaches C sources in third-party subdirectories, triggering CRoaring's `run.c` diagnostic under the existing `-Werror` policy.
2. Bundled OpenBLAS is now forced to build with `NOFORTRAN=1`, selecting its C-translated LAPACK implementation, which contains incompatible `float*` to `double*` calls rejected by GCC 14.

Both are compilation failures before VSAG tests. The pre-C++11 ABI build with GCC 10.2.1 completed and produced an archive; the C++11 ABI build with GCC 14.2.1 stopped.

### Minimal reproducer

On an x86_64 Linux host, use the C++11 distribution image with GCC 14.2.1. From a clean `v1.0.0` checkout:

```bash
docker build -t vsag-builder-cxx11 -f docker/Dockerfile.dist_cxx11_x86 .
docker run --rm -u "$(id -u):$(id -g)" -v "$PWD:/work" -w /work \
  vsag-builder-cxx11 bash -ec '
    export COMPILE_JOBS=96
    export CMAKE_INSTALL_PREFIX=/tmp/vsag
    export EXTRA_DEFINED="-DVSAG_USE_SYSTEM_DEPS:STRING=OFF -DENABLE_INTEL_MKL=OFF -DENABLE_CXX11_ABI=ON -DENABLE_LIBCXX=OFF -DENABLE_TESTS=OFF -DENABLE_MOCKIMPL=OFF"
    make clean-release
    make release
  '
```

This isolates compilation from the distribution script's subsequent tests. The observed run used these settings. The floating `latest` tag is not a reproducible image identifier; the image version used for this report is recorded below.

### Environment

[...]

PR #2952, opened 2026-09-14
chore(build): upgrade OpenBLAS to 0.3.34

wxyucs:
## Change Type

- [x] Improvement/Refactor
- [x] CI/Build/Infra

## What Changed

Upgrade bundled OpenBLAS from 0.3.24 to the latest stable [0.3.34 release](https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.34), published July 16, 2026 and reverified through `releases/latest`. Update both upstream URLs, the download-cache filename, and English/Chinese examples to `VSAG_THIRDPARTY_OPENBLAS_0_3_34`.

The maintainer reports that their investigation traced the recurring runner build bug to the old OpenBLAS version not recognizing the GitHub runner CPU. Refreshing the bundled dependency is motivated by that CPU-support finding and the build trackers below. The retained logs independently establish build failures, including missing architecture/configuration macros in one occurrence, but do not identify the affected CPU model or prove which upstream change fixes each occurrence. Parallel-build OOM is not an established explanation for these reports.

Independently computed archive SHA-256: `cd7e129868320cc2d033afa920e31202dfe0b8066a5b66661900ccc0f197dfed`. Its MD5, `ae9a0c2945f91b6f91747fd89a301d36`, matches upstream's release notes. No downstream patches require rebasing.

## Linked Issue

Refs #2507. Refs #2439. Refs #2559.

| Tracker | Scope of this PR |
| --- | --- |
| #2439 (open) | Closest match to the maintainer's CPU-support explanation. [OpenBLAS 0.3.24 notes](https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.24) document Cooper Lake/Sapphire Rapids compilation and dynamic detection fixes, feature-based fallback, and AMX activation. #2459 already delivered 0.3.23 → 0.3.24 to `main`; this PR advances 0.3.24 → 0.3.34. Later occurrences in the tracker still need validation, so no automatic closure. |
| #2507 (open) | Cross-workflow bundled-build tracker, including release branches and a confirmed 0.3.24 failure. This `main`-only upgrade is a candidate remedy for the CPU-support failure; it does not establish resolution of every occurrence or release-branch coverage. |
| #2559 (open) | Independent work to retain nested build diagnostics. A version bump does not implement log printing/upload or satisfy this issue. |
| #2514 / #2599 (closed) | Historical TSAN occurrences consolidated into #2507; serialization/system-library mitigations do not establish the underlying cause. |
[...]

PR #2971, opened 2026-09-17
fix(build): repair 1.0 manylinux dependency builds

wxyucs:
## Change Type

- [x] Bug fix
- [x] CI/Build/Infra

## What Changed

Repair the **1.0** manylinux dependency compilation failures in #2970. Backport main's compiler-option isolation from #2807 (`cd075cd2`) so CRoaring retains its own Release optimization while VSAG keeps `-Ofast` and `-Werror`. Preserve ABI definitions and restore VSAG's compile/link options after dependency declarations.

Backport the pinned OpenBLAS **0.3.34** upgrade from #2952 (`1a1b7d9d`). Its C LAPACK `spstf2`/`spstrf` implementations use the float-specific `smaxloc_`, correcting the incompatible pointer calls in 0.3.24. Retain `NOFORTRAN=1`, runtime linkage, CPU selection, strict-FP build-tool flags, threading settings, and existing build diagnostics. Update the SHA-256, versioned override/cache names, override tests, and English/Chinese offline-build examples together.

Add generated-command regression coverage for dependency isolation, retained VSAG optimization/warnings, ABI definitions, language scoping, and restored link options.

## Linked Issue

Fixes: #2970

## Test Evidence

- `make test-cmake` and `git diff --check` pass.
- GCC 15.2 reproduces CRoaring `run.c` failing with `-Ofast -Werror` and passing with `-O3 -Werror`.
- Both affected OpenBLAS 0.3.24 C units fail on the reported pointer mismatch; both 0.3.34 units pass with `-Werror=incompatible-pointer-types`.
- Full Release configuration with bundled dependencies and tests enabled passes for both libstdc++ ABIs. Generated commands retain VSAG flags and matching ABI definitions while isolating dependency options; CRoaring, fmt, and Catch2 build for both ABIs.
- OpenBLAS archive SHA-256 matches the pin and offline cache extraction succeeds. The optimized bundled ExternalProject build/install/link passes with `TARGET=GENERIC`, the default full `DYNAMIC_ARCH=1` dispatch list, `NOFORTRAN=1`, `USE_THREAD=0`, and `USE_LOCKING=1`; nested compilation is capped at eight jobs.
- Both ABI consumers link and run against the new static OpenBLAS archive: all eight VSAG BLAS/LAPACKE wrappers, upper/lower `spstf2` and blocked/unblocked `spstrf` checks, and the existing five `BlasFunction` tests (38 assertions per ABI) pass. Wrapper checks pass with native Cooperlake and baseline Prescott dispatch. fmt/CRoaring C++ consumers also link and run for both ABIs. `ldd` shows no system OpenBLAS/LAPACKE or Fortran-runtime dependency.

GCC 14.2.1 and Docker/Podman are unavailable locally, so the exact manylinux container is not reproduced. Full VSAG tests/coverage and the platform matrix remain for CI. Exact clang-format/tidy 15 are unavailable locally; normal CI is requested. Distribution tests remain unchanged.

## Compatibility, Documentation, and Risk

This PR targets **1.0 only**. No VSAG API changes. OpenBLAS contains numerical/kernel changes, so focused linkage checks do not replace the complete platform and distribution matrix; no performance claim is made. `DEVELOPMENT.md` and both offline-build guides reflect the new pin. Revert this commit to restore the prior dependency version and flag behavior.


tests/cmake/compile_flag_scope_test.cmake
```diff
@@ -0,0 +1,107 @@
...
+string (RANDOM LENGTH 12 ALPHABET 0123456789abcdef fixture_suffix)
+set (fixture_source "${VSAG_SOURCE_DIR}/tests/cmake/compile_flag_scope_fixture")
+set (fixture_build "${CMAKE_CURRENT_BINARY_DIR}/vsag-compile-flag-scope-${fixture_suffix}")
+
+execute_process (
+    COMMAND ${CMAKE_COMMAND}
+            -S ${fixture_source}
+            -B ${fixture_build}
+            -G Ninja
+            -DCMAKE_BUILD_TYPE=Release
```
LHT129, line 28: [suggestion] The test hardcodes `-G Ninja`, which makes `make test-cmake` fail on systems without Ninja installed. The Makefile target does not check for Ninja availability.

Consider either:
1. Detecting the available generator (e.g., `cmake --help` to list generators, fall back to "Unix Makefiles"), or
2. Adding a `command -v ninja` guard in the Makefile target and skipping this test with a warning when Ninja is absent.

wxyucs: This fixture deliberately retains the Ninja requirement from the main-branch regression test in #2807: it inspects `build.ninja` to verify that VSAG link options are restored. The maintained development images install Ninja and PR CI already uses it. Falling back would require a separate generator-specific assertion, while skipping would hide this regression coverage. I am keeping that existing test prerequisite for this scoped 1.0 backport; broader generator support can be handled separately.

LHT129 (commented):
I have completed a thorough review of this PR. The changes are well-structured and address two distinct issues cleanly:

**Compiler option isolation (`vsag_include_without_directory_options`)**
- The macro correctly saves and restores directory-level `COMPILE_OPTIONS` and `LINK_OPTIONS`, preventing VSAG flags (`-Ofast`, `-Werror`) from leaking into third-party dependency builds (e.g., CRoaring).
- The `vsag_add_compile_flag` update to use `$<$<COMPILE_LANGUAGE:C,CXX>:${flag}>` is the right fix to prevent flags from reaching ASM targets.
- The test fixture (`compile_flag_scope_test.cmake`) provides thorough coverage: it verifies that third-party targets are isolated from VSAG flags, that VSAG-owned C/CXX/ASM targets receive the correct flags, and that link options are properly restored.

**OpenBLAS 0.3.24 → 0.3.34 upgrade**
- The version bump addresses the `smaxloc_` incompatible-pointer-type issue in 0.3.24 C LAPACK units.
- Switching from MD5 to SHA256 for the URL hash is a welcome security improvement.
- The `thirdparty_override_test.cmake` updates correctly validate that the old `VSAG_THIRDPARTY_OPENBLAS_0_3_24` pin is ignored while the new `VSAG_THIRDPARTY_OPENBLAS_0_3_34` pin takes precedence.
- Documentation updates in `DEVELOPMENT.md`, `offline_build.md` (both EN and ZH) are consistent and complete.

No blocking issues found. The test coverage is comprehensive and the implementation is clean.

LHT129 (approved):
LGTM

PR #2973, opened 2026-09-18
fix(mci): keep maximal cliques whole and inline the filter bitmap

### LINKED ISSUE #2972: [bug](hgraph): MCI truncates every maximal clique to mci_clique_max
### Affected module

module/index

### Affected area

not applicable

### Interface

cpp

### VSAG version

main @ `7dc90ad5` (`v1.0.0-182-g7dc90ad5`)

### Summary

The MCI companion of `hgraph` clips every enumerated maximal clique to `mci_clique_max` members, so a built index only ever stores cliques of exactly that size. `mci_clique_max` is used simultaneously as the minimum size of the maximal-clique enumeration (`clique_threshold`) and as the truncation cap of the saved clique, which makes it impossible to keep the larger cliques the enumeration actually produces.

The reference MCI implementation treats the same parameter (`cliqueMax`) as a lower bound only and stores each maximal clique in full, so the two implementations build indexes with very different clique-union degrees from the same graph and parameters.

### Minimal reproducer

A candidate pool whose maximal clique is larger than `mci_clique_max` is stored clipped. `BuildMCICliques` (`src/algorithm/mci/mci_builder.cpp`) is called here with `candidate_limit = 6` and `clique_max = 3` over 7 equidistant points, so the only maximal clique inside each candidate pool has size 7:

```cpp
    constexpr uint64_t total = 7;
    constexpr uint64_t dim = 7;             // e_i basis vectors: every pair is at distance 2
    constexpr uint64_t candidate_limit = 6;
    std::vector<float> vectors(total * dim, 0.0F);
    for (uint64_t id = 0; id < total; ++id) {
        vectors[id * dim + id] = 1.0F;
    }
    // every node lists all other 6 nodes as its candidate pool
    ...
    params.candidate_limit = candidate_limit;
    params.clique_max = 3;
    params.alpha = 1.2F;                    // distance limit 2 * 1.2 > 2, so the pool is a clique

    const auto cliques = vsag::BuildMCICliques(vectors.data(), graph, params, &allocator);
    // actual:   every clique has size 3
    // expected: at least one clique has size candidate_limit + 1 == 7
```

The same leak exists on the public path: with `"mci_mcs": 200, "mci_clique_max": 50`, every clique stored by `GetStats()["mci_max_clique_size"]` is exactly 50, while the same graph under the reference implementation yields cliques of 50..201 members (mean ≈ 99.6).

### Environment

```shell
- OS: Ubuntu 26.04
- vsag version: v1.0.0-182-g7dc90ad5
```

### Expected behavior

`mci_clique_max` should act as the minimum size a maximal clique must have to be kept (as `cliqueMax` does in the reference implementation), and an enumerated maximal clique should be stored in full, naturally bounded by the candidate pool (`mci_mcs + 1`). The parameter is also documented as "Maximum clique size during full build" in `docs/docs/{en,zh}/src/advanced/hgraph_mci_companion.md`, which contradicts the intended semantics.

### Actual behavior

`src/algorithm/mci/mci_builder.cpp` clips the enumeration result at both save sites:

```cpp
if (clique.size() > params.clique_max) {
    clique.resize(params.clique_max);
}
```

and the generic (quantized / non-float) build path does the same in `normalize_clique` (`src/algorithm/hgraph/hgraph_mci.cpp`). Because `clique_threshold` is derived from the same value, no clique can ever exceed `mci_clique_max`.

Impact on a 5M-vector, 1536-dim multi-label dataset (`mci_mcs=200, mci_clique_max=50, base=rabitq/precise=sq8`):

| metric | vsag | reference implementation |
| --- | --- | --- |
| clique sizes | always exactly 50 | 50..201 (mean 99.56) |
| mean clique-union degree | 430 | 1501 |
| recall@100, `seed_ratio=0.1`, `ef_search=1000` | 0.6313 | 0.7943 |

### References

- `src/algorithm/mci/mci_builder.cpp:153` (`clique_threshold`) and the two `clique.resize(params.clique_max)` sites
- `src/algorithm/hgraph/hgraph_mci.cpp` (`normalize_clique` in `build_mci_clique_index`)
- `src/algorithm/hgraph/hgraph_mci_test.cpp`: the unit test "MCI builder treats clique_max only as a storage cap" asserts `clique.size() <= clique_max`, i.e. it currently pins the buggy behavior.

### Additional context

Discovered while comparing the vsag MCI companion against the reference maximal-clique-index implementation on the same KNN graph. Fixing the truncation (keeping `clique_max` as a minimum and storing cliques in full) restores the degree distribution and recall: mean clique-union degree 1610.6 vs 1500.9 for the reference, and recall@100 at `seed_ratio=0.1`, `ef_search=1000` improves from 0.6313 to 0.7909 (reference 0.7943).

LightWant:
Fixes #2972

## Summary

Two changes to the `hgraph` MCI companion. The first restores the intended meaning of `mci_clique_max`, which is a correctness bug reported in #ISSUE_NUMBER; the second is an opt-in fast path for filtered MCI searches that roughly doubles its QPS.

1. `fix(mci)`: `mci_clique_max` is a lower bound, and an enumerated maximal clique is stored in full (bounded only by the candidate pool, `mci_mcs + 1`).
2. `perf(mci)`: a new optional `Filter::GetValidBitmap()` that the MCI searcher merges into its visited marks, so a probe resolves "filtered out" and "already visited" with one memory access and no virtual `CheckValid()` call.

## 1. Clique truncation

`BuildMCICliques()` clipped every clique right before storing it (`clique.resize(params.clique_max)`), while the same `clique_max` value was already used as the minimum size of the maximal-clique enumeration. Every stored clique therefore had exactly `mci_clique_max` members, and the clique-union degree of the built index was ~3.5x smaller than the reference implementation's for the same graph and parameters. The generic build path used for quantized data clipped cliques the same way in `normalize_clique`.

`clique_max` now only sets the enumeration threshold, matching `cliqueMax` in the reference MCI implementation. The only remaining bound on a stored clique is its candidate pool, so the defensive resize keeps `mci_mcs + 1` as its limit. Docs and the `HGraphMCIParameters` comment are updated, and the existing unit test that pinned the old behavior (`clique.size() <= clique_max`) is replaced.

Measured on a 5M-vector, 1536-dim multi-label dataset (`mci_mcs=200`, `mci_clique_max=50`, `base=rabitq`, `precise=sq8`), against the reference implementation on the same KNN graph:

| clique-union degree | reference | before | after |
| --- | --- | --- | --- |
| p10 | 306 | 50 | 318 |
| p50 | 1062 | 182 | 1125 |
| p90 | 2958 | 1055 | 3276 |
| mean | 1500.9 | 429.9 | 1611.7 |

`degree_min` is 50 in both cases, i.e. `mci_clique_max` is respected as a lower bound in the quantized path as well (see the third commit, which applies the same threshold rule there as in the float path).

| recall@100, `seed_ratio=0.1` | reference (`ef=1280`) | before (`ef=1000`) | after (`ef=1000`) |
| --- | --- | --- | --- |
| recall | 0.7943 | 0.6313 | 0.7931 |

Build cost is unchanged (647 s vs 772 s before the fix on 64 threads, 11.73 GB peak RSS).

The dataset in these measurements stores RaBitQ base codes with SQ8 precise codes, so the degree and recall figures above exercise the generic (quantized) enumeration path; the float path (`mci_builder.cpp`) applies the same threshold rule and is covered by the unit tests below.

## 2. Bitmap fast path

`Filter` gains an optional

```cpp
[[nodiscard]] virtual const uint8_t* GetValidBitmap(uint64_t* size) const;
```

with one byte per id in the same id space `CheckValid()` is called with, non-zero meaning valid. The MCI searcher merges that bitmap into `MCIMergedVisitMarks`, so `try_visit()` answers "filtered out" and "already visited" with a single array access instead of a visited check plus a virtual `CheckValid()` call per probed vector.

Correctness of the id space: the searcher indexes the bitmap with inner ids, while a user filter observes labels. `InnerIdWrapperFilter` therefore forwards the wrapped bitmap only when the label table maps every live inner id to itself, via the new cached `LabelTable::IsIdentityMapping()` that every label mutation invalidates. Searches whose labels differ from their inner ids keep the callback path. A test builds an index with shifted labels and asserts that a bitmap claiming the *rejected* ids is ignored, so the results still follow `CheckValid()`.

5M-vector, 1536-dim multi-label dataset, 64 threads, 2000 queries, `mci_seed_ratio=5`. Two alternating runs per configuration (QPS on this machine depends on the external load, so the mean latency is the more stable column):

| `ef_search` | QPS before | QPS after | speedup | mean latency | recall (identical) |
| --- | --- | --- | --- | --- | --- |
| 20 | 1669 / 1701 | 3282 / 3431 | 1.97x / 2.02x | 27.2 / 26.9 ms -> 16.0 / 15.8 ms | 0.89845 |
| 160 | 1221 / 1221 | 2501 / 2609 | 2.05x / 2.14x | 39.0 / 38.6 ms -> 21.3 / 21.3 ms | 0.95047 |
| 600 | 455 / 466 | 920 / 967 | 2.02x / 2.07x | 108.2 / 108.3 ms -> 57.7 / 58.1 ms | 0.98334 |

The marks are per-search scratch allocated from the allocator the search was given, so nothing outlives the search: an earlier revision cached them in a `thread_local` and the unit-test leak check caught the allocation still being live when the index (and its allocator) was destroyed. Making them per-search did not cost anything measurable, as the table above shows.

Recall and the returned ids are bit-identical with and without the fast path at every `ef_search`. The statistic `mci_bitmap_fast_path` reports whether a search used the bitmap.

## Tests

`unittests_algorithm "[mci]"` (134 assertions in 17 cases) and `unittests_impl "[LabelTable],[InnerIdWrapperFilter]"` (1210 assertions in 15 cases) pass, plus `./scripts/format/format-cpp.sh` (clang-format 15) is clean.

New/updated coverage:

- `MCI builder keeps maximal cliques larger than clique_max` (replaces the test that asserted the truncation) and `MCI builder stores a full candidate-pool clique`: both fail on the unmodified builder and pass with the fix.
- `HGraph MCI inlines an identity-mapped filter bitmap`: asserts the fast path is taken and returns exactly the same ids as the callback path.
- `HGraph MCI ignores a bitmap whose ids are not inner ids`: asserts the fast path is disabled and the rejected ids stay out of the result.
- `LabelTable reports identity label mappings`: covers insert, relabel, move, pre-allocated capacity and deserialize invalidation.
- `InnerIdWrapperFilter forwards a bitmap only for identity labels`.


src/impl/label_table/label_table.h
```diff
@@ -341,7 +345,44 @@ class LabelTable {
     InnerIdType
```
LHT129, line 345: [note] `IsIdentityMapping()` reads `label_table_[id]` in a loop without holding any lock. While the `labels_version_` atomic with acquire-release ordering provides happens-before for writes that completed before the version bump, there is no mutual exclusion with a concurrent `Insert()` or `Resize()` that may be modifying `label_table_` (e.g. resizing the underlying vector storage). If a search thread calls `IsIdentityMapping()` (via `InnerIdWrapperFilter::GetValidBitmap()`) while a build thread is inserting labels, this is a data race.

In VSAG's current architecture, searches and label mutations are typically serialized at a higher level, so this is unlikely to manifest in practice. If that invariant ever changes, consider either protecting the read with a shared lock or documenting that `IsIdentityMapping()` must not be called concurrently with label mutations.

LHT129 (commented):
Thanks for this PR. The changes are well-structured and the test coverage is thorough. Here is a summary of the review:

**Clique truncation fix (mci_builder.cpp, hgraph_mci.cpp):** Correctly restores the intended semantics of `mci_clique_max` as a lower bound rather than a storage cap. The defensive resize to `candidate_limit + 1` is the right safety net. The updated unit tests properly validate that maximal cliques larger than `clique_max` survive intact.

**Bitmap fast path (mci_searcher.cpp, filter.h, inner_id_wrapper_filter.h):** Clean design. The `MCIMergedVisitMarks` merging filter validity and visited state into one byte array is an elegant optimization. The identity-mapping guard in `InnerIdWrapperFilter::GetValidBitmap()` correctly prevents misuse when labels differ from inner ids. The `thread_local` storage for `merged_marks` is consistent with the existing `MCIEpochMarks` pattern.

**LabelTable::IsIdentityMapping (label_table.h):** Good caching strategy with version-based invalidation. The `MarkLabelsMutated()` calls are correctly placed at every mutation point (insert, relabel, move, resize, deserialize).

**Tests:** Comprehensive — covers the clique fix, the bitmap fast path with identity-mapped labels, the bitmap rejection path with shifted labels, and the `IsIdentityMapping` caching behavior across insert/relabel/move/resize/deserialize.

One `[note]`-level comment was left inline regarding a potential data race in `IsIdentityMapping()` — unlikely to manifest given VSAG's architecture but worth being aware of.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -66,6 +66,54 @@ struct MCIEpochMarks {
...
+                        MCIMergedVisitMarks& merged_marks) {
+    if (mci_param.valid_bitmap == nullptr or mci_param.valid_bitmap_size < total) {
+        return nullptr;
+    }
+    merged_marks.Reset(mci_param.valid_bitmap, total);
+    return mci_param.valid_bitmap;
+}
+
 bool
 mci_check_overtime(const InnerSearchParam& inner_search_param, QueryContext* ctx) {
```
LHT129, line 134: [note] `PrepareMergedVisitMarks` returns the raw `mci_param.valid_bitmap` pointer, but callers only use the return value as a non-null sentinel to select the fast path in `try_visit`. The actual merged marks live in `merged_marks.state`, not in the bitmap. This is correct but subtle — a future reader might mistakenly think the bitmap itself was modified. Consider returning `bool` or a dedicated sentinel type to make the intent clearer.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -66,6 +66,54 @@ struct MCIEpochMarks {
...
+struct MCIMergedVisitMarks {
+    std::vector<uint8_t> state;
+
+    void
+    Reset(const uint8_t* valid_bitmap, uint64_t total) {
+        if (state.size() < total) {
+            state.resize(total);
+        }
+        // The destination is reached through a local pointer: indexing the vector member directly
+        // makes the compiler reload its data pointer on every iteration and keeps the loop scalar.
```
LHT129, line 94: [suggestion] The `Reset()` loop normalizes the bitmap byte-by-byte with a branch per element. For large `total` values this loop runs on every search that uses the bitmap fast path. Consider whether the bitmap provider could guarantee 0/1 values directly (so the copy is just a `memcpy`), or whether a SIMD path (e.g. `_mm256_cmpeq_epi8` against zero) could amortize the normalization cost. The current scalar loop is fine for correctness, but the normalization cost eats into the ~2x speedup the fast path otherwise delivers.

include/vsag/filter.h
```diff
@@ -85,6 +85,26 @@ class Filter {
...
+      * non-zero byte means the id is not filtered out. Exposing it lets index implementations
+      * inline the per-candidate check instead of paying a virtual call for every probed vector.
+      * Providers whose id space is not the inner-id space must keep the default nullptr, and
+      * callers must confirm *size covers the ids they are about to index.
+      *
+      * @param size receives the number of ids covered by the returned bitmap
+      * @return pointer to the bitmap, or nullptr when this filter cannot provide one
+      */
+    [[nodiscard]] virtual const uint8_t*
+    GetValidBitmap(uint64_t* size) const {
```
LHT129, line 102: [note] The `GetValidBitmap` API documents that `size` receives the number of ids covered by the bitmap, but callers in `hgraph_mci.cpp` compare it against `total_count` as a safety check (`bitmap_size >= total_count`). This is the correct usage, but the docstring could be more explicit that callers must validate `*size >= N` before indexing the bitmap with ids in `[0, N)`. The current wording "callers must confirm *size covers the ids they are about to index" already says this — just noting that the check in `try_mci_search` is the right pattern for future implementors to follow.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -66,6 +66,54 @@ struct MCIEpochMarks {
...
+            destination[id] = valid_bitmap[id] == 0 ? 0 : 1;
+        }
+    }
+
+    // Returns true only for a valid id that was not visited yet; 2 records the visit.
+    [[nodiscard]] bool
+    TryVisit(InnerIdType id) {
+        uint8_t& value = state[id];
+        if (value != 1) {
+            return false;
```
LHT129, line 110: [suggestion] `TryVisit` indexes `state[id]` without a bounds check. While the current callers guarantee `id < total` and `Reset()` ensures `state.size() >= total`, adding an explicit `id < state.size()` guard would make the invariant locally verifiable and protect against future misuse (e.g. if `Reset` is called with a smaller `total` than a previous call, stale entries beyond `total` remain in `state` but would not be accessed under the current contract).

Consider:
```cpp
[[nodiscard]] bool
TryVisit(InnerIdType id) {
    if (id >= state.size()) {
        return false;
    }
    uint8_t& value = state[id];
    if (value != 1) {
        return false;
    }
    value = 2;
    return true;
}
```

src/impl/searcher/mci_searcher.cpp
```diff
@@ -131,6 +179,11 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
     SearchCandidateQueue candidates(allocator);
     visited_nodes.Reset(total);
     visited_cliques.Reset(view.total_clique_count);
+    thread_local MCIMergedVisitMarks merged_marks;
```
LHT129, line 199: [note] The `thread_local MCIMergedVisitMarks` in both `search_precise_float_csr` and `MCISearcher::Search` grows its internal `state` vector to the largest `total` ever seen by that thread but never shrinks. If a thread alternates between searching a large index and a small one (e.g. after a merge that reduces `total`), the vector retains its peak allocation. `Reset()` only overwrites the first `total` entries, which is correct for correctness, but the memory overhead persists. This is unlikely to matter in practice since `total` is typically stable within a process lifetime, but worth being aware of.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -933,8 +946,12 @@ HGraph::build_mci_clique_index(const void* vectors) {
         normalized.assign(clique.begin(), clique.end());
         std::sort(normalized.begin(), normalized.end());
         normalized.erase(std::unique(normalized.begin(), normalized.end()), normalized.end());
-        if (normalized.size() > this->mci_parameters_.clique_max) {
-            normalized.resize(this->mci_parameters_.clique_max);
+        // mci_clique_max is a minimum size, not a storage cap: the maximal clique enumerated for
+        // this seed is kept in full. Its size cannot exceed the candidate pool (mcs neighbours plus
+        // the seed), so this bound only rejects an inconsistent result.
+        const auto clique_cap = std::min<uint64_t>(this->mci_parameters_.mcs, total - 1) + 1;
```
LHT129, line 959: [note] The `clique_cap` fix in `normalize_clique` correctly removes the storage truncation to `clique_max`, but the `clique_min` threshold passed to `run_local_ccr_mce` (line 996 in the PR branch, computed as `std::min(K_MCI_MIN_CLIQUE_SIZE, candidate_limit + 1, total)` where `K_MCI_MIN_CLIQUE_SIZE = 2`) still ignores `mci_clique_max`. In the float path (`mci_builder.cpp`), `clique_threshold` is `std::max(2, std::min(clique_max, candidate_limit + 1, total))`, which respects `clique_max` as a lower bound. The generic/quantized path should similarly use `mci_clique_max` in `clique_min` so that the enumeration threshold is consistent across both paths.

Suggested fix:
```cpp
const auto clique_min = std::min<uint64_t>(
    {std::max<uint64_t>(K_MCI_MIN_CLIQUE_SIZE, this->mci_parameters_.clique_max),
     candidate_limit + 1,
     total});
```

src/impl/label_table/label_table.h
LightWant: Fixed in 733a3e5e.

The read is now protected on the MCI path instead of relying on the higher-level serialization: `try_mci_search()` fetches the bitmap — and therefore the `IsIdentityMapping()` scan — while holding `label_lookup_mutex_` in shared mode, in the same scope that already reads labels for seed collection. That also covers the bitset branch, whose `CheckValid()` calls read labels too.

The bitmap fetch moved ahead of `persistent_codes_mutex_`, so the lock order this function already used (label lock, then codes lock) is unchanged and no new nesting was introduced.

I also documented the requirement on `IsIdentityMapping()` itself: `labels_version_` only protects the cached answer, not the scan, so callers have to exclude concurrent mutations — the search paths do that with the shared label lock.

src/impl/searcher/mci_searcher.cpp
LightWant: Agreed it is subtle; I kept the pointer return. `PrepareMergedVisitMarks` validates the pair (`valid_bitmap != nullptr && bitmap_size >= total`) and returns the validated bitmap or `nullptr`, so the return value reads as "the fast path is armed with this bitmap", while the marks themselves live in the clearly named `merged_marks` out-parameter, and `try_visit` only tests `valid_bitmap != nullptr`.

Happy to switch it to `bool` plus reading `mci_param.valid_bitmap` directly as the normalization source — a four-line change with no behavioural difference — but I left it alone here to keep this iteration limited to the changes that were required.

src/impl/searcher/mci_searcher.cpp
LightWant: No further change: the loop is already vectorized in the shipped build, and it was scalar in the earlier revision for a different reason.

1. The provider contract is deliberately "any non-zero byte means valid", so the searcher normalizes instead of `memcpy`-ing. Requiring 0/1 would make correctness depend on an undocumented convention: a provider returning e.g. 255 for valid would be read as *visited* by the merged encoding (2 = visited). Normalizing keeps the API tolerant of any provider that follows the documented rule.
2. The cost is already amortized. In this TU the loop compiles to `movdqu`/`pcmpeqb`/`pand`/`movups`, i.e. SSE2 at 16 bytes per iteration — not AVX-512, because only `src/simd/` is built with arch flags while `mci_searcher.cpp` uses the baseline x86-64 flags. It only became vectorized after hoisting `marks.data()` into a local: indexing the `std::vector` member directly made the compiler reload its data pointer every iteration (`cmpb`/`setne` scalar loop), which cost roughly a quarter of the fast path's benefit (2407 -> 3089 QPS at `ef_search=20` on the 5M dataset). That hoist and its comment are part of the PR.

include/vsag/filter.h
LightWant: No change: the docstring requirement and the check in `try_mci_search()` are meant to pair. The provider guarantees nothing about size; the consumer checks `bitmap_size >= total_count` and only then indexes ids in `[0, total_count)`, and `hgraph_mci.cpp` documents that as the pattern for implementors.

I kept the wording as is because it already states the contract ("callers must confirm *size covers the ids they are about to index"), but I am happy to spell out the `*size >= N` form explicitly if you think that reads better.

src/impl/searcher/mci_searcher.cpp
LightWant: Applied in 42c6ad30: `TryVisit` now begins with

```cpp
if (not state.has_value() or id >= state->size()) {
    return false;
}
```

The guard came in together with storing the marks in a `std::optional<Vector<uint8_t>>` — needed to rebind the cached buffer to the index allocator — where "no state yet" is a legal state, so the bounds check is required rather than merely defensive.

src/impl/searcher/mci_searcher.cpp
LightWant: Acknowledged, and kept as is deliberately: retaining the peak allocation is what makes the fast path cheap — the marks are per-thread scratch that is rewritten on every search, and reallocating them per search would add an allocation plus page faults to every query.

In the current revision the buffer is bound to the *index* allocator rather than the query allocator, so a rebind happens only when a thread searches a different index, and `TryVisit` bounds-checks against the live size, so entries beyond `total` left over from a larger index are never read. The trade-off to be aware of is that the memory goes back to the allocator when the thread exits, not when a search ends; if that ever becomes a problem for small indexes after a merge, the alternative is a per-search local `Vector<uint8_t>` at the cost of one allocation per query.

src/algorithm/hgraph/hgraph_mci.cpp
LightWant: Fixed in 733a3e5e — the generic path now derives its threshold the same way as the float path:

```cpp
const auto clique_min = std::min<uint64_t>(
    {std::max<uint64_t>(K_MCI_MIN_CLIQUE_SIZE, this->mci_parameters_.clique_max),
     candidate_limit + 1,
     total});
```

so `mci_clique_max` is a lower bound for quantized data again, instead of having been only the (now removed) truncation cap.

Note that this also changes which cliques a quantized index enumerates, so the 5M measurements in the description (degree distribution and the filtered A/B) are being re-run against an index rebuilt with this change; I will refresh those numbers once they are in.

LHT129 (commented):
This is a well-structured PR with two clear changes: fixing the clique truncation bug and adding the bitmap fast path for filtered MCI searches. The test coverage is thorough (134 assertions in 17 MCI cases, 1210 assertions in 15 LabelTable/InnerIdWrapperFilter cases), and the performance data in the PR description shows the expected ~2x QPS improvement with bit-identical recall.

The existing inline comments cover the main review findings — the `IsIdentityMapping()` concurrency contract, the bitmap normalization loop performance, and the `PrepareMergedVisitMarks` return-type clarity. The latest commit already addressed the `TryVisit` bounds check and the `clique_min` consistency issue that were noted earlier.

No blocking issues from my side. The PR description includes detailed benchmarks and correctness validation, which is appreciated.

LHT129 (commented):
Reviewed commit 733a3e5e.

This PR makes two well-scoped changes:

1. **clique_max semantics fix** — Correctly changes `mci_clique_max` from a storage cap to a lower bound on maximal clique size, matching the reference MCI implementation. Both the float path (`mci_builder.cpp`) and the generic path (`hgraph_mci.cpp`) now use `clique_cap = candidate_limit + 1` as the defensive upper bound, and `clique_threshold`/`clique_min` as the enumeration lower bound derived from `clique_max`.

2. **Bitmap fast path for filtered MCI searches** — Adds `Filter::GetValidBitmap()` to the public API, `InnerIdWrapperFilter` forwarding gated on `LabelTable::IsIdentityMapping()`, and `MCIMergedVisitMarks` to merge filter validity and visit tracking into a single byte array per probe. The `labels_version_` atomic with acquire-release ordering correctly gates the cached identity answer.

The 7 review comments from the previous round have all been addressed in this commit:
- `IsIdentityMapping()` concurrency: now protected by `label_lookup_mutex_` on the MCI search path
- `PrepareMergedVisitMarks` sentinel return: acknowledged as intentional
- `Reset()` loop normalization: acknowledged as already vectorized in shipped builds
- `GetValidBitmap` size contract: documented in the API and checked at the call site
- `TryVisit` bounds check: added in 42c6ad30
- `thread_local` caching: acknowledged as intentional for performance
- `clique_min` threshold in generic path: fixed in 733a3e5e

The test coverage is thorough: unit tests for `LabelTable::IsIdentityMapping` across insert/relabel/move/resize/deserialize, `InnerIdWrapperFilter::GetValidBitmap` for identity and non-identity label tables, MCI builder clique size assertions, and end-to-end HGraph MCI search tests verifying the bitmap fast path produces identical results to the callback path and correctly falls back when labels are not identity-mapped.

No new substantive issues found.

src/impl/label_table/label_table.h
```diff
@@ -341,7 +345,49 @@ class LabelTable {
...
+    /**
+     * @brief Whether this table maps every live inner id to itself, i.e. the label space and the
+     * inner-id space coincide.
+     *
+     * Components that reuse an id-space-indexed structure as an inner-id-indexed one (for example
+     * the validity bitmap that the MCI searcher inlines) must check this first, because the label
+     * a filter observes is not necessarily the inner id it is probed with. Only the live ids
+     * `[0, GetTotalCount())` take part: the table pre-allocates capacity, and those spare slots are
+     * never probed. The answer is cached and recomputed after any label mutation.
+     *
```
LHT129, line 365: [note] The `IsIdentityMapping()` docstring now correctly documents that "scanning the table requires concurrent label mutations to be excluded" and that "Search paths therefore call this while holding the label lock in shared mode." This is the right contract, and the MCI path in `hgraph_mci.cpp` follows it by calling `GetValidBitmap()` inside the `label_lookup_mutex_` shared-lock scope.

One subtlety worth noting for future readers: `IsIdentityMapping()` is a `const` public method on `LabelTable`, so nothing in the type system prevents a caller from invoking it without the lock. The comment is the only guard. If the codebase ever adds a static analysis annotation for lock requirements (e.g. `GUARDED_BY`), this method would be a good candidate.

LHT129 (commented):
This is a well-structured PR with two clear, independently motivated changes:

**1. Clique truncation fix** — Correctly restores `mci_clique_max` to its intended semantics as a lower bound (matching the reference MCI implementation). The fix in `mci_builder.cpp` (float path), `hgraph_mci.cpp` (generic/quantized path), and the updated docs are all consistent. The replacement tests (`MCI builder keeps maximal cliques larger than clique_max` and `MCI builder stores a full candidate-pool clique`) properly validate the new behavior and would have caught the old bug.

**2. Bitmap fast path** — The `Filter::GetValidBitmap()` API is cleanly designed with a safe default. The `InnerIdWrapperFilter` gating on `LabelTable::IsIdentityMapping()` is the right correctness guard, and the negative test (`HGraph MCI ignores a bitmap whose ids are not inner ids`) convincingly proves the guard works. The `IsIdentityMapping()` caching with version-based invalidation is well-implemented, and the `MarkLabelsMutated()` calls appear to cover all mutation paths (Insert, UpdateLabel, Move, Resize, ShrinkToFit, Deserialize, MergeOther).

The ~2x QPS improvement with bit-identical results is compelling evidence that the fast path works correctly.

The existing inline comments cover the remaining discussion points (lock documentation, normalization loop, allocation strategy). No new blocking issues from my side.

wxyucs (approved):
lgtm

Issue #2975, opened 2026-09-18
[bug](build): repair 1.0 header and FILE deleter compilation with gcc 15


wxyucs:
### Affected area

build

### Interface

cpp

### VSAG version

Release branch `1.0`, commit `78470e70ecf098d3b22988c39188c50ca5e6f083`.

### Summary

The optimized build with GCC 15.2 fails because several headers use fixed-width integer types without including `<cstdint>`. After correcting the includes, the FILE deleter in streaming serialization also triggers `-Werror=ignored-attributes`.

### Minimal reproducer

```sh
make release COMPILE_JOBS=8 VSAG_ENABLE_TESTS=ON VSAG_ENABLE_MOCKIMPL=ON
```

A standalone public-header reproducer is:

```sh
printf '#include "vsag/attribute.h"\n' | g++ -std=c++17 -Iinclude -x c++ -fsyntax-only -
```

The same check exposes the missing include in `vsag/binaryset.h`. Both public-header checks reproduce with `_GLIBCXX_USE_CXX11_ABI=0` and `1`.

### Environment

Actual `bash scripts/check_environment.sh` output after preparing fixes:

```text
- OS: Ubuntu 26.04
- vsag version: v1.0.0-15-g78470e70-dirty
- compiler version: c++ (Ubuntu 15.2.0-16ubuntu1) 15.2.0
```

x86_64 Linux, CMake 4.2.3, GNU Make 4.4.1, system OpenBLAS 0.3.32, LAPACKE and libaio. The default AUTO dependency policy selects system OpenBLAS. This reproduction uses the installed GCC 15 toolchain; it is not a GCC 14 manylinux/container reproduction.

### Expected behavior

The Release build compiles with tests enabled, and the public headers compile independently.

### Actual behavior

The first original build failure is `extern/diskann/DiskANN/include/natural_number_map.h:37`: `uint64_t does not name a type [-Wtemplate-body]`. The build also reports undeclared `uint64_t` in `src/attr/expression.h`; independent probes fail for `include/vsag/attribute.h` and `include/vsag/binaryset.h`.

After adding the missing includes, `src/storage/tlv_section.cpp:277` fails with `-Werror=ignored-attributes` because `decltype(&std::fclose)` is used as a template argument for `std::unique_ptr`.

### References

- `extern/diskann/DiskANN/include/natural_number_map.h:37`
- `include/vsag/attribute.h:17`
- `include/vsag/binaryset.h:32`
[...]

PR #2976, opened 2026-09-18
fix(build): repair 1.0 GCC 15 header compilation

wxyucs:
## Change Type

- [x] Bug fix
- [x] CI/Build/Infra

## Linked Issue

Fixes: #2975

## What Changed

GCC 15.2 rejects the 1.0 Release build when fixed-width integer types lack a direct `<cstdint>` include and when the attributed `std::fclose` pointer is used as a template argument. Add the four required includes and use the existing main-branch FILE deleter, preserving temporary-file cleanup and error handling. Include `expression.h` first and assert its integer width before other headers in the existing visitor test to cover header self-containment.

The attribute/expression includes and deleter are narrow backports from #2681; the BinarySet include comes from #2699. The sole DiskANN change is the standard-header include. No algorithms or compiler diagnostic policies change.

## Test Evidence

- [x] Standalone C++17 compilation of all four affected headers with `_GLIBCXX_USE_CXX11_ABI=0` and `1`.
- [x] `tlv_section.cpp` and `natural_number_map.cpp` compiled with generated Release flags under both ABI settings.
- [x] `make test-cmake`: all three suites pass.
- [x] Mock implementation tests: 1 case, 2 assertions pass.
- [x] Full-library attribute, expression visitor and streaming tests: 27 cases, 402 assertions pass.
- [x] Factory functional tests: 13 cases, 3,480 assertions pass.
- [x] `git diff --check`.
- [x] Exact clang-format 15: [fork CI passed](https://github.com/wxyucs/vsag/actions/runs/35325549340) at the current PR commit.

The original `make release COMPILE_JOBS=8 VSAG_ENABLE_TESTS=ON VSAG_ENABLE_MOCKIMPL=ON` passes, including the shared/static libraries and all test executables. The affected unit tests were rebuilt and rerun after the formatting correction. Bundled Release OpenBLAS 0.3.34 and CRoaring 3.0.1 also build successfully with `VSAG_USE_SYSTEM_DEPS=OFF`, tests enabled, and unchanged compiler flags (dependency targets only; not a second full-library build).

Local environment: Ubuntu 26.04 x86_64, GCC/G++ 15.2.0, CMake 4.2.3, default AUTO dependency policy using system OpenBLAS 0.3.32 and LAPACKE. No GCC 14 manylinux or container reproduction is claimed. ABI0 checks cover affected headers and objects, not a complete ABI0 distribution build. Exact clang-format 15 passed in CI. Clang-tidy 15 is running in [fork CI](https://github.com/wxyucs/vsag/actions/runs/35325554587); compatible local tools are unavailable. No local coverage percentage is claimed.

## Compatibility Impact

Public declarations and object layouts are unchanged. The private FILE deleter still calls `std::fclose` through the same RAII lifetime.

No expected runtime performance or concurrency change.

## Documentation Impact

- [x] No docs update needed; this repairs compilation without changing user behavior.

## Risk and Rollback

Low risk: direct standard includes and an existing main-branch deleter. Revert the single commit to roll back.

## Checklist

- [x] Linked the relevant issue.
- [x] Added header regression coverage and exercised existing affected tests.
- [x] Considered API/ABI compatibility.
- [x] Commit follows Conventional Commits with human DCO and AI attribution.


LHT129 (commented):
All changes look correct and minimal. Summary of review:

**`#include <cstdint>` additions (4 headers):** Correct fix for GCC 15 no longer transitively providing fixed-width integer types. Each header uses `uint64_t` or related types and now properly includes `<cstdint>`.

**`src/storage/tlv_section.cpp` — `file_closer` struct:** The replacement of `std::unique_ptr<std::FILE, decltype(&std::fclose)>` with a custom `file_closer` functor is the right approach. GCC 15 rejects the attributed `std::fclose` pointer as a template argument; wrapping it in a stateless callable avoids the issue cleanly. The RAII lifetime and error handling are preserved unchanged.

**`src/attr/expression_visitor_test.cpp` — header self-containment test:** Including `attr/expression.h` first and asserting `sizeof(uint64_t) == 8` before other includes is a reasonable regression guard for the `<cstdint>` fix. No issues.

No blocking concerns. The changes are narrow, well-targeted, and do not alter any public API, object layout, or runtime behavior.

src/attr/expression_visitor_test.cpp
```diff
@@ -13,6 +13,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "attr/expression.h"
+
+// Check the header's integer type before another include can provide it.
```
LHT129, line 18: [note] The comment says "Check the header's integer type before another include can provide it", but the `static_assert` actually checks `sizeof(uint64_t) == 8`, which verifies the width of the type rather than whether the header provides it. If `<cstdint>` were missing, compilation would already fail at `#include "attr/expression.h"` (since `expression.h` uses `std::numeric_limits` etc. that depend on `<cstdint>`).

Consider rewording the comment to more accurately describe what the assert verifies, e.g.:
```cpp
// Verify uint64_t is 8 bytes wide before other headers could potentially redefine it.
static_assert(sizeof(uint64_t) == 8);
```

src/storage/tlv_section.cpp
```diff
@@ -37,6 +37,13 @@ namespace vsag {
 
 namespace {
 
+struct file_closer {
+    void
```
LHT129, line 41: [suggestion] `file_closer::operator()` is used as a `std::unique_ptr` deleter, which is invoked during destruction. If `std::fclose` fails (returns `EOF`), the deleter should not throw, as throwing from a destructor context would call `std::terminate`. Consider marking it `noexcept`:

```cpp
struct file_closer {
    void
    operator()(std::FILE* file) const noexcept {
        std::fclose(file);
    }
};
```

This matches the common pattern used in the standard library (e.g., `std::default_delete` has a `noexcept` `operator()`).

src/attr/expression_visitor_test.cpp
wxyucs: The include itself is the header self-containment regression, and the assertion additionally verifies the 64-bit width before subsequent headers are included. Placing a declaration here also prevents clang-format 15 from moving the expression header after the other includes, which happened in the initial CI run. The comment describes that ordering purpose; it does not imply that later headers should redefine uint64_t. I will retain the current narrow test to avoid another source revision for this optional wording note. The amended test passes locally, and exact clang-format 15 passes in CI.

src/storage/tlv_section.cpp
wxyucs: std::fclose reports failure through its return value; it does not throw a C++ exception when it returns EOF. This deleter only calls std::fclose and preserves the existing cleanup behavior. An explicit noexcept annotation would document that property, but is not needed to correct this compilation failure. I will keep the exact deleter already merged on main in #2681 for this narrow 1.0 backport, without another source revision for the optional annotation.

LHT129 (approved):
LGTM

PR #2978, opened 2026-09-19
perf(mci): stop the clique traversal once it discovers nothing new

### LINKED ISSUE #2977: [improve](hgraph): MCI keeps expanding candidates that discover nothing
### Affected module

module/index

### Affected area

not applicable

### Interface

cpp

### VSAG version

main @ `897715d5`

### Summary

The MCI search traversal expands every candidate it collects, even when an expansion adds neither a new candidate nor a new result. Cliques overlap heavily, so once the reachable neighbourhood has been scored the remaining expansions only walk members that are already visited.

Measured on a 5M-vector quantized index (`mci_mcs=200`, `mci_clique_max=50`, `ef_search=600`, `mci_seed_ratio=10`), where a vector belongs to 22.4 cliques and an average clique has 102 members:

| | recall@100 | mean latency | cliques expanded per query |
| --- | --- | --- | --- |
| current | 0.9869 | 66 ms | 11k-27k |
| with a progress-based stop | 0.9856 | 31 ms | 0.7k-2.0k |

The traversal scores only 0.68 points per expanded clique. Nothing bounds it either: `hops_limit` defaults to `UINT32_MAX` and `apply_hops_limit()` replaces any configured value that is `<= ef_search` with `UINT32_MAX`, so the only exits are the candidate queue draining and `timeout_ms`.

### Expected behavior

The traversal should stop once its frontier cannot add anything new, instead of walking the cliques of every collected candidate to the end of the candidate queue.

### Actual behavior

Every candidate is expanded. For the filters whose seeds already cover every valid vector the whole expansion phase is pure overhead: recall is bit-identical with and without it, while the latency is 4-11x higher (e.g. a filter with 2485 valid vectors: 87.0 ms vs 9.9 ms at `ef_search=600`).

### References

- `src/impl/searcher/mci_searcher.cpp` (`MCISearcher::Search`, the clique expansion loop)
- `src/algorithm/hgraph/hgraph_search.cpp` (`apply_hops_limit`)

### Additional context

A pull request implements a stop after 32 consecutive candidates that discovered nothing new, for the generic (quantized) search path, with a unit test and a documentation section. The float-CSR path (`mci_raw_float_csr`) can adopt the same rule once it has test coverage.

LightWant:
Fixes #2977

## Summary

`MCISearcher::Search` expands every candidate it collects, even when an expansion adds neither a candidate nor a result. Cliques overlap heavily — on a 5M-vector quantized index a vector belongs to 22.4 cliques and an average clique has 102 members — so once the reachable neighbourhood has been scored the remaining expansions only walk already-visited members. That index expanded 11k-27k cliques per query while scoring 0.68 points per expanded clique.

This stops the expansion after 32 consecutive candidates that discovered nothing new (`K_MCI_IDLE_CANDIDATE_LIMIT`). Nothing else changes: no lower-bound stop is added, the seed phase and the candidate ordering are untouched.

## Measurements

5M-vector quantized index, `mci_mcs=200`, `mci_clique_max=50`, `ef_search=600`, `mci_seed_ratio=10`, 2000 queries, 64 threads. Same binary and index, alternating runs:

| | recall@100 | mean latency | p50 | cliques expanded per query |
| --- | --- | --- | --- | --- |
| before | 0.98687 | 66.4 / 65.2 ms | 58.4 ms | 11k-27k |
| after | 0.98562 | 30.6 / 31.6 ms | 15.0 ms | 0.7k-2.0k |

Per query class:

- Filters whose seeds already cover every valid vector (`seed_count == valid_count`) keep **bit-identical recall** while their latency drops 4-11x, and their clique expansions drop 9-23x. For example the filter with 2485 valid vectors: 87.0 ms -> 9.9 ms and 16124 -> 697 expansions; the one with 9905 valid vectors: 99.1 ms -> 18.4 ms.
- Filters whose valid set is larger than the seed budget expand roughly as before and lose 0.1-1.7 recall points, which is where the aggregate -0.0013 comes from.
- An idle limit of 8 was measured as well: recall 0.97890, mean 22.8 ms (a more aggressive trade, not used here).

With the stop, a filtered MCI search at `ef_search=600` goes from 1.5x slower than routing the same queries to brute force (43.7-44.0 ms mean, recall 0.98960) to 1.4x faster than it, at -0.004 recall (0.98562).

## Also measured, not included

A lower-bound stop (a separate top-k heap, breaking once the closest unexpanded candidate is worse than the k-th result) had no measurable effect on this workload: MCI keeps discovering points better than the current k-th result until its frontier saturates, so only the progress-based stop matters here.

## Test

`HGraph MCI stops expanding cliques that discover nothing`: a 512-vector fp32 index with `mci_mcs=64`, a filter whose 400 valid vectors are all reached by the seeds, and `ef_search=512` so the candidate queue can hold them all. It asserts the MCI route, that the seeds covered the whole valid set, that the five returned ids are exactly the five nearest valid vectors (the stop must not drop results), and that the traversal stays under 150 clique expansions. On that index the traversal expands **314 cliques without the stop and 38 with it** (measured by disabling the constant), so the bound separates the two behaviours with a wide margin.

`./unittests_algorithm "[mci]"` passes (15 cases).

## Docs

New "Search Traversal and Early Stop" section in `docs/docs/en/src/advanced/hgraph_mci_companion.md` and `docs/docs/zh/src/advanced/hgraph_mci_companion.md`.

## Notes

- The float-CSR search path (raw float precise codes, `mci_raw_float_csr`) is not covered here: that path could not be reached from a unit test, so changing it would land untested. The generic path covers quantized indexes, which is where MCI serves the large filtered sets this is about.
- QPS on the measurement machine varies with external load, so the table reports mean latency; the two configurations were measured in alternating runs to keep the comparison fair.


src/impl/searcher/mci_searcher.cpp
```diff
@@ -66,6 +66,14 @@ struct MCIEpochMarks {
     }
 };
 
+// A candidate whose cliques are all visited contributes nothing, yet the traversal still walks
+// them: measured 0.7 newly scored points per expanded clique on a 5M quantized index. Stop the
+// expansion once this many consecutive candidates have discovered nothing new. At ef_search=600
+// this removed 9-23x of the clique expansions for the filters whose seeds already cover every valid
```
LHT129, line 72: [note] The `search_precise_float_csr` function (anonymous namespace) does not include the same idle-candidate early-stop logic that was added to `MCISearcher::Search`. The PR description acknowledges this — the float-CSR path cannot be reached from a unit test, so changing it would land untested. This is a reasonable trade-off for now, but if that path becomes testable in the future, applying the same early-stop would be worthwhile for consistency and to avoid a latent performance gap.

LHT129 (commented):
Thanks for this PR — the idle-candidate early-stop is a clean, well-measured optimization. The implementation is minimal and correct, and the test covers the key scenario well. I have one note below.

LHT129 (commented):
Reviewed the changes in `mci_searcher.cpp` and the corresponding test.

**Summary**: The idle-candidate early stop is a clean, low-risk optimization. The logic is correct:
- `accepted` tracks how many new points were actually inserted into the candidate queue during a candidate's expansion.
- After 32 consecutive candidates that discover nothing new, the traversal stops.
- The early stop bounds work, not recall — the heap already contains the best results found so far, and candidates are processed in increasing distance order.

**Test**: The new test case is well-designed — it creates a scenario where every valid vector is already seeded, so the expansion phase can only rediscover visited points. It verifies the MCI route, seed coverage, result correctness (top-5 nearest), and that hops stay bounded. The 150-hop bound cleanly separates the two behaviors (38 vs 314 cliques).

**Docs**: The new "Search Traversal and Early Stop" section in both English and Chinese docs clearly explains the mechanism and the tradeoff.

No issues found. The PR description already acknowledges that the `search_precise_float_csr` path is intentionally left unchanged because it cannot be reached from a unit test.

PR #2979, opened 2026-09-19
fix(pyramid): align RangeSearch radius filtering and ef_search with HGraph (#2963)

### LINKED ISSUE #2963: [bug](pyramid): align RangeSearch radius filtering and ef_search semantics with HGraph
### Affected module

module/index

### Affected area

Pyramid range search

### Interface

C++

### VSAG version

Not specified

### Summary

Pyramid range search has two semantic inconsistencies compared with HGraph:

1. Pyramid may return results whose exact distance exceeds the user-provided radius.
2. Pyramid does not increase `ef_search` to `limited_size`, which can unnecessarily restrict the candidate set for a single category path.

The second behavior may not strictly violate the public API because `limited_size` is defined as the maximum number of returned results and graph search is approximate. However, it differs from HGraph and can produce unexpectedly small result sets.

### Issue 1: results beyond the user-provided radius may be returned

For example, when the user provides:

```text
radius = 0.8
```

Pyramid may return results with exact distances in:

```text
(0.8, 0.88]
```

The current flow appears to be:

1. Pyramid changes the internal search radius to `radius * 1.1`.
2. Graph search, FLAT-node search, and final trimming after reorder do not restore the original radius.
3. The final result is trimmed only against the enlarged `search_param.radius`.
4. Therefore, candidates whose exact distances fall within
   `(radius, radius * 1.1]` can be returned.

This affects both:

- `Pyramid::RangeSearch(...)`
- The `RANGE_SEARCH` branch of `Pyramid::SearchWithRequest(...)`

HGraph behaves differently: before returning the final result, it trims candidates using `request.radius_ + THRESHOLD_ERROR`.

### Expected behavior

The enlarged radius may be used internally to improve recall, but the final result should be filtered using the original user-provided radius, with the documented numerical tolerance if applicable.

`RangeSearch(...)` and the `RANGE_SEARCH` branch of `SearchWithRequest(...)` should follow the same final-radius semantics.

### Actual behavior

The enlarged internal radius is also used for final trimming, so Pyramid may expose candidates outside the radius requested by the user.

### Issue 2: RangeSearch does not raise `ef_search` to `limited_size`

Consider:

```text
limited_size = 100
ef_search = 30
```

For a query targeting a single category path, Pyramid directly uses
`ef_search` and trims the merged node results for each query path to
`ef_search`. As a result, at most 30 candidates may survive for that path,
even when 100 or more points satisfy the radius.

HGraph instead uses:

```text
max(ef_search, limited_size)
```

and therefore expands the search window to at least 100 candidates in this
example.

### Expected behavior or clarification

Pyramid should either:

- use `max(ef_search, limited_size)` for range search, consistently with
  HGraph; or
- explicitly document that `ef_search` can cap the candidate set below
  `limited_size`, even for a single path with many qualifying points.

This does not imply that an approximate graph index must always return
`limited_size` results. The concern is the avoidable candidate cap and the
semantic difference between Pyramid and HGraph.

### Suggested fixes

1. Preserve the original user radius separately from the enlarged internal
   search radius.
2. After reorder, apply final trimming using the original radius plus the
   documented tolerance.
3. Apply the same rule to `RangeSearch(...)` and `SearchWithRequest(...)`.
4. Consider raising the effective Pyramid range-search `ef_search` to
   `max(ef_search, limited_size)`, or document the current behavior.
5. Add regression tests covering:
   - exact distances in `(radius, radius * 1.1]`;
   - both Pyramid range-search entry points;
   - reorder and non-reorder configurations;
   - FLAT and graph nodes;
   - a single path where `limited_size > ef_search`.

### Related issues

- #2319 — HGraph range-search filtering after reorder
- #2565 and #2566 — Pyramid range-result trimming defects

thlurte:
### What problem does this PR solve?

Fixes #2963

Aligns `Pyramid`'s range-search radius filtering and candidate discovery window with `HGraph`:

1. **Radius Leakage Filtering**:
   - `Pyramid::RangeSearch` and `Pyramid::SearchWithRequest(RANGE_SEARCH)` previously enlarged the internal search radius to `radius * 1.1F` for graph exploration, but `search_impl` trimmed candidates against `search_param.radius` (`radius * 1.1F`). Consequently, candidates with exact distances in `(radius, radius * 1.1]` leaked into the returned dataset.
   - Now passes the user-specified radius to `search_impl` and trims against `range_search_radius + THRESHOLD_ERROR` (`1e-5`), directly aligning with [`hgraph_search.cpp:655`](https://github.com/antgroup/vsag/blob/main/src/algorithm/hgraph/hgraph_search.cpp#L655).

2. **Candidate Truncation (`limited_size > ef_search`)**:
   - When `limited_size > ef_search`, Pyramid previously passed `ef_search` directly, prematurely capping candidates below `limited_size`.
   - Now sets `search_param.ef = limited_size > 0 ? std::max(static_cast<int64_t>(parsed_param.ef_search), limited_size) : parsed_param.ef_search;`, matching [`hgraph_search.cpp:591`](https://github.com/antgroup/vsag/blob/main/src/algorithm/hgraph/hgraph_search.cpp#L591).

3. **Regression Tests**:
   - Added `Pyramid RangeSearch radius filtering and ef_search alignment` in `tests/test_pyramid.cpp` covering:
     - Strict radius cutoff rejecting points in `(radius, radius * 1.1]` across both `RangeSearch` and `SearchWithRequest`.
     - Both reordered and non-reordered configurations.
     - Candidate expansion when `limited_size > ef_search`.

### Type of change

- [x] Bug Fix (non-breaking change which fixes an issue)
- [ ] New Feature (non-breaking change which adds functionality)
- [ ] Breaking Change (fix or feature that could cause existing functionality not to work as expected)
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Performance Improvement
- [x] Test cases
- [ ] Python SDK


LHT129 (commented):
Reviewed the changes in `pyramid.cpp`, `pyramid.h`, and `tests/test_pyramid.cpp`. The fix correctly addresses the radius leakage issue (#2963) by:

1. Passing the user-specified radius (not `radius * 1.1`) to `search_impl` and trimming candidates against `range_search_radius + THRESHOLD_ERROR`, matching HGraph behavior at `hgraph_search.cpp:655`.

2. Using `std::max(ef_search, limited_size)` for `search_param.ef` when `limited_size > 0`, preventing premature candidate capping, matching `hgraph_search.cpp:591`.

3. The new `range_search_radius` parameter defaults to `std::nullopt`, preserving backward compatibility for the KNN search path.

The test coverage is thorough: it validates both `RangeSearch` and `SearchWithRequest` entrypoints, covers both reordered and non-reordered configurations, and includes the `limited_size > ef_search` case.

No blocking issues found. LGTM.

src/algorithm/pyramid/pyramid.h
```diff
@@ -132,7 +132,7 @@ class Pyramid : public InnerIndexInterface {
     }
 
     explicit Pyramid(const ParamPtr& param, const IndexCommonParam& common_param)
-        : Pyramid(std::dynamic_pointer_cast<PyramidParameters>(param), common_param){};
+        : Pyramid(std::dynamic_pointer_cast<PyramidParameters>(param), common_param) {};
```
LHT129: [note] This whitespace change (`){};` → `) {};`) is purely cosmetic and unrelated to the radius filtering / ef_search fix. It adds noise to the diff and could cause trivial merge conflicts with other PRs touching this line. Consider reverting it to keep the diff focused on the actual fix.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -781,7 +782,9 @@ Pyramid::RangeSearch(const DatasetPtr& query,
     CHECK_ARGUMENT(parsed_param.hierarchy_op == PyramidSearchParameters::HierarchyOp::SINGLE,
                    "multi-hierarchy search (union/intersection) is not yet implemented");
     InnerSearchParam search_param;
-    search_param.ef = parsed_param.ef_search;
+    search_param.ef = limited_size > 0
+                          ? std::max(static_cast<int64_t>(parsed_param.ef_search), limited_size)
+                          : parsed_param.ef_search;
     search_param.radius = radius * RADIUS_EPSILON;
```
LHT129: [suggestion] The internal search radius is still enlarged to `radius * RADIUS_EPSILON` (1.1x) via `search_param.radius`, while HGraph passes the raw `request.radius_` directly (hgraph_search.cpp:593). Although the result trimming now correctly uses the user-specified radius, the graph exploration window remains 10% wider than HGraph's. This may cause Pyramid to visit more nodes than necessary during range search.

If the intent is to fully align with HGraph, consider setting `search_param.radius = radius` (without the RADIUS_EPSILON multiplier) and relying solely on `range_search_radius` for the final cutoff. If the 1.1x expansion is intentional for Pyramid's multi-layer structure, a comment explaining why would help future readers.

src/algorithm/pyramid/pyramid.cpp
```diff
@@ -874,7 +878,10 @@ Pyramid::SearchWithRequest(const SearchRequest& request) const {
     } else {
         CHECK_ARGUMENT(parsed_param.hierarchy_op == PyramidSearchParameters::HierarchyOp::SINGLE,
                        "multi-hierarchy search (union/intersection) is not yet implemented");
-        search_param.ef = parsed_param.ef_search;
+        search_param.ef =
+            request.limited_size_ > 0
+                ? std::max(static_cast<int64_t>(parsed_param.ef_search), request.limited_size_)
+                : parsed_param.ef_search;
         search_param.radius = request.radius_ * RADIUS_EPSILON;
```
LHT129: [suggestion] Same RADIUS_EPSILON concern as in RangeSearch above: `search_param.radius` is still set to `request.radius_ * RADIUS_EPSILON` (1.1x). The result trimming via `range_search_radius` fixes the leakage, but the graph exploration radius remains wider than HGraph's approach where `search_param.radius = request.radius_` directly.

src/algorithm/pyramid/pyramid.h
thlurte: Reverted the whitespace change.

src/algorithm/pyramid/pyramid.cpp
thlurte: Thanks for adding a bit of clarity on this, your point is valid so i removed `RADIUS_EPSILON` and set `search_param.radius = radius` directly to align the graph exploration window with HGraph (`hgraph_search.cpp:593`) without visiting extra nodes.

src/algorithm/pyramid/pyramid.cpp
thlurte: Updated this here as well — set `search_param.radius = request.radius_` directly across `SearchWithRequest` to match `RangeSearch` and `HGraph`.

PR #2981, opened 2026-09-19
feat: add SINDI publish timestamp filtering

### LINKED ISSUE #2980: [feat](sindi): add publish timestamp filtering
### Affected module

module/index

### Affected area

not applicable

### Motivation

SINDI and SINDI_V2 need native publication-time filtering based on Unix
timestamps. Applications should be able to prune unrelated posting windows
before KNN candidate evaluation while combining time, host, tombstone, and
user filters with AND semantics.

### Proposed solution

Add named int64 metadata support to `Dataset` and use the following SINDI
metadata contract:

- Base documents provide Unix seconds through `publish_time_stamp`.
- `0` represents a document without a publication time.
- Queries use `publish_time_stamp` for one UTC day, or
  `publish_time_stamp_begin` and `publish_time_stamp_end` for an inclusive
  range.
- Internally, timestamps are normalized to int32 UTC epoch days.
- Time-ordered documents are divided into logical `window_size` partitions,
  with optional host grouping inside each partition.
- Missing-time documents form one final partition and are excluded from time
  queries.
- Standard and streaming serialization preserve time partitions, per-document
  days, and optional host directories.
- Time-aware mutable indexes remain build-once and reject later `Add()` calls.

The Dataset API is added through non-virtual forwarding methods so the existing
`Dataset` vtable is unchanged.

### Alternatives considered

A user callback filter can enforce time membership after candidates are read,
but cannot prune unrelated posting windows. Maintaining separate indexes for
time periods increases index-management and storage costs.

breaking change

The public C++ API addition preserves the `Dataset` vtable. However, the
existing string-path date-filter contract and serialized date payload are
replaced without backward-compatibility handling.

### References

- `include/vsag/dataset.h:415`
- `src/algorithm/sindi_metadata_filter.h:196`
- `src/algorithm/sindi_metadata_filter.cpp:760`
- `src/algorithm/sindi_metadata_filter.cpp:934`
- `docs/docs/en/src/indexes/sindi.md:178`

### Additional context

Range search, iterators, and incremental additions to an existing time-aware
index are outside this issue. Issue #2863 tracks broader deferred date-filter
workflows and should remain open.

## Related issues

- #2863 — [feat](sindi): complete deferred date-filtering workflows
- #2499 — [feat](sindi): add date-window filtering

---

Roxanne0321:
## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## Linked Issue

- Fixes: #2980

## What Changed

- Add vtable-preserving named int64 metadata APIs to `Dataset`, including ownership, deep-copy, append, and pointer-alias handling.
- Replace SINDI and SINDI_V2 string date buckets with Unix-second `publish_time_stamp` metadata normalized to int32 UTC epoch days.
- Add exact-day and inclusive-range routing, missing-time partitions, per-partition host directories, and precise boundary filtering.
- Persist the new time metadata through standard and streaming serialization, including reader-backed SINDI_V2 loading.
- Update English and Chinese Dataset, SINDI, SINDI_V2, and serialization documentation.

## Test Evidence

- [ ] `make fmt`
- [ ] `make lint`
- [ ] `make test`
- [x] Other (describe below)

Test details:

```text
cmake --build build --target unittests_algorithm unittests --parallel 6
  passed

./build/tests/unittests_algorithm '[time_filter]' --reporter compact
  6 test cases, 406 assertions passed

./build/tests/unittests_algorithm '[host_filter]' --reporter compact
  18 test cases, 428 assertions passed

./build/tests/unittests 'Dataset Named Int64 Metadata Test' --reporter compact
  1 test case, 14 assertions passed

clang-format 15 --dry-run --Werror <changed C++ files>
  passed

clang-tidy 15 -p build <changed production C++ files>
  passed

git diff --check
  passed

node .github/scripts/check-docs.mjs
  not run: Node.js is unavailable in the local environment; changed links and anchors were checked manually
```

## Compatibility Impact

- API/ABI compatibility: `Dataset::Int64Metadata` and `GetInt64Metadata` are additive non-virtual methods, preserving the existing `Dataset` vtable. The previous SINDI string date-filter contract and date payload are replaced without backward compatibility.
- Behavior changes: SINDI and SINDI_V2 KNN time filtering now accepts Unix seconds via `publish_time_stamp`; time-aware mutable indexes remain build-once. Range search and iterators remain unchanged.

- Performance impact: time queries prune unrelated logical partitions and store per-document time values as 4-byte epoch days; no end-to-end benchmark was collected for this PR.
- Concurrency/thread-safety impact: none; metadata is built or restored before concurrent search.

## Documentation Impact

- [ ] No docs update needed
- [x] Updated docs:
  - [ ] `README.md`
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [x] Other: `docs/docs/{en,zh}/src/api/dataset.md`, `indexes/{sindi,sindi_v2}.md`, and `advanced/new_serialization.md`

## Risk and Rollback

- Risk level: medium
- Rollback plan: revert commit `f186ec8f8020fc438d1f411693030a621923216b` to restore the previous date-filter implementation and serialization identifiers.

## Checklist


src/storage/serialization_tags.h
```diff
@@ -45,7 +45,7 @@ enum class StreamSerializationTag : uint32_t {
     PYRAMID_PATHS = 19,
     SINDI_HOST_METADATA = 20,
     SINDI_V2_TERM_LAYOUT = 21,
-    SINDI_DATE_METADATA = 22,
+    SINDI_TIME_METADATA = 22,
```
LHT129, line 48: The `SINDI_TIME_METADATA` tag reuses the same numeric value (22) as the removed `SINDI_DATE_METADATA`. This means any previously-serialized SINDI/SINDI_V2 index that contains a `SINDI_DATE_METADATA` block (tag 22) will be unreadable with this new code — the deserializer will try to interpret old date-metadata bytes as time-metadata bytes and fail validation.

If backward compatibility with existing serialized indexes is not required, this is fine. But if there are any persisted indexes in the wild using the old date metadata, consider either keeping `SINDI_DATE_METADATA = 22` and assigning a new tag (e.g. 23) for `SINDI_TIME_METADATA`, or documenting explicitly that this is a breaking format change.

src/dataset_impl.cpp
```diff
@@ -42,6 +42,24 @@ Dataset::Paths(const std::string& hierarchy_name, std::vector<std::vector<std::s
     return dataset->Paths(hierarchy_name, std::move(paths));
 }
 
+DatasetPtr
+Dataset::Int64Metadata(const std::string& name, const int64_t* values) {
```
LHT129, line 52: The `Dataset::Int64Metadata` and `Dataset::GetInt64Metadata` base-class implementations use `dynamic_cast<DatasetImpl*>(this)`. This creates a hidden coupling: any third-party `Dataset` subclass that does not inherit from `DatasetImpl` will either throw (`Int64Metadata`) or silently return `nullptr` (`GetInt64Metadata`).

Since these are documented as non-virtual to preserve vtable compatibility, the `dynamic_cast` pattern undermines that goal — a downstream user who subclasses `Dataset` directly (without going through `DatasetImpl`) gets broken behavior with no compile-time warning.

Consider making these pure virtual (and accepting the vtable change), or adding a protected virtual hook that `DatasetImpl` overrides, so the contract is explicit.

src/algorithm/sindi_metadata_filter.h
```diff
@@ -277,26 +273,26 @@ class SindiDateFilter {
...
 
-    // Exact date predicates are evaluated per document, so date queries scan every term in each
-    // routed window before applying the date filter.
+    // Exact time predicates are evaluated per document, so time queries scan every term in each
+    // routed window before applying the time filter.
     [[nodiscard]] static bool
-    RequiresFullTermScan(const SindiDateSearchRoute& route,
+    RequiresFullTermScan(const SindiTimeSearchRoute& route,
                          uint32_t window_id,
                          uint32_t window_size);
```
LHT129, line 297: `SindiTimeFilter::RequiresFullTermScan` unconditionally returns `true` when `route.has_time` is true, meaning every time-filtered query disables term-level posting pruning and scans every term in each selected window. This is a significant performance difference from host-only queries, which can skip windows entirely when the host range covers the whole window.

The doc comment on the declaration explains the rationale (exact day checks are per-document), but the performance impact is worth calling out more prominently in the public docs — a user switching from host-only to time filtering may see substantially higher latency.

Consider whether a fast-path is possible when a time partition's `[min_day, max_day]` is fully contained within `[query_begin, query_end]` (i.e., the per-document day check is a no-op for that partition).

src/algorithm/sindi_metadata_filter.h
```diff
@@ -277,26 +273,26 @@ class SindiDateFilter {
...
                        int64_t max_window_id);
 
-    // Exact date predicates are evaluated per document, so date queries scan every term in each
-    // routed window before applying the date filter.
+    // Exact time predicates are evaluated per document, so time queries scan every term in each
+    // routed window before applying the time filter.
     [[nodiscard]] static bool
-    RequiresFullTermScan(const SindiDateSearchRoute& route,
+    RequiresFullTermScan(const SindiTimeSearchRoute& route,
                          uint32_t window_id,
```
LHT129, line 295: `SindiTimeFilter::RequiresFullTermScan` is declared `static` but the equivalent `SindiHostFilter::RequiresFullTermScan` is a `const` member function. The inconsistency is minor but worth noting — if `RequiresFullTermScan` ever needs to access filter state (e.g., to check partition day ranges for the fast-path optimization), the `static` qualifier would need to be removed.

src/algorithm/sindi_metadata_filter.h
```diff
@@ -311,65 +307,68 @@ class SindiDateFilter {
...
     }
 
     [[nodiscard]] bool
     Enabled() const {
-        return host_plan_.Enabled() or date_plan_.Enabled();
+        return host_plan_.Enabled() or time_plan_.Enabled();
     }
 
     [[nodiscard]] uint32_t
     SourceIndex(uint32_t ordered_position) const {
```
LHT129, line 339: The `SindiMetadataBuildPlan::SourceIndex` method delegates to `time_plan_.SourceIndex` when `time_plan_.Enabled()` is true, and falls back to `host_plan_.SourceIndex` otherwise. This is correct but the dual-path dispatch is subtle — both build plans have their own `order_` vectors with different sort criteria (time-partition-then-host vs. host-only).

Consider adding a brief comment explaining that when time metadata is present, the time plan owns the canonical ordering and the host plan's `RecordSuccess` is still called (via the `SindiMetadataBuildPlan::RecordSuccess` forwarding) but its ordering is not used for source index lookup.

LHT129 (commented):
Thanks for this PR. This is a well-structured refactoring that replaces the old string-based date-bucket filtering with Unix-timestamp-based publish-time filtering. The code quality is high — good test coverage, thorough validation, and clean separation between the time and host filter paths.

**Summary of review comments:**

1. **Serialization tag reuse (tag 22)**: `SINDI_TIME_METADATA` reuses the same numeric tag as the removed `SINDI_DATE_METADATA`. This is a breaking format change — any previously-serialized index with date metadata will fail to deserialize. Worth documenting explicitly if this is intentional.

2. **`Dataset::Int64Metadata` base-class `dynamic_cast`**: The non-virtual base-class implementations use `dynamic_cast<DatasetImpl*>`, which silently breaks for any third-party `Dataset` subclass. Consider making these virtual or adding a protected hook.

3. **`RequiresFullTermScan` always true for time queries**: Every time-filtered query disables term-level posting pruning, which is a significant performance difference from host-only queries. The docs mention this but the impact could be more prominent. A fast-path when a partition is fully contained in the query range is worth considering.

4. **`static` vs `const` inconsistency**: `SindiTimeFilter::RequiresFullTermScan` is `static` while `SindiHostFilter::RequiresFullTermScan` is a `const` member — minor but worth aligning.

5. **Dual-path `SourceIndex` dispatch**: The delegation from `SindiMetadataBuildPlan::SourceIndex` to `time_plan_` vs `host_plan_` is correct but subtle. A brief comment would help future readers.

**What this PR does well:**
- Clean separation of time and host filter paths in `SindiMetadataFilter`
- Comprehensive validation in both `PrepareBuild` and `Deserialize` (range checks, ordering checks, coverage checks)
- Good test coverage: routing, serialization, validation, boundary conditions, combined time+host, partition tolerance for failed documents
- Proper handling of edge cases: missing timestamps (value 0), overflow, negative values, conflicting query selectors
- Documentation updated in both English and Chinese
- The `Int64Metadata` Dataset API follows the existing pattern for `UInt32Metadata` and `StringMetadata`

No blocking issues found. The comments above are suggestions for consideration.

LHT129 (commented):
Code review for PR #2981: feat: add SINDI publish timestamp filtering

## Summary

This PR replaces the old string-based hierarchical date bucket filtering (YYYY/MM/DD) with Unix timestamp-based publish time filtering using int64 metadata. The changes are thorough and well-structured across 19 files. The core logic is correct and the test coverage is updated appropriately.

## Issues noted

Two inline comments have already been raised by other reviewers:
1. **Serialization tag reuse** (tag 22): The `SINDI_TIME_METADATA` tag reuses the same numeric value as the removed `SINDI_DATE_METADATA`, making old serialized data unreadable.
2. **`dynamic_cast<DatasetImpl*>` pattern**: The non-virtual `Int64Metadata`/`GetInt64Metadata` methods use `dynamic_cast`, creating hidden coupling with `DatasetImpl`.

## Additional note

In `PrepareBuild`, the first sort uses `source_timestamps` (raw int64 seconds) as the tiebreaker for dated documents rather than `source_days_` (int32 epoch days). This means documents with different timestamps but the same UTC day can be split across adjacent partitions, causing `min_day`/`max_day` to overlap. While `Classify` handles this correctly, using `source_days_` for the tiebreaker would produce non-overlapping partition day ranges.

## Verdict

The PR is well-implemented. The two issues already raised are the main concerns to address. No critical bugs found.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +777,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
-        if (source_quarters[lhs] != source_quarters[rhs]) {
-            return source_quarters[lhs] < source_quarters[rhs];
+        const bool lhs_missing = plan.source_days_[lhs] == SINDI_MISSING_EPOCH_DAY;
+        const bool rhs_missing = plan.source_days_[rhs] == SINDI_MISSING_EPOCH_DAY;
+        if (lhs_missing != rhs_missing) {
+            return not lhs_missing;
         }
-        if (source_host_ids != nullptr and source_host_ids[lhs] != source_host_ids[rhs]) {
-            return source_host_ids[lhs] < source_host_ids[rhs];
+        if (not lhs_missing and source_timestamps[lhs] != source_timestamps[rhs]) {
```
LHT129, line 800: [note] In PrepareBuild, the first sort uses source_timestamps (raw int64 seconds) as the tiebreaker for dated documents, rather than source_days_ (int32 epoch days). Documents with different timestamps but the same UTC day can be split across adjacent partitions, causing min_day/max_day to overlap. While Classify handles this correctly, using source_days_ for the tiebreaker would produce non-overlapping partition day ranges.

src/storage/serialization_tags.h
Roxanne0321: Intentional for this PR. The agreed scope treats publish-time filtering as the first supported date format and explicitly does not read or migrate old string-date payloads. The PR Compatibility Impact calls out that the old date payload is not backward compatible, so tag 22 is deliberately redefined instead of allocating a compatibility tag. No code change. Final commit: `78364c92`; all checks on the prior code revision passed, and the amended comment-only revision is running CI now.

src/algorithm/sindi_metadata_filter.h
Roxanne0321: The public SINDI docs already state that exact publish-time filtering disables term-level posting pruning and can scan more postings. A partition-containment fast path is not safe as a local check: failed or skipped inserts can make adjacent logical partitions share one physical window, so candidates outside the selected day or inner ranges can coexist in that window and still require exact filtering. Keeping full scan for time predicates is intentional. No code change. Final commit: `78364c92`; all checks on the prior code revision passed, and the amended comment-only revision is running CI now.

src/algorithm/sindi_metadata_filter.h
Roxanne0321: Added a comment clarifying that the time-aware plan owns canonical timestamp/host ordering and that otherwise the standalone host plan owns it; the unused plan remains disabled and ignores success forwarding. Amended in `78364c92`. Validation: clang-format 15 dry-run and `git diff --check` passed; the new CI run is in progress.

src/dataset_impl.cpp
Roxanne0321: Preserving the existing `Dataset` vtable is an explicit requirement. Making the API pure virtual or adding a virtual hook would change that vtable and break existing implementations at the ABI level. The built-in implementation is supported via `DatasetImpl`; custom subclasses retain old behavior: the setter reports `UNSUPPORTED_INDEX_OPERATION`, while the getter returns `nullptr` so queries without the new metadata continue to work. No code change. Final commit: `78364c92`; all checks on the prior code revision passed, and the amended comment-only revision is running CI now.

src/algorithm/sindi_metadata_filter.h
Roxanne0321: Intentional: `SindiTimeFilter::RequiresFullTermScan` depends only on the self-contained search route, whereas the host implementation needs filter object state to inspect host ranges. Keeping the time helper static makes that lack of state dependency explicit. No code change. Final commit: `78364c92`; all checks on the prior code revision passed, and the amended comment-only revision is running CI now.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Intentional per the publish-time contract: documents are globally ordered by the original Unix timestamp before window-sized logical partitioning. Adjacent partitions may therefore share the same UTC day; the design requires non-strict range monotonicity and `Classify` selects all intersecting partitions. A focused test (`SINDI adjacent time partitions can share a UTC day`) covers this. Changing to day ordering would discard the confirmed second-level ordering. No code change. Final commit: `78364c92`; all checks on the prior code revision passed, and the amended comment-only revision is running CI now.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -1341,30 +1259,36 @@ void
...
 }
 
 void
 SindiMetadataFilter::Clear() {
     host_filter_.Clear();
-    date_filter_.Clear();
+    time_filter_.Clear();
 }
 
 SindiMetadataSearchRoute
```
LHT129, line 1280: [note] `TimeRouteFilter` stores a raw pointer `inner_ranges_` that borrows from `SindiTimeSearchRoute::inner_ranges`. The struct-level comment on `SindiTimeSearchRoute` documents this lifetime dependency, but the design is fragile: if a `TimeRouteFilter` instance is ever copied, moved, or stored beyond the scope of the owning `SindiTimeSearchRoute`, the pointer dangles and `CheckValid` / `contains_value` will read freed memory. Consider copying the ranges into the filter (the vector is typically small — bounded by the number of time partitions) or using `std::shared_ptr` to make the ownership explicit.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +777,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
-        if (date_buckets[i].empty()) {
-            plan.source_buckets_[i] = MISSING_DATE_BUCKET;
-            source_quarters[i] = MISSING_DATE_QUARTER;
-        } else {
-            plan.source_buckets_[i] = parse_date_bucket(date_buckets[i]);
-            source_quarters[i] = date_bucket_to_quarter(plan.source_buckets_[i]);
+        const auto timestamp = source_timestamps[i];
+        CHECK_ARGUMENT(timestamp >= 0, "SINDI publish_time_stamp base values must be non-negative");
+        if (timestamp == 0) {
+            plan.source_days_[i] = SINDI_MISSING_EPOCH_DAY;
```
LHT129, line 785: [note] `timestamp == 0` is treated as a sentinel meaning "no publish timestamp" (`SINDI_MISSING_EPOCH_DAY`). This means a genuine Unix epoch timestamp of exactly 1970-01-01 00:00:00 UTC is indistinguishable from a missing timestamp. For publish-time data this is unlikely to matter in practice, but it is a constraint worth documenting in the public API (e.g. the `Int64Metadata("publish_time_stamp", ...)` docstring) so callers know that zero is reserved.

LHT129 (commented):
Code review for PR #2981 (supplementary to LHT129's existing review).

LHT129 has already covered the major design points: tag 22 reuse, dynamic_cast coupling, RequiresFullTermScan performance, raw pointer lifetime in TimeRouteFilter, timestamp==0 sentinel semantics, and the sort ordering choice. The author has addressed most of these with clarifications in commit 78364c92. I will focus on remaining observations.

**1. `SindiTimeFilter::ApplyWindowRoute` — implicit non-empty invariant on `window_ranges`**

At `src/algorithm/sindi_metadata_filter.cpp:1050`, `route.window_ranges.front().first` and `route.window_ranges.back().second` are accessed without an explicit empty check. This is safe because `Classify` only sets `kind = WINDOW` after populating `window_ranges` from non-empty `inner_ranges`. However, the invariant is non-local — a future refactor of `Classify` could break it silently. Consider adding `CHECK_ARGUMENT(not route.window_ranges.empty(), ...)` or a comment at the call site.

**2. `SindiMetadataBuildPlan::RecordSuccess` forwards to both plans unconditionally**

At `src/algorithm/sindi_metadata_filter.h:348-351`, `RecordSuccess` calls both `host_plan_.RecordSuccess(ordered_position)` and `time_plan_.RecordSuccess(ordered_position)`. This is correct because only one plan is enabled at a time (the disabled plan's `RecordSuccess` is a no-op). But it means both plans accumulate `successful_counts_` and `successful_days_` vectors even though only one will be used. This is negligible overhead but worth noting for future readers.

**3. `SindiTimeFilter::PrepareBuild` — `host_dictionary_.Encode` called on member during const method**

At `src/algorithm/sindi_metadata_filter.cpp:775`, `PrepareBuild` (a const method) calls `host_dictionary_.Encode(...)`. This is safe because `Encode` is itself const and writes to output parameters (`plan.source_host_ids_`, `plan.new_hosts_`). But it is subtle — a reader might expect a const method not to interact with mutable dictionary state. The `Encode` method signature makes this clear, but a brief comment would help.

**4. Test coverage is comprehensive**

The test file covers: time+host combined routing, exact-day and range queries, validation of query/base values (negative, zero, overflow), partition tolerance at boundaries with failed documents, adjacent partitions sharing a UTC day, serialization round-trip validation, streaming serialization with block erasure, and incremental-Add rejection for time-aware indexes. This is thorough.

**5. Documentation**

The 8 documentation files (en/zh) have been updated from date-bucket to publish-time terminology. The changes are consistent and complete.

**Overall**: The implementation is solid. The design tradeoffs (non-virtual Int64Metadata, tag 22 reuse, RequiresFullTermScan always-true for time queries) are explicitly documented and intentional. No blocking issues found beyond what LHT129 has already raised and the author has addressed.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Documented in `207b83d2`. The public `Dataset::Int64Metadata` Doxygen note and both English and Chinese Dataset API pages now state that base `publish_time_stamp == 0` is reserved for a missing timestamp and never matches a time selector. Validation: clang-format 15 and `git diff --check` passed; the local Node.js docs checker was unavailable, so the new remote Docs Check will verify the documentation set.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Fixed in `207b83d2`. `TimeRouteFilter` now owns a copy of `inner_ranges` instead of borrowing a raw pointer from the search route, so copying, moving, or retaining the filter cannot leave that range state dangling. The regression assertion clears the source route ranges after filter creation and verifies that filtering still uses the captured ranges. Validation: clang-format 15 and `git diff --check` passed; `[time_filter]` passed 6 test cases and 407 assertions.

Roxanne0321:
Follow-up review handled in `207b83d2`:

1. Added an explicit non-empty `window_ranges` check before `ApplyWindowRoute` accesses `front()` / `back()`, with a regression assertion for malformed WINDOW routes.
2. Kept the dual `RecordSuccess` forwarding unchanged. Both build-plan implementations immediately return when `enabled_ == false`, so the disabled plan does not accumulate counts or days; only the active plan mutates.
3. Kept `host_dictionary_.Encode` unchanged. It is a `const` method that reads the committed lookup and writes only the plan-owned output arguments; dictionary mutation remains isolated to `Commit`.

The two new inline notes were also addressed: `TimeRouteFilter` now owns its inner ranges, and the reserved zero publish timestamp is documented in the public Dataset API. Local validation: clang-format 15, `git diff --check`, and `[time_filter]` (6 cases, 407 assertions) passed.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +776,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
+            continue;
         }
+        const auto day = timestamp / SINDI_SECONDS_PER_DAY;
+        CHECK_ARGUMENT(day <= std::numeric_limits<int32_t>::max(),
+                       "SINDI publish_time_stamp exceeds the supported epoch day range");
+        plan.source_days_[i] = static_cast<int32_t>(day);
     }
 
     std::sort(plan.order_.begin(), plan.order_.end(), [&](uint32_t lhs, uint32_t rhs) {
```
LHT129, line 794: [suggestion] The sort lambda at line 794 captures `source_timestamps` (a raw pointer obtained from `base->GetInt64Metadata()` at line 760) by reference via `[&]`. While currently safe because the plan is consumed synchronously before `PrepareBuild` returns, this creates an implicit lifetime dependency between the lambda and the Dataset argument. If the build plan were ever stored or the sort were deferred (e.g., for async build pipelines), the captured pointer would dangle. Consider capturing `source_timestamps` by value instead — `[=]` or explicit `[source_timestamps, &plan]` would make the dependency explicit and self-documenting.

src/algorithm/sindi_metadata_filter.h
```diff
@@ -311,65 +306,70 @@ class SindiDateFilter {
...
+        // host plan owns the ordering; the unused plan remains disabled and ignores successes.
+        return time_plan_.Enabled() ? time_plan_.SourceIndex(ordered_position)
                                     : host_plan_.SourceIndex(ordered_position);
     }
 
     void
     RecordSuccess(uint32_t ordered_position) {
         host_plan_.RecordSuccess(ordered_position);
-        date_plan_.RecordSuccess(ordered_position);
+        time_plan_.RecordSuccess(ordered_position);
```
LHT129, line 350: [note] `SindiMetadataBuildPlan::RecordSuccess` unconditionally forwards to both `host_plan_` and `time_plan_`. When `time_plan_.Enabled()` is true, the host plan is disabled and its `RecordSuccess` is a no-op (returns immediately). When `time_plan_.Enabled()` is false, the time plan is disabled and its `RecordSuccess` is also a no-op. This is correct, but the dual-dispatch is subtle — a reader might wonder whether calling both plans is intentional or a bug. A brief comment noting that each plan guards itself via `enabled_` would clarify the intent.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +776,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
+        if (partition_count > 0) {
             partitions_.emplace_back(allocator_);
             auto& partition = partitions_.back();
-            partition.quarter = plan.group_quarters_[group_begin];
             partition.begin = inner_cursor;
-            partition.end = inner_cursor + quarter_count;
+            partition.end = inner_cursor + partition_count;
+            const auto day_range = std::minmax_element(document_days_.begin() + partition.begin,
+                                                       document_days_.begin() + partition.end);
+            partition.min_day = *day_range.first;
```
LHT129, line 888: [note] The `day_range` computation uses `std::minmax_element` on `document_days_` within the partition bounds. This requires at least one element in the range, which is guaranteed by the `partition_count > 0` guard above. However, the `minmax_element` result is used directly without a null check — if `partition_count` were ever 0 (e.g., due to a future refactoring that removes the guard), this would be undefined behavior. The existing `partition_count > 0` check makes this safe, but the dependency between the check and the `minmax_element` call is not locally obvious within the block.

LHT129 (commented):
## Code Review Summary

I've reviewed all 19 changed files (C++ source, headers, tests, and documentation). The existing 14 review comments have been addressed. Below are additional observations not covered by prior comments.

### Review Dimensions

- **Correctness**: The epoch-day partitioning logic, timestamp-to-day conversion, and partition selection in `Classify` are sound. The missing-time sentinel (`timestamp == 0`, `SINDI_MISSING_EPOCH_DAY = -1`) is handled consistently across build, classify, serialize, and deserialize paths.
- **Style**: Consistent with the existing codebase conventions. The `append_merged_range` and `contains_value` helpers are clean abstractions.
- **Test Coverage**: The test file covers exact timestamp match, range queries, combined time+host filtering, serialization roundtrip, streaming serialization, missing-time error handling, and adjacent partition day overlap. Good coverage of the primary paths.
- **Performance**: `RequiresFullTermScan` returning true for time queries is a deliberate tradeoff (per-document day checks disable term-level pruning), already discussed in existing comments.
- **Security**: No security concerns identified.
- **Maintainability**: The rename from "date" to "time" is thorough across all files. The `SindiMetadataBuildPlan` delegation pattern (time_plan_ vs host_plan_) is clean.

### Inline Comments

See inline comments below for specific suggestions.

LHT129 (commented):
## Code Review Summary

I reviewed all 19 changed files in this PR, which replaces SINDI/SINDI_V2 string-based date bucket filtering with Unix timestamp-based publish time filtering using int64 metadata. The implementation is thorough and well-structured.

### What this PR does well
- Clean rename from "date" to "time" throughout the codebase (filters, plans, routes, serialization)
- Proper ownership fix: `TimeRouteFilter` now owns `inner_ranges_` by value instead of borrowing a raw pointer
- Clear sentinel design: `timestamp == 0` → missing, `SINDI_MISSING_EPOCH_DAY = -1`
- Good validation in `Deserialize`: partition ordering, day range consistency, missing partition isolation
- Helpful error message in `SindiMetadataFilter::Classify` when time queries are used on non-time indexes
- Documentation updated in both English and Chinese

### Issues already raised and addressed
17 inline comments cover the key concerns: tag reuse (intentional), `dynamic_cast` pattern (intentional for vtable), `RequiresFullTermScan` always true for time queries, sort tiebreaker using raw timestamps, `inner_ranges_` dangling pointer (fixed), `timestamp == 0` sentinel documentation (added), and several others.

### New observations
No critical issues found. The existing comments cover the significant design decisions and trade-offs. The PR is ready for merge from a correctness standpoint.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: No code change. Capturing `source_timestamps` by value would only copy the pointer; it would not extend the Dataset storage lifetime. The comparator is a local temporary consumed synchronously by `std::sort` inside `PrepareBuild` and is never stored in the build plan. If sorting is ever deferred, that change will need an ownership redesign regardless of the lambda capture syntax.

src/algorithm/sindi_metadata_filter.h
Roxanne0321: Clarified in `09997766`. `RecordSuccess` now states that exactly one plan is enabled and the disabled plan ignores the forwarded call. Validation: clang-format 15 dry-run and `git diff --check` passed.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: No code change. The `minmax_element` call is lexically inside `if (partition_count > 0)`, and its end iterator is computed as `begin + partition_count`, so the non-empty-range invariant is local and explicit. The returned values are iterators rather than nullable pointers; adding a duplicate guard would not improve safety while the existing guard is present.

Roxanne0321: No code change. The `minmax_element` call is lexically inside `if (partition_count > 0)`, and its end iterator is computed as `begin + partition_count`, so the non-empty-range invariant is local and explicit. The returned values are iterators rather than nullable pointers; adding a duplicate guard would not improve safety while the existing guard is present.

LHT129 (commented):
Review complete. All substantive issues identified during this review cycle have already been captured by the 21 existing inline comments on commit 09997766, which cover:

- **Serialization compatibility**: `SINDI_TIME_METADATA` tag reuse (tag 22) — intentional, documented as breaking change
- **Dataset vtable ABI**: `dynamic_cast<DatasetImpl*>` pattern — intentional, preserves vtable compatibility
- **Performance**: `RequiresFullTermScan` always returns true for time-filtered queries — noted, with suggestion for future fast-path optimization
- **API consistency**: `static` vs member function inconsistency between `SindiTimeFilter` and `SindiHostFilter` — noted
- **Build plan delegation**: `SourceIndex` dual-path dispatch — clarified with comments
- **Sort tiebreaker**: timestamp vs epoch-day ordering — intentional, documented with test coverage
- **Memory safety**: `TimeRouteFilter` raw pointer lifetime — fixed (now owns a copy of `inner_ranges_`)
- **API contract**: `timestamp == 0` sentinel — documented in public API and docs
- **Lambda capture**: `source_timestamps` captured by reference — no change needed (synchronous consumption)
- **Dual dispatch**: `RecordSuccess` forwarding to both plans — clarified with comments
- **Iterator safety**: `minmax_element` on non-empty range — no change needed (lexically guarded)

The implementation correctly replaces string-based date filtering with Unix timestamp-based publish time filtering across SINDI and SINDI_V2 indexes. The core logic — timestamp-to-epoch-day conversion, time-ordered partitioning, UTC-day range intersection during Classify, and per-document exact-day filtering via TimeRouteFilter — is sound. Serialization, deserialization, streaming, and legacy format paths are all consistently updated. Test coverage includes exact-day matching, range queries, combined host+time queries, serialization round-trips, streaming serialization, and edge cases (missing timestamps, boundary conditions, Add rejection for time-aware indexes).

No new substantive issues to report.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -1013,62 +908,65 @@ SindiDateFilter::CommitBuild(SindiDateBuildPlan&& plan, uint64_t element_count)
...
+    const auto* query_begin = query->GetInt64Metadata(SINDI_PUBLISH_TIME_BEGIN_METADATA_NAME);
+    const auto* query_end = query->GetInt64Metadata(SINDI_PUBLISH_TIME_END_METADATA_NAME);
+    CHECK_ARGUMENT((query_begin == nullptr) == (query_end == nullptr),
+                   "SINDI publish_time_stamp_begin and publish_time_stamp_end must be provided "
+                   "together");
+    CHECK_ARGUMENT(  // NOLINT(readability-simplify-boolean-expr)
+        not(query_time != nullptr and query_begin != nullptr),
+        "SINDI publish_time_stamp cannot be combined with a time range");
+    const auto to_day = [](int64_t timestamp) {
+        CHECK_ARGUMENT(timestamp > 0, "SINDI query timestamps must be greater than zero");
```
LHT129, line 954: [note] The `to_day` lambda used for query timestamps rejects `timestamp == 0` with `CHECK_ARGUMENT(timestamp > 0, ...)`, while base data validation at line 782 uses `CHECK_ARGUMENT(timestamp >= 0, ...)` to accept `0` as the missing-timestamp sentinel. The asymmetry is intentional (base `0` means "missing", query `0` is invalid), and the Dataset API docstring documents that `publish_time_stamp == 0` is reserved. This is correct but worth keeping in mind if the sentinel value ever changes.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -1166,16 +1076,16 @@ SindiDateFilter::NextMatchingWindow(const SindiDateSearchRoute& route,
...
 bool
-SindiDateFilter::RequiresFullTermScan(const SindiDateSearchRoute& route,
+SindiTimeFilter::RequiresFullTermScan(const SindiTimeSearchRoute& route,
                                       uint32_t window_id,
                                       uint32_t window_size) {
     if (not route.enabled or route.kind != SindiHostRouteKind::WINDOW) {
         return false;
     }
-    if (route.has_date_bucket or route.has_date_range) {
+    if (route.has_time) {
```
LHT129, line 1085: [note] `SindiTimeFilter::RequiresFullTermScan` unconditionally returns `true` when `route.has_time` is true, meaning every time-filtered query disables term-level posting pruning and scans every term in each selected window. This is a significant performance difference from host-only queries, which can skip terms whose postings do not intersect the window. The comment in the header explains that "exact time predicates are evaluated per document, so time queries scan every term in each routed window before applying the time filter." This is a known design tradeoff — the per-document `TimeRouteFilter` cannot be pushed down to the term level because document days are stored per document, not per posting.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +776,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
-        if (source_quarters[lhs] != source_quarters[rhs]) {
-            return source_quarters[lhs] < source_quarters[rhs];
+        const bool lhs_missing = plan.source_days_[lhs] == SINDI_MISSING_EPOCH_DAY;
+        const bool rhs_missing = plan.source_days_[rhs] == SINDI_MISSING_EPOCH_DAY;
+        if (lhs_missing != rhs_missing) {
+            return not lhs_missing;
         }
-        if (source_host_ids != nullptr and source_host_ids[lhs] != source_host_ids[rhs]) {
-            return source_host_ids[lhs] < source_host_ids[rhs];
+        if (not lhs_missing and source_timestamps[lhs] != source_timestamps[rhs]) {
```
LHT129, line 800: [note] The first sort in `PrepareBuild` uses `source_timestamps` (raw int64 Unix seconds) as the tiebreaker for dated documents rather than `source_days_` (int32 epoch days). Documents with different timestamps but the same UTC day can be assigned to adjacent partitions, causing `min_day`/`max_day` ranges of adjacent partitions to overlap on the same day. The `Classify` method handles this correctly by selecting all partitions whose `[min_day, max_day]` intersects the query range (via `lower_bound` on `max_day` and early break on `min_day > query_end`), and the test `"SINDI adjacent time partitions can share a UTC day"` validates this behavior. This is an intentional design choice that prioritizes stable timestamp ordering over strict day-boundary partitioning.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -1119,31 +1027,33 @@ SindiDateFilter::Classify(const DatasetPtr& query, uint32_t window_size) const {
...
 
 void
-SindiDateFilter::ApplyWindowRoute(const SindiDateSearchRoute& route,
+SindiTimeFilter::ApplyWindowRoute(const SindiTimeSearchRoute& route,
                                   int64_t& min_window_id,
                                   int64_t& max_window_id) {
     if (not route.enabled or route.kind != SindiHostRouteKind::WINDOW) {
         return;
     }
+    CHECK_ARGUMENT(not route.window_ranges.empty(),
```
LHT129, line 1049: [note] The new `CHECK_ARGUMENT(not route.window_ranges.empty(), ...)` guard in `ApplyWindowRoute` is a good defensive addition. Without it, calling `front()` or `back()` on an empty `window_ranges` vector would be undefined behavior. This can happen if a time query matches no partitions — in that case `Classify` returns a route with `kind == EMPTY` and `ApplyWindowRoute` returns early, but the explicit check adds an extra layer of safety for any future code path that might reach this point with an enabled but empty route.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -1341,30 +1260,36 @@ void
...
-        route.kind = route.date_route.kind;
+    if (time_filter_.HasMetadata()) {
+        route.time_route = time_filter_.Classify(query, window_size);
+        route.kind = route.time_route.kind;
     } else {
+        const bool has_time_query =
+            query->GetInt64Metadata(SINDI_PUBLISH_TIME_METADATA_NAME) != nullptr or
+            query->GetInt64Metadata(SINDI_PUBLISH_TIME_BEGIN_METADATA_NAME) != nullptr or
+            query->GetInt64Metadata(SINDI_PUBLISH_TIME_END_METADATA_NAME) != nullptr;
+        CHECK_ARGUMENT(not has_time_query,
```
LHT129, line 1291: [note] The `CHECK_ARGUMENT(not has_time_query, ...)` guard in `SindiMetadataFilter::Classify` is a good usability improvement — it catches the case where a user sends a time-filtered query to an index built without `publish_time_stamp`, producing a clear error message instead of silently ignoring the time filter. The check covers all three query metadata names (`publish_time_stamp`, `publish_time_stamp_begin`, `publish_time_stamp_end`).

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -891,29 +742,30 @@ SindiDateBuildPlan::RecordSuccess(uint32_t ordered_position) {
...
+    SindiTimeBuildPlan plan(allocator_);
     CHECK_ARGUMENT(base->GetUInt32Metadata(SINDI_LEGACY_HOST_METADATA_NAME) == nullptr,
                    "numeric SINDI host_id metadata is unsupported; use string metadata host");
-    const auto* date_buckets = base->GetPaths(SINDI_DATE_PATH_NAME);
-    if (date_buckets == nullptr) {
+    const auto* source_timestamps = base->GetInt64Metadata(SINDI_PUBLISH_TIME_METADATA_NAME);
+    if (source_timestamps == nullptr) {
         return plan;
     }
+    CHECK_ARGUMENT(window_size > 0, "SINDI window_size must be greater than zero");
```
LHT129, line 764: [note] The `CHECK_ARGUMENT(window_size > 0, ...)` guard was added to both `PrepareBuild` and `Classify`. This is a good defensive check since a zero `window_size` would cause division by zero in `(dated_count - 1) / window_size + 1` and other arithmetic issues. The check is placed early, before any expensive work.

src/algorithm/sindi_metadata_filter.h
```diff
@@ -277,26 +272,26 @@ class SindiDateFilter {
...
                        int64_t current_window_id,
                        int64_t max_window_id);
 
-    // Exact date predicates are evaluated per document, so date queries scan every term in each
-    // routed window before applying the date filter.
+    // Exact time predicates are evaluated per document, so time queries scan every term in each
+    // routed window before applying the time filter.
     [[nodiscard]] static bool
-    RequiresFullTermScan(const SindiDateSearchRoute& route,
+    RequiresFullTermScan(const SindiTimeSearchRoute& route,
```
LHT129, line 294: [note] `SindiTimeFilter::RequiresFullTermScan` is declared `static` (line 294 in the header, line 1079 in the .cpp), while `SindiHostFilter::RequiresFullTermScan` is a `const` member function. The inconsistency is minor — the time variant does not access any member state (it only reads from the route parameter), so `static` is technically correct. However, if `RequiresFullTermScan` ever needs to access filter state (e.g., to check partition day ranges for the window), the `static` qualifier would need to change. Consider making it a `const` member function for consistency with `SindiHostFilter`.

src/algorithm/sindi_metadata_filter.cpp
```diff
@@ -924,77 +776,120 @@ SindiDateFilter::PrepareBuild(const DatasetPtr& base) const {
...
         }
-        if (quarter_count > 0) {
+        if (partition_count > 0) {
             partitions_.emplace_back(allocator_);
             auto& partition = partitions_.back();
-            partition.quarter = plan.group_quarters_[group_begin];
             partition.begin = inner_cursor;
-            partition.end = inner_cursor + quarter_count;
+            partition.end = inner_cursor + partition_count;
+            const auto day_range = std::minmax_element(document_days_.begin() + partition.begin,
```
LHT129, line 886: [note] In `CommitBuild`, `std::minmax_element` is called on `document_days_.begin() + partition.begin` to `document_days_.begin() + partition.end`. This is inside `if (partition_count > 0)` which guarantees the range is non-empty (since `partition_count` is the sum of `successful_counts_` for groups in this partition, and `partition.end - partition.begin == partition_count`). The returned iterators are dereferenced directly without a null/empty check. This is safe given the invariant, but the relationship between `partition_count > 0` and the range being non-empty is indirect — a reader must trace back through `successful_counts_` to verify. A local assertion like `assert(partition.end > partition.begin)` before the `minmax_element` call would make the invariant explicit at the point of use.

LHT129 (commented):
## Review Summary

This PR replaces SINDI's string-based date filtering with Unix timestamp-based `publish_time_stamp` filtering using new `Int64Metadata` Dataset APIs. The change is substantial (19 files, +1065/-1047) and well-structured.

### What was reviewed
- Core time filter logic (`sindi_metadata_filter.cpp/.h`)
- Dataset API additions (`dataset.h`, `dataset_impl.cpp/.h`)
- SINDI/SINDI_V2 integration (`sindi.cpp`, `sindi_v2.cpp`)
- Serialization tag changes (`serialization_tags.h`)
- Unit tests (`sindi_metadata_filter_test.cpp`, `dataset_impl_test.cpp`)
- Documentation (8 doc files)

### Key observations
1. **Backward compatibility**: The `SINDI_TIME_METADATA` tag reuses value 22 from the removed `SINDI_DATE_METADATA`. Old date-payload indices are intentionally not readable — this is documented in the PR description as a deliberate scope decision.
2. **Performance**: `RequiresFullTermScan` returns `true` for all time queries, disabling term-level posting pruning. This is a known tradeoff since per-document time filtering cannot be pushed to the term level.
3. **Defensive checks**: Good additions of `window_size > 0` guards, `window_ranges.empty()` check in `ApplyWindowRoute`, and `has_time_query` validation for host-only indices.
4. **Test coverage**: Comprehensive tests cover exact day matching, range matching, time+host combined filtering, serialization roundtrip, streaming serialization, edge cases (missing timestamps, invalid ranges, conflicting queries), and the "adjacent partitions sharing a UTC day" scenario.

### Issues found
All inline comments are `[note]` level — no blocking issues. The existing review comments from previous rounds have been addressed with code changes, documentation, or explicit design decisions.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no additional change needed. Both build and query paths validate `window_size` before performing partition arithmetic.

src/algorithm/sindi_metadata_filter.h
Roxanne0321: No code change. The distinction is intentional: the time helper depends only on the self-contained route, so `static` makes its lack of object-state dependency explicit; the host implementation needs filter state. A future implementation that needs time-filter state can change the signature then.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no code change. The asymmetry is intentional: base value `0` is the documented missing-value sentinel, while query selectors must identify an actual positive Unix timestamp. The public Dataset documentation now records this contract.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no additional change needed. The guard intentionally rejects all three time selector forms on an index built without publish-time metadata.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: No code change. The invariant is local: inside `if (partition_count > 0)`, the code assigns `partition.begin = inner_cursor` and immediately assigns `partition.end = inner_cursor + partition_count` before calling `minmax_element`. Therefore `partition.end > partition.begin` follows directly, and a duplicate assertion would not add a separate safety guarantee.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no additional change needed. The explicit non-empty check was added defensively even though normal classification returns `EMPTY` before this path when no partition matches.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no code change. Raw timestamp ordering is intentional, and overlapping UTC-day bounds are handled by selecting every intersecting partition. `SINDI adjacent time partitions can share a UTC day` provides focused regression coverage.

src/algorithm/sindi_metadata_filter.cpp
Roxanne0321: Confirmed; no code change. This is the known time-query tradeoff already documented on `RequiresFullTermScan`: exact day predicates are evaluated per document after scanning the routed window, so term-level pruning cannot safely replace that filtering.

wxyucs (approved):
lgtm

PR #2983, opened 2026-09-20
fix(hgraph): prevent stall in low-selectivity filtered KNN (#2974)

### LINKED ISSUE #2974: [bug](hgraph): low-selectivity filtered KNN can stall a query for minutes at high ef_search
### Affected module

module/index

### Affected area

not applicable

### Interface

cpp

### VSAG version

main @ `7dc90ad5` (`v1.0.0-182-g7dc90ad5`), see "Additional context" for a caveat about unrelated local changes.

### Summary

With `ef_search=1000` and a highly selective filter (`Filter::ValidRatio() ≈ 0.02`), a single KNN query in the plain HGraph path can keep one search thread at 100% CPU for more than 12 minutes and never return. The same workload at `ef_search <= 600` finishes in seconds, and the same command has also completed at `ef_search=1000` (302-308 QPS), so the stall is intermittent rather than deterministic.

### Minimal reproducer

I could not reduce this to a small dataset yet; this is the workload on which it appears.

```
index: hgraph, dtype float32, metric cosine, dim 1536, 5,000,000 base vectors
  base_quantization_type=rabitq, precise_quantization_type=sq8,
  store_raw_vector=false, rabitq_one_bit_search=true, use_reorder=true,
  use_mci=true, mci_mcs=200, mci_clique_max=50, mci_alpha=1.2,
  max_degree=48, ef_construction=600

search parameters:
  {"hgraph":{"ef_search":1000,"parallelism":1,"rabitq_one_bit_search":true,
             "brute_force_threshold":0.0,"use_mci":true,"mci_seed_ratio":5.0,
             "hgraph_valid_ratio_threshold":0.02}}

filter: AND over per-label bitmaps of the base set; per-query ValidRatio() ≈ 0.02
workload: 2000 queries, 64 search threads, topk=100, no timeout_ms
```

The affected query is a plain `Index::KnnSearch(query, 100, parameters, filter)`. MCI is enabled, but the stalling query takes the `search_one_graph` fallback (no MCI frame is present in the backtrace), which is expected for queries whose `ValidRatio()` is at or above `hgraph_valid_ratio_threshold`.

### Environment

```shell
- OS: Ubuntu 26.04
- vsag version: v1.0.0-182-g7dc90ad5
```

AMD EPYC 9T24 (192 cores), 1.1 TB RAM, dataset held in a block-memory index file.

### Expected behavior

A filtered KNN search either honours a bound (`hops_limit`, `timeout_ms`) or terminates promptly. One query should not occupy a search thread for minutes.

### Actual behavior

Two of four attempts at `ef_search=1000` stalled. `gdb -p <pid>` on a stalled run (after ~20 minutes) shows 127 of 128 threads parked in the thread pool and exactly one thread running:

```
#0 vsag::(anonymous namespace)::AccumulativeFilterSearchSkipStrategy::ShouldVisit()
        at src/utils/filter_search_skip_strategy.cpp:54
#1 vsag::BasicSearcher::visit(...)          at src/impl/searcher/basic_searcher.cpp:67
#2 vsag::BasicSearcher::search_impl<KNN_SEARCH>(...)
        at src/impl/searcher/basic_searcher.cpp:771
#3 vsag::BasicSearcher::Search(...)         at src/impl/searcher/basic_searcher.cpp:86
#4 vsag::HGraph::search_one_graph<KNN_SEARCH>(...)
        at src/algorithm/hgraph/hgraph_search.cpp:328
#5 vsag::HGraph::SearchWithRequest(...)     at src/algorithm/hgraph/hgraph_search.cpp:784
#6 vsag::HGraph::KnnSearch(...)             at src/algorithm/hgraph/hgraph_search.cpp:62
```

The busy thread's CPU time kept increasing with no result, so the run was killed.

Code paths that look relevant:

- The search loop is `while (not candidate_set->Empty())` (`basic_searcher.cpp:735`) whose only explicit bound is `if (hops >= inner_search_param.hops_limit) break;`. `apply_hops_limit()` (`hgraph_search.cpp:62-74`) replaces the limit with `UINT32_MAX` whenever the configured `hops_limit <= ef_search`, and `HGraphSearchParameters::hops_limit` defaults to `UINT32_MAX`, so with default parameters nothing bounds the traversal except the candidate queue draining.
- The early exit is `if ((-current_node_pair.first) > lower_bound and top_candidates->Size() >= ef) break;`, i.e. it requires `ef` results that the filter accepts. `candidate_set` itself is unbounded (`StandardHeap<true, false>(alloc, -1)`).
- `visit()` marks every neighbour as visited but only queues those that satisfy `skip_strategy->ShouldVisit() || filter->CheckValid(neighbor)`, so each iteration does make progress in the visited list. A minutes-long stall therefore looks like a very large (possibly near-full-graph) traversal of 5M nodes with 48 neighbours each, rather than a zero-progress spin — but a single query taking 3600x the average latency (`ef_search=1000` averages ~170 ms/query when it completes) still seems wrong.
- A `ValidRatio() ≈ 0.02` filter also means the accepted neighbours are sparse, so I would like to know whether `skip_strategy` can end up declining essentially every neighbour for a long stretch, and whether that is expected to be bounded.

I tried to capture the runaway state by aborting once `hops` reached 2,000,000 and printing `top_candidates->Size()`, `lower_bound` and `is_finite(lower_bound)`. The stall did not reproduce in that run (it completed with recall 0.98641 at 302 QPS), so the hop count of a stalling query is still unknown.

### References

- `src/impl/searcher/basic_searcher.cpp:65-73` (`visit`), `:735-780` (search loop)
- `src/utils/filter_search_skip_strategy.cpp:47-60` (`AccumulativeFilterSearchSkipStrategy`)
- `src/algorithm/hgraph/hgraph_search.cpp:62-74` (`apply_hops_limit`), `:310`, `:385`, `:1021-1066` (`search_one_graph` call sites)

### Additional context

- Frequency: 2 of 4 attempts stalled for >12 minutes and were killed (two independent runs, same command); the other two completed at `ef_search=1000` with 302 and 308 QPS. At `ef_search <= 600` the same command never stalled across many runs.
- The stall was observed on a local build that also carries unrelated work-in-progress changes in `src/algorithm/hgraph/hgraph_search.cpp` (fused RaBitQ split work). Those changes do not touch `skip_ratio`, `skip_strategy_type`, `search_one_graph`, `try_mci_search` or the brute-force routing, and every frame in the backtrace above lives in a file that those changes leave untouched (`git status` is clean for `basic_searcher.cpp` and `filter_search_skip_strategy.cpp`). I have not yet reproduced it on a clean `origin/main` build, so please treat the root cause as unconfirmed.
- Questions:
  1. Is the plain search path expected to bound its traversal when the filter is highly selective (e.g. by `valid_count`, or by capping `candidate_set`)?
  2. Should `search_impl` assert that the loop is making progress, so that this class of stall fails loudly instead of hanging a thread?
  3. Should the effective default `hops_limit = UINT32_MAX` be reconsidered for filtered searches?

thlurte:
### What problem does this PR solve?

Fixes #2974

When a KNN search uses a low-selectivity filter where the total valid candidates in the dataset or reachable component are fewer than `ef_search`, the searcher loop condition:
```cpp
if ((-current_first_node_pair.first) > lower_bound && top_candidates->Size() == ef) {
    break;
}
```
is never met because `top_candidates->Size()` cannot reach `ef`.

Moreover, candidate insertion logic (`top_candidates->Size() < ef || lower_bound > dist`) continually pushed neighbors into `candidate_set` regardless of distance. This caused the searcher to traverse the entire graph (millions of nodes/edges), stalling queries for minutes at 100% CPU.

### What is changed and how it works?

1. **Unrewarded Hops Safeguard**:
   - Tracks consecutive hops where the candidate distance exceeds `lower_bound` without adding or improving `top_candidates`.
   - Capped at `max_unrewarded_hops = std::max<uint32_t>(ef, 500)`.
   - Resets to `0` whenever a valid candidate is pushed to `top_candidates`.
2. **Empty Traversal Safeguard**:
   - When no valid candidates exist yet (`top_candidates->Empty()`), caps empty exploration at `max_empty_hops = std::max<uint32_t>(10000, 10 * ef)`.
3. **Searcher Coverage**:
   - Applied across all 3 `search_impl` overloads in `BasicSearcher` as well as `ParallelSearcher`.
4. **Regression Tests**:
   - Added unit tests in `basic_searcher_test.cpp` verifying bounded traversal when valid nodes < `ef`, zero valid nodes, and recall preservation across non-valid intermediate bridge nodes.

### Check List

- [x] Relevant tests added or updated
- [x] All unit & regression tests passing locally
- [x] Code style checked with `clang-format`
- [x] DCO Signed-off


src/impl/searcher/basic_searcher.cpp
```diff
@@ -253,6 +279,7 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
                 if (check_func(id) and
```
LHT129, line 280: [suggestion] The `empty_hops` counter is never reset to 0, unlike `unrewarded_hops` which is reset whenever a valid candidate is pushed to `top_candidates`. While this is safe in practice because `top_candidates` is monotonic (only grows, never shrinks during the main loop), it creates a latent fragility: if future changes ever remove entries from `top_candidates` mid-loop, `empty_hops` could carry stale state. Consider resetting `empty_hops = 0` alongside `unrewarded_hops = 0` for consistency and defense-in-depth.

This applies to all 3 `search_impl` overloads in `basic_searcher.cpp` and the one in `parallel_searcher.cpp`.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -219,14 +219,40 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
...
+                ++empty_hops;
+                if (empty_hops >= max_empty_hops) {
+                    if (reasoning != nullptr) {
+                        reasoning->SetTermination(ReasoningContext::kTerminationHopsLimitReached);
+                    }
+                    break;
+                }
+            } else if (-candidate_set->Top().first > lower_bound) {
+                if (top_candidates->Size() >= ef) {
+                    if (reasoning != nullptr) {
```
LHT129, line 243: [note] When the empty-hops safeguard triggers (`empty_hops >= max_empty_hops`), the termination reason is recorded as `kTerminationHopsLimitReached`, which is the same reason used when `hops >= inner_search_param.hops_limit`. These are semantically different conditions: one means "no valid candidates found after extensive exploration", the other means "the absolute hop budget was exhausted". If downstream consumers of the reasoning report need to distinguish these cases, consider introducing a distinct termination reason (e.g. `kTerminationNoValidCandidates`).

This applies to all 3 `search_impl` overloads in `basic_searcher.cpp`.

src/impl/searcher/parallel_searcher.cpp
```diff
@@ -258,8 +262,21 @@ ParallelSearcher::search_impl(const GraphInterfacePtr& graph,
         node_pair[0] = current_first_node_pair;
```
LHT129, line 264: [note] Unlike the `basic_searcher.cpp` overloads, the `parallel_searcher.cpp` safeguard paths do not record a termination reason via `reasoning->SetTermination(...)`. This appears to be because `ParallelSearcher::search_impl` does not have a `reasoning` variable wired up from `QueryContext`. If reasoning support is added to the parallel searcher in the future, these new termination paths should also record their reason.

This applies to the `empty_hops >= max_empty_hops` and `unrewarded_hops >= max_unrewarded_hops` break paths.

src/impl/searcher/parallel_searcher.cpp
```diff
@@ -151,6 +151,10 @@ ParallelSearcher::search_impl(const GraphInterfacePtr& graph,
     uint32_t hops = 0;
     uint32_t dist_cmp = 0;
     uint32_t count_no_visited = 0;
+    uint32_t unrewarded_hops = 0;
+    uint32_t empty_hops = 0;
+    const uint32_t max_unrewarded_hops = std::max<uint32_t>(ef, 500);
+    const uint32_t max_empty_hops = std::max<uint32_t>(10000, 10 * ef);
```
LHT129, line 159: [suggestion] The new safeguard logic in `ParallelSearcher::search_impl` does not report termination reasons via `ReasoningContext`, unlike the equivalent changes in `BasicSearcher::search_impl`.

In `basic_searcher.cpp`, both the `max_empty_hops` and `max_unrewarded_hops` break paths call `reasoning->SetTermination(...)` before breaking. The `ParallelSearcher` lacks a `reasoning` variable entirely in this scope, so these termination paths silently break without recording the reason.

This creates an inconsistency in the reasoning/debugging output between the basic and parallel searchers. Consider either:
1. Adding a `reasoning` variable (derived from `ctx`) and calling `SetTermination` in the break paths, or
2. At minimum, documenting why the parallel searcher intentionally omits this.

The `reasoning` variable can be obtained the same way as in `BasicSearcher`:
```cpp
auto* reasoning = ctx == nullptr ? nullptr : ctx->reasoning_ctx;
```

src/impl/searcher/basic_searcher.cpp
```diff
@@ -219,14 +219,40 @@ BasicSearcher::search_impl(const GraphInterfacePtr& graph,
...
         ++hops;
         if (hops >= inner_search_param.hops_limit) {
             break;
         }
         if constexpr (mode == InnerSearchMode::KNN_SEARCH) {
-            if (-candidate_set->Top().first > lower_bound && top_candidates->Size() >= ef) {
-                break;
+            if (top_candidates->Empty()) {
+                ++empty_hops;
+                if (empty_hops >= max_empty_hops) {
```
LHT129, line 234: [note] The `max_empty_hops` safeguard uses `kTerminationHopsLimitReached` as the termination reason, but this is semantically different from the actual `hops_limit` check at the top of the loop.

When `empty_hops >= max_empty_hops` triggers, the traversal was not terminated by the user-configured `hops_limit` — it was terminated by the new empty-traversal safeguard. Using the same reason string for both makes it harder to distinguish between these two termination paths in reasoning/debugging output.

Consider adding a dedicated constant (e.g., `kTerminationEmptyTraversalLimitReached`) or a more specific reason string to differentiate this case from a genuine hops-limit termination.

src/impl/searcher/basic_searcher.cpp
thlurte: Addressed in 7c8e6d0: reset `empty_hops = 0` alongside `unrewarded_hops = 0` across all `BasicSearcher` overloads and `ParallelSearcher`.

LHT129 (commented):
Reviewed the updated commits (e54b5e43). All issues raised in the previous review round have been addressed:

1. **`empty_hops` reset**: `empty_hops = 0` is now consistently reset alongside `unrewarded_hops = 0` across all `BasicSearcher` overloads and `ParallelSearcher`.
2. **Distinct termination reason**: The empty-traversal safeguard now uses the dedicated `kTerminationEmptyTraversalLimitReached` constant instead of reusing `kTerminationHopsLimitReached`.
3. **Parallel searcher reasoning**: `ParallelSearcher::search_impl` now derives `reasoning` from `QueryContext` and records termination reasons in both safeguard break paths, consistent with `BasicSearcher`.

The core logic — unrewarded hops safeguard and empty traversal safeguard — is correctly implemented across all 3 `BasicSearcher::search_impl` overloads and `ParallelSearcher::search_impl`. The three new test cases provide good coverage for the low-selectivity, zero-candidate, and bridge-node scenarios.

No new issues found.

PR #2984, opened 2026-09-20
perf(hgraph): accelerate multi-core search via compact visit and SQ8 Batch4 SIMD

159753a52:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Improvement/Refactor
- [ ] Documentation
- [ ] CI/Build/Infra

## What Changed
- **Visit Fast-Path & Inlining**: Enforced `always_inline` for `BasicSearcher::visit`, eliminating per-call function boundary overhead; implemented branch-reduced in-place compact collection for un-filtered neighbor visits (`count += not vl->TestAndSet(id)`), avoiding branch mispredictions in hot loops; replaced non-inlined `PrefetchLines(..., 64)` with `PrefetchImpl<1>`, issuing direct hardware prefetch hints (`prefetcht0`).
- **SQ8 Uniform Batch4 Distance Kernel**: Restructured `SQ8UniformComputeCodesIPBatch4` across AVX-512 and AVX2; unpacked query vectors once per block and shared across 4 candidate codes to amortize memory load and unpack instructions.
- **Candidate Queue Optimization**: Allocated temporary `candidate_set` on the stack across `search_impl` overloads, eliminating per-query heap control-block allocations.
- **Two-Level Parameter Caching**: Introduced thread-local strong-typed parameter cache and JSON AST cache for `HGraphSearchParameters`, avoiding repetitive lexical and syntax parsing for identical search queries.
- **Deferred Statistics Serialization**: Wrapped single-query search statistics in `DeferredStatisticsDataset`, deferring JSON string formatting and memory allocation until explicit `GetStatistics()` invocation. Guaranteed memory safety via C++ multi-inheritance destruction order (`DatasetImpl` buffer freed before `ResultAllocatorOwner`).
- **ELP Optimizer State Machine & Regression Tests**: Re-instantiated Optimizer on every ELP run to isolate stale parameter cards; dynamically bound object baseline stride (`prefetch_stride_code_`); installed winning candidate immediately upon successful dimension; added deterministic regression tests using virtual `MockRun`.

## Test Evidence
- [x] `make fmt`
- [x] `make lint`
- [x] `make test`
- [x] Other (describe below)

Test details:
```text
- 35 modified files passed clang-format-15 --dry-run -Werror
- Deterministic optimizer regression test suite added in basic_searcher_test.cpp:
  * Baseline preservation under no improvement (tested on initial stride 1, 4, 16)
  * Winning candidate installation (50% improvement installs stride 4)
  * Rejected candidates progression without looping (tested with max_attempts_ guard)
- Verified clean three-way merge against antgroup/vsag:main (897715d5) via git merge-tree (0 conflicts)
```

## Compatibility Impact
- API/ABI compatibility: Fully backward compatible. Deprecated `prefetch_stride_visit_` accepts parameters gracefully.
- Behavior changes: None. Floating point recall remains identical within standard tolerances across SIFT, GloVe, and MNIST benchmarks.

- Performance impact: Improved.
  * GloVe-25 (32 threads): +26.48% QPS (immutable), +26.62% QPS (mutable), P99 latency reduced by ~25.5%.
  * SIFT-128 (32 threads): +32.19% QPS (immutable), +29.27% QPS (mutable), P99 latency reduced by ~28% ~ -34%.
- Concurrency/thread-safety impact: None. `VisitedList` is exclusively borrowed by the coordinator thread during graph traversal; no concurrent writes across threads.

## Documentation Impact
- [x] No docs update needed
- [ ] Updated docs:

## Risk and Rollback
- Risk level: Low
- Rollback plan: Revert commit if regressions are observed.

## Checklist
- [ ] I have linked the relevant issue (required for `kind/bug` and `kind/feature`; see "Linked Issue" above)


src/impl/searcher/basic_searcher.cpp
```diff
@@ -57,14 +57,21 @@ BasicSearcher::visit(const GraphInterfacePtr& graph,
...
-    for (uint32_t i = 0; i < neighbors.size(); i++) {
-        if (i + prefetch_stride_visit_ < neighbors.size()) {
-            vl->Prefetch(neighbors[i + prefetch_stride_visit_]);
+    if (not filter) {
+        for (const auto id : neighbors) {
+            vl->Prefetch(id);
+        }
+        for (const auto id : neighbors) {
+            to_be_visited_id[count_no_visited] = id;
+            count_no_visited += not vl->TestAndSet(id);
```
LHT129, line 66: [suggestion] The `visit` function uses `not vl->TestAndSet(id)` in the no-filter fast path, which relies on implicit `bool`-to-`uint32_t` conversion (`true` → 1, `false` → 0). While this is well-defined in C++, the `not` alternative token may be surprising to readers more familiar with `!`. Consider using `!` for consistency with the rest of the codebase.

The compact-collection pattern `to_be_visited_id[count_no_visited] = id; count_no_visited += !vl->TestAndSet(id);` is correct: unvisited IDs are written at sequential positions and `count_no_visited` accurately tracks the count. Already-visited IDs leave a stale value at their slot, but those slots are beyond the returned `count_no_visited` boundary and are never read.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -1250,7 +1285,18 @@ HGraph::SearchWithRequest(const SearchRequest& request) const {
...
                 search_result->Pop();
             }
-            dataset_results->Statistics(mci_result.MakeStatistics(stats).Dump());
+            if (reasoning_ctx) {
+                dataset_results->Statistics(mci_result.MakeStatistics(stats).Dump());
+            } else {
+                auto metadata = mci_result;
+                metadata.result.reset();
+                dataset_results = std::make_shared<DeferredStatisticsDataset>(
+                    std::move(static_cast<DatasetImpl&>(*dataset_results)),
```
LHT129, line 1294: [suggestion] The `DeferredStatisticsDataset` constructor takes `DatasetImpl&& data` and the call site passes `std::move(static_cast<DatasetImpl&>(*dataset_results))`. After the move, the original `DatasetImpl` subobject is in a moved-from state, and then `dataset_results` (the `shared_ptr`) is reassigned to point to the new `DeferredStatisticsDataset`. This is correct because the old `shared_ptr` is overwritten — the moved-from `DatasetImpl` is immediately destroyed when the last reference drops.

However, the `static_cast<DatasetImpl&>` is fragile: if `dataset_results` ever points to a type that inherits from `DatasetImpl` but overrides its move constructor, slicing could occur. Consider adding a `static_assert` or a comment noting that `dataset_results` is always a plain `DatasetImpl` (not a subclass) at this point.

src/datacell/graph_datacell.h
```diff
@@ -336,20 +336,20 @@ GraphDataCell<IOTmpl>::GetNeighbors(InnerIdType id, Vector<InnerIdType>& neighbo
     }
     if (is_support_delete_) {
         neighbor_count &= remove_flag_mask_;
-        Vector<InnerIdType> shared_neighbor_ids(neighbor_count, this->allocator_);
+        neighbor_ids.resize(neighbor_count);
```
LHT129, line 339: [note] The `GetNeighbors` refactoring eliminates the temporary `shared_neighbor_ids` vector and its associated allocation by reusing `neighbor_ids` directly for both the raw read and the in-place filtered result. The `valid_count` index tracks the compacted output within the same buffer, and the final `resize(valid_count)` truncates to the correct length. This is a clean improvement that removes one heap allocation per call.

One subtlety: `neighbor_ids.resize(neighbor_count)` before `ReadAt` may allocate if the existing capacity is insufficient, but since `neighbor_ids` is typically a caller-provided `Vector<InnerIdType>` that was already sized to `maximum_degree_`, this resize is usually a no-op in practice.

src/utils/json_parameter_cache.h
```diff
@@ -0,0 +1,65 @@
...
+
+inline bool
+IsJsonParameterCacheable(const std::string& parameters) {
+    constexpr uint64_t kMaxCachedJsonParameterBytes = 4096;
+    return static_cast<uint64_t>(parameters.size()) <= kMaxCachedJsonParameterBytes;
+}
+
+inline const JsonType*
+GetCachedJsonParameter(const std::string& parameters) {
+    if (not IsJsonParameterCacheable(parameters)) {
```
LHT129, line 34: [note] `GetCachedJsonParameter` returns a pointer to a `thread_local` cache entry. The pointer is valid only until the next call to `GetCachedJsonParameter` on the same thread with different parameters — at which point `cache.json` is overwritten. This is safe for the current usage pattern where the pointer is consumed immediately (e.g., in `GetOrParseJsonParameter`), but callers that store the pointer across subsequent `GetCachedJsonParameter` calls would observe a use-after-overwrite.

Consider adding a brief comment on the returned pointer's lifetime constraint, or returning by value (with the understanding that `JsonType` may be expensive to copy for the hot path).

src/impl/searcher/basic_searcher.cpp
```diff
@@ -57,14 +57,21 @@ BasicSearcher::visit(const GraphInterfacePtr& graph,
         graph->GetNeighbors(current_node_pair.second, neighbors);
     }
 
-    for (uint32_t i = 0; i < neighbors.size(); i++) {
-        if (i + prefetch_stride_visit_ < neighbors.size()) {
```
LHT129, line 61: [suggestion] `__attribute__((always_inline))` is a GCC/Clang-specific extension and will not compile on MSVC. VSAG targets multiple platforms including Windows. Consider using a portable macro such as `VSAG_ALWAYS_INLINE` (or `ALWAYS_INLINE`) that expands to the appropriate compiler attribute, or use `inline` alone and rely on LTO for cross-TU inlining.

If `visit` is only called within this translation unit and the compiler already inlines it, the attribute may be unnecessary. Please verify with a disassembly check on the relevant build configurations.

src/impl/searcher/basic_searcher.cpp
```diff
@@ -57,14 +57,21 @@ BasicSearcher::visit(const GraphInterfacePtr& graph,
...
-            vl->Prefetch(neighbors[i + prefetch_stride_visit_]);
+    if (not filter) {
+        for (const auto id : neighbors) {
+            vl->Prefetch(id);
+        }
+        for (const auto id : neighbors) {
+            to_be_visited_id[count_no_visited] = id;
+            count_no_visited += not vl->TestAndSet(id);
         }
```
LHT129, line 67: [suggestion] In the no-filter fast path, `count_no_visited += not vl->TestAndSet(id)` relies on implicit `bool`-to-`uint32_t` conversion. While this is well-defined by the standard, it is an unusual idiom. Consider making the intent explicit:

```cpp
if (not vl->TestAndSet(id)) {
    to_be_visited_id[count_no_visited++] = id;
}
```

This avoids relying on implicit conversion semantics and makes the control flow clearer to readers.

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -31,6 +31,40 @@
...
+class DeferredStatisticsDataset final : private ResultAllocatorOwner, public DatasetImpl {
+public:
+    DeferredStatisticsDataset(DatasetImpl&& data,
+                              std::shared_ptr<Allocator> allocator,
+                              std::function<std::string()> statistics)
+        : ResultAllocatorOwner(std::move(allocator)),
+          DatasetImpl(std::move(data)),
+          statistics_(std::move(statistics)) {
+    }
+
```
LHT129, line 51: [note] `DeferredStatisticsDataset::GetStatistics()` is `const` but calls `statistics_()` which mutates the internal state of the `SearchStatistics` object captured by `stats_owner`. The `const` qualifier is technically correct (no members of `DeferredStatisticsDataset` are mutated), but the captured `shared_ptr<SearchStatistics>` is not `const`, so `MakeStatistics` may modify the statistics object. This is fine for single-threaded access, but if multiple threads call `GetStatistics()` concurrently on the same `DeferredStatisticsDataset`, there could be a data race inside `MakeStatistics`. The test in `hgraph_statistics_test.cpp` verifies concurrent reads are consistent, but this relies on `MakeStatistics` being effectively read-only after the search completes. Consider documenting this invariant.

src/algorithm/hgraph/hgraph_build.cpp
```diff
@@ -1123,8 +1123,9 @@ HGraph::elp_optimize() {
     param.is_inner_id_allowed = nullptr;
     searcher_->SetMockParameters(bottom_graph_, basic_flatten_codes_, pool_, param, dim_);
     // TODO(ZXY): optimize PREFETCH_DEPTH_CODE and add default value for the others
-    optimizer_->RegisterParameter(RuntimeParameter(PREFETCH_STRIDE_CODE, 1, 10, 1));
-    optimizer_->RegisterParameter(RuntimeParameter(PREFETCH_STRIDE_VISIT, 1, 10, 1));
+    optimizer_ = std::make_shared<Optimizer<BasicSearcher>>(index_common_param_);
```
LHT129, line 1126: [note] The `PREFETCH_STRIDE_VISIT` optimizer parameter registration was removed from `elp_optimize()`. The PR description states `BasicSearcher::visit` no longer consumes `prefetch_stride_visit_`, so this is intentional. However, the `prefetch_stride_visit_` member and its `SetRuntimeParameters` handling are kept for backward compatibility. If `prefetch_stride_visit_` is truly dead in the visit hot path, consider adding a deprecation warning or removing it in a follow-up cleanup PR to avoid confusion for future maintainers.

src/utils/json_parameter_cache.h
```diff
@@ -0,0 +1,65 @@
...
+        return nullptr;
+    }
+
+    struct JsonParameterCache {
+        std::string parameters;
+        std::optional<JsonType> json;
+    };
+    thread_local JsonParameterCache cache;
+
```
LHT129, line 43: [note] `GetCachedJsonParameter` returns a raw pointer into a `thread_local` cache. `GetOrParseJsonParameter` dereferences it and returns a `const JsonType&`. Callers that store this reference across subsequent `GetOrParseJsonParameter` calls could observe the cached value change underneath them. In the current PR, all callers consume the reference immediately (within the same expression or statement), so this is safe. Consider adding a comment noting that the returned reference must not outlive the next call to `GetOrParseJsonParameter` with a different string.

src/impl/basic_optimizer.cpp
```diff
@@ -39,6 +39,7 @@ Optimizer<OptimizableOBJ>::Optimize(std::shared_ptr<OptimizableOBJ> obj) {
             current_params[param.name_] = param.Cur();
             auto set_status = obj->SetRuntimeParameters(current_params);
             if (not set_status) {
+                param.Next();
                 continue;
             }
```
LHT129, line 45: [note] Good catch: when `SetRuntimeParameters` returns false, the optimizer previously did not advance `param.Next()`, which could cause an infinite loop if a parameter value was persistently rejected. Adding `param.Next()` here ensures the optimizer makes forward progress. The regression tests in `basic_searcher_test.cpp` ("Rejected candidates progress without looping") validate this fix.

PR #2988, opened 2026-09-22
perf(mci): coverage-driven seeds, saturation guard and expansion early stop

### LINKED ISSUE #2987: [improve](hgraph): MCI loses to brute force above ~0.5% selectivity because the seed budget ignores the valid set
### Describe the issue

MCI only beats an exact scan while the seed phase covers the whole valid set. The seed budget is
`ceil(sqrt(N) * mci_seed_ratio)`, which is **independent of the filter**, so it stops covering the
filter above a certain selectivity and the search falls back to the clique expansion:

* the expansion costs **3-8 us per clique** and runs **17k-54k cliques (1.8M-5.6M member probes)
  per query** on gist-1m;
* those probes discover almost nothing: measured **7.4-288 hops per newly scored point**, and
  exactly **0 new distance evaluations** once the seed phase already covered the valid set;
* the seed phase, by contrast, is exact and cheap: `~0.5 ms + 0.65 us * valid`, i.e. seeding a
  point costs ~10x less than the expansion it replaces.

Measured on gist-1m (N=1e6, d=960, ef=600, topk=100, 2000 queries per selectivity band,
`mci_seed_ratio=5` so the budget is 5000 seeds, i.e. 0.5% selectivity):

| selectivity | valid | MCI | brute force | MCI recall | BF recall |
|---|---|---|---|---|---|
| 0.50% | 4,955 | 3.7 ms | 11.6 ms | 0.9725 | 0.9725 |
| 0.99% | 9,942 | 138.3 ms | 13.3 ms | 0.9630 | 0.9657 |
| 1.98% | 19,820 | 151.9 ms | 13.1 ms | 0.9633 | 0.9660 |
| 0.51% (clustered labels) | 5,132 | 122.7 ms | 7.2 ms | 0.9735 | 0.9756 |
| 0.97% (clustered labels) | 9,727 | 131.1 ms | 9.1 ms | 0.9449 | 0.9705 |

So above roughly **0.5% selectivity** MCI is both slower than an exact scan and slightly less
accurate, purely because the seed budget is not derived from the filter.

### Expected behaviour

* raise the seed budget from the valid-set size whenever that fits a cap, so the seed phase can be
  complete (it then returns the exact top-k and the expansion can be skipped entirely);
* bound the remaining expansion for the bands where full coverage is too expensive.

Related: #2977 (clique expansion that discovers nothing), #2972.

LightWant:
### Summary

MCI loses to an exact scan above roughly **0.5% selectivity** because the seed budget
`ceil(sqrt(N) * mci_seed_ratio)` ignores the filter: the lossless saturation guard cannot trigger,
so the query falls back to the clique expansion tail (17k-54k cliques, 3-8 us each, 1.8M-5.6M member
probes) which discovers almost nothing (7.4-288 hops per newly scored point, and **0 new distance
evaluations** once the seed phase is already complete).

This PR makes the seed phase able to be complete, keeps it exact when it is, and prunes the
remaining expansion.

### Changes

| item | change |
|---|---|
| saturation guard | the seed list reports whether it enumerates every valid point; the searcher then returns the exact top-k and skips the expansion (`mci_seed_saturated`) |
| seed budget | new `mci_seed_coverage` (default 1.0) and `mci_seed_max_count` (default 32768): budget = `max(sqrt budget, mci_seed_coverage * valid)` when it fits the cap |
| expansion early stop | new `mci_expansion_idle_window` (default 32) and `mci_expansion_lower_bound` (default false) replace the experimental env knobs; both criteria are now also applied to the precise-float CSR path, which had no early stop |
| distance kernel | seed distances are evaluated in batches of 64, matching HGraph's brute force (a single-id call pays per-call overhead once per point) |
| accounting | result heap is sized by the user's top-k (`rerank_topk`); distances are counted once per search with the flatten's real backend (the old code hardcoded `FP32`) |

### Measurements (gist-1m, N=1e6, d=960, ef=600, topk=100, 2000 queries per band)

| selectivity | valid | before | after | recall vs brute force |
|---|---|---|---|---|
| 1.0% (indep) | 9,942 | 0.96301 / 138.3 ms | **0.96566 / 5.0 ms** | equal |
| 2.0% (indep) | 19,820 | 0.96325 / 151.9 ms | **0.96602 / 10.3 ms** | equal |
| 0.51% (cluster) | 5,132 | 0.97352 / 122.7 ms | **0.97560 / 3.4 ms** | equal |
| 0.97% (cluster) | 9,727 | 0.94495 / 131.1 ms | **0.97055 / 5.2 ms** | equal |
| 1.95% (cluster) | 19,505 | 0.95440 / 139.2 ms | **0.96846 / 10.4 ms** | equal |

Aggregate (all 12 bands): indep `0.97150 / 75.92 ms / 673 QPS` -> `0.97156 / 21.83 ms / 2168 QPS`;
cluster `0.97007 / 95.51 ms / 547 QPS` -> `0.97314 / 25.15 ms / 1944 QPS`. Bands above 5%
selectivity are unchanged (the budget falls back to `sqrt(N) * mci_seed_ratio`).

### Tests

`[mci],[MCISearcher]` 21 cases / 169 assertions, `[hgraph]` 72 cases / 4261 assertions,
`*parameter*` 100 cases / 25836 assertions pass in the Debug build (leak check active);
`clang-format-15 --dry-run -Werror` is clean. New tests cover the coverage budget
(`mci_seed_count == valid_count`, exact top-k, and the disabled path) and the idle window
(strictly fewer hops with an earlier stop).

### Notes

* Based on #2973 (the bitmap fast path / `clique_max` fix this builds on), whose head branch lives
  in a fork and therefore cannot be used as a base ref here. While #2973 is open the diff below
  includes its 3 commits; the new work is the last commit (`perf(mci): cover the valid set before
  expanding cliques`) and the diff shrinks to that commit automatically once #2973 lands.
* The idle expansion stop generalises #2978 / #2977 — if that PR lands first, only the parameter
  plumbing here is redundant and this PR can drop it.
* The lower-bound criterion stays opt-in (default false): it is a heuristic (a clique member can be
  closer to the query than the candidate that led to it) and it measured no effect on gist-1m /
  openai-5M.

Fixes #2987

## Per-route `ef_search`

`ef_search` was a single value for both routes, so it had to be sized for the worst band. The
banded router now accepts one breadth per route; each falls back to `ef_search` when unset, so
existing parameter strings and callers are unaffected:

| Key | Route | Default |
| --- | --- | --- |
| `hgraph_ef_search` | plain HGraph walk (single-query, range, batch/fused) | `ef_search` |
| `mci_ef_search` | MCI hybrid expansion | `ef_search` |

```json
{"hgraph": {"ef_search": 600, "mci_ef_search": 100}}
```

The asymmetry is measured: in the gist-1m ef sweep the covered band held its recall at every
`ef` (0.9984 / 0.9725 on indep / cluster), while the uncovered HGraph band needed `ef` above 320
(0.706 -> 0.953 from ef 160 to ef 600). One shared `ef` therefore pays for the worst band on
every query; a large `ef_search` with a small `mci_ef_search` is what makes the "<= 2% to MCI"
routing policy cheap.

The hybrid sizes itself as `ef = max(mci_ef_search, rerank_topk)`, so a value below the user's
`k` cannot shrink the result set, and the caller's `search_param` is left untouched.
`apply_hops_limit` now compares against the smaller of the two effective breadths, so narrowing
one route can no longer silently discard a user hop limit.

Tests: `[HGraphParameter]` covers the fallback matrix (only `ef_search`; one route narrowed;
both narrowed; `< 1` rejected). An integration test on the MCI and HGraph routes asserts that
`{"ef_search": 4096, "mci_ef_search": 16}` and `{"ef_search": 16}` produce identical hop counts,
and likewise for `hgraph_ef_search`; reverting the wiring to `params.ef_search` makes that
assertion fail, so the test is sensitive to the regression it guards.


src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -548,23 +562,71 @@ HGraph::try_mci_search(const SearchRequest& request,
...
-    mci_param.seed_count = scaled_seed_count >= static_cast<double>(total_count)
+    uint64_t seed_budget = scaled_seed_count >= static_cast<double>(total_count)
                                ? total_count
                                : std::max<uint64_t>(1, static_cast<uint64_t>(scaled_seed_count));
+    // Coverage-driven budget: seeding every valid point makes the seed phase exact (the searcher
+    // then skips the whole clique expansion), which beats the expansion tail whenever
+    // `0.5 ms + per_distance * valid` does. The budget is only raised when the target fits into
+    // `mci_seed_max_count`, so a wide filter never degenerates into an exact scan.
+    if (params.mci_seed_coverage > 0.0F and not bitset_seed_source and request.filter_ != nullptr) {
+        const int64_t* valid_labels = nullptr;
```
LHT129: [suggestion] The `GetValidIds()` call for coverage budget calculation happens outside the `label_lookup_mutex_` shared lock, while `collect_seed_inner_ids()` and `GetValidBitmap()` both run inside it. If a label mutation occurs between the coverage budget calculation and the seed collection, `valid_label_count` may already be stale, and the budget could be raised for a filter that no longer has that many valid labels. Consider moving the `GetValidIds()` call inside the lock scope as well, or document why the race is benign (e.g., the budget is a hint and an overestimate is harmless).

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -548,23 +562,71 @@ HGraph::try_mci_search(const SearchRequest& request,
...
+        int64_t valid_label_count = 0;
+        request.filter_->GetValidIds(&valid_labels, valid_label_count);
+        if (valid_label_count > 0) {
+            const auto covered =
+                static_cast<uint64_t>(std::ceil(static_cast<double>(params.mci_seed_coverage) *
+                                                static_cast<double>(valid_label_count)));
+            const bool fits_cap = params.mci_seed_max_count <= 0 or
+                                  covered <= static_cast<uint64_t>(params.mci_seed_max_count);
```
LHT129: [suggestion] `mci_seed_max_count` is typed `int64_t` with `<= 0` as the "unlimited" sentinel, but the JSON parser already validates `>= 0`. The `<= 0` check in `try_mci_search` (line 581) would also treat `0` as unlimited, which is inconsistent with `mci_expansion_idle_window` where `0` means disabled. Consider using `0` consistently as "unlimited" (and removing the `< 0` path), or adding a named constant like `kMciSeedMaxCountUnlimited` to make the sentinel explicit.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -125,12 +212,22 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
                          Allocator* allocator) {
     const auto candidate_limit =
         std::max<int64_t>(inner_search_param.topk, static_cast<int64_t>(inner_search_param.ef));
-    auto result_heap = DistanceHeap::MakeInstanceBySize<true, true>(allocator, candidate_limit);
+    // The returned heap is truncated to the user's top-k by the caller; `candidate_limit`
```
LHT129: [note] The `result_limit` change (using `rerank_topk` or `topk` instead of `max(topk, ef)`) is a good optimization — the result heap only needs to hold the final top-k, not the full expansion frontier. However, `search_precise_float_csr` still uses `candidate_limit` (which is `max(topk, ef)`) for the `candidates.Reset()` call. The two limits serve different purposes (result heap vs. expansion frontier), and the naming makes this clear. Worth noting that if `rerank_topk > max(topk, ef)`, the result heap could be larger than the frontier, but this is an unlikely configuration.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -66,6 +95,64 @@ struct MCIEpochMarks {
...
+    }
+
+    void
+    Reset(const uint8_t* valid_bitmap, uint64_t total) {
+        if (state.size() < total) {
+            state.resize(total);
+        }
+        // The destination is reached through a local pointer: indexing the vector member directly
+        // makes the compiler reload its data pointer on every iteration and keeps the loop scalar.
+        auto* destination = state.data();
```
LHT129, line 118: [note] `MCIMergedVisitMarks::Reset` normalizes the bitmap with `valid_bitmap[id] == 0 ? 0 : 1`, treating any non-zero byte as valid. This matches `Filter::CheckValid` semantics but means the bitmap provider must not use multi-valued bytes (e.g., encoding distances or categories). The `InnerIdWrapperFilter::GetValidBitmap` doc comment already states the contract, but it may be worth adding a brief note here that the bitmap is treated as a boolean mask.

src/algorithm/mci/mci_builder.cpp
```diff
@@ -156,6 +156,12 @@ BuildMCICliques(const float* vectors,
     const auto thread_count = std::max<uint64_t>(1, std::min<uint64_t>(params.thread_count, total));
     const auto max_saved_cliques =
         std::min<uint64_t>(candidate_limit, static_cast<uint64_t>(params.max_degree + 2));
+    // clique_max is the MINIMUM size the enumerator must keep: every maximal clique found inside
```
LHT129, line 159: [note] The `clique_max` semantic change from "storage cap" to "minimum size of a maximal clique to keep" is well-motivated and the new `clique_cap = candidate_limit + 1` correctly bounds the stored clique size. The `hgraph_mci.cpp` build path mirrors this change with `clique_min = max(K_MCI_MIN_CLIQUE_SIZE, clique_max)`. Both the float path (`mci_builder.cpp`) and the quantized path (`hgraph_mci.cpp`) are updated consistently. The test "MCI builder treats clique_max only as a storage cap" has been renamed and rewritten to verify the new semantics.

src/impl/label_table/label_table.h
```diff
@@ -341,7 +345,49 @@ class LabelTable {
...
+     * inner-id space coincide.
+     *
+     * Components that reuse an id-space-indexed structure as an inner-id-indexed one (for example
+     * the validity bitmap that the MCI searcher inlines) must check this first, because the label
+     * a filter observes is not necessarily the inner id it is probed with. Only the live ids
+     * `[0, GetTotalCount())` take part: the table pre-allocates capacity, and those spare slots are
+     * never probed. The answer is cached and recomputed after any label mutation.
+     *
+     * Concurrency: scanning the table requires concurrent label mutations to be excluded, because a
+     * mutation can reallocate the storage being read. `labels_version_` only protects the cached
```
LHT129, line 367: [note] `IsIdentityMapping()` uses a version-based cache with `labels_version_` / `identity_version_` / `identity_mapping_`. The scan of `label_table_` inside the method is documented to require the caller to hold `label_lookup_mutex_` in shared mode, which `try_mci_search` already does. The `MarkLabelsMutated()` calls are correctly placed after every write to `label_table_`. The `labels_version_` starts at 1 (not 0) so that the initial `identity_version_` of 0 always triggers a recomputation on first call — a clean pattern.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -288,17 +436,33 @@ MCISearcher::Search(const CliqueDataCellPtr& cliques,
 
     auto computer_lease = AcquireQueryComputer(flatten, query, ctx);
     const auto& computer = computer_lease.computer;
+    // Distances are counted once at the end of the search, so the datacell must not aggregate
```
LHT129, line 444: [suggestion] The `distance_ctx_storage` copies the entire `QueryContext` (including the `stats` pointer), then disables `track_distance_evaluations` on the copy. This correctly prevents double-counting distances during seed batch evaluation (since `Search` already counts them at the end via `ctx->stats->AddDistance`). However, the copy also duplicates `allocator` and other fields, which is harmless but slightly wasteful. Consider using a more targeted approach: pass a null `QueryContext*` to `flatten->Query` for the seed batch (since the distances are counted at the end anyway), or explicitly document why the full copy is needed.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -118,6 +126,10 @@ collect_bitset_seed_inner_ids(const FilterPtr& inner_filter,
             accumulated_remainder -= sample_count;
         }
     }
+    if (enumerates_all_valid != nullptr) {
```
LHT129, line 149: [note] The `enumerates_all_valid` condition in `collect_bitset_seed_inner_ids`: `id_step == 1 and inner_ids.size() < seed_count`. This correctly identifies the case where every inner id was probed (`id_step == 1`, no stride) and the scan exhausted all ids before filling the budget (`inner_ids.size() < seed_count`). This means all valid points were collected. The logic is correct but the `<` operator is easy to misread — consider adding a brief comment like "the loop terminated because it ran out of ids, not because the budget was filled."

src/impl/filter/inner_id_wrapper_filter.h
```diff
@@ -45,6 +45,19 @@ class InnerIdWrapperFilter : public Filter {
         filter_impl_->GetValidIds(valid_ids, count);
     }
 
+    [[nodiscard]] const uint8_t*
```
LHT129, line 48: [note] `InnerIdWrapperFilter::GetValidBitmap` correctly gates on `label_table_.IsIdentityMapping()` before forwarding the wrapped filter's bitmap. This is the key safety check that prevents a label-indexed bitmap from being misinterpreted as an inner-id-indexed one. The `IsIdentityMapping()` cache (with version-based invalidation via `MarkLabelsMutated()`) ensures this check is cheap on the hot path. The test "HGraph MCI ignores a bitmap whose ids are not inner ids" validates this gate end-to-end.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -321,22 +486,44 @@ MCISearcher::Search(const CliqueDataCellPtr& cliques,
         return nullptr;
     };
     uint32_t dist_cmp = 0;
-    auto try_visit = [&](InnerIdType inner_id) -> bool {
-        if (inner_id >= total or visited_nodes.Get(inner_id)) {
+    auto try_mark = [&](InnerIdType inner_id) -> bool {
```
LHT129, line 497: [note] The split of `try_visit` into `try_mark` + `score_marked` in `MCISearcher::Search` is a clean refactor that enables the seed batch optimization: marks are checked first (under the lock/bitmap), and distance computation is deferred to the batch `flush_seed_batch` call. This avoids the per-call overhead of `flatten->Query` for single-id calls during seed evaluation. The `search_precise_float_csr` path does not need this split because it computes distances directly without going through `flatten->Query`.

LHT129 (commented):
## Code Review for PR #2988

### [suggestion] `search_precise_float_csr` lacks seed batching optimization

In `src/impl/searcher/mci_searcher.cpp`, the main `Search()` function batches seed distance computations in groups of 64 (`kSeedBatch`) via `try_mark`/`score_marked`/`flush_seed_batch` to amortize per-call overhead (dispatch, layout acquire, statistics). However, `search_precise_float_csr` still calls `try_visit` one seed at a time:

```cpp
// search_precise_float_csr, ~line 299:
if (try_visit((*mci_param.seed_inner_ids)[offset])) {
    ++seeds;
}
```

This means the CSR path pays the per-call overhead once per seed point instead of once per 64, which can be significant when the seed budget is large (e.g., `mci_seed_max_count` defaults to 32768). The fix would be to apply the same `kSeedBatch` pattern here: split `try_visit` into `try_mark` + `score_marked`, accumulate ids in a batch buffer, and flush every 64.

### [note] Overall assessment

The rest of the PR looks solid:
- **Bitmap fast path** (`MCIMergedVisitMarks`): merges filter validity and visited state into one byte array, avoiding virtual `CheckValid()` calls per probe. Correctly gated behind `GetValidBitmap()` returning a non-null bitmap.
- **Saturation guard**: when seeds enumerate all valid points, clique expansion is skipped entirely — correct optimization.
- **Expansion early stop**: idle window and lower-bound heuristics are well-isolated in `MCIEarlyStopKnobs` with env-var overrides for A/B testing.
- **`clique_cap = candidate_limit + 1`**: fixes a bug where `mci_clique_max` (a minimum size) was used as a storage cap, which could truncate valid cliques.
- **`LabelTable::IsIdentityMapping()`**: version-based caching with atomic counter is concurrency-safe.
- **Result heap sizing**: switched from `candidate_limit` to `rerank_topk`-aware sizing, reducing memory for the common case.

No correctness or security issues found.

LHT129 (commented):
I have completed a thorough review of this PR. The code quality is high and the changes are well-structured.

1. **Coverage-driven seed budget** (`mci_seed_coverage` / `mci_seed_max_count`): The logic for raising the seed budget to cover all valid points when the filter selectivity is low is sound. The `mci_seed_max_count` cap correctly prevents degenerate exact scans on wide filters.

2. **Saturation guard** (`enumerates_all_valid` / `seeds_cover_all_valid_points`): The mechanism to detect when the seed phase has enumerated every valid point and skip the clique expansion is correct and well-integrated into both the main search path and the precise-float CSR path.

3. **Expansion early stop** (`mci_expansion_idle_window` / `mci_expansion_lower_bound`): The idle-window heuristic is properly applied to both search paths (the CSR path previously had no early stop). The lower-bound criterion is correctly opt-in (default false).

4. **Bitmap fast path** (`GetValidBitmap` / `MCIMergedVisitMarks`): The merged visit marks avoid a virtual `CheckValid()` call per probed vector, and the `IsIdentityMapping()` gate in `InnerIdWrapperFilter` correctly prevents label-indexed bitmaps from being misinterpreted as inner-id-indexed ones. The version-based cache in `LabelTable` keeps the check cheap.

5. **Seed distance batching**: Batching seed distances in groups of 64 matches HGraph's brute-force pattern and avoids per-call overhead.

6. **Distance accounting fix**: Using `flatten->backend_` instead of hardcoded `FP32` for distance counting, and avoiding double-counting via `distance_ctx_storage`.

7. **`clique_max` semantic correction**: The change from "storage cap" to "minimum size of a maximal clique to keep" is consistently applied across both the float path (`mci_builder.cpp`) and the quantized path (`hgraph_mci.cpp`), with the new `clique_cap = candidate_limit + 1` correctly bounding stored clique size.

8. **Test coverage**: New tests cover the coverage budget, exact top-k, disabled path, idle window, bitmap fast path gating, and `IsIdentityMapping`. The renamed clique_max test correctly validates the new semantics.

The existing inline comments from the previous reviewer cover the remaining discussion points (lock scope for `GetValidIds`, sentinel value consistency, etc.). I have no additional blocking concerns.

src/algorithm/hgraph/hgraph_mci.cpp
LHT129, line 149: [note] The `enumerates_all_valid` condition `id_step == 1 and inner_ids.size() < seed_count` correctly identifies that all valid ids were collected when the loop exhausted all samples before filling the budget. However, when `sample_count == total` (every id was probed, which happens when `id_step == 1 and id_remainder == 0`) and the budget is exactly filled by valid ids (`inner_ids.size() == seed_count`), the flag is set to `false` even though every valid id was collected. This is a missed saturation opportunity (the expansion still runs and produces the same result), not a correctness issue. Consider also checking `id_remainder == 0` or computing `sample_count == total` before the loop to cover this edge case.

LHT129 (commented):
Additional review findings:

LightWant:
Thanks for the review — all four points are addressed in `18b84321`.

**1. [P1] bitset coverage false positive — confirmed and fixed.** Your repro reproduces exactly: with
`total=6000` and a budget of 8 the loop runs `sample_count=4096` iterations with `id_step=1` and
`id_remainder=1904`, so 1904 positions are never probed even though `id_step == 1`. The condition is
now `sample_count == total and inner_ids.size() < seed_count` (the scan must really have visited every
inner id, and the budget must not have cut it short).

Regression test added: a 6000-vector index, a blacklist filter that keeps exactly `{0, 3}` valid, a
query sitting on id 3 and `mci_seed_ratio=0.1` (budget 8). It asserts `mci_seed_count == 1` (only id 0
is probed), `mci_seed_saturated == "false"` and that the returned neighbour is id 3 rather than id 0.
I verified the test **fails against the previous condition** (`1 test case failed`) and passes with the
fix, so it does guard the behaviour rather than merely documenting it.

**2. [P2] float-to-integer conversion UB — fixed.** The target is now computed and range-checked as a
`double` against `mci_seed_max_count` (or `total_count` when the cap is disabled) before any cast, so
no out-of-range conversion can happen; `mci_seed_coverage = 1e30` now simply keeps the sqrt budget.

**3. [P2] result heap vs conjugate-graph enhancement — reverted.** You are right about the call
chain: `hgraph_search.cpp` raises `base_search_param.topk` to at least `LOOK_AT_K` before the search so
that `EnhanceResult()` has candidates to work with, and sizing the MCI result heap by `rerank_topk`
(possibly 1) removed them. The sizing change measured no benefit (recall was bit-identical and latency
was within noise), so both search paths use `candidate_limit` again and the enhancement keeps its
candidates.

**4. [P2] batched `mci_seed_saturated` — fixed.** It is now accumulated over sub-queries with OR
semantics, exactly like `mci_raw_float_csr`, and emitted into the batch statistics; a two-query batch
test asserts `true` for a filter whose seed phase covers everything.

On the ABI item: that comes from #2973 (old headers compiled against the new shared library), which is
already merged, so it is not introduced by this PR.

Checks: `[mci],[MCISearcher]` 23 cases / 182 assertions, `[hgraph]` 74 cases / 4274 assertions pass;
`clang-format-15 --dry-run -Werror` clean. The branch was also rebased onto current `main` after #2973
was squash-merged, so the diff is now this change set plus the review-fix commit.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -299,6 +356,15 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
...
+                if (++idle_candidates >= early_stop.idle_limit) {
+                    break;
+                }
+            } else {
+                idle_candidates = 0;
+            }
+        }
     }
 
     if (ctx != nullptr and ctx->stats != nullptr) {
```
LHT129, line 378: [suggestion] `search_precise_float_csr` still hardcodes `DistanceEvaluationBackend::FP32` in its final `AddDistance` call (line ~370), while `MCISearcher::Search` correctly uses `flatten->backend_` (line ~660). The PR description states "distances are counted once per search with the flatten's real backend (the old code hardcoded FP32)", but this fix was only applied to one of the two search paths.

When the precise float CSR path is taken (e.g. the index stores raw float vectors alongside quantized codes), the distance evaluation backend in statistics will always be reported as FP32 regardless of the actual flatten backend. Consider passing the flatten's backend through `MCISearcherParam` (similar to how `used_precise_float_csr` is passed back) so both paths report the correct backend.

Current code in `search_precise_float_csr`:
```cpp
ctx->stats->AddDistance(SearchStatistics::DistancePhase::APPROXIMATE,
                        DistanceEvaluationBackend::FP32,
                        dist_cmp);
```

Fixed code in `MCISearcher::Search`:
```cpp
ctx->stats->AddDistance(
    SearchStatistics::DistancePhase::APPROXIMATE, flatten->backend_, dist_cmp);
```

src/algorithm/hgraph/hgraph_mci.cpp
LightWant: Fixed in 7e3757b2: the coverage budget now runs inside the same `label_lookup_mutex_` shared scope as `collect_seed_inner_ids()` and `GetValidBitmap()`, so the valid-label list it reads and the list the collector walks can no longer be observed inconsistently when a label mutation races the search.

src/algorithm/hgraph/hgraph_mci.cpp
LightWant: Keeping it as is, but the sentinel is now explicit in the parameter comment (`0 = unlimited` for `mci_seed_max_count`). The parser rejects negative values, so `<= 0` in the search path only ever means `0`. The difference from `mci_expansion_idle_window` (0 = disabled) is intended: one is a budget cap, the other an early-stop window. Happy to introduce a named constant if you prefer it.

src/impl/searcher/mci_searcher.cpp
LightWant: The `result_limit` sizing was reverted in 18b84321: it measured no benefit, and as you noted it risks removing the extra candidates the conjugate-graph enhancement consumes. Both paths use `candidate_limit` again, so there is no `result_limit` left to reconcile with `candidates.Reset()`.

src/impl/searcher/mci_searcher.cpp
LightWant: Agreed the bitmap is treated as a boolean mask; that contract is documented on `InnerIdWrapperFilter::GetValidBitmap()` and this code is part of #2973, which is already merged. An extra note in `MCIMergedVisitMarks::Reset` would be a separate follow-up rather than a change in this PR.

src/impl/label_table/label_table.h
LightWant: Also part of the merged #2973 (label-table identity cache); nothing to change here.

src/impl/searcher/mci_searcher.cpp
LightWant: The copy happens once per `Search()` call, not per distance -- the per-id calls only reuse the `distance_ctx` pointer -- so it is a single small struct copy. Passing `nullptr` instead would drop the datacell's `io_cnt` accounting for those distances, so I kept the copy; `track_distance_evaluations = false` on the copy is what prevents double counting, and the search still reports the total through `AddDistance(dist_cmp)`.

src/algorithm/hgraph/hgraph_mci.cpp
LightWant: Superseded by the P1 fix: the condition is now `sample_count == total and samples_done == sample_count` with a comment explaining why `id_step == 1` alone was wrong (it still skips `id_remainder` positions). Correction to my earlier comment: the new regression test asserts the collector contract (probed count + saturation), not the reachability of the skipped point -- that assertion turned out to be flaky because it depends on the sampled clique build. The contract assertions are deterministic and still fail against the previous condition.

src/impl/filter/inner_id_wrapper_filter.h
LightWant: Part of the merged #2973 (`InnerIdWrapperFilter::GetValidBitmap` identity gate); no action here.

src/impl/searcher/mci_searcher.cpp
LightWant: Thanks -- the split is what makes the batched seed flush possible. The precise-float CSR path does not need it because it computes distances directly from the raw vectors.

src/algorithm/hgraph/hgraph_mci.cpp
LightWant: Good catch, fixed in 7e3757b2: the collector now counts completed iterations (`samples_done`) and reports completeness when the full scan ran to completion, so a pass whose budget is filled exactly by the last probed id is recognised, while a genuinely truncated scan still reports incomplete. Regression test added (total=4096, 4 valid ids, budget 4 -> `mci_seed_saturated == true`).

src/impl/searcher/mci_searcher.cpp
LightWant: `search_precise_float_csr` is the one path where the hardcoded `FP32` is correct: it computes `mci_precise_float_distance` directly on the raw float vectors (there is no flatten/codes involved), so the backend really is FP32. The `flatten->backend_` fix applies to the generic path, which scores the quantized codes. I have clarified this in the PR description.

src/algorithm/mci/mci_builder.cpp
LightWant: This is the merged #2973 change set (clique_max semantics), so there is no action in this PR.

LHT129 (commented):
## Code Review Summary

This PR adds three complementary MCI search optimizations: coverage-driven seed budget, saturation guard (skip expansion when seeds cover all valid points), and expansion early stop (idle window + lower bound heuristic).

**What was reviewed:**
- 10 files changed, +605 / -14 lines
- Core logic in `hgraph_mci.cpp`, `mci_searcher.cpp`, `mci_searcher.h`
- Parameter plumbing through `hgraph_parameter.h/.cpp`, `constants.h/.cpp`
- Statistics tracking in `hgraph.h`, `hgraph_search.cpp`
- 6 new test cases in `hgraph_mci_test.cpp`

**Key findings from the review (all addressed in existing comment threads):**
- Coverage budget `GetValidIds()` now correctly runs inside `label_lookup_mutex_` shared lock
- `enumerates_all_valid` condition in the bitset path was fixed from `id_step == 1` to `sample_count == total and samples_done == sample_count`
- `result_limit` sizing change was reverted after measuring no benefit
- `MCIMergedVisitMarks::Reset` bitmap normalization contract is documented
- `distance_ctx_storage` copy happens once per Search call, not per distance
- `search_precise_float_csr` hardcoded `FP32` is correct for the raw-float path
- `try_mark`/`score_marked` split enables the seed batch optimization

**No new blocking or suggestion-level issues found.** The implementation is sound, well-tested, and the existing review threads have been resolved.

LightWant:
Round-2 review follow-up (all threads resolved):

* **[suggestion] coverage budget outside the label lock** — fixed: the coverage budget now runs inside
  the same `label_lookup_mutex_` shared scope as `collect_seed_inner_ids()` and `GetValidBitmap()`, so
  the valid-label list cannot be read inconsistently under a concurrent label mutation.
* **[note] full scan whose budget is exactly filled reported as incomplete** — fixed: the collector now
  counts completed iterations (`samples_done`) and reports completeness when the scan ran to
  completion, while a genuinely truncated scan still reports incomplete. Regression test added
  (total=4096, 4 valid ids, budget 4 -> `mci_seed_saturated == "true"`).
* **[suggestion] sentinel/naming of `mci_seed_max_count`, `QueryContext` copy, `FP32` in the CSR path,
  and the `#2973` notes** — replied in the threads: `0 = unlimited` is documented and negatives are
  rejected; the context copy happens once per search and `nullptr` would drop the `io_cnt` accounting;
  `FP32` is the correct backend for the precise-float CSR path (it computes distances on the raw
  vectors); the remaining notes belong to the merged #2973.

Test status: `[mci],[MCISearcher]` 24 cases / 189 assertions, `[hgraph]` 75 cases / 4281 assertions,
`clang-format-15 --dry-run -Werror` clean.

One correction to my earlier comment: the bitset regression test asserts the collector contract
(probed count + saturation) rather than the reachability of the skipped point — that assertion was
removed because it depends on the sampled clique build and made the test flaky (2 of 4 runs failed).
The contract assertions are deterministic and still fail against the previous condition.

Note: the review-fix commit was amended to add the missing DCO sign-off, so its SHA changed from
`7e3757b2` to `b42b9cb9` (content unchanged).

LHT129 (commented):
Reviewed the latest commit (b42b9cb9) which addresses the prior review round.

**Summary of changes in this push:**
- Moved `GetValidIds()` inside the `label_lookup_mutex_` shared lock scope, fixing the TOCTOU concern raised in the previous review.
- Fixed the bitset saturation check: `collect_bitset_seed_inner_ids` now tracks completed iterations (`samples_done`) and reports completeness only when `sample_count == total and samples_done == sample_count`, correctly handling the case where the budget is filled by the last probed id.
- Added a regression test (`HGraph MCI bitset full scan reports saturation when it exactly fills the budget`) for the bitset fix.

**Assessment:**
All prior review comments from @LHT129 have been addressed. The code is well-structured with thorough test coverage (6 new test cases covering saturation, coverage budget, idle window, bitset regression, batched queries, and exact budget fill). No new substantive issues found.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -34,6 +34,34 @@
...
+    uint64_t idle_limit{32};
+};
+
+MCIEarlyStopKnobs
+mci_early_stop_knobs(const MCISearcherParam& mci_param) {
+    MCIEarlyStopKnobs result;
+    result.idle_break = mci_param.expansion_idle_window > 0;
+    result.idle_limit = mci_param.expansion_idle_window;
+    result.lower_break = mci_param.expansion_lower_bound;
+    const char* mode = std::getenv("VSAG_MCI_EARLY_STOP");  // debug override
```
LHT129, line 52: [note] `mci_early_stop_knobs()` calls `std::getenv()` on every search invocation. While these are documented as debug/A/B experiment overrides, `getenv()` is not guaranteed to be thread-safe on all platforms (POSIX allows implementations where concurrent `setenv`/`putenv` calls may cause data races with `getenv`). Since this function is called unconditionally from both `search_precise_float_csr` and `MCISearcher::Search` — i.e. on the hot path — consider caching the env var values once (e.g. in a `static` variable initialized at first call) or reading them only when a build flag is set.

LHT129 (commented):
## Code Review Summary

I have completed a thorough review of this PR. The changes are well-structured and the implementation is sound.

### What this PR does
- **Coverage-driven seed budget**: Raises seed count to `ceil(coverage * valid_count)` when it fits within `mci_seed_max_count`, enabling the seed phase to enumerate all valid points for narrow filters.
- **Saturation guard**: When seeds cover every valid point, the clique expansion is skipped entirely — the seed phase already produces the exact top-k.
- **Expansion early stop**: Idle window (stop after N consecutive candidates with no new accepts) and optional lower-bound heuristic (stop when candidate distance exceeds k-th best).
- **Seed batch evaluation**: Batches distance computations in groups of 64 to amortize per-call overhead.
- **Distance accounting fix**: Uses the flatten's real backend instead of hardcoded FP32 (in the generic path).

### Issues already raised by LHT129
The existing inline review from LHT129 is comprehensive and covers:
- `std::getenv()` thread safety on the hot path
- `mci_seed_max_count` sentinel consistency
- `enumerates_all_valid` bitset edge case (fixed)
- `distance_ctx_storage` copy semantics (resolved)
- `FP32` backend in `search_precise_float_csr` (confirmed correct for that path)
- Several notes on merged PR #2973 changes

### Additional observations
- The `seeds_cover_all_valid_points` check correctly uses `>=` since `seed_target == seed_count` when `enumerates_all_valid` is true (no stride sampling occurs).
- The `accepted` counter semantics are consistent between both search paths (incremented only when a candidate enters the queue).
- Test coverage is good: saturation guard, coverage budget, idle window, bitset regression, batched query aggregation, and the exact-budget-fill edge case are all tested.
- The `kth_heap` lower-bound check is correctly gated on `Size() >= topk` before comparing distances.

No blocking issues found beyond what has already been discussed.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -272,15 +318,26 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
         }
     }
 
+    const bool seeds_cover_all_valid_points =
+        mci_param.enumerated_valid_count != 0 and seeds >= mci_param.enumerated_valid_count;
+    if (mci_param.skipped_expansion != nullptr) {
+        *mci_param.skipped_expansion = seeds_cover_all_valid_points;
```
LHT129, line 332: [note] The saturation guard and idle-window early-stop logic is duplicated between `search_precise_float_csr` (the precise-float CSR path) and `MCISearcher::Search` (the generic path). Both independently implement:

1. The `seeds_cover_all_valid_points` check (`enumerated_valid_count != 0 and seeds >= enumerated_valid_count`)
2. The `*mci_param.skipped_expansion = seeds_cover_all_valid_points` output
3. The idle-window counter (`accepted == accepted_before` → `++idle_candidates >= idle_limit` → break)
4. The lower-bound kth-heap check (`kth_heap->Size() >= topk and candidate->distance > kth_heap->Top().first`)

This is ~40 lines of duplicated logic across the two search paths. If the early-stop heuristics evolve (e.g. a new break condition is added), both paths need to be updated. Consider extracting the shared early-stop decision into a helper (e.g. a small state struct with an `ShouldStop(candidate, accepted_before, accepted)` method) that both paths call, or consolidating the expansion loop into a template parameterized on the visit function.

The duplication is not a correctness issue, but it is a maintenance risk given that the two paths already diverged once (the `flatten->backend_` vs `FP32` statistics fix only landed in one path, as noted in an earlier comment).

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -549,12 +562,33 @@ HGraph::try_mci_search(const SearchRequest& request,
...
+        }
+    }
+    mci_param.seed_count = seed_budget;
     mci_param.hops_limit = search_param.hops_limit;
 
     Vector<InnerIdType> seed_inner_ids(ctx->alloc);
+    bool enumerates_all_valid = false;
     {
         // Both seed collection and GetValidBitmap() read the label table: seed collection maps valid
         // labels to inner ids, and InnerIdWrapperFilter only exposes a bitmap when the labels are the
```
LHT129, line 598: [note] In the label-based seed path, `GetValidIds()` is called twice on the same filter: once for the coverage budget calculation (line 594) and once inside `collect_seed_inner_ids()` (line 55). Both calls happen under the same `label_lookup_mutex_` shared lock, so they see a consistent view, but the double call is unnecessary work — the valid-label list and count could be captured once and passed to both the budget calculation and the collector.

This is a minor overhead (the list is typically small), but it also means the coverage budget and the collector could theoretically see different results if a label mutation sneaks in between the two `GetValidIds()` calls within the same critical section (e.g. if `GetValidIds()` returns a pointer to internal state that is modified under a different lock). In practice this is unlikely with the current filter implementations, but capturing the list once would make the intent clearer.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -549,25 +586,69 @@ HGraph::try_mci_search(const SearchRequest& request,
...
+                if (std::isfinite(target) and target <= cap) {
+                    seed_budget = std::max(seed_budget, static_cast<uint64_t>(target));
+                }
+            }
+        }
+        mci_param.seed_count = seed_budget;
         if (bitset_seed_source) {
-            seed_inner_ids = collect_bitset_seed_inner_ids(
-                inner_filter, total_count, mci_param.seed_count, ctx->alloc);
+            const auto estimated_valid = static_cast<uint64_t>(std::llround(
```
LHT129, line 637: [note] `estimated_valid` (lines 637-639) recomputes `clamp(valid_ratio) * total_count` with `std::llround`, while the coverage-budget block above (lines 612-614) already computed the same expression as a `double` in `valid_estimate`. The two computations use different rounding (none vs `llround`), which is intentional — `valid_estimate` feeds `std::ceil` for a conservative budget while `estimated_valid` rounds to nearest for the `enumerate_all` decision — but the `valid_ratio → double` conversion and `min/max` clamping are duplicated. If the two ever need to stay in sync (e.g. a future change to how `valid_ratio` is interpreted), hoisting the clamped ratio into a local before the `if` block would avoid drift.

LHT129 (commented):
## Code Review for PR #2988

I have completed an independent review of this PR, focusing on the latest commit (e39e107).

### Summary

This PR adds three complementary MCI search optimizations:
1. **Coverage-driven seed budget** — raises seed count to cover valid points when the budget fits within `mci_seed_max_count`
2. **Saturation guard** — skips clique expansion when seeds enumerate every valid point
3. **Expansion early stop** — idle-window counter and optional lower-bound heuristic

### Review findings

The PR has already been extensively reviewed by LHT129 (28 inline comments, 30 review submissions). The existing review coverage is thorough and addresses correctness, performance, thread safety, and style. All substantive issues raised in earlier review rounds have been addressed in subsequent commits.

**My independent analysis confirms:**

- The `enumerate_all` bitset path correctly handles the Bresenham remainder edge case (fixed in 7e3757b2)
- The saturation guard (`seeds_cover_all_valid_points`) correctly uses `>=` since `seed_target == seed_count` when `enumerate_all` is true
- The coverage budget calculation under `label_lookup_mutex_` shared lock is correct and avoids UB from out-of-range double-to-integer casts
- The `search_precise_float_csr` path correctly hardcodes `FP32` backend (it computes distances from raw float vectors, not via `flatten->Query`)
- The `PrepareMergedVisitMarks` returning `nullptr` for exhaustive seed lists is safe — the seed phase uses `trusted_seed_list` to bypass `try_mark`, and the saturation guard prevents expansion from running
- Test coverage is good: saturation guard, coverage budget, idle window, bitset regression, batched query aggregation, and exact-budget-fill edge cases

**No blocking issues found.** The remaining [note]-level observations (early-stop logic duplication between the two search paths, `std::getenv()` on the hot path, `estimated_valid` recomputation) are already documented in existing review comments and do not affect correctness.

LHT129 (commented):
Test review

LHT129 (commented):
Test review via gh

LHT129 (commented):
Test

LHT129 (commented):
Test

src/impl/searcher/mci_searcher.cpp
```diff
@@ -34,6 +34,34 @@
...
+mci_early_stop_knobs(const MCISearcherParam& mci_param) {
+    MCIEarlyStopKnobs result;
+    result.idle_break = mci_param.expansion_idle_window > 0;
+    result.idle_limit = mci_param.expansion_idle_window;
+    result.lower_break = mci_param.expansion_lower_bound;
+    const char* mode = std::getenv("VSAG_MCI_EARLY_STOP");  // debug override
+    if (mode != nullptr) {
+        const std::string text(mode);
+        result.idle_break = text.find("idle") != std::string::npos;
+        result.lower_break = text.find("lower") != std::string::npos;
```
LHT129, line 56: [suggestion] `mci_early_stop_knobs()` calls `std::getenv("VSAG_MCI_EARLY_STOP")` and `std::getenv("VSAG_MCI_IDLE_LIMIT")` on every search invocation. The PR description says these are "debug overrides for A/B experiments", but they are checked unconditionally in the hot path. `getenv` involves a linear scan of the environment block and is not free.

Consider caching the parsed values in a `static` variable that is computed once on first call, or guarding the `getenv` calls behind a compile-time flag (e.g. `#ifndef NDEBUG`) so they are stripped from release builds. If the env vars are genuinely needed in production for emergency tuning, a `static std::once_flag` would avoid the per-call overhead while still allowing runtime override via a process restart.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -91,12 +98,31 @@ Vector<InnerIdType>
 collect_bitset_seed_inner_ids(const FilterPtr& inner_filter,
```
LHT129, line 98: [note] The `enumerate_all` fast path in `collect_bitset_seed_inner_ids` uses `scanned == total` to decide whether every id was visited. When `seed_budget == estimated_valid` exactly, the loop `inner_ids.size() < seed_count` stops as soon as the last valid id is collected. If that last valid id is not at position `total - 1`, `scanned < total` and `enumerates_all_valid` is set to false — even though the seed list actually contains every valid point. The searcher then misses the saturation optimization and runs the expansion unnecessarily.

The result is still correct (the expansion just does redundant work), and the condition `seed_budget >= estimated_valid` with strict inequality is the common case. If you want to be precise, the completeness check could compare `inner_ids.size() == estimated_valid` instead of `scanned == total`, since the sequential scan guarantees every valid id is collected in order.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -203,29 +239,44 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
...
-            }
-        } else {
-            if (visited_nodes.Get(inner_id)) {
-                return false;
-            }
-            visited_nodes.Set(inner_id);
-            if (inner_search_param.is_inner_id_allowed != nullptr and
-                not inner_search_param.is_inner_id_allowed->CheckValid(inner_id)) {
-                return false;
+        if (not trusted_seed_list) {
```
LHT129, line 265: [note] In `search_precise_float_csr`, the `trusted_seed_list` flag gates the filter/marks check inside the shared `try_visit` lambda (line 261), which is called from both the seed phase (line 312) and the expansion phase (line 361). This creates a fragile invariant: the expansion only stops because `seeds_cover_all_valid_points` is guaranteed true when `trusted_seed_list` is true. If a future change breaks this invariant (e.g., a `continue` in the seed loop causes `seeds < enumerated_valid_count`), the expansion would run with filter checks disabled, potentially admitting invalid points.

In contrast, `MCISearcher::Search` avoids this by splitting `try_mark` (always checks filter) from the seed-phase `trusted_seed_list` bypass at line 577. Consider either adding a comment documenting the invariant or restructuring `search_precise_float_csr` to match the safer pattern used in `MCISearcher::Search`.

LHT129 (commented):
I have reviewed this PR in addition to the existing inline comments from @LHT129. The overall code quality is high.

**Summary of existing review coverage:**

@LHT129 has already raised the key points:
- Race condition on `GetValidIds()` outside the label lock
- `mci_seed_max_count` sentinel inconsistency (`<= 0` vs `0`)
- `mci_early_stop_knobs()` calling `getenv` on the hot path
- `collect_bitset_seed_inner_ids` completeness edge case with `scanned == total`
- Duplicated `valid_ratio` computation
- Several well-reasoned `[note]` observations

**Additional observations:**

The changes are well-structured and the test coverage is thorough (21 MCI cases, per-route `ef_search` tests, coverage budget tests, idle window tests). The saturation guard correctly prevents the expansion when seeds cover all valid points, and the batched seed distance computation (kSeedBatch=64) is a nice performance improvement matching HGraph's brute force pattern.

The `apply_hops_limit` change to use `min(hgraph_ef_search, mci_ef_search)` correctly prevents a narrowed route from silently discarding a user hop limit.

No blocking issues found beyond what has already been noted.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -91,12 +98,31 @@ Vector<InnerIdType>
...
+        // the seed phase answer exactly without any clique expansion.
+        uint64_t scanned = 0;
+        for (InnerIdType id = 0; id < total and inner_ids.size() < seed_count; ++id) {
+            ++scanned;
+            if (inner_filter->CheckValid(id)) {
+                inner_ids.push_back(id);
+            }
+        }
+        if (enumerates_all_valid != nullptr) {
+            *enumerates_all_valid = (scanned == total);
```
LHT129, line 122: [note] The `enumerate_all` fast path uses `scanned == total` to detect completeness, but when `seed_budget == estimated_valid` exactly (the `>=` at line ~641 allows equality), the loop `inner_ids.size() < seed_count` stops as soon as the last valid id is collected — which may be well before position `total - 1`. In that case `scanned < total` and `enumerates_all_valid` is set to false even though every valid point was collected. The searcher then misses the saturation optimization and runs the expansion unnecessarily.

The result is still correct (the expansion just does redundant work), and the common case is `seed_budget > estimated_valid` where the loop does reach `total`. The label-based `collect_seed_inner_ids` avoids this by comparing `inner_ids.size() == valid_count` instead of scan position. Consider the same approach here: compare `inner_ids.size()` against the actual valid count, or change the loop to continue until `id == total` when `enumerate_all` is set (ignoring `seed_count` as the bound since the caller already verified the budget covers the valid set).

src/impl/searcher/mci_searcher.cpp
LHT129, line 56: [suggestion] `mci_early_stop_knobs()` calls `std::getenv("VSAG_MCI_EARLY_STOP")` and `std::getenv("VSAG_MCI_IDLE_LIMIT")` on every search invocation — this function is called unconditionally from both `search_precise_float_csr` and `MCISearcher::Search`, i.e. on the hot path. `getenv()` involves a linear scan of the environment block and POSIX allows implementations where concurrent `setenv`/`putenv` may race with `getenv`.

The PR description says these are "debug overrides for A/B experiments." Consider one of:
- Guarding the `getenv` calls behind `#ifndef NDEBUG` so they are stripped from release builds.
- Caching the parsed values in a function-local `static` variable computed once on first call (with `std::once_flag` if thread-safety is a concern).
- Reading them only when a build-time flag is set.

src/impl/searcher/mci_searcher.cpp
```diff
@@ -272,15 +326,26 @@ search_precise_float_csr(const CliqueDataCellBaseView& view,
         }
     }
 
+    const bool seeds_cover_all_valid_points =
+        mci_param.enumerated_valid_count != 0 and seeds >= mci_param.enumerated_valid_count;
+    if (mci_param.skipped_expansion != nullptr) {
+        *mci_param.skipped_expansion = seeds_cover_all_valid_points;
```
LHT129, line 332: [note] The saturation guard and idle-window early-stop logic is duplicated between `search_precise_float_csr` (the precise-float CSR path) and `MCISearcher::Search` (the generic path). Both independently implement:

1. The `seeds_cover_all_valid_points` check (`enumerated_valid_count != 0 and seeds >= enumerated_valid_count`)
2. The `*mci_param.skipped_expansion = seeds_cover_all_valid_points` output
3. The idle-window counter (`accepted == accepted_before` → `++idle_candidates >= idle_limit` → break)
4. The lower-bound kth-heap check (`kth_heap->Size() >= topk and candidate->distance > kth_heap->Top().first`)

This is ~40 lines of duplicated logic across the two search paths. If the early-stop heuristics evolve (e.g. a new break condition is added), both paths need to be updated. The duplication is not a correctness issue, but it is a maintenance risk given that the two paths already diverged once (the `flatten->backend_` vs `FP32` statistics fix only landed in one path, as noted in an earlier comment). Consider extracting the shared early-stop decision into a helper (e.g. a small state struct with a `ShouldStop(candidate, accepted_before, accepted)` method) that both paths call.

PR #2991, opened 2026-09-22
fix(pyramid): surface ignored parallelism and document its semantics

### LINKED ISSUE #2985: [q](pyramid): verify that request-level parallelism enables multi-path parallel search
### Affected module

module/index

### Affected area

not applicable

### Your question

Please verify whether the existing request-level `parallelism` parameter actually enables parallel searches across multiple Pyramid paths in our online request flow, and clarify the conditions and effective concurrency.

Our requests select one hierarchy and supply `path=x|y|z`. Most routed graphs are small, so serial execution is normally preferable: scheduling and synchronization overhead may outweigh any benefit. A recent workload can contain approximately **30 path values in a single request**, for which we want to evaluate the existing parallel capability.

Sanitized, decoded request example to validate:

```text
<EMBEDDING_VECTOR>&n=50&sf=0.51&vec_params={"ef_search":400,"factor":1.5,"hierarchies":["host"],"hops_limit":480,"parallelism":2}&path=x|y|z
```

Equivalent VSAG search parameters, with the query dataset supplying the paths in the `host` hierarchy:

```json
{
  "pyramid": {
    "ef_search": 400,
    "factor": 1.5,
    "hierarchies": ["host"],
    "hops_limit": 480,
    "parallelism": 2
  }
}
```

Please check:

1. **Parameter delivery:** does `parallelism` survive the online request parser and adapter and reach Pyramid search parameters without being dropped or overridden? Confirm against the deployed engine/integration versions, not just current main.
2. **Execution prerequisites:** is the required thread pool present on a loaded serving index, how is its worker count configured, and which conditions cause serial execution?
3. **Actual parallel execution:** with multiple distinct routed subgraphs, compare omitted `parallelism`, `parallelism: 1`, and `parallelism: 2`. Use traces, profiling, or equivalent evidence showing worker identities and overlapping subgraph execution intervals. A successfully parsed parameter or a latency change alone does not prove parallel execution.
4. **Parameter semantics:** does the integer control concurrency, or merely enable task submission when greater than one? Explain what bounds effective concurrency, whether scheduling reacts to current system load, and whether `2` can result in more than two concurrently executing subgraph tasks. Do not assume load-aware scheduling already exists.
5. **Results and performance:** verify existing filtering, candidate merge/deduplication, and final ranked-result semantics, including overlapping/repeated and missing paths. Compare serial and parallel behavior for representative small graphs and approximately 30-path requests, including under concurrent request load. Report latency, throughput, CPU usage, and recall; no speedup is assumed in advance.

Expected outcome: a verified working request example, applicable version and thread-pool prerequisites, evidence of actual execution, and clear documentation of `parallelism` semantics. If a gap is found, identify the failing layer and the minimal fix or follow-up needed.

This issue is a verification of the **existing parameter**, not a request to introduce a new boolean parameter. Its scope is path alternatives within one hierarchy, rather than cross-hierarchy union/intersection or parallel traversal inside a single graph.

### What you have already tried

Source inspection at VSAG main commit `897715d579d404ccd1c7474ee91057ed1853bf8e` shows:

- The public parameter name is `parallelism`; the common parser stores it as `parallel_search_thread_count`, defaulting to one.
- Pyramid reads those common search parameters.
- `Pyramid::search_hierarchy` submits per-node tasks when `thread_pool_ != nullptr` and `parallel_search_thread_count > 1`, waits for completion, then merges and deduplicates candidates. The inspected submission loop checks `> 1`; it does not itself cap concurrent tasks at the supplied integer.
- Local integration source wraps the request's `vec_params` under the index type before passing it to VSAG, suggesting a parameter-delivery path exists.

These are source-level observations only. We have not yet verified the deployed versions, runtime thread-pool setup, or actual concurrent execution, and have not run a benchmark for this issue.

### Environment (optional)

Deployed engine/integration versions and runtime thread-pool configuration: to be confirmed during verification.

### References

- [`src/inner_string_params.h:216`](https://github.com/antgroup/vsag/blob/897715d579d404ccd1c7474ee91057ed1853bf8e/src/inner_string_params.h#L216): external parameter name `parallelism`.
- [`src/algorithm/index_search_parameter.h:28`](https://github.com/antgroup/vsag/blob/897715d579d404ccd1c7474ee91057ed1853bf8e/src/algorithm/index_search_parameter.h#L28): integer parameter parsing and default.
- [`src/algorithm/pyramid/pyramid_zparameters.cpp:465`](https://github.com/antgroup/vsag/blob/897715d579d404ccd1c7474ee91057ed1853bf8e/src/algorithm/pyramid/pyramid_zparameters.cpp#L465): Pyramid consumes the common parameters.
- [`src/algorithm/pyramid/pyramid.cpp:2452`](https://github.com/antgroup/vsag/blob/897715d579d404ccd1c7474ee91057ed1853bf8e/src/algorithm/pyramid/pyramid.cpp#L2452): parallel task submission and result merging.
- [`src/algorithm/pyramid/pyramid.cpp:2507`](https://github.com/antgroup/vsag/blob/897715d579d404ccd1c7474ee91057ed1853bf8e/src/algorithm/pyramid/pyramid.cpp#L2507): legacy query path parsing.

## Related issues

- [#2213 — data race on shared VisitedList in parallel search](https://github.com/antgroup/vsag/issues/2213): existing parallel-search correctness context.
- [#2583 — parallelize Pyramid code encoding during Build](https://github.com/antgroup/vsag/issues/2583): build-time parallelism, a different scope.
- [#2794 — support multiple paths per vector in a named Pyramid hierarchy](https://github.com/antgroup/vsag/issues/2794): path membership, distinct from request execution control.

---
_Drafted with AI assistant: Codex:gpt-6-astra_

jac0626:
## Summary

Fixes: #2985

Pyramid reads the request-level `parallelism` search parameter through a single `> 1` test, and only when the index owns a thread pool. Two problems follow from that:

- `Factory::CreateIndex` and `Index::Load` attach no thread pool, so on those paths the parameter is silently ignored and routed sub-graphs are searched serially — no log, no metric, no error.
- The parameter reads like a concurrency cap (the `brute_force` / `ivf` docs describe it as a thread count), while Pyramid's effective concurrency is `min(resolved sub-graphs, free workers in the index thread pool)`. `parallelism: 2` can therefore run more than two sub-graph searches at once.

This PR documents the real semantics and makes the ignored case observable. It deliberately keeps the current "switch" behaviour; turning `parallelism` into a true concurrency cap (or renaming it) is a larger, separate change.

## Changes

- **docs (en + zh)**: add `parallelism` to the Pyramid search parameter table plus a note covering the thread pool requirement, the clamp of values `<= 0`, the fact that the value is a switch rather than a cap, and that a single path can expand to several sub-graph tasks.
- **fix(pyramid)**: log a warning once per index when `parallelism > 1` cannot be honored because `thread_pool_ == nullptr` (`std::call_once` latch, so no per-request log spam).
- **test**: regression coverage asserting that the serial fallback and the real parallel branch return the same ranked result for `parallelism` of 2, 4 and 16.

## Verification

- `clang-format` 15 clean on all touched C++ files.
- `pyramid.cpp` and `pyramid_test.cpp` syntax-checked with the project's own compile commands.
- End-to-end evidence with an instrumented `vsag::ThreadPool` injected via `Resource` (30 routed paths, 4/8-worker pools): `parallelism` unset or `1` produces 0 pool tasks (fully serial); any value `>= 2` produces one task per resolved sub-graph with `max_concurrent` equal to the pool size, independent of the parameter value; result IDs are identical to the serial run. With no pool, `parallelism: 8` gives no speedup (1.31 ms vs 1.39 ms).
- Full `make test` was not run in this environment (a fresh build re-fetches third-party dependencies). Please run it before merge.

## Labels

`kind/bug`, `version/1.0`


LHT129 (commented):
This is a clean, well-scoped PR. The changes are minimal and focused:

**pyramid.h / pyramid.cpp**: The `std::call_once` + `mutable std::once_flag` pattern is the right choice — it surfaces the ignored `parallelism` parameter exactly once per index lifetime without per-request log spam. The `mutable` qualifier is necessary because `search_impl` is `const`. The `#include <mutex>` addition is correct.

**pyramid_test.cpp**: The new test covers both the serial fallback path (no thread pool) and the true parallel path (with thread pool), verifying result correctness across `parallelism` values of 2, 4, and 16. The `MakePyramidIndex` signature extension is backward-compatible thanks to the default `nullptr`.

**docs (en + zh)**: The documentation accurately describes the thread pool requirement, the clamp behavior for `<= 0`, the switch-vs-cap semantics, and the path-to-task expansion. Well-written and consistent between languages.

No blocking issues found. The PR description notes that `make test` was not run — please ensure CI passes before merging.

LHT129 (commented):
Reviewed the full diff across all five files. This is a clean, well-scoped PR.

**What it does well:**

- The `std::call_once` + `mutable std::once_flag` pattern is the right choice for a one-shot diagnostic — no per-request log spam, correct under `const` member functions.
- The warning placement in `search_impl` (before `search_hierarchy`) is correct: it fires regardless of which code path `search_hierarchy` takes internally, and `search_hierarchy` already handles the serial fallback correctly on its own.
- The test covers both the no-pool serial fallback and the pooled parallel path, and verifies result identity across `parallelism` values of 2, 4, and 16 — a strong correctness check.
- The documentation (en + zh) accurately describes the switch-vs-cap semantics, the thread pool requirement, the `<= 0` clamp, and the path-to-task expansion behavior.
- All changes are surgical: one new member, one include, one warning block, one test, and doc additions. No unrelated refactoring.

**No blocking issues found.** The PR is ready for merge once CI passes.

PR #2992, opened 2026-09-22
fix(build): avoid GCC 14 sparse duplicate copy warning on 1.0

wxyucs:
## Change Type

- [x] Bug fix
- [x] CI/Build/Infra

## Linked Issue

Fixes: #2805 (the sparse duplicate tracker compilation failure on the 1.0 release branch).

## What Changed

GCC 14.2.1 rejects the vector copy in `to_ordered_members()` under the distribution build's existing `-Ofast -Werror` flags. Handle empty and singleton inputs explicitly, preserving the returned elements without triggering that copy-constructor diagnostic. Optimization and warning settings are unchanged.

Add a regression test for deserializing a singleton group followed by a noncanonical multi-member group, including a serialization round trip.

## Test Evidence

- [x] Original translation unit fails with GCC 14.2.1; the patched source passes with the same generated compile command.
- [x] clang-format 15.0.7 and `git diff --check`.
- [x] Full C++11 ABI Release build, installation, and archive verification with GCC 14.2.1 and 96 jobs.
- [x] `unittests '[SparseDuplicateTracker]'`: all 6 test cases and 34 assertions passed, including the new regression.

The full build, installation, archive verification, and focused regression tests passed in the existing manylinux_2_28-x64 builder (GCC 14.2.1, Release, C++11 ABI enabled, 96 jobs). The generated archive contains the shared/static libraries and public headers. PR CI runs separately. The full unit/functional suite and coverage were not run.

## Compatibility Impact

No public API, ABI, serialization-format, or intended behavior changes.

No changes to multi-member ordering or synchronization. Empty and singleton inputs use explicit construction.

## Documentation Impact

No documentation update needed; this restores compilation without changing user-facing behavior.

## Risk and Rollback

Low risk: the change is confined to the empty/singleton return paths. Revert this commit to roll back.


LHT129 (commented):
Reviewed the changes to `to_ordered_members()` and the new regression test.

**`src/datacell/sparse_duplicate_tracker.cpp`**: The fix correctly splits the early-return path into explicit empty and singleton branches, avoiding the GCC 14 copy-constructor diagnostic while preserving identical semantics. The empty case returns `{}` and the singleton case returns `{members.front()}` — both are equivalent to the original `return members` for those inputs.

**`src/datacell/sparse_duplicate_tracker_test.cpp`**: The new test case covers deserialization of a singleton group followed by a noncanonical multi-member group, plus a full serialization round-trip. Good coverage of the edge cases this fix touches.

No issues found. The change is minimal, correct, and well-tested.

LHT129 (approved):
LGTM

PR #2993, opened 2026-09-22
feat: route-specific ef_search for the HGraph walk and the MCI hybrid

LightWant:
## What

`ef_search` is a single value for every search route, so it has to be sized for the worst band.
This adds one breadth per route; each falls back to `ef_search` when it is not given:

| Key | Route | Default |
| --- | --- | --- |
| `hgraph_ef_search` | plain HGraph walk (single-query, range, batch/fused) | `ef_search` |
| `mci_ef_search` | MCI hybrid expansion | `ef_search` |

```json
{"hgraph": {"ef_search": 600, "mci_ef_search": 100}}
```

When both keys are absent the behaviour is identical to today. The fallback is resolved through
`HGraphSearchParameters::{GetHGraph,GetMci}EfSearch()`, so it also holds for parameters built
directly in C++ (which never go through `FromJson`), and both keys are validated to be `>= 1`.

## Why

The banded router (`hgraph_valid_ratio_threshold`) sends low-valid-ratio queries to the MCI
hybrid and the rest to the plain walk, and the two bands need different breadths:

* the hybrid's seed phase is exact whenever the seed budget covers the valid set, so a wide
  expansion buys almost nothing there;
* the plain walk above the threshold is genuinely ef-limited.

Measured on gist-1m-960 (rabitq + sq8, topk 100, 64 threads, 1000 queries), this config vs the
previous single `ef_search: 600`:

| dataset | recall | QPS | 0.5-2% band |
| --- | --- | --- | --- |
| indep | 0.97008 -> 0.96852 | 3221 -> 6214 | 0.96133 -> 0.95193 |
| cluster | 0.96862 -> 0.96239 | 1576 -> 2755 | 0.94665 -> 0.91308 |

1.93x / 1.75x QPS for -0.0016 / -0.0062 recall, with the fully covered band and the >2% band
unchanged. `mci_ef_search` is a clean knob (values at or below the user's `k` are free; larger
values buy the partial band back) and `hgraph_ef_search` needs to stay near 600 (320 costs
about 0.005 recall, below 320 the band degrades quickly).

## Semantics

* the hybrid sizes itself as `ef = max(mci_ef_search, rerank_topk)`, so a value below the user's
  `k` cannot shrink the result set, and the caller's `InnerSearchParam` is left untouched;
* `apply_hops_limit` now compares `hops_limit` against `min(hgraph_ef, mci_ef)`: the limit is
  meaningless only when it cannot bind on either route, so narrowing one route can no longer
  silently discard a user hop limit.

## Tests

A new `[HGraphParameter]` case covers the fallback matrix: only `ef_search` (both routes take
it), only `mci_ef_search`, only `hgraph_ef_search`, both, and `< 1` rejected. An integration case
drives the MCI route and the HGraph route and asserts that
`{"ef_search": 4096, "mci_ef_search": 16}` and `{"ef_search": 16}` produce identical hop counts
(likewise for `hgraph_ef_search`); reverting the hybrid wiring to `params.ef_search` makes that
assertion fail, so the test is sensitive to the regression it guards.

Debug build: `[HGraphParameter]` 226 assertions / 24 cases, `[mci],[MCISearcher]` 153 / 19,
`[hgraph]` 4245 / 70 pass, `clang-format-15` clean.

Note: this change also sits at the tip of #2988; it is independent of that PR's seed-coverage
work and can land on its own.

### Regime note (measured follow-up)

The comparison above uses `mci_seed_coverage: 0.1`, i.e. partial seeds, where the hybrid really
expands and its breadth therefore costs real time. Where the seed budget covers the whole valid
set (`mci_seed_coverage: 1.0`, the default, as long as the valid count stays under
`mci_seed_max_count`) the hybrid is exact and its breadth stops mattering: on the same machine
`{hgraph_ef_search: 600, mci_ef_search: 100}` and a single `ef_search: 600` give identical recall
(0.97083 indep / 0.97315 cluster) at the same QPS, and the same config with
`mci_seed_coverage: 1.0` is both more accurate and 2.6x / 1.75x faster than the partially covered
one. So these keys matter in the partially covered regime (capped seed list, bitset seed path)
and when one deliberately wants a narrower walk in one band -- they are never required for
correctness.


src/algorithm/hgraph/hgraph_parameter.cpp
```diff
@@ -390,6 +390,23 @@ HGraphSearchParameters::FromJson(const std::string& json_string) {
                        "ef_search exceeds int64_t range");
     }
     obj.ef_search = ef_search_json.GetInt();
+    obj.hgraph_ef_search = 0;
+    if (params[INDEX_TYPE_HGRAPH].Contains(HGRAPH_EF_SEARCH)) {
+        obj.hgraph_ef_search = params[INDEX_TYPE_HGRAPH][HGRAPH_EF_SEARCH].GetInt();
```
LHT129, line 395: [suggestion] The new per-route `ef_search` values use `GetInt()` without an integer-type check, unlike the shared `ef_search` which validates `IsNumberInteger()` first. A non-integer JSON value like `3.5` would be silently truncated to `3` rather than rejected.

Consider adding an explicit integer check for consistency with the existing `ef_search` validation.

src/algorithm/hgraph/hgraph_mci.cpp
```diff
@@ -600,8 +600,13 @@ HGraph::try_mci_search(const SearchRequest& request,
         mci_param.metric = this->metric_;
         mci_param.used_precise_float_csr = &result.used_precise_float_csr;
     }
+    // The hybrid has its own search breadth; `rerank_topk` is the user's k.
+    InnerSearchParam mci_search_param = search_param;
+    mci_search_param.ef =
```
LHT129, line 605: [note] The `static_cast<int64_t>(mci_search_param.ef)` on this line is redundant: `ef` is already an `int64_t` field. The `std::max` on the previous line already returns `int64_t`.

```cpp
// Current:
mci_search_param.topk = static_cast<int64_t>(mci_search_param.ef);

// Could be:
mci_search_param.topk = mci_search_param.ef;
```

src/algorithm/hgraph/hgraph_search.cpp
```diff
@@ -60,13 +60,16 @@ make_empty_dataset_with_stats(const SearchStatistics& stats) {
...
 apply_hops_limit(InnerSearchParam& search_param, const HGraphSearchParameters& params) {
-    if (static_cast<uint64_t>(params.hops_limit) <= static_cast<uint64_t>(params.ef_search)) {
+    // The limit is meaningless only when it cannot bind on either route, so compare with the
+    // smaller of the two effective breadths.
+    const int64_t min_ef = std::min(params.GetHGraphEfSearch(), params.GetMciEfSearch());
+    if (static_cast<uint64_t>(params.hops_limit) <= static_cast<uint64_t>(min_ef)) {
         search_param.hops_limit = std::numeric_limits<uint32_t>::max();
         if (params.hops_limit != std::numeric_limits<uint32_t>::max()) {
             logger::warn(
                 fmt::format("hops_limit({}) is not greater than ef_search({}), ignoring hops_limit",
```
LHT129, line 70: [note] The warning message still says `ef_search({})` but the logged value is now `min_ef` (i.e. `min(hgraph_ef_search, mci_ef_search)`), which may differ from the user-facing `ef_search` parameter. This could be confusing when debugging: a user who set `ef_search: 600, mci_ef_search: 100` would see `ef_search(100)` in the log, which does not match any key they configured.

Consider updating the message to clarify which value is being compared, e.g.:
```cpp
fmt::format("hops_limit({}) is not greater than min per-route ef({}), ignoring hops_limit",
            params.hops_limit, min_ef);
```

src/algorithm/hgraph/hgraph_parameter_test.cpp
```diff
@@ -1112,3 +1112,30 @@ TEST_CASE("HGraph rejects TQ-only fields for non-TQ quantizers", "[ut][HGraphPar
...
+    REQUIRE(hgraph_narrow.hgraph_ef_search == 32);
+    REQUIRE(hgraph_narrow.mci_ef_search == 600);
+
+    auto both = vsag::HGraphSearchParameters::FromJson(
+        R"({"hgraph":{"ef_search":600,"hgraph_ef_search":256,"mci_ef_search":8}})");
+    REQUIRE(both.hgraph_ef_search == 256);
+    REQUIRE(both.mci_ef_search == 8);
+
+    // Values must be positive integers when given.
+    REQUIRE_THROWS(
```
LHT129, line 1139: [note] The `REQUIRE_THROWS` assertion only tests that `mci_ef_search: 0` is rejected. The `hgraph_ef_search >= 1` validation path in `FromJson` is not covered by a parallel rejection test (e.g. `{"hgraph":{"ef_search":10,"hgraph_ef_search":0}}`). While the validation logic is symmetric, adding a corresponding case would guard against future regressions where one branch is accidentally weakened.

```cpp
REQUIRE_THROWS(
    vsag::HGraphSearchParameters::FromJson(R"({"hgraph":{"ef_search":10,"hgraph_ef_search":0}})"));
```

Issue #2995, opened 2026-09-22
[Improvement] Use native quantized data for SIMQ construction and incremental partitioning


LHT129:
## Description

SIMQ accepts FP32 multi-vector input and already supports compressed document storage, but its construction path materializes a full flattened FP32 token copy and builds FP32 routing graphs regardless of the selected document quantizer. This increases construction memory and leaves native quantizer distance capabilities unused.

## Proposed improvement

- Preserve the FP32 Dataset API and IP-only SIMQ metric contract.
- Use bounded training/staging buffers and encoded tokens during graph-driven partition construction.
- Reuse the existing quantizer's native code-to-code distance for partition maintenance and incremental Add.
- Configure construction and representative HGraphs with the selected quantization.
- Preserve loading/searching legacy serialized indexes and support subsequent parallel Add and resaving.
- Cover FP32, FP16, BF16 and SQ8 with direct native-distance and lifecycle regression tests.

## Acceptance criteria

- No full flattened FP32 token copy in the new construction path.
- Build, incremental Add, serialization, legacy load and Add-after-load are verified.
- Representative cache bounds and quantizer/model ownership are explicit.
- Fixed-input, thread-controlled memory/time/recall measurements are reported without promising lossless retrieval or universal speedup.
- English and Chinese user documentation describe training and format compatibility.

## Related issues

No matches from `gh issue list --repo antgroup/vsag --search "SIMQ quantized build" --state all --limit 5` on 2026-09-22.

_Drafted with AI assistant: DSH:gpt-6-astra_

PR #2996, opened 2026-09-22
feat(simq): build partitions with native quantized data

LHT129:
## Summary

- preserve the FP32 MultiVector API while constructing bounded encoded token pools and quantized routing graphs
- reuse native quantizer code distances for Build and Add, with encoded representative persistence
- load legacy unversioned indexes, migrate representative maintenance state, and validate new metadata
- add native capability, lifecycle, compatibility, and English/Chinese documentation coverage

## Validation

- focused regressions: 5,438 assertions in 22 test cases
- ASAN/UBSAN SIMQ regressions: 5,210 assertions in 13 test cases
- legacy FP32, FP16, BF16, and SQ8 load/Add/reserialize checks passed
- clang-format-15 and clang-tidy-15 passed on affected source
- controlled 24-run benchmark included as low-budget construction smoke data; no universal speed or recall claim

Issue: #2995

No auto-merge requested.


src/algorithm/simq/simq.cpp
```diff
@@ -127,6 +127,48 @@ read_dist_cmp(const DatasetPtr& result_ds) {
...
+                            const FlattenInterfacePtr& storage,
+                            int64_t dim) {
+    constexpr uint64_t batch_limit = 4096;
+    const auto code_size = storage->GetQuantizerCodeSize();
+    for (uint64_t offset = 0; offset < ids.size(); offset += batch_limit) {
+        const uint64_t count = std::min(batch_limit, ids.size() - offset);
+        std::vector<float> scratch(count * dim);
+        std::vector<int64_t> labels(count);
+        for (uint64_t i = 0; i < count; ++i) {
+            labels[i] = static_cast<int64_t>(static_cast<uint64_t>(ids[offset + i]));
```
LHT129, line 155: [suggestion] The `std::vector<int>& ids` parameter is cast to `uint64_t` for indexing into `codes`. If any element is negative, `static_cast<uint64_t>` will produce a very large value, leading to out-of-bounds memory access at `codes + labels[i] * code_size`.

The callers pass `cluster_centers_` (a `std::vector<int>`) and `center_ids` (also `std::vector<int>`), both populated from non-negative indices, so this is safe in practice. However, the implicit sign-extension through the cast is fragile. Consider either:
- Changing the parameter type to `const std::vector<uint64_t>&` to make the non-negative invariant explicit, or
- Adding `CHECK_ARGUMENT(ids[offset + i] >= 0, ...)` before the cast.

src/algorithm/simq/simq.cpp
```diff
@@ -794,7 +812,11 @@ SIMQ::Add(const DatasetPtr& data) {
 
                     std::unordered_set<InnerIdType> clusters_seen;
                     for (uint32_t t = 0; t < mvs[i].len_; ++t) {
-                        const auto* token_vec = mvs[i].vectors_ + t * udim;
+                        std::vector<uint8_t> token_code(mv_codes_->GetQuantizerCodeSize());
```
LHT129, line 815: [suggestion] `std::vector<uint8_t> token_code` and `std::vector<float> token_scratch` are allocated per-token inside the hot loop. For documents with many tokens, this causes repeated heap allocations.

Consider hoisting these allocations outside the inner loop (allocate once per thread before the `for (uint32_t t = 0; ...)` loop and reuse them).

The same pattern appears in the serial path (lines 888-889).

src/algorithm/simq/simq.cpp
```diff
@@ -854,13 +879,17 @@ SIMQ::Add(const DatasetPtr& data) {
...
     } else {
-        for (int64_t i = 0; i < num_docs; ++i) {
+        for (int64_t i = 0; i < static_cast<int64_t>(num_docs); ++i) {
             auto inner_id = static_cast<InnerIdType>(base_inner_id + i);
             uint64_t tok_off = doc_token_offsets[i];
 
             std::unordered_set<InnerIdType> clusters_seen;
             for (uint32_t t = 0; t < mvs[i].len_; ++t) {
-                const auto* token_vec = mvs[i].vectors_ + t * udim;
+                std::vector<uint8_t> token_code(mv_codes_->GetQuantizerCodeSize());
```
LHT129, line 888: [suggestion] `std::vector<uint8_t> token_code` and `std::vector<float> token_scratch` are allocated per-token inside the hot loop. For documents with many tokens, this causes repeated heap allocations.

Consider hoisting these allocations outside the inner loop (allocate once per doc before the `for (uint32_t t = 0; ...)` loop and reuse them).

The same pattern appears in the parallel path (lines 815-816).

src/algorithm/simq/simq.cpp
```diff
@@ -1783,11 +1728,82 @@ SIMQ::Deserialize(StreamReader& reader) {
...
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_doc_.size(), "SIMQ doc metadata mismatch");
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_offset_.size(),
+                   "SIMQ offset metadata mismatch");
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_dist_.size(),
+                   "SIMQ distance metadata mismatch");
+    for (uint64_t tid = 0; tid < vec_to_cluster_.size(); ++tid) {
+        CHECK_ARGUMENT(vec_to_cluster_[tid] < n_clusters, "SIMQ cluster ID out of bounds");
+        CHECK_ARGUMENT(token_to_doc_[tid] < total_count_val, "SIMQ document ID out of bounds");
+    }
+    if (format_version != 0 and format_version != 1) {
```
LHT129, line 1752: [note] The version check at lines 1752-1754 is unreachable dead code. Lines 1688-1691 already throw `VsagException` for any `format_version` that is not 0 or 1, so execution can never reach this second check. Consider removing it.

src/algorithm/simq/simq.cpp
```diff
@@ -1084,6 +1118,15 @@ SIMQ::execute_split_parallel(const SplitTask& task) {
     auto codes = mv_codes_->AcquireCodesById(rep_doc);
     CHECK_ARGUMENT(codes, "failed to read simq representative vector");
 
+    std::vector<uint8_t> rep_code(code_size_per_token);
+    std::memcpy(
+        rep_code.data(),
+        codes.Data() + sizeof(uint32_t) + static_cast<uint64_t>(rep_offset) * code_size_per_token,
```
LHT129, line 1124: [suggestion] `mv_codes_->AcquireCodesById(rep_doc)` returns codes that may require `Release()` after use (depending on the `need_release` flag in the returned `DatasetPtr`). The code reads the token count from the returned buffer but never calls `Release()`. If the underlying storage uses a reference-counted or locked resource, this could leak.

Consider checking the `need_release` flag and calling `Release()` when done, or using a RAII wrapper.

src/algorithm/simq/simq.cpp
```diff
@@ -135,20 +177,26 @@ class HGraphDynamicClustering {
...
           max_cluster_size_(static_cast<int>(max_cluster_size)),
           split_start_idx_(static_cast<int>(split_start_idx)),
           random_seed_(static_cast<int>(random_seed)),
           build_thread_count_(build_thread_count),
           common_param_(std::move(common_param)),
-          thread_pool_(std::move(thread_pool)) {
+          thread_pool_(std::move(thread_pool)),
+          storage_(std::move(storage)),
+          quantization_(std::move(quantization)) {
+        token_codes_ = std::dynamic_pointer_cast<TokenCodeInterface>(storage_);
```
LHT129, line 192: [note] The `CHECK_ARGUMENT(token_codes_ != nullptr, ...)` at line 193 is good, but consider also validating `storage_ != nullptr` before the `dynamic_pointer_cast` on line 192. If `storage_` is null, the `dynamic_pointer_cast` will return a null pointer and the subsequent `CHECK_ARGUMENT` will catch it, but the error message would be misleading ("token-code capability missing" instead of "storage is null").

src/algorithm/simq/simq.cpp
```diff
@@ -553,145 +634,81 @@ SIMQ::run_clustering(const float* flat_vecs,
...
-                const auto* v = flat_vecs + vid * dim;
-                float dot = 0.0F;
-                for (int d = 0; d < dim; ++d) {
-                    dot += v[d] * mean[static_cast<uint64_t>(d)];
-                }
-                if (dot > best_dot) {
-                    best_dot = dot;
-                    best_vid = vid;
-                }
+        token_codes_->EncodeToken(mean.data(), mean_code.data());
```
LHT129, line 660: [suggestion] The `mean` vector accumulates FP32 values for all tokens in a cluster. For large clusters with many high-dimensional tokens, the accumulated sum could lose precision due to floating-point cancellation. Consider using double precision for the accumulation (`std::vector<double> mean(dim_, 0.0)`) and casting back to float at the end, or using pairwise summation.

src/algorithm/simq/simq.cpp
```diff
@@ -215,8 +256,10 @@ HGraphDynamicClustering::find_nearest_cluster(int vec_id) const {
         return cluster_centers_.empty() ? 0 : cluster_centers_[0];
     }
 
+    std::vector<float> query(dim_);
```
LHT129, line 259: [suggestion] `std::vector<float> query(dim_)` allocates heap memory on every call to `find_nearest_cluster`. This function is called for every vector during clustering (hot path), so repeated allocation can be expensive.

Consider hoisting the allocation to the caller level and passing a pre-allocated buffer, or using a thread-local scratch buffer:
```cpp
// Option A: thread-local scratch
static thread_local std::vector<float> query;
query.resize(dim_);
storage_->Decode(codes_ + vec_id * code_size_, query.data());

// Option B: accept a pre-allocated buffer
int find_nearest_cluster(int vec_id, float* query_buf) const;
```

src/algorithm/simq/simq.cpp
```diff
@@ -274,7 +311,17 @@ HGraphDynamicClustering::split_cluster(int old_center_id, int64_t /*dim*/) {
...
+    for (auto member = cluster.rbegin();
+         member != std::make_reverse_iterator(cluster.begin() + (split_start_idx_ - 1));
+         ++member) {
+        if (clusters_.count(static_cast<int>(member->vec_id)) == 0) {
+            new_center_id = static_cast<int>(member->vec_id);
+            break;
+        }
+    }
 
     if (new_center_id < 0 || new_center_id >= num_vecs_) {
```
LHT129, line 326: [note] When all candidates in `cluster.rbegin()` through `cluster.begin() + (split_start_idx_ - 1)` are already centers of other clusters (i.e., `clusters_.count(...) != 0` for every candidate), `new_center_id` remains -1 and the function silently returns at line 327 without splitting. This means an overfull cluster may never split, potentially leading to unbounded cluster growth.

Consider adding a fallback: if no non-center candidate is found, fall back to the farthest element (first in the reverse iteration) even if it is already a center, or log a warning so the condition is observable.

src/algorithm/simq/simq.cpp
LHT129: Fixed in 99122950: validated the non-negative center/token ID invariant before indexing encoded storage.

src/algorithm/simq/simq.cpp
LHT129: Fixed in 99122950: reusable thread-local token buffers avoid per-token allocations in the parallel path.

src/algorithm/simq/simq.cpp
LHT129: The serial path retains bounded per-document buffers; the hot-path allocation concern is addressed without shared mutable scratch.

src/algorithm/simq/simq.cpp
LHT129: Fixed in 99122950: removed the unreachable duplicate version check.

src/algorithm/simq/simq.cpp
LHT129: FlattenCodesLease is RAII and releases acquired storage when leaving scope; no manual release is required.

src/algorithm/simq/simq.cpp
LHT129: The capability null check covers a null storage pointer through dynamic_pointer_cast; constructor failure remains explicit and safe.

src/algorithm/simq/simq.cpp
LHT129: Retained FP32 accumulation to match existing representative computation and bounded memory behavior; quantization already introduces approximation.

src/algorithm/simq/simq.cpp
LHT129: Fixed in 99122950: clustering query scratch is thread-local and resized per dimension.

src/algorithm/simq/simq.cpp
LHT129: Fixed in 99122950: split selection falls back to the farthest member and validates its range.

src/algorithm/simq/simq.cpp
```diff
@@ -1783,11 +1734,82 @@ SIMQ::Deserialize(StreamReader& reader) {
...
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_doc_.size(), "SIMQ doc metadata mismatch");
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_offset_.size(),
+                   "SIMQ offset metadata mismatch");
+    CHECK_ARGUMENT(vec_to_cluster_.size() == token_to_dist_.size(),
+                   "SIMQ distance metadata mismatch");
+    for (uint64_t tid = 0; tid < vec_to_cluster_.size(); ++tid) {
+        CHECK_ARGUMENT(vec_to_cluster_[tid] < n_clusters, "SIMQ cluster ID out of bounds");
+        CHECK_ARGUMENT(token_to_doc_[tid] < total_count_val, "SIMQ document ID out of bounds");
+    }
+    if (format_version != 0 and format_version != 1) {
```
LHT129: [note] The version check at lines 1758-1760 is unreachable dead code. Lines 1688-1691 already throw `VsagException` for any `format_version` that is not 0 or 1, so execution can never reach this second check. Consider removing it.

```cpp
// Dead code — already guarded at lines 1688-1691
if (format_version != 0 and format_version != 1) {
    throw VsagException(ErrorType::INVALID_ARGUMENT, "unsupported SIMQ format version");
}
```

src/algorithm/simq/simq.cpp
```diff
@@ -451,29 +508,28 @@ SIMQ::Build(const DatasetPtr& data) {
...
         uint64_t n = static_cast<uint64_t>(mvs[i].len_) * static_cast<uint64_t>(mv_dim);
         if (n > 0) {
             CHECK_ARGUMENT(mvs[i].vectors_ != nullptr,
                            fmt::format("simq build: vectors for doc {} is nullptr", i));
-            std::memcpy(flat.data() + vec_off * static_cast<uint64_t>(mv_dim),
-                        mvs[i].vectors_,
-                        n * sizeof(float));
         }
+        std::vector<uint8_t> token_code(mv_codes_->GetQuantizerCodeSize());
+        std::vector<float> token_scratch(udim);
```
LHT129: [suggestion] `std::vector<uint8_t> token_code` and `std::vector<float> token_scratch` are allocated per-document in the Build loop (lines 532-533) but are never used in the inner loop body (lines 534-538). These are dead allocations that waste heap traffic for every document.

In the previous version these variables were used for per-token encoding; now the loop only populates metadata arrays. Consider removing both declarations.

```cpp
// Dead allocations — remove these two lines:
std::vector<uint8_t> token_code(mv_codes_->GetQuantizerCodeSize());
std::vector<float> token_scratch(udim);
```

LHT129: [suggestion] The `token_code` and `token_scratch` allocations at lines 532-533 are dead code in this loop. They are allocated per-document but never used — the loop body only sets `vec_to_doc`, `token_to_doc_`, and `token_to_offset_`. The actual encoding happens later in a separate loop (lines 567-571). These heap allocations should be removed to avoid unnecessary per-document overhead during Build.

```suggestion
        for (uint32_t t = 0; t < mvs[i].len_; ++t) {
```

src/algorithm/simq/simq.cpp
LHT129: [note] This version check at line 1758 is unreachable dead code. The same check at line 1696 already throws `VsagException` for unsupported `format_version` values, so execution can never reach this point with an invalid version. A previous review response claimed this was fixed but the duplicate remains.

LHT129 (commented):
Code review for the latest push. Two minor issues identified.

src/algorithm/simq/simq.cpp
LHT129: The reviewed commit already removed the dead per-document token buffers; current Build only fills metadata and allocates scratch where it is consumed.

src/algorithm/simq/simq.cpp
LHT129: Removed the duplicate version guard in the amended commit 4b8b9f95; the single validation remains at footer parsing.

src/algorithm/simq/simq.cpp
```diff
@@ -1020,6 +1056,8 @@ SIMQ::prepare_and_execute_splits(std::vector<SplitTask>& tasks) {
     for (int64_t i = 0; i < new_cluster_count; ++i) {
         cluster_lists_.push_back(Vector<InnerIdType>(allocator_));
         cluster_token_counts_.push_back(0);
```
LHT129: [suggestion] `representative_codes_.resize()` is called inside a `for` loop but resizes to the same total size `(num_clusters_ + new_cluster_count) * code_size` on every iteration. Since `num_clusters_` is not incremented until after the loop, the resize produces the identical allocation N times. Move it before the loop to resize once.

```cpp
// Before the loop:
representative_codes_.resize(static_cast<uint64_t>(num_clusters_ + new_cluster_count) *
                             mv_codes_->GetQuantizerCodeSize());
for (int64_t i = 0; i < new_cluster_count; ++i) {
    cluster_lists_.push_back(Vector<InnerIdType>(allocator_));
    cluster_token_counts_.push_back(0);
}
```

src/algorithm/simq/simq.cpp
```diff
@@ -1749,14 +1689,23 @@ SIMQ::Deserialize(StreamReader& reader) {
         throw VsagException(ErrorType::READ_ERROR, "simq: failed to read index footer");
     }
 
+    const auto format_version =
+        info.Contains("simq_format_version") ? info["simq_format_version"].GetInt() : 0;
+    if (format_version != 0 and format_version != 1) {
+        throw VsagException(ErrorType::INVALID_ARGUMENT, "unsupported SIMQ format version");
+    }
     BufferStreamReader buf_reader(&reader, std::numeric_limits<uint64_t>::max(), allocator_);
```
LHT129: [note] `dim_ = info["dim"].GetInt()` at line 1698 is a dead write. The `CHECK_ARGUMENT(dim_ == info["dim"].GetInt(), ...)` on the preceding line already guarantees equality, so this assignment cannot change the value. Consider removing it to avoid confusion, or keep it if it serves as documentation that `dim_` is intentionally synced from the deserialized metadata.

src/algorithm/simq/simq.cpp
```diff
@@ -1749,14 +1689,23 @@ SIMQ::Deserialize(StreamReader& reader) {
...
 
+    CHECK_ARGUMENT(dim_ == info["dim"].GetInt(), "SIMQ deserialize dimension mismatch");
     dim_ = info["dim"].GetInt();
 
     if (info.Contains(INDEX_PARAM) && info[INDEX_PARAM].IsString()) {
         auto inner = JsonType::Parse(info[INDEX_PARAM].GetString());
         SIMQParameter tmp_param;
         tmp_param.FromJson(inner);
```
LHT129: [note] Similarly, `quantization_type_ = tmp_param.quantization_type` at line 1705 is a dead write. The `CHECK_ARGUMENT(quantization_type_ == tmp_param.quantization_type, ...)` on the preceding line already guarantees equality. Same consideration as the `dim_` case above.

src/algorithm/simq/simq.cpp
```diff
@@ -1242,6 +1175,7 @@ SIMQ::coarse_search(const float* query_tokens,
         coarse_seen_buf_.assign(n_docs, false);
     }
     coarse_dirty_.clear();
+    std::vector<bool> touched(n_docs, false);
```
LHT129: [suggestion] Allocating `std::vector<bool> touched(n_docs, false)` on every `coarse_search` call adds heap allocation overhead on the search hot path. The fix for the zero-score collision is correct (quantized distances can be exactly 0), but the allocation should be amortized.

Consider adding a `coarse_touched_` member (e.g. `Vector<bool> coarse_touched_` using the allocator) that is resized once and cleared per call, similar to how `coarse_score_buf_` and `coarse_seen_buf_` are already managed. This avoids the per-search allocation while keeping the correct zero-score semantics.

```cpp
// In coarse_search, replace:
std::vector<bool> touched(n_docs, false);
// with:
coarse_touched_.assign(n_docs, false);
```

src/algorithm/simq/simq.cpp
```diff
@@ -553,145 +636,81 @@ SIMQ::run_clustering(const float* flat_vecs,
...
-            const int64_t end = std::min(start + chunk_size, num_tokens);
-            if (start >= num_tokens) {
-                break;
+SIMQ::build_rep_hgraph(const uint8_t* token_codes, int64_t dim) {
+    std::vector<std::vector<InnerIdType>> members(num_clusters_);
+    for (uint64_t tid = 0; tid < vec_to_cluster_.size(); ++tid) {
+        members[vec_to_cluster_[tid]].push_back(static_cast<InnerIdType>(tid));
+    }
+    const auto code_size = mv_codes_->GetQuantizerCodeSize();
+    std::vector<InnerIdType> representatives(num_clusters_);
```
LHT129: [note] The old `build_rep_hgraph` used `thread_pool_` to compute cluster representatives in parallel when `num_clusters_ > 10`. The new implementation is purely serial (`for (int64_t c = 0; c < num_clusters_; ++c)`), which decodes every token in every cluster to compute the mean, then encodes the mean and compares against all members.

For large numbers of clusters or high-dimensional vectors, this serial loop may become a build-time bottleneck. The per-cluster work is independent (each cluster computes its own mean and representative), so it is trivially parallelizable. Consider restoring thread-pool parallelism for the representative selection loop, or document why serial execution is preferred here (e.g. the quantized code path makes parallelism less beneficial, or the FP32 decode dominates and is already bounded).

src/algorithm/simq/simq.cpp
LHT129: [suggestion] `std::vector<bool> touched(n_docs, false)` is allocated on every `coarse_search` call. The existing `coarse_score_buf_` and `coarse_seen_buf_` are already member variables that are lazily grown and reused across queries. Consider making `touched` a member variable (e.g. `mutable std::vector<bool> coarse_touched_buf_`) for consistency and to avoid a per-query O(n_docs) allocation in the search hot path.

```cpp
// In simq.h, add:
mutable std::vector<bool> coarse_touched_buf_;

// In coarse_search, replace:
//   std::vector<bool> touched(n_docs, false);
// with:
if (coarse_touched_buf_.size() < n_docs) {
    coarse_touched_buf_.assign(n_docs, false);
} else {
    std::fill(coarse_touched_buf_.begin(), coarse_touched_buf_.begin() + n_docs, false);
}
// Then use coarse_touched_buf_ instead of touched, and reset via coarse_dirty_ at the end.
```

LHT129 (commented):
## Review Summary

This is a follow-up review on commit `4b8b9f95`. The existing review threads have already covered the substantive issues thoroughly. Below is a summary of the remaining unaddressed items from prior review rounds that are still present in this commit:

### Still present (previously raised)

1. **`representative_codes_.resize()` inside a loop** (`src/algorithm/simq/simq.cpp:1059`): Called `new_cluster_count` times with the same final size. Should be moved outside the loop — resize once to the final capacity before the loop.

2. **Dead writes in `Deserialize`** (`src/algorithm/simq/simq.cpp:1700,1708`): `dim_ = info["dim"].GetInt()` and `quantization_type_ = tmp_param.quantization_type` are dead stores after `CHECK_ARGUMENT` already confirmed equality. These assignments have no effect.

3. **Per-call `std::vector<bool> touched` allocation in `coarse_search`** (`src/algorithm/simq/simq.cpp:1178`): Allocated on every query. Could be moved to a mutable member buffer (similar to `coarse_seen_buf_`) and grown lazily.

4. **`build_rep_hgraph` representative selection is now sequential** (`src/algorithm/simq/simq.cpp:650`): The old code used `thread_pool_` for parallel mean-and-best computation across clusters. The new code iterates clusters sequentially. For indexes with many clusters this may regress build time.

### Positive observations

- The `TokenCodeInterface` abstraction is clean and correctly separates encode/compute concerns.
- The serialization compatibility path (legacy → v1 migration) is well-designed with appropriate validation.
- Test coverage is good: lifecycle test across all 4 quantization types, serialization version rejection, and unit tests for the `TokenCodeInterface` on `MultiVectorDataCell`.
- The bounded 4096-token training sample avoids the previous O(total_tokens × dim) FP32 allocation.
- The `CHECK_ARGUMENT` guards throughout Deserialize provide strong corruption detection.

No new issues identified beyond those already in discussion.

